PR #1, opened 2015-06-12
A better setup for CI

ryanhiebert:


EmilStenstrom:
Excellent additions. Didn't think about the order of the dict.

As someone who hasn't run tox before, how do I run the tests? The ideas with the run_tests.py file was to make test running easy.

ryanhiebert:
You install tox with `pip install tox` (or some variation of it depending on whether it's a global install, user install, virtualenv, etc), then just run `tox` from anywhere in the repository. The rest is configured in the `tox.ini` file.

ryanhiebert:
Tox deals with setting up the virtualenvs for each version of python that you want to test with, and it's even able to set up a build matrix of virtualenvs for different versions of Django, etc.

EmilStenstrom:
Ok! Could you add a couple of lines to the Readme too then? How to install What you need to run the tests and then run them. Just to make participation easy.

Then lets merge this.

PR #4, opened 2019-11-22
Reviving the project

moalhaddar:


EmilStenstrom:
Hi! Big thanks for contributing your changes back to the project. I've now gone through all the code you've added!

1. I've added mark_safe to the template output, making the project compatible with new versions of django. I've also added all current versions of python and django to the test setup is available in Travis: https://travis-ci.org/EmilStenstrom/django-components
2. I don't want the change how Media is specified. I use the same format as is specified here: https://docs.djangoproject.com/en/2.2/topics/forms/media/
3. I've improved the documentation based on your example! Thanks!

The only thing I don't understand is the child component. Could you give an example of how you use it and why it is good to have?

EmilStenstrom:
@mohammedalhaddar Hi! Did you see my last question above?

mohammedalhaddar:
> @mohammedalhaddar Hi! Did you see my last question above?

Sorry for the late response. Been a really busy week.

The child component tag can be used to mock the behavior of react, something like:

```
<component>
    <child></child>
</component>
```
But there is a problem with that.
When rendering the parent component, we cannot pass the children to the library.

So the solution i've made is to define the children separately. Suppose that i have a *Card* component that is expected to receive a prop called title

`{% component name="Card" title="This is my title" %} 
`
The problem here is that we cannot add much customization for the title like html tags.

So a child component tag can be used to build a complex block as follows

```
{% child_component  as custom_title %}
   <div>
        <b>This is my title</b>
    </div>
{% end_child_component %}
```

Then we can pass *custom_title* as  a prop to the main component as

`{% component name="Card" title=custom_title %} `

EmilStenstrom:
Thanks for your explaination! I'll have to think about this, if this is a good addition to django-components. Very interesting! :)

EmilStenstrom:
Hi again!

There's now support for components with nested subparts, similar to how components work in react. I didn't quite use your API, but I think this ended up really nice anyways. Here's an example component:

```htmldjango
<div class="calendar-component">
    <div class="header">
        {% slot "header" %}Calendar header{% endslot %}
    </div>
    <div class="body">
        {% slot "body" %}Calendar body{% endslot %}
    </div>
</div>
```

When using the component, you specify what slots you want to fill and where you want to use the defaults from the template. It looks like this:

```htmldjango
{% component_block %}
    {% slot "body" %}Today's date is <span>{{ date }}</span>{% endslot %}
{% endcomponent_block %}
```

Since to header block is unspecified, it's taken from the base template. If you put this in a template, and send in date=2020-06-06, this is what's rendered:

```htmldjango
<div class="calendar-component">
    <div class="header">
        Calendar header
    </div>
    <div class="body">
        Today's date is <span>2020-06-06</span>
    </div>
</div>
```

Easy to use an intuitive. Thanks for the inspiration!

PR #8, opened 2020-07-08
Update README

BradleyKirton:
I noticed in the example the template loads django_components which I think should be component_tags.
I also updated the input name in the `def template` method in the example from date to context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ class Calendar(component.Component):
             "date": date,
         }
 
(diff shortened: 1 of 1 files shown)


EmilStenstrom:
@BradleyKirton I really should try to find a way to make sure the code in the README is tested somehow. Thank you for being my unit test! :) Let me know if there's anything else that would make the library better!

PR #13, opened 2020-11-02
Add Django classifier to PyPI

telenieko:
setup.py does not set the `Framework :: Django` classifier which makes it less discoverable by potential new users.


EmilStenstrom:
Thanks! Great little contribution!

Issue #16, opened 2020-12-13
Handling missing context_variables in slotted templates and passing kwargs in {% component_block %}


rbeard0330:
Sorry for putting both of these in one issue, but they're both so minor it seemed excessive to open separate issues:

1.  In slotted components, trying to include a variable that isn't available in the context raises an exception.  The reason seems to be that the render path for slotted components uses `nodelist.render()`, which automatically binds the template, rather than `template.render()`, which does not.  The problem goes away if the template is manually bound.

2.  Using {% component_block variable="XYZ" %}, it is currently an error if the component's `context` function inserts additional variables into the context (e.g., default parameters).  The issue is that if there are any arguments specified in the tag, `ComponentNode.render()` runs the entire context through the dictcomp that resolves `FilterExpression`s to generate `extra_context`.  This can be fixed by passing anything that doesn't have a `resolve` method through the dictcomp unchanged.

I put together a PR to demonstrate the issues and fix them--will send shortly.

EmilStenstrom:
Great finds! I'll add some comments to the PR, but looks really good! Thanks for helping me make django-components better!

EmilStenstrom:
Just released 0.4.3 with your fixes!

PR #17, opened 2020-12-13
Missing context variables and kwargs in component_block

rbeard0330:
These changes address both issues referenced in #16.  Happy to explain any changes or make any further edits you think appropriate.

### DIFF
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -85,7 +85,9 @@ class Component(with_metaclass(MediaDefiningClass)):
                 else:
                     nodelist.append(node)
 
-            return nodelist.render(Context(context))
+            render_context = Context(context)
+            with render_context.bind_template(template.template):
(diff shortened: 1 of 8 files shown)


EmilStenstrom (changes requested):
These changes look great! The only change I'd like to see is that the tests for this are separate from the rest of the tests. I'm trying hard to really use unit tests in this project, so that each test only tests one specific thing. Would you mind moving the tests to a separate test class with specific templates just for this?

rbeard0330:
Will do!

rbeard0330:
I updated the tests in the last two commits.  I also made an additional substantive change to pass extra_context to Component.render() in one place where it wasn't being passed.

EmilStenstrom:
Looks great, thanks for helping me make django-components better! :) And let me know if you encounter any additional roadblocks or annoyances.

Issue #12, opened 2020-10-14
`{% csrf_token %}` not available inside component HTML


Buroni:
Firstly thanks for creating this useful library.

I've noticed that `{% csrf_token %}` isn't rendered when used inside component template HTML. Other default Django template tags seem to work, so I'm guessing this is because the component HTML doesn't have access to Django's `RequestContext`.

A work-around is to just pass down the `{% csrf_token %}` inside a `slot` from the parent component, however this becomes repetitive. I'm wondering if it's currently possible to give the component access to the necessary context. Cheers!

EmilStenstrom:
Excellent question, this is something I have totally missed. I'll have a look and see if I can find a way to propagate the requestcontext down into a component, I think that is what people would expect.

I have lots of other things going on, but I'll see if I can find some time during the weekend. A PR would be appreciated though :)

Buroni:
Thanks @EmilStenstrom I'll try to dig into the library a bit at the end of the week.

EmilStenstrom:
I have looked at this for a couple of hours with no luck. The problem is that I can't manage to get a good testcase going. There are sooo many moving parts involved with csrf, and most current tests are at the unit test level. If you have the time and know how to create a failing test for csrf_token, please help me out!

EmilStenstrom:
@Buroni I just released django-reusable-components 0.5 to pypi, in which @rbeard0330 reworked context handling. Would you mind trying to call csrf_token inside a component again to see of things work now?

PR #18, opened 2020-12-22
Rework of context handling

rbeard0330:
Hey, sorry to drop a bunch of changes on you, but I encountered an issue, had to dig in deep to figure out what was going on, and then needed to change a bunch of stuff to fix my bug.  Here's an overview:

My original problem came when I tried to pass an object as a variable into a `component_block` tag, then have the context method process it into a handful of context variables.  Long story short, this wasn't working, because the context method sometimes gets called twice, with the outputs of the first call being the inputs of the second call.  That works fine if the context method doesn't transform its arguments, but fails when the context variables are different than the component params.

My fix was essentially just to have a single place for calling the context method (in the ComponentNode's render method), then to pass the context itself into the component's render method.  I think this is more straightforward and fixed my problem nicely.  As a side benefit, I think it will fix the CSRF problem in #12, at least when component_block is used.  Previously, the non-slot portions of the component were being rendered with a bare `Context`, not a `RequestContext`.  I reworked it so that the component renders itself with the context (presumably a RequestContext) that the template engine passes to it, so hopefully that will help.  I think using a component tag still won't work, because that tag doesn't get a context.  You could fix that by making it a `takes_context` tag, but I didn't make that change.

I also made some assorted tweaks as I was studying the code to help me follow better.  I left those in this PR, but you can obviously disregarded them if you don't think they're helpful.

I added a bunch of tests to look at various permutations of context, and everything passes.  That said, I can't swear that these changes won't break someone's code in some subtle way.

I hope these are helpful, and I'm happy to provide more explanation about anything I did.

### DIFF
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -1,5 +1,4 @@
 from django.forms.widgets import MediaDefiningClass
-from django.template import Context
 from django.template.base import NodeList, TextNode
 from django.template.loader import get_template
 from django.utils.safestring import mark_safe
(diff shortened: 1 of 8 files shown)


EmilStenstrom:
This looks great. Thanks for refactoring my code and making it clearer. Getting rid of getfullargspec is quite an accomplishment among many!

Few things before this can be merged cleanly (I really want this in, all of it!):
* Would you mind rebasing against master? This PR continues on your last PR, so all the old commits are still there.
* The reason CI is breaking is because of flake8. You can get it locally by running `pip install flake8` and `flake8` in the root of the project. 
* Minor: in test_context `{%endcomponent_block %}` should probably be `{% endcomponent_block %}` with a space.

How to fix the flake8 errors:
* For the "line too long" errors, just add a # NOQA comment at the end of the line, they are fine.
* Remove the double linebreak at the end
* Remove the python 2 compatibiliy stuff that imports the now unused getfullargspec

This is great! :)

rbeard0330:
Thanks!  I'll get these changes made tomorrow.  Best holiday wishes!

EmilStenstrom:
Hi again! The changes to fix the lints look great, but I'm still seeing old commits in the list above. Isn't "Add context tests" the first one in this PR?

rbeard0330:
Sorry, I'm the absolute worst at git.  Hopefully that did it...

EmilStenstrom:
Now released as django-components 0.5 on pypi! Happy to have you onboard! 🥳

PR #19, opened 2020-12-29
Minor bugfix

rbeard0330:
One more minor bug that I ran across.  If you have two `component_block` tags in the same template, and the second doesn't have any slots, you get a key error--the first tag adds the COMPONENT_CONTEXT_KEY subdict to the context, but nothing ever adds the second component as a key to the subdict before the render method tries to look that key up.


EmilStenstrom:
Great find! I like the code I see. It doesn't seem to run on older python (see red Travis link above) due to the f-strings in the tests. Mind having a look at that?

EmilStenstrom:
Thanks! Released 0.5.1 to pypi. 🚀

PR #20, opened 2021-01-12
Outer context modifications

rbeard0330:
Hi Emil,

I've been getting a ton of use out of this library and have come up with a few suggestions based on the rough patches that I've encountered.  This PR is focused on giving users more control over how the component can access the outer context, and also adding flexibility in how arguments are passed to the component.  I have a couple other ideas that aren't quite camera-ready... I'll mention them at the end of this.

This is a pretty large block of changes, so here's what's going on in detail:
* Both `component_block` and `component` now allow the user to decide whether or not the component should be able to access the outer context.  The syntax is modelled on the syntax for the [include tag in base Django](https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#include): by default, the outer context is passed, but you can append `only` to the end of the token to screen the component from the outer context.
* Both tags can now parse any combination of positional/keyword args and pass them to the context method.
* As a convenience, users now have access to an outer_context property in the Component context() method, which you can use to derive info from the outer context without having to pass the specific values in to the tag.
* To make things a little easier to work with, I made some significant changes to how slots are handled.  Previously, each slot rendered itself and included itself in the context, then the component pull the rendered strings out.  As I revised it, the slots no longer use their render methods.  Instead, the component just builds a single nodelist out of its slots and the rest of the template and renders it in one go.  Apart from (hopefully) simplifying the code, the main impact is that you can now have multiple instances of the same slot name in a block, and the component will just append all of the blocks into the template slot.

The other two ideas I wanted to get your thoughts on are both in branches on my fork, but not yet very well tested or documented:
* Declarative syntax for creating components ([here](https://github.com/rbeard0330/django-components/tree/new_subclass)).  The idea here was to remove some of the boilerplate in making new components.  Essentially, you subclass the new version of Component, assigning a name for the component, a template, and the component props as class variables.  Then there's an __init_subclass__ method that sanity-checks the component values and registers the component.  There's also an overrideable context method that assembles all the arguments and puts them in the context.  This should be fully backwards-compatible.
* This is a bit silly, but it would be nice to be able to pass several args to a component without having insanely long lines.  A potential fix ([here](https://github.com/rbeard0330/django-components/tree/extra_args)) is to have another tag (I called it component args) that the parser can harvest arguments from.  Not sure it's worth complicating the syntax for such a marginal benefit, but that's your call.

As always, happy to discuss/modify/whatever as you think best.

Cheers,
Robert

### DIFF
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -1,5 +1,7 @@
+from itertools import chain
+
 from django.forms.widgets import MediaDefiningClass
-from django.template.base import NodeList, TextNode
+from django.template.base import NodeList
 from django.template.loader import get_template
 from django.utils.safestring import mark_safe
 from six import with_metaclass
@@ -41,32 +43,24 @@ class Component(with_metaclass(MediaDefiningClass)):
 
         return mark_safe("\n".join(self.media.render_js()))
 
-    def slots_in_template(self, template):
+    @staticmethod
+    def slots_in_template(template):
         return NodeList(node for node in template.template.nodelist if is_slot_node(node))
 
     def render(self, context, slots_filled=None):
         slots_filled = slots_filled or {}
         template = get_template(self.template(context))
         slots_in_template = self.slots_in_template(template)
+        for default_slot in slots_in_template:
+            if default_slot.name not in slots_filled:
+                slots_filled[default_slot.name] = default_slot.nodelist
 
-        # If there are no slots, then we can simply render the template
-        if not slots_in_template:
-            return template.template.render(context)
+        # Replace slot nodes with their nodelists, then combine into a single, flat nodelist
+        node_iterator = ([node] if not is_slot_node(node) else slots_filled[node.name]
+                         for node in template.template.nodelist)
+        flattened_nodelist = NodeList(chain.from_iterable(node_iterator))
 
-        # Otherwise, we need to assemble and render a nodelist containing the nodes from the template, slots that were
-        # provided when the component was called (previously rendered by the component's render method) and the
-        # unrendered default slots
-        nodelist = NodeList()
-        for node in template.template.nodelist:
-            if is_slot_node(node):
-                if node.name in slots_filled:
-                    nodelist.append(TextNode(slots_filled[node.name]))
-                else:
-                    nodelist.extend(node.nodelist)
-            else:
-                nodelist.append(node)
-
-        return nodelist.render(context)
+        return flattened_nodelist.render(context)
 
     class Media:
         css = {}
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -24,9 +24,8 @@ except ImportError:
 
 
 class Component(with_metaclass(MediaDefiningClass)):
-    __cached_slot_nodes = None
 
-    def __init__(self, component_name):
+    def __init__(self, component_name='unnamed component'):
         self.__component_name = component_name
 
     def context(self):
@@ -55,7 +54,7 @@ class Component(with_metaclass(MediaDefiningClass)):
         return {node.name: node.nodelist for node in template.template.nodelist if is_slot_node(node)}
 
     def render(self, context, slots_filled=None):
-        slots_filled = dict(slots_filled) or {}
+        slots_filled = slots_filled or {}
 
         template = get_template(self.template(context))
         slots_in_template = self.slots_in_template(template)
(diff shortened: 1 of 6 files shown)


EmilStenstrom:
Hi! This looks like a great addition. Since the PR is huge, and contains many different changes, I will need some time to look at it. I’ll be back! :)

django_components/component.py
```diff
@@ -24,9 +24,8 @@ class TokenType:
 
 
 class Component(with_metaclass(MediaDefiningClass)):
-    __cached_slot_nodes = None
 
-    def __init__(self, component_name):
+    def __init__(self, component_name='unnamed component'):
```
EmilStenstrom, line 28: Why would you want to create a component without a name?

EmilStenstrom (commented):
This is SO MUCH CODE! :) I would really appreciate if you make smaller and more focused commits, preferably with descriptive comments. Would make my life easier! 

That said... I'm very happy with where it's going! Lots of old cruft cleaned up, and you can tell that you are well versed in writing template tags! One minor comment, then I'm happy to merge this.

EmilStenstrom:
About the declarative syntax: Could we take that in a separate PR? I'm having trouble visualizing how a component with this syntax is specified.

About the extra tag for long lines. I've like to avoid adding another tag for that. In similar situations I've encapsulated all the component data in an object, and passed that around instead. Would that be possible in your case?

django_components/component.py
rbeard0330: The reason I made this change was to let the dependency-rendering tags create components to access their media without needing to pass names for them.  Since the name is only used for informational purposes, it seemed easier to create a default rather than making changes there.

Understood on the need for more focused commits--I'm self-taught and self-employed, so being a good collaborator is definitely an area I can grow.  I appreciate your patience and the feedback.

EmilStenstrom: Hmm... I'm still not sure I follow. Doesn't the component registering happen before dependency rendering? If it does, shouldn't the dependency rendering code be able to get the names and the classes from the registry? Kinda funny that I'm a bit lost in the library I once wrote :) 

I'm only giving the feedback about focused commits because I've got that feedback myself, and found it very useful. I usually am all over the place when coding, fixing things as they come up, and therefore ending up with loads of changed code when it's time to commit. Two strategies I've found to deal with this:

- **My strategy**: I've learnt to add a step after my coding frenzies where I pick out changes line by line into commits before pushing a PR. I'm happy with the Tower App for this, but the Github App is good too. Command line is horrible for this.
- **Better strategy**: The best programmers I've worked with are using a "parking list". It's an open text file where they note down things they should "fix later". So when they encounter changes not related to the current thing they are working on, they add it there and stay focused on their original task. That specific task is what goes into the commit message (or ideally, the reason that task needs fixing). Seems simple, but requires immense self control that I hope to reach sometime :) 

Anyway, all these things are nice extras in the scheme of things. I'm super thankful that you are taking time from your busy day to help me make django-components great. I have a secret dream to merge it into Django one day :)

rbeard0330: Thanks, this is really helpful advice.  I'd love to see this library get folded into the mothership... it's really strange that there are so many different ways to include and inherit and modify templates in Django, but there's this huge gap for slots.

Anyways, I just pushed some changes that make the name argument mandatory.  I'd forgotten this nuance before I got back into the code, but the reason it's a little more complicated than just using the registry name is that component classes can be aliased in the registry.  I took the easy way out of just using the class `__name__` rather than trying to ensure that the names are unique--let me know if you want me to take a different approach.

rbeard0330:
> About the declarative syntax: Could we take that in a separate PR? I'm having trouble visualizing how a component with this syntax is specified.
> 
> About the extra tag for long lines. I've like to avoid adding another tag for that. In similar situations I've encapsulated all the component data in an object, and passed that around instead. Would that be possible in your case?

Completely understand on the second point, and there are definitely plenty of workarounds (or I guess just close your eyes and let there be a 200-character line...).

I'll work on teeing up the syntax proposal.  Before I do that, though, I've been tinkering a bit with Media classes to try to avoid including dependencies that aren't used.  I think I've got it working, but let me polish it up a little bit to be sure.

EmilStenstrom:
Thank you! Released Django-components 0.6.1 which includes your changes!

PR #21, opened 2021-01-28
Fixing bug in last PR

rbeard0330:
I'm embarrassed to report that my last PR interferes with Django's test instrumentation.  Specifically, if you try to use the `assertTemplateUsed` test assertion for `{% component %}`-rendered templates, it will fail (and the `assertTemplateNotUsed` assertion could pass erroneously).  I don't think this was ever working 100% right for `{% component_block %}`, but moving regular component renders from being a simple tag to using the `Component.render()` method has made them also not work properly.

The root cause is that the TemplateUsed/NotUsed assertions are driven by calls to `Template.render()` (which is a wrapper around `Template._render()`.  For `TestCase` (but not `SimpleTestCase`), Django monkey-patches `Template._render()` to emit a signal every time that method is called, which the test framework tracks.  For slotted components, `Component.render()` has always just built a nodelist and rendered it directly rather than going through a template, so those calls haven't been made.  My last couple rounds of changes have been pushing more and more components through that rendering path.

I think the fix is not too difficult.  The only problem with using `Template.render()` for slotted components is that the template has the wrong nodelist.  We can't directly modify the template's nodelist, since Django caches them, but we can do a shallow copy of the template, drop in the new nodelist, and render the copied component.  (I don't think a deepcopy is necessary, and it's also extraordinarily expensive, as each template has a reference to the entire template engine.)

Most of this PR is setting up some basic regression tests.  Django doesn't set up the TemplateUsed machinery for a `SimpleTestCase`, so I did (a very rudimentary version of) the setup for the test.  The alternative would be to create a dummy project so that the tests could run in a real `TestCase`.

### DIFF
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 import warnings
 from itertools import chain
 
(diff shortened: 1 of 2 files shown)


EmilStenstrom:
@rbeard0330 You are awesome! I won't never have found this :)

EmilStenstrom:
Just released 0.6.2 to PyPI if you want to use this using pip.

PR #25, opened 2021-02-05
Readme changes

danjac:


EmilStenstrom:
Inspired by your text I split it up into one part that just says you should add your components in the components.py file, and a meatier section in the end that describes more about how it works. Thanks!

Issue #26, opened 2021-02-09
"context" is empty in template method


danjac:
So I have this component:

```
class Svg(component.Component):
    def context(self, name: str, css_class: str = "", title: str = "", **attrs) -> Dict:
        return {"name": name, "css_class": css_class, "title": title, **attrs}

    def template(self, context: Dict) -> str:
        print(context)
        return f"svg/_{context['name']}.svg"
```

The template name is assigned dynamically depending on the name of the SVG:

{% component "svg" name="download" %}

However the template() method raises a KeyError, as the key "name" is not in the context: it's empty.

What is the order of execution here? Is template() called before context() ? If so, how do I access variables to be able to set the template name dynamically?

danjac:
Looks like there's some optimization happening here in component.py: 

        component_template = get_template(self.template({}))

So basically the "context" argument is pointless as it's *always* going to be {}, and is misleading as it gives the impression that the template context is available at rendering time.

If that's the intention, it kind of makes having "template" a method pointless as well: why not just make it a class attribute?

EmilStenstrom:
@rbeard0330 Is this related to your latest optimization?

EmilStenstrom:
@danjac This is a bug. The component you've written should work as you intended it too. It's actually a good example to add a test for. I have a lot on my plate right now, but give me a couple of days and I'll have a look.

rbeard0330:
> @rbeard0330 Is this related to your latest optimization?

Yes, definitely.  I think I mentioned in the PR that the change was breaking for this kind of use--fundamentally, you can't compile the component when the rest of the template is compiled if the identity of the component template changes each time the template is rendered, and you'll need to compile it fresh every render.  A couple thoughts:

* The native `{% include %}` tag can cover a lot of potential use cases by passing it a dynamic template to load.
* The easiest fix would probably be to add a flag as either a setting or a class attribute of each `Component` that would control whether the component tries to compile itself at compile time (with an empty context) or waits until render time and uses the full context.
* Agreed that, as currently written, the `get_template` method doesn't really need to be a method, but removing it or changing its signature would break existing code.

danjac:
Native "include" probably covers the use-case best here I think, and I wouldn't want to add complexity or performance regressions for this use case.

It would probably be a good idea to remove the "context" arg for template if it doesn't really do anything, or at least make it optional for backward compatibility.

EmilStenstrom:
[...]

PR #27, opened 2021-02-12
Cache compiled templates within component by template_name

rbeard0330:
As discussed in #26.

Only difference from what was discussed earlier is that I didn't invoke the compilation method when the template is compiled.  It would require catching KeyErrors that are thrown when a client tries to look up something they expect to be in the context, and I don't think it will actually improve responsiveness--I'm not sure I have this right, but I think Django compiles a template immediately before the first time it renders it, so it's part of the first request-response cycle either way.

I used `lru_cache` for the caching, and defined a do-nothing decorator to avoid breaking Python 2.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ import os
 
 from setuptools import find_packages, setup
 
-VERSION = '0.8'
+VERSION = '0.8.1'
 
 setup(
     name="django_reusable_components",


django_components/component.py
```diff
@@ -1,5 +1,13 @@
 import warnings
 from copy import copy
+try:
+    from functools import lru_cache
+except ImportError:
+    def lru_cache(maxsize=None):  # noqa
+        def decorator(func):
+            return func
+        return decorator
+
```
EmilStenstrom, line 10: I removed support for Python 2, so this can be removed!

django_components/component.py
```diff
@@ -55,15 +64,16 @@ def render_js_dependencies(self):
     def slots_in_template(template):
         return {node.name: node.nodelist for node in template.template.nodelist if is_slot_node(node)}
 
-    def compile_instance_template(self, slots_for_instance):
+    @lru_cache(maxsize=128)
```
EmilStenstrom, line 67: Is there a rationale for setting this to 128? Should be the number of unique times all your components get called with different values for template_name right? 

Maybe we could make this configurable via the COMPONENTS setting that @danjac added?

tests/test_component.py
```diff
@@ -32,7 +32,7 @@ class Media:
 
         comp = SimpleComponent("simple_component")
         context = Context(comp.context(variable="test"))
-        comp.compile_instance_template({})
+        #comp.compile_instance_template({})
```
EmilStenstrom, line 35: This should be removed, not commented out, right?

EmilStenstrom (commented):
I'm very happy with this! We even got the simpler interface for creating Components in Python back too! I added a couple of comments, but they are minor.

django_components/component.py
EmilStenstrom: You should be able to get tests to pass by rebasing your code from latest master and force pushing to your branch.

django_components/component.py
rbeard0330: It's the default value, but I think it does make sense to make it configurable.  I would bet that it won't matter for 95% of users, because they'll have < 10 templates ever used, but people who care at all will probably care a lot.

EmilStenstrom: Fully agree. So we add a default to COMPONENTS and always read from there?

EmilStenstrom:
I'm not sure that rebase worked. I see a merge commit and my own commits above. Anyways, and can select the commits I want in the list above and review them.

About the code! This looks great. I think we need to add a default for TEMPLATE_CACHE_SIZE in app_settings the same way the two other properties are set there. Otherwise we're good to go. 

Thanks again! :)

EmilStenstrom:
@rbeard0330 I messed up my git game, but your code is merged, I promise :) Thanks! Also released version 0.9 of the library with your fixes.

Issue #31, opened 2021-02-20
Class decorator for registering components


danjac:
So with Django admin I can do this:
```

@admin.register(MyModel):
class MyModelAdmin(admin.ModelAdmin):
    ....


```
It would be nice to have similar functionality for django-components:
```

@component.registry.register("button")
class Button(component.Component):
    ...
```

EmilStenstrom:
Agreed, would be a nice addition 👍

rbeard0330:
I wrote this for myself as a convenience (I think it's a one-liner), but I actually think the better API is to auto register each subclass of component. In the admin case, you often don't want to register a model with the admin, so it makes sense to have users opt-in. An unregistered component doesn't do anything, so making the user register it at all is boilerplate that it would be best to eliminate if possible (other than allowing the user to specify the component name, although you could snake-case the class name as a default option).

I have a version of this for internal use that I can set up as a PR, hopefully today.

danjac:
The problem is that I might not want to auto-register each component subclass : I might have for example a base component class that isn't used itself but has multiple subclasses. You would need some way to exclude this superclass. You could add some "abstract" attribute (as with models) but this adds its own complexity and documentation requirements.

Registering a component is also a pretty well known pattern e.g. with cusom template tags as well as the Django admin. 

The counter-argument though are Django models themselves: Django automatically picks up models for migrations etc without having to register each model.

EmilStenstrom:
Another case where having unregistered components would make sense, is if someone wrote a package with a set of pre-made components that you could import and use at will. You won't automatically want all components registered just because you enabled the package.

I do agree that auto-registering is what most people would want, most of the time. It's also one thing less for beginners to think about, and possibly get wrong... Tricky.

I'm leaning towards sticking with a registry after all. As danjac says, it should be familiar to Django users from template tags or admin. Updating the readme to include a decorator in the basic examples would surely help beginner get things right...

PR #38, opened 2021-02-27
Class decorator and test

danjac:
This adds a simple class decorator to fix https://github.com/EmilStenstrom/django-components/issues/31. Basically this:

```
from django_components import component

@component.register("my-component")
class MyComponent(component.Component):
    ...
```

### DIFF
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -97,3 +97,20 @@ def is_slot_node(node):
 
 # This variable represents the global component registry
 registry = ComponentRegistry()
+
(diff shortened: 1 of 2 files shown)


EmilStenstrom:
Looks great! I though that we maybe should have a test of some error cases, but nvm. Code is simple enough.

EmilStenstrom:
I also updated the README to point to the decorator syntax as the default. This is available in django-components 0.9.1 in PyPI.

Issue #37, opened 2021-02-27
Slot is not executed in conditional {% if %} block


danjac:
I have this component template:

```
<div>
...
{% if show_extra %}
{% slot extra_text %}{% endslot %}
{% endif %}
</div>
```

And in another template:
```

{% component_block "mycomponent" show_extra=True %}
{% slot extra_text %}show me!{% endslot %}
{% endcomponent_block %}

```
The text "show me!" is not rendered, just the slot default content. Overriden {% block %} content on the other hand is rendered inside a conditional.

danjac:
I wonder if this is related to https://github.com/EmilStenstrom/django-components/issues/34: the current implementation of slots as a simple list doesn't work in when the slots themselves are executed conditionally.

EmilStenstrom:
Interesting, you are a great tester for this project! :) I think we should do whatever block is doing, makes sense.

jereze:
Hi,
I noticed the same behavior.

EmilStenstrom:
@jereze Would you mind helping out fixing it? A PR that adds a test that shows the error would be a great start for instance.

rbeard0330:
> I wonder if this is related to #34: the current implementation of slots as a simple list doesn't work when the slots themselves are executed conditionally.

I think this is right.  The current implementation looks at the components nodelist, identifies any slot nodes *in that nodelist*, then replaces them with either the provided content or the default content.  In both of these issues, the problem is that there is a slot nested within some other node (either an `IfNode` or another slot).  If that happens, then the `SlotNode` isn't removed when the template is compiled, and it renders itself using its default content.

You could potentially fix this by making the search for slots recursive.  In conjunction with that, I think you'd also want to modify the slot's own nodelist rather than replacing it like we do currently, to minimize the chance that we violate some other node's expectations.

Issue #33, opened 2021-02-20
slot.super?


danjac:
Suppose I have a component with a slot:

```
<div>
  {% slot  header %}some default content...{% endslot %}
</div>

```
There doesn't appear to be a way to do something like {{ block.super }} i.e.

```
{% component_block "my-component" %}
{% slot header %}
some new content 
{{ slot.super }}
{% endslot %}
```

EmilStenstrom:
Great suggestion, I'm 👍 for adding this.

rbeard0330:
An inelegant, but easy, way to support this would be to make a variant of the slot tag that doesn't have a body and just adds the default content. On its own, this wouldn't do anything, but you can use the same slot name multiple times, so you could build up a final slot that looks the way you want through a combination of "real" slot tags and the new default slot tag.

EmilStenstrom:
@rbeard0330 I would like to stick as close to how `{% block %}` works as possible as this is what I think people expect, and therefore might make django-components easiest to learn. So I think supporting this use case would mean to add a dict to the context of each slot with the name `slot` and the content `{"super": "Default content for this slot"}`. That would mean `{{ slot.super }}` returns the correct thing.

PR #56, opened 2021-04-29
Improved slot handling

### LINKED ISSUE #34: Nested slots don't work
Example:

```
  {% slot form_footer %}
  <div class="flex items-center justify-end">
    {% slot form_buttons %}{% endslot %}
  </div>
  {% endslot %}
```

I use the component thus:

```
{% component_block "form" %}
{% slot form_buttons %}
<button>Submit</button>
{% endslot %}
{% endcomponent_block %}

```
The content of the form_buttons is not overridden, i.e. it's empty. Only the outer content of form_footer is rendered:

```
  <div class="flex items-center justify-end">
  </div>

```

rbeard0330:
Emil,

This should fix #34 and #37.  The general approach is to allow `SlotNode`s to render themselves, instead of removing them from the template and replacing them with the correct nodelist at the time the template is compiled.  Now, the first time the component is rendered, it loads its template, searches through all the nodes in the template (including through descendants), and gives each `SlotNode` a reference to itself (i.e., the parent component).  Then, when it is time for the `SlotNode` to render itself, it checks to see if it has a parent reference and, if it does, if the parent has an overriding nodelist for it to use.  If an override exists, it renders that, otherwise it renders its default nodelist.

I think it should be possible to extend this approach to handle #33 (slot.super()) as well, but I haven't fully thought that through yet. 

As always, let me know of comments/questions/suggestions.

Best,
Robert


rbeard0330:
I just pushed a couple additional commits.  These address two issues:

1. I started encountering performance problems with the test suite.  Some investigation showed that the deepcopy of the template nodelist in Component.get_processed_template() was picking up a node that had a reference to the whole template engine.  Since we only need to copy the nodelists, I wrote a helper function to recursively copy the nodes in the nodelist, along with their subsidiary nodelists.

2. I also added an implementation of `{{ slot.super }}`.  To avoid concurrency issues, a new `SlotNode` is instantiated for each render, and it's receives a reference to the context for the render.  Then the new `SlotNode` can use the saved context and its default nodelist to respond to super calls.

EmilStenstrom:
THIS IS FANTASTIC!!! :D Thank you so much for this. Give me a couple of days to go through the changes you made. You made my day! :)

tests/test_templatetags.py
```diff
@@ -356,3 +356,253 @@ def test_nested_component_templates_all_shown_as_used(self):
...
+            </custom-template>
+        """,
+        )
+
+    def test_multiple_super_calls(self):
+        template = Template(
+            """
+            {% load component_tags %}
+            {% component_block "test" %}
+                {% slot "header" %}First: {{ slot.super }}; Second: {{slot.super }}{% endslot %}
```
EmilStenstrom, line 541: Minor: "{{slot" => "{{ slot"

tests/test_templatetags.py
```diff
@@ -356,3 +356,253 @@ def test_nested_component_templates_all_shown_as_used(self):
...
+        )
+
+    def test_super_is_scoped_to_slot(self):
+        template = Template(
+            """
+            {% load component_tags %}
+            {% component_block "test" %}
+                {% slot "header" %}Override header{% endslot %}{{ slot.super }}
+                {% slot "main" %}{{ slot.super }}{% endslot %}{{ slot.super }}
+                {% slot "footer" %}{{ slot.super }}{% endslot %}{{ slot.super }}
```
EmilStenstrom, line 565: I think the behaviour here should be to raise an error instead of erroring silently?

EmilStenstrom (commented):
Now I've finally looked through it! I added some questions and minor things, but overall this looks great! Mind having a look so we can get this merged? :)

django_components/component.py
```diff
@@ -73,45 +75,54 @@ def slots_in_template(template):
...
-        defined_slot_names = set(slots_in_template.keys())
-        filled_slot_names = set(self.slots.keys())
-        unexpected_slots = filled_slot_names - defined_slot_names
-        if unexpected_slots:
-            if settings.DEBUG:
+    def get_processed_template(self, template_name):
+        """Retrieve the requested template and add a link to this component to each SlotNode in the template."""
+
+        source_template = get_template(template_name)
+        component_template = copy(source_template)
```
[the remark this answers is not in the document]
rbeard0330: I think it is, but let me think about this and try to write a test to capture the problem if there is one.

tests/templates/conditional_template.html
```diff
@@ -0,0 +1,6 @@
+{% load component_tags %}
+{% if branch == 'a' %}
+    <p id="a">{% slot 'a' %}Default A{% endslot %}</p>
+{% elif branch == 'b' %}
+    <p id="b">{% slot 'b' %}Default B{% endslot %}</p>
+{% endif %}
```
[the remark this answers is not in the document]
rbeard0330: Will do!

tests/test_templatetags.py
rbeard0330: Thanks, fixed.

tests/test_templatetags.py
rbeard0330: Makes sense and will do.

rbeard0330: Actually, unresolving this after spending more time with it.  This pattern is just a specific case of having any content in a component block that isn't also in a slot block.  It's actually hard to raise an error for bad `slot.super` calls specifically because anything that isn't a `SlotNode` is just discarded during parsing.  We'd have to inspect the skipped nodes and decide whether or not to raise the error.  Would it be better to treat any orphan nodes as `TemplateSyntaxError`s (either all the time or only in debug mode)?  Or debug mode warnings?

EmilStenstrom: I'm totally fine with leaving it as it is. I think it would be nicer, but not worth a lot of effort imho!

EmilStenstrom: Marking this as resolved!

django_components/component.py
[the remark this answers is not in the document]
EmilStenstrom: @rbeard0330 I don't want to rush, is this conversation still unresolved, or did your commit at the end resolve this? I don't know if you are waiting for me to check again or not :)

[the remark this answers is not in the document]
rbeard0330: All set.  Just pushed some commits with the new test and some slight tweaking of this section.

tests/test_templatetags.py
rbeard0330: I ended up putting this in and playing around with the parsing code generally, but it's a separate commit that we can easily leave out if you prefer.

EmilStenstrom:
Looks great! Thanks again for your hard work. I'll merge this right now and release version 0.13 🥳 🎊 🎆

PR #65, opened 2021-05-29
Rework slot management to avoid nodelist copying

### LINKED ISSUE #64: Duplicate node / deepcopy bug in 0.13
Hi,

Hitting some issues running 0.13 - our CI tests generate the following error - which recurses into a long set of `deepcopy` operations, failing eventually within the pickle `failure` at the bottom. I haven't dug into the code but think there may perhaps be a bug in the `duplicate_node` that is copying additional python objects that it shouldn't be.

```python
  File "/home/runner/work/datapane-hosted/datapane-hosted/dp-server/.venv/lib/python3.9/site-packages/django_components/component.py", line 146, in duplicate_node
    setattr(clone, nodelist_name, type(nodelist)(nodelist_contents))
AttributeError: can't set attribute

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/runner/work/datapane-hosted/datapane-hosted/dp-server/.venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/runner/work/datapane-hosted/datapane-hosted/dp-server/.venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 204, in _get_response
    response = response.render()
...
  File "/home/runner/work/datapane-hosted/datapane-hosted/dp-server/.venv/lib/python3.9/site-packages/django_components/component.py", line 151, in duplicate_node
    return deepcopy(source_node)
  File "/opt/hostedtoolcache/Python/3.9.5/x64/lib/python3.9/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/opt/hostedtoolcache/Python/3.9.5/x64/lib/python3.9/copy.py", line 270, in _reconstruct
    state = deepcopy(state, memo)
  File "/opt/hostedtoolcache/Python/3.9.5/x64/lib/python3.9/copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "/opt/hostedtoolcache/Python/3.9.5/x64/lib/python3.9/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
...
  File "/opt/hostedtoolcache/Python/3.9.5/x64/lib/python3.9/copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "/opt/hostedtoolcache/Python/3.9.5/x64/lib/python3.9/copy.py", line 161, in deepcopy
    rv = reductor(4)
TypeError: cannot pickle '_thread.lock' object
```

rbeard0330:
See attached for some revisions that hopefully fix #64.

The general approach is for each component to add a context variable that is a dict with the names of the slots that have been provided to it and the relevant nodelists.  When each slot is rendered, it checks for its name in the dict, and renders either the override or the default nodelist.

If the dict isn't in the context, it means that the slot is either being rendered outside of a component or without access to the context provided by the component.  The current approach is to call this a TemplateSyntaxError in debug mode and to just render the slot's contents in production.  Alternatively, you could introspect your way up the call stack looking for the component that's making the render call, but that seems too weird to be a good idea.  I left that code in here in inactive form so you can see it, but I think it's probably best to just drop it.

### DIFF
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -1,10 +1,9 @@
 import warnings
-from copy import copy, deepcopy
 from functools import lru_cache
 
 from django.conf import settings
 from django.forms.widgets import MediaDefiningClass
-from django.template.base import Node, NodeList, TokenType
+from django.template.base import Node, TokenType
 from django.template.loader import get_template
 from django.utils.safestring import mark_safe
 
@@ -12,6 +11,7 @@ from django.utils.safestring import mark_safe
 from django_components.component_registry import AlreadyRegistered, ComponentRegistry, NotRegistered  # noqa
 
 TEMPLATE_CACHE_SIZE = getattr(settings, "COMPONENTS", {}).get('TEMPLATE_CACHE_SIZE', 128)
+ACTIVE_SLOT_CONTEXT_KEY = '_DJANGO_COMPONENTS_ACTIVE_SLOTS'
 
 
 class SimplifiedInterfaceMediaDefiningClass(MediaDefiningClass):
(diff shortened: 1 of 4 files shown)


EmilStenstrom:
@rbeard0330 Thank you so much for this. Is there a reason why you don't want to raise a TemplateSyntaxError in production? Removing that distinction and always raising on error seems simpler and more consistent to me. Removing the copying is likely good for performance too, especially for more complicated templates.

Looking at stack frames seems complicated, so I fully agree with just raising an error, Simpler is usually better.

There are a couple of unrelated changes there too that I'm unsure of the reason for, but nothing that I'm against.

All in all, I think we should drop the special casing of error handling, and ship this!

EmilStenstrom:
Also, I want to make you a full owner of this project, with the rights to commit to master and make releases. Would you be interested?

rbeard0330:
Sorry for the delay on getting your changes in... I didn't have a compelling reason to be laxer in non-debug mode, so fine with handling it that way.

> Also, I want to make you a full owner of this project, with the rights to commit to master and make releases. Would you be interested?

I'd definitely be interested in this!  I'll probably still want to run all my code by you anyways though.  :)

EmilStenstrom:
@rbeard0330 There are not deadlines in open source :) 

How do I best contact you? Zoom/Messenger/E-mail? What do you prefer?

rbeard0330:
Email is best.

PR #85, opened 2021-08-28
Added template loader for component templates automatic discovery.

Real-Gecko:
Added a template loader and modified README accordingly.


Real-Gecko:
Third commit refactors `Component` class to make it easier to use and and to resemble Django design patterns (views mostly). If you don't like this you can cherry pick only the commit that adds template loader.

EmilStenstrom:
@Real-Gecko Big thanks for making a PR, much appreciated! Let me go through it right away to see if I understand things correctly.

**First commit**: I'm not ready to switch to poetry for dependency management. Maybe later, and definitely in a separate PR.

**Third commit**: I don't agree with the refactoring of the Component class. I agree that setting the template_name variable is easier than returning it from a function, but in changing that we break backwards compatibility with all existing code, and we can't have that. I'm open for this, but I'd like this to be done without breaking backwards compatibility (checking to see if a template_name property exists dynamically?)

I'll add inline comments to the second commit below:

README.md
```diff
@@ -22,23 +22,45 @@ Read on to learn about the details!
 
 Install the app into your environment:
 
-> ```pip install django_components```
+```> pip install django_components```
```
EmilStenstrom, line 25: Why did you change this?

README.md
```diff
@@ -22,23 +22,45 @@ Read on to learn about the details!
...
 
 Then add the app into INSTALLED APPS in settings.py
 
-```
+```python
```
EmilStenstrom, line 29: Great catch, had missed this!

README.md
```diff
@@ -22,23 +22,45 @@ Read on to learn about the details!
...
 INSTALLED_APPS = [
     ...,
     "django_components",
     ...
 ]
 ```
 
+Modify `TEMPLATES` section of settings.py as follows:
+- Remove `'APP_DIRS': True,`
```
EmilStenstrom, line 38: Does this really have to be removed? As I understand it it just adds more directories to look into for templates. I can imagine people needing this setting for other reasons.

README.md
```diff
@@ -22,23 +22,45 @@ Read on to learn about the details!
...
+        ...,
+        'OPTIONS': {
+            'context_processors': [
+                ...
+            ],
+            'loaders':[(
+                'django.template.loaders.cached.Loader', [
+                    'django.template.loaders.filesystem.Loader',
+                    'django.template.loaders.app_directories.Loader',
+                    'django_components.template_loader.Loader',
```
EmilStenstrom, line 52: This is the only line that's needed right? I have an idea of how to make this easier to follow, but I can fix that in a separate commit.

README.md
```diff
@@ -157,8 +179,9 @@ class Calendar(component.Component):
             "date": date,
         }
 
+    # Note that Django will seek for templates inside `[your app]/components` dir
```
EmilStenstrom, line 177: Nitpick: seek -> look

django_components/template_loader.py
```diff
@@ -0,0 +1,12 @@
+"""
+Wrapper for loading templates from "components" directories in INSTALLED_APPS
+packages.
```
EmilStenstrom, line 3: Suggestion to change this to: Template loader that loads templates from each Django app's "components" directory.

EmilStenstrom (changes requested):
Great work! Would you minding answering a couple of questions?

Real-Gecko:
> I'm not ready to switch to poetry for dependency management. Maybe later, and definitely in a separate PR.

Will revert that, but take a look at `poetry` it's quite cool and easy to use, especially assuming that this project has only dev dependencies.

Real-Gecko:
> I agree that setting the template_name variable is easier than returning it from a function, but in changing that we break backwards compatibility with all existing code, and we can't have that.

This will not break any backwards compatibility as `Component.template` method was raising `NotImplementedError`:
```python
   def template(self, context):
        raise NotImplementedError("Missing template() method on component")
```
That means that every class inherited from `Component` had to implement this method, so if somewhere in your code you call `template` method you'll call it from child class anyways. As for calling it from parent class there's already a code that handles the issue in `render`:
```python
    def render(self, context):
        ...
        if hasattr(self, 'template'):
            warnings.warn(
                f'{self.__class__.__name__}: `template` method is deprecated, set `template_name` or override `get_template_name` instead',
                DeprecationWarning
            )
            template_name = self.template(context)
        else:
            template_name = self.get_template_name(context)
```

README.md
Real-Gecko: Emmm, will revert that one.

README.md
Real-Gecko: Yep, otherwise Django will complain. `APP_DIRS = True` enables `django.template.loaders.app_directories.Loader` and as we have to add our own loader we must put it in `loaders` explicitly, `APP_DIRS` becomes redundant. Refer to this section https://docs.djangoproject.com/en/3.2/ref/templates/api/#django.template.loaders.cached.Loader and above for clarification.

README.md
Real-Gecko: Refer to this section https://docs.djangoproject.com/en/3.2/ref/templates/api/#django.template.loaders.cached.Loader

README.md
Real-Gecko: :D

django_components/template_loader.py
Real-Gecko: That's just a copy paste from https://github.com/django/django/blob/main/django/template/loaders/app_directories.py :D

Real-Gecko:
I've done updates you requested.

EmilStenstrom:
> > I agree that setting the template_name variable is easier than returning it from a function, but in changing that we break backwards compatibility with all existing code, and we can't have that.
> 
> This will not break any backwards compatibility as `Component.template` method was raising `NotImplementedError`:

Ah, I see how you were thinking now!

Still, you are deprecating the old names, template() and context() and replace them with new names. How come?

EmilStenstrom:
> 
> I've done updates you requested.

Looks good! This PR is ready to be merged if we move out the refactorings to a separate PR?

Real-Gecko:
Alrighty, I've reverted all changes to component refactor. Commit history is a little bit messy, feel free to squash commits on merge.

Real-Gecko:
> Ah, I see how you were thinking now!
> 
> Still, you are deprecating the old names, template() and context() and replace them with new names. How come?

Well if those methods will be unused it's OK to inform developers that they're deprecated. If you run tests with Django 3.2 you'll notice deprecation warnings from Django itself :D

PR #87, opened 2021-08-30
Add a github action to automate releases to PyPI

### LINKED ISSUE #69: Automate releases to PyPI with GitHub Actions
In an attempt to make this project less dependent on me I'd like to automate releases to PyPI. I think a good way to do this is to move the PyPI account from my personal to a new one, put the secrets for the new one as GitHub environment variables, and make a GitHub Action that uses them to push new releases to PyPI. That way, everyone with GitHub access to django-components will also be able to make releases.

hanifbirgani:
This pr adds a github action file to automatically publish tagged releases to PyPI server. fixes #69

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,0 +1,46 @@
+name: Publish to PyPI 
+
+on:
+  push:
+    tags:
+      - '*'
+
+  # Allows you to run this workflow manually from the Actions tab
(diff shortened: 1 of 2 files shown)


EmilStenstrom:
I just made the first release (0.16) using the GitHub Action you created. Worked great! https://github.com/EmilStenstrom/django-components/actions/workflows/publish-to-pypi.yml

PR #94, opened 2021-09-11
Move duplicated isort config from setup.cfg to pyproject.toml

hanifbirgani:
There were duplicated config for isort in two files, this PR merge them into pyproject.toml

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,10 +1,3 @@
 [flake8]
 ignore = E302,W503
 max-line-length = 119
-
-[isort]
-line_length = 119
-multi_line_output = 5
-include_trailing_comma = True
-known_first_party =
-	django_components
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,3 +19,5 @@ exclude = '''
 [tool.isort]
 profile = "black"
 multi_line_output = 3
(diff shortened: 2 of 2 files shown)


setup.cfg
```diff
@@ -1,10 +1,3 @@
 [flake8]
 ignore = E302,W503
 max-line-length = 119
-
-[isort]
-line_length = 119
-multi_line_output = 5
```
EmilStenstrom, line 7: This changes multi_line_output from 5 to 3, and removes the line-length setting from isort. Is that intentional?

hanifbirgani: Yes.
1. To resolve conflicts between isort and black [multi_line_output should be set to 3](https://pycqa.github.io/isort/docs/configuration/black_compatibility.html).
2. Currently, black uses standard pep-8 line_length=79. We should sync the line_length in all of our formatting tools (isort/black/flake8). We can set all of them to 119 if you prefer.
3. The main reason for this PR was cleaning duplicated configs in two files.

EmilStenstrom: Removing duplication is great. 

I think we should have the same line-length everywhere. I prefer 119, but it isn’t that important.

PR #102, opened 2021-10-28
Aligning component js with html in example

spollard:
Perhaps fixing an example. Does the js end up in the head even when the component_js_dependencies is at the end of the body?


EmilStenstrom:
@spollard Thanks for fixing the README, these little things are more imporant than you would think!

PR #128, opened 2022-01-24
Implement slot exports

rbeard0330:
Sorry for delay on this--holiday season plus school closures have been making it hard.

Here's a working implementation of exportable slots.  This uses the syntax we talked about earlier in the issue thread, so please take a look and let me know if you still have reservations.

At a high level, the changes are:
* Expand parsing of slot tags to collect the new variables.
* Modify Component class so that it knows what exported names it's supposed to be looking for in the context when it renders itself, and also to override the existing default content for any exported slots that provide their own defaults.
* Add another context variable that keeps track of slots that were provided but haven't been used.  This is where components look for exported slots that have been filled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,6 +314,36 @@ Produces:
 </div>
 ```
 
+Normal slots can only be filled with "local" content, meaning content that is included between the `component_block` tags that create the component that has the slot.  If you need more flexibility, you can "export" a slot. This can be useful when you're creating a nested component. Exporting a slot is as simple as including the slot in the right component and adding an `export_as` argument to the tag. For example, if we wanted to create a dashboard component that uses our calendar component, we could do it with a template like this:
+
+```
+<div id="dashboard">
+{% component_block "calendar" date="2020-06-06" %}
+    {% slot header export_as=calendar_header %}{% endslot %}
+    {% slot body %}Here are your to-do items for today:{% endslot %}
+{% endcomponent_block %}
+<ol>
+{% for item in items %}
+    <li>{{ item }} </li>
+{% endfor %}
+</ol>
+</div>
+```
+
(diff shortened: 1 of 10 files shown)


EmilStenstrom:
Hi! No worries about any "delays": This is voluntary work on a random open source project. I'm happy for all the time you can give.

Thanks for the explanatory text in the PR, helped me understand the implications of this. I think nested slots is an important problem to solve for this project.

I'm still not fond of the syntax 🤔 . There's something about it that doesn't make sense in my head, and I keep re-reading the examples and explanations to understand what it's doing. After think about this for a while, I think we should go with the {% slot %} and {% fill %} combination. This does break backwards compatibility, and also requires that people change their code when upgrading, but I think it's worth it to get it right.

Here are the two examples from your documentation commit using this syntax:
```
<div id="dashboard">
{% component_block "calendar" date="2020-06-06" %}
    {% fill header %}
        {% slot calendar_header %}{% endslot %}
    {% endfill %}
    {% fill body %}Here are your to-do items for today:{% endfill %}
{% endcomponent_block %}
<ol>
{% for item in items %}
    <li>{{ item }} </li>
{% endfor %}
</ol>
</div>
```

And...

```
<div id="dashboard">
{% component_block "calendar" date="2020-06-06" %}
    {% fill header %}
        {% slot calendar_header %}Welcome to your dashboard!{% endslot %}
    {% endfill %}
    {% fill body %}Here are your to-do items for today:{% endfill %}
{% endcomponent_block %}
...
```

This syntax just makes sense to me. I can fully understand what's happening without any other explanation. It also doesn't require learning any new concept when going from the normal `slot` and `fill` concepts to the nested form. Things just work.

**What are your thoughts on this syntax?** Does it make sense to you too? I understand this is more work, and I'm sure I've missed some implication of this.

EmilStenstrom:
Also ping @hanifbirgani for thoughts. Is changing the syntax worth it for more readable nested slots?

hanifbirgani:
In my view, having more readable syntax is one of the most required features in component based systems, so I root for readable syntax 👍

EmilStenstrom:
Closed in favour of: https://github.com/EmilStenstrom/django-components/issues/133

PR #144, opened 2022-04-28
Fix #143: ability to preload components

### LINKED ISSUE #143: Ability to choose which components JS/CSS to load
Hi there,

We have a page that initially doesn't render any component, but is able to insert into its DOM some HTML fragments (through htmx or hotwired or stuff like that), which themselves may carry some components.

With current `django-components` capabilities, we have two options when calling `{% component_js_dependencies %}` :
* If the `ComponentDependencyMiddleware` is disabled, JS from all components will be loaded, which is suboptimal
* If the middleware is enabled, no component JS will be loaded, as the page template doesn't hold any `{% component mycomponent %}` ; that forces us to load JS with the classic `<script src="{% static 'path/to/component.js' %}">`

My first thought would be to add optional arguments to `component_js_dependencies`/`component_css_dependencies`, listing components that have a chance to appear on this page. Do you think it's the way to go? If so, I can work on a PR :wink:

David-Guillot:


EmilStenstrom:
@David-Guillot Let me know when you are ready for a review! :)

David-Guillot:
Yeah I'll need some more time because I'm writing more tests and I just found out that the approach I took make some duplicate rendering as I didn't take the middleware into account. Reworking... :slightly_smiling_face:

David-Guillot:
Actially @EmilStenstrom I think I have here a first reliable version. I has to reuse the `<!-- _RENDERED -->` HTML comment in order to keep it simple. Let me know what you think about it :wink:

David-Guillot:
@EmilStenstrom I'd be happy to get your review on this one :wink:

PR #153, opened 2022-07-12
Component name as variable

housUnus:
I have removed the exception raising when a variable without quotes is given, and keep the variable name until the render function where I'm resolving it if it's a String (which means it's a variable, not a component) then I'm moving the slot parts which relies on the component from __init__ to render after getting the component.

### DIFF
--- a/django_components/templatetags/component_tags.py
+++ b/django_components/templatetags/component_tags.py
@@ -204,12 +204,8 @@ class ComponentNode(Node):
         self.context_args = context_args or []
         self.context_kwargs = context_kwargs or {}
         self.component, self.isolated_context = component, isolated_context
+        self.slots = slots
 
-        # Group slot notes by name and concatenate their nodelists
-        self.component.slots = defaultdict(NodeList)
-        for slot in slots or []:
-            self.component.slots[slot.name].extend(slot.nodelist)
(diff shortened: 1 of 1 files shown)


housUnus:
Hi, i don't actually know what to add exactly but:

 In my case I had to call a page that needs a template card, the template depends on the caller wich sends the card type ('card' or 'blank')
#app/ceate.html page:
{% include 'includes/CRUD/create.html' with card_type="card" use_title=True %}

#includes/crud_create.html
```
{% component_block card_type %}
....
{% endcomponent_block %}
```

#component.py
```
@component.register("card")
class Card(component.Component):
    template_name = "includes/cards/default.html"

@component.register("card_blank")
class Blank_Card(component.Component):
    template_name = "includes/cards/blank.html"
```

and the inclusion of the card is successfully done with the name of the card I'm specifying in the variable card_type.

EmilStenstrom:
Hi again! I added the tests for this, and cleaned up the code a bit while I was at it. Thank you for this suggestion!

I just released a new version, with your changes, of django_components to PyPI: https://github.com/EmilStenstrom/django-components/

housUnus:
Thank you, Glad I've contributed something :).

PR #164, opened 2022-07-23
🐛Prevent potential null elements

### LINKED ISSUE #163: because {% component_js_dependencies %} loads all the components, so best to check to prevent null elements
since all the components JS are loaded via ` {% component_js_dependencies %} `

we need to make sure that the element selected in the js scripts actually exist in the page, esp if conditional loading is not used.

so instead of 

```javascript
/* In a file called [your app]/components/calendar/script.js */
(function(){
    document.querySelector(".calendar-component").onclick = function(){ alert("Clicked calendar!"); };
})()
```

it should be

```javascript
/* In a file called [your app]/components/calendar/script.js */
(function(){
    if (document.querySelector(".calendar-component")) {
        document.querySelector(".calendar-component").onclick = function(){ alert("Clicked calendar!"); };
    }
})()
```

simkimsia:
because {% component_js_dependencies %} loads all the components regardless

fix #163

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,9 @@ Then you need a javascript file that specifies how you interact with this compon
 ```js
 /* In a file called [project root]/components/calendar/script.js */
 (function(){
-    document.querySelector(".calendar-component").onclick = function(){ alert("Clicked calendar!") }
+    if (document.querySelector(".calendar-component")) {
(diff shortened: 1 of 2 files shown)


EmilStenstrom:
Thanks!

PR #177, opened 2022-09-10
(Docs) Fix param name of autodiscovery -> autodiscover

mands:
The docs say to set `autodiscovery: False` in `settings.COMPONENTS` to disable autodiscovery, but in the `app_settings` code the field is called `autodiscover`.

Cheers!


EmilStenstrom:
Doh! Great find, thanks!

PR #168, opened 2022-08-01
Issue loading media when RENDER_DEPENDENCIES True

rafae56038:
image
I have an issue, using COMPONENTS = {"RENDER_DEPENDENCIES": True} in settings, specific error is response["Content-Type"], this issue is for media loading.

PR #207, opened 2023-01-05
Issue loading media when RENDER_DEPENDENCIES True

hjalves:
This PR should replace #168.

I kept the previous commit and just added a test case. I'm not sure if this is the best way to do this, though.

Any feedback appreciated. Thanks!


EmilStenstrom:
Thanks! This is now released as part of https://pypi.org/project/django-components/0.24/

Issue #239, opened 2023-03-20
Very problematic behavior after upgrading to 0.26.1


spapas:
I've got a simple component named card. Here's the template 

```html
<div class="card mb-3 {{ color }} {{ border_color }}">
  <h5 class="card-header {{ header_color }} {{ header_text_color }}">
    {{ title }}</h5>

  <div class="card-body {{ color }} {{ text_color }}">
    {% slot "body" %}{% endslot %}
  </div>
</div>
```

and the function:

```py
@component.register("card")
class Card(component.Component):
    template_name = "components/card.html"

    # This component takes one parameter, a date string to show in the template
    def get_context_data(  # noqa PLR0913
        self, title, color="", text_color="", header_color="", header_text_color="", border_color=""
    ):
        return {
            "title": title,
            "color": color,
            "text-color": text_color,
            "header_color": header_color,
            "header_text_color": header_text_color,
            "border_color": border_color,
        }

    class Media:
        css = ""
        js = ""
```

Now, this was working flawlessly when I was using {% slot %} to render my template. However, after I upgraded to the new version and used {% fill %} instead it works *only* when I *include* a partial template with component on my template! 

So, Let's suppose I have a template named foo.html and I want to include this component:

```
{% component_block "card" "Test"  %}
{% fill "body" %}
  <div>test</div>
{% endfill %}
{% endcomponent_block %}
```

If I add this *directly* on foo.html I'll get the following error:

```
Traceback (most recent call last):
  File "C:\progr\py3\ehcg\venv\lib\site-packages\django\contrib\staticfiles\handlers.py", line 80, in __call__
    return self.application(environ, start_response)
  File "C:\progr\py3\ehcg\venv\lib\site-packages\django\core\handlers\wsgi.py", line 131, in __call__
    response = self.get_response(request)
  File "C:\progr\py3\ehcg\venv\lib\site-packages\django\core\handlers\base.py", line 140, in get_response
    response = self._middleware_chain(request)
  File "C:\progr\py3\ehcg\venv\lib\site-packages\django\core\handlers\exception.py", line 58, in inner
    response = response_for_exception(request, exc)
  File "C:\progr\py3\ehcg\venv\lib\site-packages\django\core\handlers\exception.py", line 141, in response_for_exception
    response = handle_uncaught_exception(
  File "C:\progr\py3\ehcg\venv\lib\site-packages\django\core\handlers\exception.py", line 182, in handle_uncaught_exception
[...]

PR #247, opened 2023-03-23
Add regression test for #239 (broke extends tags) + minor tweaks

lemontheme:


lemontheme:
This was such a small change I first tried just pushing straight to master. Turns out that branch is protected or something. Kind of unsure why it didn't work this time, seeing as I was able to bump the version tag a couple of days back.

Anyway, no matter. PRs are probably safer than YOLO pushing to master. Something strange is going on with that 'contributors readme update' action though. Like, why is it running Python tests... for changes to a Markdown file?! =p

tests/templates/incrementer.html
EmilStenstrom, line 1: I think this might have been added by mistake?

EmilStenstrom:
Yeah, I put the branch protection to avoid code going directly to master without passing review. Maybe we should by default review each other's code? 

Yeah, I have no idea what's going on with the readme update thingie. Maybe we should remove it, and that part of the readme, altogether?

lemontheme:
> avoid code going directly to master without passing review

Yeah, that's probably a good idea..

> Maybe we should by default review each other's code?

Sure, sounds good. Always or mostly always? Take for example the patch release that arose from the issue in the title – would that have warranted a review? 

---

As for the readme, I'd daresay we can drop it. It's obviously a nice gesture toward contributors but I'm unsure about the added value. We can also keep it but replace the action with a [contrib.rocks](https://contrib.rocks/preview?repo=angular%2Fangular-ja)-generated image. 

What I would definitely do, though, is move the contributors panel to the bottom of the readme, in keeping with convention. When I was new to django-components, I remember clicking away from the readme a couple of times once I'd hit the contributors panel, thinking that was it. Only later did I notice there was plenty left in the scrollbar and I discovered the docs went on way longer. (I realize how dumb that sounds, but it's the truth.)

EmilStenstrom:
> Sure, sounds good. Always or mostly always? Take for example the patch release that arose from the issue in the title – would that have warranted a review?

I'm not a fan of strict rules, so mostly always is fine by me!

What is the incrementer.html file btw? I see the conversation is resolved?

There's a sidebar to the right where everyone can see all the contributors, so lets skip it and all the admin around it.

lemontheme:
> I see the conversation is resolved.

Sorry about that! Was trying to get rid of that 'started a review' thing. 

To answer your question, that template file was already part of the tests. It's nothing new. Just added a newline, is all.

> so lets skip it and all the admin around it.

Keeping things simple. I like it!

PR #257, opened 2023-04-06
Fix #250 (3.6 failing) by removing postponed eval of type hints

### LINKED ISSUE #250: Tests have failed on Python 3.6 for a while, but have been hidden by misconfiguration
I just found out that tox-gh-actions has silently not ran any tests for Python 3.6 for a while. It seems it doesn't find the right environment and then just quits and reports success. 

Running "tox" locally confirms that django-components is currently broken in Python 3.6. Annoying. 

I've played around with CI settings, and tests now fail when no environment is found, which I think is a more truthful state to be in, but of course annoying to have failing tests.

I've asked [a question in the tox-gh-actions repo](https://github.com/ymyzk/tox-gh-actions/issues/146) to see if there's something obvious I've done incorrectly.

lemontheme:


django_components/component.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import annotations
+# from __future__ import annotations
```
EmilStenstrom, line 1: Maybe remove this comment instead of commenting it out?

EmilStenstrom (approved):
Looks good!

lemontheme:
Hm... Looks like those run-tests are frozen again.

EmilStenstrom:
@lemontheme FYI: The stray running tests was because branch protections are by name, and I had changed the name of the tests when troubleshooting. Have now updated the tests in settings, and everything seems to work as expected with not "Waiting for status to be reported" tests! :)

EmilStenstrom:
Should we do a point release to get 3.6 working again?

lemontheme:
Yeah, let's do a point release. Do you want to care of that, seeing as you're here? =p

lemontheme:
Btw, we should maybe quickly squash those last three commits? Was going to merge-and-squash once the checks had passed

EmilStenstrom:
I'm on it! :)

lemontheme:
Probably too late to easily squash the commits from this PR branch. Slightly embarrassing to see stuff like this in the master commit history. 🙈

I believe there's an option to squash commits when merging a PR. Perhaps we should do that by default in the future?

Screenshot 2023-04-07 at 09 34 30

EmilStenstrom:
Yeah, I’ll make sure to squash next time!

Issue #260, opened 2023-04-11
Change default staticfiles configuration to avoid .py files being accessible publicly


EmilStenstrom:
Today we suggest the following configuration:

> Modify STATICFILES_DIRS (or add it if you don't have it) so django can find your static JS and CSS files:
> 
> ```
> STATICFILES_DIRS = [
>     ...,
>     BASE_DIR / "components",
> ]
> ```

This means we will get automatic bundling of static files via the collectstatic management command. But since all .py files are stored in the same components directory, we will bundle the .py files too. This could be a security issue, if sensitive information is included in the .py file.

The suggested fix is to ignore all the .py files, by subclassing and replacing the staticfiles app, so that it ignored .py files.

lemontheme:
On it!

lemontheme:
PR above has been merged and released in v0.27. Going forward, users can choose what level of risk is acceptable to them.

If you're looking for an easy fix to stop .py and .html files being collected by collectstatic, modify your INSTALLED_APPS as follows:

```
INSTALLED_APPS = [
    ...,
    #  "django.contrib.staticfiles",  # <-- remove
    "django_components",
    "django_components.safer_staticfiles",  # <-- add
    ...,
]
```

More details are provided in the new Security Notes section in the README.

VojtechPetru:
:1st_place_medal: Thanks for being so quick with the fix guys! :clap:

PR #261, opened 2023-04-11
Introduce safer_staticfiles app as user-friendly security measure for #260

lemontheme:
This PR introduces a second installable app: _safer_staticfiles_ – so-named because it prevents .html and .py files from being collected by the collectfiles management command. (See #260 for why this matters.)

Usage:

```python
INSTALLED_APPS = [
    ...,
    # Replaced by django_components.safer_staticfiles as of v0.27:
    #   "django.contrib.staticfiles",
    "django_components",
    "django_components.safer_staticfiles",
    ...,
]
```

Notes:

- 'safer_staticfiles' seems like a good name, because I'm sure we can do more, but this is already better than nothing.
- The new app is placed inside its own apps.py inside a separate package because this makes it installable as 'django_components.safer_staticfiles', which is nicer than 'django_components.apps.SaferStaticFilesConfig'.
- As an alternative, I also looked into defining a custom django.contrib.staticfiles.finders.Finder subclass. All in that seemed like more work. Plus, the Finder API is undocumented, which means we'd have to monitor more closely for silent breakages.
- The approach taken here is opt-in, but I've added a 'security' section to the docs detailing why it should be followed.


EmilStenstrom (approved):
Looks good to me!

lemontheme:
Bumped to v0.27 (seemed better not to have patch versions in the Release Notes) and pushed

PR #264, opened 2023-04-13
Fix/263 performance regression

### LINKED ISSUE #263: Large performance issues when using the {% component_block ... %} syntax
Hello :wave: 

I've noticed a large performance issues when using `component_block` syntax.

For example, I've got a page with component
```
{% component_block 'form' form=form %}
    {% fill 'buttons' %}
        <button>Create</button>
    {% endfill %}
{% endcomponent_block %}
```
-> the average page load speed is around 4.1s (!).

But when I change it to
```
{% component 'form' form=form %}
```
The load speed drops to normal (around 100-150ms).

I'm using v0.27 of `django-components`.

lemontheme:
Fixes #263. 

Cause: deep-copying of dict of lists of fillnodes created a deeeeep stack of further calls to copy.deepcopy(). 
Solution: Replaced deep-copying with a for-loop that copies only the lists (instead of everything inside them). 

Issue reporter @VojtechPetru verified the solution resolves the issue by testing on the same real-life code where the issue was first observed.

### DIFF
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -1,4 +1,3 @@
-import copy
 from typing import (
     TYPE_CHECKING,
     ClassVar,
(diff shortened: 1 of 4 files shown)


EmilStenstrom:
Wow, nice find and fix! ⭐

Issue #231, opened 2023-03-02
Use default slots to save space


timothyis:
Related to #189, I think it would be great to have a "default" slot. Then you can fill a component_block without using the slot syntax if a component only has one slot and save a lot of lines. E.g., 

```django
{% component_block "box" %}
	<h1>hello</h1>
{% endcomponent_block %}
```

with the component HTML:
```django
<div>
	{% slot default %}
	{% endslot %}
</div>
```

This would help with a lot of components that are nested and become messy. For example, we have a table component:
```django
{% component_block "table" columns="sm|default (md)|lg (todo)" only %}
	{% slot "rows" %}
		{% component_block "table-row" classes=classes %}
			{% slot "columns" %}
				{% component_block "table-column" %}
					{% slot value %}
						<span>Some value</span>
					{% endslot %}
				{% endcomponent_block %}
				{% component_block "table-column" %}
					{% slot value %}
						<span>Some value</span>
					{% endslot %}
				{% endcomponent_block %}
			{% endslot %}
		{% endcomponent_block %}
	{% endslot %}
{% endcomponent_block %}
```

Would become:
```django
{% component_block "table" columns="sm|default (md)|lg (todo)" only %}
	{% component_block "table-row" classes=classes %}
		{% component_block "table-column" %}
			<span>Some value</span>
		{% endcomponent_block %}
		{% component_block "table-column" %}
			<span>Some value</span>
		{% endcomponent_block %}
	{% endcomponent_block %}
{% endcomponent_block %}
```

You can imagine the amount of space this will save when it has more rows/columns too.

This model also follows closely with how frontend JS frameworks work (without slots), so it will be easier for those of us starting there to understand.

EmilStenstrom:
I think this makes a lot of sense, and would be open to a PR to add this. 

Your thoughts @lemontheme?

lemontheme:
@timothyis is right about other JS frameworks doing it this way. If we _really_ want to follow their example we should make it optional for slots to specify a name _if and only if_ there is only one slot the template. (I think that's preferable to having a semi-magical 'default' label.)

For example, the component template would look like this:

```
<div>
	{% slot %}  {# look Ma, no name  #}
	{% endslot %}
</div>
```

I can more or less see how we'd make this work on the implementation side. :)

EmilStenstrom:
I like the idea of using concepts from JS frameworks if that makes the code clearer. But in this case I'm not sure it does. 

Say you see a component call like this:
```django
[...]

PR #269, opened 2023-04-17
Implement #231 – default slot filling

lemontheme:
This PR adds default slot filling as described in #231.

Here's an example (copied from the updated docs):

### Example

The template:

```htmldjango 
<div class="calendar-component">
    <div class="header">
        {% slot "header" %}Calendar header{% endslot %}
    </div>
    <div class="body">
        {% slot "body" default %}Today's date is <span>{{ date }}</span>{% endslot %}
    </div>
</div>
```

Including the component (notice how the `fill` tag is omitted):

```htmldjango
{% component_block "calendar" date="2020-06-06" %}
    Can you believe it's already <span>{{ date }}</span>??
{% endcomponent_block %}
```

The rendered result (exactly the same as before):

```html
<div class="calendar-component">
    <div class="header">
        Calendar header
    </div>
    <div class="body">
        Can you believe it's already <span>2020-06-06</span>??
    </div>
</div>
```

---

### Main changes:

- New `default` option added to `slot` tags.
- Component_block parsing updated to deal with 'bare'/'implicit' (still looking for the right term) filling.

Additional changes:

- Tests
- Updated the README + minor cosmetic changes
- Added example to sampleproject

To discuss:

- With the new `default` keyword, we're overloading the 'default' concept slightly, since it's also used in the `<fill_alias>.super` usage (i.e. where we used to have `slot.super`). In one of my code comments, I argue it might be better to drop the second use of 'default' and use `<fill_alias>.original` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,11 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
 
 Read on to learn about the details!
 
-# Release notes
+## Release notes
 
-*Version 0.27* includes an additional installable app *django_components.safer_staticfiles*. It provides the same behavior as *django.contrib.staticfiles* but with extra security guarantees (more info below in Security Notes). 
+*Version 0.28* introduces 'implicit' slot filling and the `default` option for `slot` tags.  
+
+*Version 0.27* adds a second installable app: *django_components.safer_staticfiles*. It provides the same behavior as *django.contrib.staticfiles* but with extra security guarantees (more info below in Security Notes). 
 
 *Version 0.26* changes the syntax for `{% slot %}` tags. From now on, we separate defining a slot (`{% slot %}`) from filling a slot with content (`{% fill %}`). This means you will likely need to change a lot of slot tags to fill. We understand this is annoying, but it's the only way we can get support for nested slots that fill in other slots, which is a very nice feature to have access to. Hoping that this will feel worth it!
 
@@ -28,11 +30,11 @@ Read on to learn about the details!
 
 *Version 0.17* renames `Component.context` and `Component.template` to `get_context_data` and `get_template_name`. The old methods still work, but emit a deprecation warning. This change was done to sync naming with Django's class based views, and make using django-components more familiar to Django users. `Component.context` and `Component.template` will be removed when version 1.0 is released.
 
-# Security notes 🚨
+## Security notes 🚨
 
 *You are advised to read this section before using django-components in production.*
 
-## Static files
+### Static files
 
 Components can be organized however you prefer.
 That said, our prefered way is to keep the files of a component close together by bundling them in the same directory.
@@ -58,7 +60,7 @@ INSTALLED_APPS = [
 If you are on an older version of django-components, your alternatives are a) passing `--ignore <pattern>` options to the _collecstatic_ CLI command, or b) defining a subclass of StaticFilesConfig.
 Both routes are described in the official [docs of the _staticfiles_ app](https://docs.djangoproject.com/en/4.2/ref/contrib/staticfiles/#customizing-the-ignored-pattern-list).  
 
-# Installation
+## Installation
 
 Install the app into your environment:
 
@@ -106,7 +108,7 @@ STATICFILES_DIRS = [
 ]
 ```
 
-## Optional
+### Optional
 
 To avoid loading the app in each template using ``` {% load django_components %} ```, you can add the tag as a 'builtin' in settings.py
 
@@ -128,7 +130,7 @@ TEMPLATES = [
 
 Read on to find out how to build your first component!
 
-# Compatiblity
+## Compatiblity
 
 Django-components supports all <a href="https://docs.djangoproject.com/en/dev/faq/install/#what-python-version-can-i-use-with-django">officially supported versions</a> of Django and Python.
 
@@ -140,7 +142,7 @@ Django-components supports all <a href="https://docs.djangoproject.com/en/dev/fa
 | 3.9            | 3.2, 4.0                 |
 | 3.10           | 4.0                      |
 
-# Create your first component
+## Create your first component
 
 A component in django-components is the combination of four things: CSS, Javascript, a Django template, and some Python code to put them all together.
 
@@ -172,7 +174,7 @@ Now you need a Django template for your component. Feel free to define more vari
 ```htmldjango
 {# In a file called [project root]/components/calendar/calendar.html #}
 <div class="calendar-component">Today's date is <span>{{ date }}</span></div>
-```
+``  `
 
 Finally, we use django-components to tie this together. Start by creating a file called `calendar.py` in your component calendar directory. It will be auto-detected and loaded by the app.
 
@@ -201,7 +203,7 @@ class Calendar(component.Component):
 
 And voilá!! We've created our first component.
 
-# Use the component in a template
+## Use the component in a template
 
 First load the `component_tags` tag library, then use the `component_[js/css]_dependencies` and `component` tags to render the component to the page.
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -172,7 +174,7 @@ Now you need a Django template for your component. Feel free to define more vari
 ```htmldjango
 {# In a file called [project root]/components/calendar/calendar.html #}
 <div class="calendar-component">Today's date is <span>{{ date }}</span></div>
-```
+``  `
```
EmilStenstrom, line 177: This must be unintentional right?

django_components/component.py
```diff
@@ -142,34 +141,60 @@ def get_updated_fill_stacks(self, context):
...
+                    f"Slot names must be unique."
+                    f"To fix, check template '{component_template.name}' "
+                    f"of component'{self._component_name}'."
                 )
+            if slot.is_default:
+                if encountered_default_slot:
+                    raise TemplateSyntaxError(
+                        "Only one component slot may be marked as 'default'. "
+                        f"To fix, check template '{component_template.name}' "
+                        f"of component'{self._component_name}'."
```
EmilStenstrom, line 175: There should be a space between component and the single quote here, right?

django_components/component.py
```diff
@@ -142,34 +141,60 @@ def get_updated_fill_stacks(self, context):
...
+                    )
+                else:
+                    encountered_default_slot = True
             slots[slot_name] = slot
         # All fill nodes must correspond to a declared slot.
+        #  Exempt from constraint are implicit fills, which match a slot marked
+        #  as 'default'
```
EmilStenstrom, line 182: Too much whitespace before the comment.

django_components/templatetags/component_tags.py
```diff
@@ -160,99 +162,177 @@ def __init__(self, slot: "SlotNode", context: Context):
...
+    #  Otherwise, we would have both
+    #  - {% slot ... default %} ...
+    #  and
+    #  - {% fill ... as <alias> %} <alias>.default {%endfill %}.
+    #  I'm afraid this risks being a source of confusion, as users might wrongly assume
+    #  that there is a link between the two separate usages of 'default'.
+    @property
+    def original(self) -> str:
+        return self.default
+
```
EmilStenstrom, line 176: Not sure I think this is worth changing concepts for. The difference here is "default slot" vs "default content" which are different concepts. Your call, but is this worth having people change their code for? Should we change the docs?

django_components/templatetags/component_tags.py
```diff
@@ -160,99 +162,177 @@ def __init__(self, slot: "SlotNode", context: Context):
...
+                fill_node = implicit_fill_node_stack.pop()
+        if self.is_required:
+            if fill_node is None:
                 raise TemplateSyntaxError(
                     f"Slot '{self.name}' is marked as 'required' (i.e. non-optional), "
                     f"yet no fill is provided. Check template '{self.template_name}'"
                 )
-        else:
-            fill_node = fill_node_stack.pop()
+        # //
```
EmilStenstrom, line 229: Strange comment, remove?

django_components/templatetags/component_tags.py
```diff
@@ -160,99 +162,177 @@ def __init__(self, slot: "SlotNode", context: Context):
...
+        if fill_node:
             nodelist = fill_node.nodelist
-
             if fill_node.alias_var is not None:
                 aliased_slot_var = UserSlotVar(self, context)
                 resolved_alias_name = fill_node.alias_var.resolve(context)
                 extra_context[resolved_alias_name] = aliased_slot_var
+        else:
+            nodelist = self.nodelist  # slot nodelist
+        # //
```
EmilStenstrom, line 240: Strange comment, remove?

django_components/templatetags/component_tags.py
```diff
@@ -160,99 +162,177 @@ def __init__(self, slot: "SlotNode", context: Context):
...
     else:
         raise TemplateSyntaxError(
-            f"{bits[0]}' tag takes only one argument (the slot name)"
-        )
-
-    if not is_wrapped_in_quotes(slot_name):
-        raise TemplateSyntaxError(
-            f"'{bits[0]}' name must be a string 'literal'."
+            "'slot' tag does not match pattern {{% slot <name> ['default'] ['required'] %}}. "
+            "Order of options is free."
```
EmilStenstrom, line 283: Free is not the right term here I think, maybe remove this line?

tests/test_templatetags.py
```diff
@@ -779,27 +902,31 @@ def tearDownClass(cls) -> None:
...
-    def test_variable_outside_fill_tag_is_error(self):
-        with self.assertRaises(TemplateSyntaxError):
-            Template(
-                """
-                {% load component_tags %}
-                {% component_block "test" %}
-                    {{ anything }}
-                {% endcomponent_block %}
+    def test_variable_outside_fill_tag_compiles_w_out_error(self):
+        # As of v0.28 this isvalid, provided the component registered under "test"
```
EmilStenstrom, line 906: "isvalid" -> "is valid"

tests/test_templatetags.py
```diff
@@ -779,27 +902,31 @@ def tearDownClass(cls) -> None:
...
-    def test_text_outside_fill_tag_is_error(self):
-        with self.assertRaises(TemplateSyntaxError):
-            Template(
-                """
-                {% load component_tags %}
-                {% component_block "test" %}
-                    Text
-                {% endcomponent_block %}
+    def test_text_outside_fill_tag_is_not_error(self):
+        # As of v0.28 this isvalid, provided the component registered under "test"
```
EmilStenstrom, line 919: "isvalid" -> "is valid"

EmilStenstrom (approved):
Looks good to me! Found some small nitpicks, but nothing serious. Tests look reasonable.

If possible, it would be easier to review PR:s when unrelated changes are put in separate commits! Then I can walk through the commits one by one, and know the purpose of that commit from the commit message.

Great job, as always! :)

django_components/templatetags/component_tags.py
lemontheme: It's probably too much detail for an error msg anyway. What I mean is that 'default' can come before or after 'required'. In ebnf terms:

```ebnf
slot tag = name, [options];
options = ["default"], ["required"] | ["required"], ["default"];
```

I'll drop it and let the examples speak for themselves.

django_components/templatetags/component_tags.py
lemontheme: I think it's best not overload terms with separate meanings, even if the differences are clear to us. On the other hand, it's probably not worth breaking people's code over.

Which of these options makes most sense to you:

- The _default_ property is maintained but a new _original_ alias is introduced. (Hell, we could even go back to using _super_ =p )
- We use a different name for the slot option. I'm not sure what that would be though.
- We leave things as they are, but add a clarifiying paragraph in the docs. Maybe create a new FAQ section at the end explaining possible sources of confusion.

lemontheme:
Thanks for reviewing. Amazing eye for detail as always!

django_components/templatetags/component_tags.py
lemontheme: Will do. (It's my way of saying 'here ends the block of code up until which the previous comment holds'.)

django_components/templatetags/component_tags.py
EmilStenstrom: I'm really not worried about this confusion. Default is a very commonly used term and being constrained to only use it in one place in the API feels limiting. It's like saying "parameter" could only be used once. But all sorts of functions take parameters.

I would say "We leave things as they are, and add things to a FAQ _IF_ it's a common confusion from users" is the best option here. But sure, it doesn't hurt to add it the FAQ if you feel strongly about it.

lemontheme: Eh, you're probably right. Until others have had the chance to use this new feature, it's too early to say what will or won't be experienced as confusing. Makes sense to hold off on updating the docs until we've received some feedback. I've dropped these lines in a subsequent commit.

lemontheme:
Btw – don't merge this just yet. Still got a couple of tests I want to add.

EmilStenstrom:
Rebased just to get the tests to run and report status checks again! Sorry if I made a mess.

lemontheme:
Thanks! Was wondering why they weren't running. No mess as far as I can tell!

I think it's safe for us to merge. 

If all goes well, a rebased #273 should start reporting success as well.

EmilStenstrom:
@lemontheme Very happy to see this merged, well done! :D Will you make a pypi release too, or wait for the other big change too?

lemontheme:
Thanks! Took a lot of sketches and diagrams to get this right. =p

It's been a while since I've looked at the bigger picture. Am I correct in assuming the other big change you refer to is the transition to a single `component` tag (to replace `component_block`)?

If so, I believe we can do a release already, independently of that change. Essentially, what I plan to do for that other change is introduce a flag in settings that will change what function get registered to which tags. The `do_component_block()` function will be registered to `{% component %}`. `do_component()` will be ignored. `{% component_block %} will cease to exist.

Once #273 is merged succesfully, we can release.

PR #271, opened 2023-04-20
Use valid script tag HTML to avoid console errors

timothyis:
This pull request updates the script tag that's placed by django-components in order to render component dependencies. Currently, the script tag is not valid and ends up causing errors in the console and breaking the next script that's defined.

image

This is currently the largest event in our brand new Sentry front-end setup, so I feel as if it might grow in production 😅

### DIFF
--- a/django_components/middleware.py
+++ b/django_components/middleware.py
@@ -6,14 +6,14 @@ from django.http import StreamingHttpResponse
 
 RENDERED_COMPONENTS_CONTEXT_KEY = "_COMPONENT_DEPENDENCIES"
 CSS_DEPENDENCY_PLACEHOLDER = '<link name="CSS_PLACEHOLDER">'
-JS_DEPENDENCY_PLACEHOLDER = '<script name="JS_PLACEHOLDER">'
+JS_DEPENDENCY_PLACEHOLDER = '<script name="JS_PLACEHOLDER"></script>'
 
 SCRIPT_TAG_REGEX = re.compile("<script")
 COMPONENT_COMMENT_REGEX = re.compile(rb"<!-- _RENDERED (?P<name>\w+?) -->")
 PLACEHOLDER_REGEX = re.compile(
     rb"<!-- _RENDERED (?P<name>\w+?) -->"
     rb'|<link name="CSS_PLACEHOLDER">'
-    rb'|<script name="JS_PLACEHOLDER">'
+    rb'|<script name="JS_PLACEHOLDER"></script>'
 )
 
 


lemontheme:
Nice! Script tags should never be void. I think it makes sense to accept these changes. 

Note that the fact you're seeing these elements means your COMPONENTS config has RENDER_DEPENDENCIES=True but hasn't included django_components.middleware.ComponentDependencyMiddleware on the MIDDLEWARE list.

@EmilStenstrom, what would you say to the idea of adding a line of JS to the placeholder element so that developers get console.warn()'ed if the middleware isn't set up right?

I'm going to defer to Emil to merge your changes ;)

timothyis:
I'm not sure if that is 100% correct, since we definitely have that defined in our middleware 🤔

lemontheme:
Huh... Well, in that case it seems you've uncovered a bug in the middleware. It _should_ replace or remove all placeholder tags.

Can I interest you in poking around the middleware module a bit? Maybe a few print statements/breakpoints while running your specific project?

timothyis:
I'd be happy to, if you don't mind telling me what you need, where and how to do it 😅 I really am not sure what I'm doing in a Python environment. 

Just to be clear, I think this is an issue when you're loading a page that has components which don't have JS scripts defined, this script loads in anticipation but is not replaced.

If there's a component on the page with a script, this will be replaced.

EmilStenstrom:
I think this change makes sense, so we could merge it right away. But as @lemontheme says: That placeholder should always be replaced, so there's a bug lurking somewhere here...

lemontheme:
> [...] this is an issue when you're loading a page that has components which don't have JS scripts defined [...]

Quick experiment. Commented out the js and css media paths on the components in sampleproject one by one and then both at the same time, while simultaneously trying different combinations of RENDER_DEPENDENCIES and middleware on/off. In this relatively simple experiment setup, there doesn't appear to be a link between missing JS scripts and the placeholder appearing. In all cases where the middleware is included, the placeholders are removed correctly.

There's something more subtle going on here... 🤔

Also experimented with adding duplicate component dependency tags to the template. The middleware seems to deal with those correctly too, so we can eliminate that as a possible cause.

---

> what you need, where and how to do it

- In the django_components/middleware.py module of your cloned fork, see if you can't place a few prints in the __call__ method of the DependencyReplacer class. When the middleware fires, this should give an indication of how many times it's matching a placeholder tag.
- To install your fork into your real-life project environment, `pip install -e /path/to/your/fork`. This will override the official version installed from pypi with a local version. The -e flag means 'editable': basically, it lets you keep making changes to the package source code.
- In case you want to poke around even easier, python has a killer built-in CLI debugger 'pdb' which you can drop into by placing breakpoint() statements in the code.

timothyis:
Sorry, I haven't had time just yet. Could we merge this PR and debug this in an issue?

lemontheme:
Sure, don't see why not. It's a minor change, after all

timothyis:
Thank you! Just allowing me to avoid Sentry emails 😅

PR #275, opened 2023-04-21
ignore .pyc files in collectstatic management command

VojtechPetru:
Ignore also _.pyc_ files when running `collectstatic` (compiled Python files inside `__pycache__` folder).

Related to https://github.com/EmilStenstrom/django-components/pull/261.


EmilStenstrom:
Good catch!

PR #276, opened 2023-04-24
Bump version

lemontheme:


lemontheme:
I know, feels pretty ridiculous, especially since it triggers two extra sets of test runs.

Reason I did it is because I keep running into strange behavior when pushing the tagged 'bump version' commit directly to master.

`git push <commit>` gets refused. `git push <tag>` works, triggers the pypi action, _but_ the commit itself doesn't seem to appear on the master branch. It's really weird. In the past, `git push <commit>` a second time _after_ the pypi action has completed fixes it. But it feels... wrong.

Any idea what's going on there?

lemontheme:
No idea. This is a part of the Github feature set that I'm not all too familiar with.

I just think it's funny. Try to push up-to-date, carefully rebased code to master, GH responds with "Nah, fam, can't have that." But do it via a tag commit and I'm pretty sure I could push whatever I want to pypi. So that's not great... 😛

PR #283, opened 2023-05-16
Invalid slot more helpful message

VojtechPetru:
Just a bit more helpful error message when trying to fill invalid slot name - try to guess typos and write a list of unfilled slots.

### DIFF
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -1,5 +1,16 @@
+import difflib
 from collections import ChainMap
-from typing import Any, ClassVar, Dict, Iterable, Optional, Tuple, Union
+from typing import (
+    Any,
+    ClassVar,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Union,
+)
 
 from django.core.exceptions import ImproperlyConfigured
 from django.forms.widgets import Media, MediaDefiningClass
@@ -196,10 +207,18 @@ class Component(metaclass=SimplifiedInterfaceMediaDefiningClass):
             )
         for fill_name in filter(None, fill_target2content.keys()):
             if fill_name not in slot_name2fill_content:
-                raise TemplateSyntaxError(
-                    f"Component '{self.registered_name}' passed fill "
-                    f"that refers to undefined slot: {fill_name}"
+                unfilled_slots = self._get_unfilled_slots(template)
+                slot_guess = self._get_slot_name_guess(
+                    fill_name, unfilled_slots
                 )
+                msg = (
+                    f"Component '{self.registered_name}' passed fill "
+                    f"that refers to undefined slot: '{fill_name}'."
+                    f"\nUnfilled slot names are: {sorted(unfilled_slots)}."
+                )
+                if slot_guess:
+                    msg += f"\nDid you mean '{slot_guess}'?"
+                raise TemplateSyntaxError(msg)
         # Return updated FILLED_SLOTS_CONTEXT map
         filled_slots_map: Dict[Tuple[SlotName, Template], FillContent] = {
             (slot_name, template): content_data
@@ -213,3 +232,25 @@ class Component(metaclass=SimplifiedInterfaceMediaDefiningClass):
             return prev_context.new_child(filled_slots_map)
         except KeyError:
             return ChainMap(filled_slots_map)
+
+    def _get_unfilled_slots(self, template: Template) -> set[SlotName]:
+        all_slots: List[SlotNode] = template.nodelist.get_nodes_by_type(
+            SlotNode
+        )
+        filled_slots: Set[SlotName] = {
+            name for name, node, _ in self.fill_content
+        }
+        return set(s.name for s in all_slots) - filled_slots
+
+    def _get_slot_name_guess(
+        self, name: SlotName, slot_names: Set[SlotName]
+    ) -> Optional[SlotName]:
+        """
+        :param name: invalid slot fill name
+        :param slot_names: names of slots in which to try to find a similar match
+        """
+        guesses: list[SlotName] = difflib.get_close_matches(
+            name, slot_names, 1
+        )
+        if guesses:
+            return guesses[0]
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -233,7 +233,7 @@ class Component(metaclass=SimplifiedInterfaceMediaDefiningClass):
         except KeyError:
             return ChainMap(filled_slots_map)
 
-    def _get_unfilled_slots(self, template: Template) -> set[SlotName]:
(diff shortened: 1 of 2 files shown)


django_components/component.py
```diff
@@ -160,14 +161,33 @@ def validate_fills_and_slots_(
...
-        if unmatchable_fills:
-            msg = (
-                f"Component '{self._component_name}' passed fill(s) "
-                f"refering to undefined slot(s). Bad fills: {list(unmatchable_fills)}."
-            )
-            raise TemplateSyntaxError(msg)
+        if not unmatchable_fills:
+            return
+        # There are some fills that don't correspond to any slot. Raise an exception with helpful error message.
+        unfilled_slots: set[str] = {s.name for s in all_slots} - fills.keys()
```
VojtechPetru, line 170: Quickly double-checking this PR now, I'm not sure this syntax is correct for older Python versions. Shouldn't it be `typing.Set[str]`? I'm surprised that the tests seem to be passing. Am I wrong?

EmilStenstrom: We're currently not running any typing checks at all, as such a small part of the library is type checked, so I'm guessing that part is just ignored. If the tests pass in 3.6, I'm happy.

tests/test_templatetags.py
```diff
@@ -349,6 +349,39 @@ class Component(component.Component):
...
+            except TemplateSyntaxError as e:
+                self.assertEqual(
+                    (
+                        "Component 'test1' passed fill(s) referring to undefined slot(s).\n"
+                        "Bad fills: ['haeder'].\n"
+                        "\n"
+                        "Did you mean: \n"
+                        "- haeder: header?\n"
+                        "Other unfilled slots are: ['footer']."
+                    ),
```
EmilStenstrom, line 380: No big deal, but I'm thinking that this might look nicer with a triple quoted string instead? To handle the indentation I tend to use textwrap.dedent.

```suggestion
                    textwrap.dedent("""
                        Component 'test1' passed fill(s) referring to undefined slot(s).
                        Bad fills: ['haeder'].
                        
                        Did you mean:
                        - haeder: header?
                        Other unfilled slots are: ['footer'].
                    """),
```

EmilStenstrom:
Looks good to me! 

@lemontheme: I'm a little worried that this might lead to merge conflicts for your many big changes. Ok to merge?

tests/test_templatetags.py
VojtechPetru: I didn't know this, nice!

EmilStenstrom: Note: You need an import too :)

VojtechPetru: Right... 😆 Thanks!

lemontheme:
I like this. It's a subtle DX improvement that doubtless stems from real-life frustration.

That said, @EmilStenstrom, I'm afraid this will indeed conflict with my changes. 🚫 Slot and fill validation has moved to a [different method and the overall strategy is quite different](https://github.com/EmilStenstrom/django-components/pull/269/files#diff-569e56c0c056d2ede4e534ad3cae3717ff8e61ba7fdd6963a23df1adc8fd86d1R175).

The good news is that the main idea of this PR is still compatible with the new slot-filling implementation – barring the fact that I no longer create a set of fills, but rather raise an exception as soon as a fill can't be matched. I think though that we can still keep this 'fail fast' approach _and_ support the behavior that this PR introduces.

@VojtechPetru, looking at the changes in the linked commit, are you able to more or less infer what's changed, or would you like me to a) give you a detailed write-up so you can make the changes yourself or b) update this PR with the necessary changes myself?

lemontheme:
[There is one more](https://github.com/EmilStenstrom/django-components/pull/269/files#diff-569e56c0c056d2ede4e534ad3cae3717ff8e61ba7fdd6963a23df1adc8fd86d1R175): a special case where a slot marked as 'required' can't be matched to either a named fill or an implicit/bare fill (the second is the main feature added by PR).

Once you hit that line, the total set of available fills is already known. Look above at `fill_target2content`. Note that a component can be provided either one or more named fills (via tags) or a *single* implicit/bare fill that gets matched to a slot marked as 'default'. Hence, the `fill_target2content` either contains a _single_ None key (implicit fill) or one or more string keys (named fills).

VojtechPetru:
> [There is one more](https://github.com/EmilStenstrom/django-components/pull/269/files#diff-569e56c0c056d2ede4e534ad3cae3717ff8e61ba7fdd6963a23df1adc8fd86d1R175): a special case where a slot marked as 'required' can't be matched to either a named fill or an implicit/bare fill (the second is the main feature added by PR).
> 
> Once you hit that line, the total set of available fills is already known. Look above at `fill_target2content`. Note that a component can be provided either one or more named fills (via tags) or a _single_ implicit/bare fill that gets matched to a slot marked as 'default'. Hence, the `fill_target2content` either contains a _single_ None key (implicit fill) or one or more string keys (named fills).

Thanks for the explanation!
I think that error message is good as is, because it already tells the developer the slot name he needs to fill. Did you have some specific improvement in mind?

lemontheme:
Great job adapting your previous implementation to the new slot-filling strategy.

To save you extra work, I just pushed a commit to your branch that takes care of a few final details, including a few improvements to the slot-filling logic revealed by the needs of your new functionality.

What follows is an overview of the changes:

Most important, I inlined the behavior of the two new methods `_get_unfilled_slots()` and `_get_slot_name_guess()` into the body of `_process_template_and_update_filled_slot_context()` (which I'll refer to as the 'main slot-filling method'). 

Reasoning:

- The return of `_get_unfilled_slots()` can be obtained from the outer loop in the main slot-filling method, without duplicating the effort of traversing the nodelist depth-first for  SlotNodes. 
- `_get_slot_name_guess()` is only ever called from one place. Seeing as it mostly wraps the difflib function, we don't win much from adding indirection.
- Additionally, because the Component class is meant to be subclassed by user code, I think it's better if we keep the class interface (the total set of methods) as focused as possible. In fact, I'm considering moving `_process_template_and_update_filled_slot_context()` _out_ of the class and just treating it as a static function.

In addition to inlining those two methods, I now postpone the check that raises an error on unfilled required slots until _after_ all fills have been tried to be matched. This allows us to generate a slightly more helpful error message in those cases.

I've also set a cutoff value for the fuzzy (edit distance) matching of unfilled fills to slots, so that false positives are less likely to occur.

Finally, seeing as `_process_template_and_update_filled_slot_context()` bundles a lot of logic into a single place, I've added some extra comments to keep everything maintainable. In the future, I'll refactor that method so that it becomes slightly less monstrous. For now, it pays to have everything in one place, since I'm still discovering optimization opportunities.

lemontheme:
Merged and released 🚀

Issue #290, opened 2023-06-19
Components directory can't be found when multiple settings files are used


bastinc:
I'm using different settings files (base.py, dev.py, prod.py) which are stored in a settings directory. The composent directory is at the root of the project.

In this case, django-components fails to register components in the components directory.

```
 raise NotRegistered('The component "%s" is not registered' % name)
django_components.component_registry.NotRegistered: The component "calendar" is not registered
```

EmilStenstrom:
This is now available in 0.35 of django-components. Thanks @dylanjcastillo!

PR #291, opened 2023-06-19
 fix: Components directory can't be found when multiple settings file are used

bastinc:
Here is a fix for  #290

### DIFF
--- a/django_components/template_loader.py
+++ b/django_components/template_loader.py
@@ -14,15 +14,9 @@ class Loader(FilesystemLoader):
         component_dir = "components"
         directories = list(get_app_template_dirs(component_dir))
 
-        if settings.SETTINGS_MODULE:
-            settings_path = Path(*settings.SETTINGS_MODULE.split("."))
-
-            path = (settings_path / ".." / component_dir).resolve()
-            if path.is_dir():
-                directories.append(path)
-
-            path = (settings_path / ".." / ".." / component_dir).resolve()
-            if path.is_dir():
+        if hasattr(settings, "BASE_DIR"):
+            path = Path(settings.BASE_DIR, component_dir)
+            if path.exists():
                 directories.append(path)
 
         return directories


django_components/template_loader.py
```diff
@@ -14,15 +14,9 @@ def get_dirs(self):
...
-        if settings.SETTINGS_MODULE:
-            settings_path = Path(*settings.SETTINGS_MODULE.split("."))
-
-            path = (settings_path / ".." / component_dir).resolve()
-            if path.is_dir():
-                directories.append(path)
-
-            path = (settings_path / ".." / ".." / component_dir).resolve()
-            if path.is_dir():
+        if hasattr(settings, "BASE_DIR"):
```
EmilStenstrom, line 17: BASE_DIR is not a standardized variable name that contains the base directory, so I don't think we can use it.

django_components/template_loader.py
```diff
@@ -14,15 +14,9 @@ def get_dirs(self):
...
+            path = Path(settings.BASE_DIR, component_dir)
+            if path.exists():
```
EmilStenstrom, line 19: You are removing .is_dir() and replacing it with .exists(), is there a reason?

django_components/template_loader.py
bastinc: Indeed

django_components/template_loader.py
```diff
@@ -14,15 +14,9 @@ def get_dirs(self):
...
-            settings_path = Path(*settings.SETTINGS_MODULE.split("."))
-
-            path = (settings_path / ".." / component_dir).resolve()
-            if path.is_dir():
-                directories.append(path)
-
-            path = (settings_path / ".." / ".." / component_dir).resolve()
-            if path.is_dir():
+        if hasattr(settings, "BASE_DIR"):
+            path = Path(settings.BASE_DIR, component_dir)
```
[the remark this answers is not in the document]
bastinc: The problem trying to find the components folder from the  settings.SETTINGS_MODULE path is that someone could have a structure more complex (settings folder -> dev folder -> settings.py) it  wouln't work either.

EmilStenstrom:
Does anyone have an idea of how to move forward on this issue? Supporting setups where the settings files are split into many files seems worthwhile, but requiring the use of the non-standard BASE_DIR is not the right solution imho.

matevz-ap:
We also have 2 settings files in our project and had trouble implementing the library, but we figured out that its due to our projects file structure. Our structure looks like this `project > project > settings > settings.py`, so none of the `path = (settings_path / ".." / ".." / component_dir).resolve()` paths can resolve it. Our idea is to maybe add  a new optional variable such as COMPONENT_DIR, that could be specified if the user has a different file structure.

EmilStenstrom:
This is now solved with this setting: https://github.com/EmilStenstrom/django-components#configure-the-module-where-components-are-loaded-from

Let me know if this doesn't solve this problem, and we'll reopen.

Issue #249, opened 2023-03-24
Create managment command to setup basic component structure


EmilStenstrom:
It would be nice with a management command that generates a component with all the supporting files, and a default directory structure. There was an attempt to do this at https://github.com/EmilStenstrom/django-components/pull/175 but unfortunately that PR is stale now. I think it might be easiest to start over.

hanifbirgani:
The boilerplate content of `componenet-name.py` is straightforward. What should be the default content of other generated files? (js/css/html)

EmilStenstrom:
Happy to hear that you are working on this! :) 

There's a discussion about a new default structure that ended up in a suggestion to change the default names of files to template.html, style.css, and index.js. 

I think it makes sense to keep the contents of the files small, but still demonstrate how you connect them. Suggestion:

template.html
```django
<div class="component-[[ insert component name ]]">
    Hello component!
</div>
```

style.css
```css
.component-[[ insert component name ]] {
    background: pink;
}
```

index.js
```js
window.addEventListener("load", (event) => {
  console.log("[[ insert component name ]] is fully loaded");
});
```

What do you think?

hanifbirgani:
@EmilStenstrom I've just added #299 PR to address this issue.

nerdoc:
I would maybe use cookiecutter for this, and define a customizable URL where to find that template. If no CC template is given, use a default

EmilStenstrom:
Interesting idea! What advantages would that give compared to this version?

nerdoc:
I wrote a similar Commando for GDAPS, a Django Plugin system, to create plugins Form templates. First I copied Django's startapp cmd, and found it very in flexible. Cookiecutter already has solved many problems, e.g. flexibly naming directories.

Templates with cc could live outside the main repo, and could get updates without adding a new version (which can be an advantage or disadvantage).

And most importantly, user could add own cc templates which could be used easily using e.g.

    ./manage.py start component my_component --URL github.com/foouser/footemplate

Sure, this could be solved locally, but online repos help to keep diversity, and sharing templates.

PR #299, opened 2023-07-11
[WIP] add `startcomponent` management command

hanifbirgani:
Initial pr to add `startcomponent` management command to the package. related #249

### DIFF
--- a/django_components/management/commands/startcomponent.py
+++ b/django_components/management/commands/startcomponent.py
@@ -10,9 +10,7 @@ class Command(BaseCommand):
 
     def add_arguments(self, parser):
         parser.add_argument(
-            "name",
-            type=str,
-            help="The name of the component to create"
+            "name", type=str, help="The name of the component to create"
         )
         parser.add_argument(
             "--path",
@@ -68,7 +66,9 @@ class Command(BaseCommand):
             verbose = kwargs["verbose"]
             dry_run = kwargs["dry_run"]
 
-            component_path = os.path.join(path or f"{base_dir}/components", name)
+            component_path = os.path.join(
+                path or f"{base_dir}/components", name
+            )
 
             if os.path.exists(component_path):
                 if force:
@@ -102,7 +102,9 @@ class Command(BaseCommand):
                     )
                     f.write(script_content.strip())
 
-                with open(os.path.join(component_path, css_filename), "w") as f:
+                with open(
+                    os.path.join(component_path, css_filename), "w"
+                ) as f:
                     style_content = dedent(
                         f"""
                         .component-{name} {{
@@ -112,7 +114,9 @@ class Command(BaseCommand):
                     )
                     f.write(style_content.strip())
 
-                with open(os.path.join(component_path, template_filename), "w") as f:
+                with open(
+                    os.path.join(component_path, template_filename), "w"
+                ) as f:
                     template_content = dedent(
                         f"""
                         <div class="component-{name}">
@@ -122,7 +126,9 @@ class Command(BaseCommand):
                     )
                     f.write(template_content.strip())
 
-                with open(os.path.join(component_path, f"{name}.py"), "w") as f:
+                with open(
+                    os.path.join(component_path, f"{name}.py"), "w"
+                ) as f:
                     py_content = dedent(
                         f"""
                         from django_components import component
--- a/django_components/management/commands/startcomponent.py
+++ b/django_components/management/commands/startcomponent.py
@@ -61,14 +61,19 @@ class Command(BaseCommand):
             js_filename = kwargs["js"]
             css_filename = kwargs["css"]
             template_filename = kwargs["template"]
-            base_dir = settings.BASE_DIR
+            base_dir = getattr(settings, "BASE_DIR", None)
             force = kwargs["force"]
             verbose = kwargs["verbose"]
             dry_run = kwargs["dry_run"]
 
-            component_path = os.path.join(
-                path or f"{base_dir}/components", name
-            )
+            if path:
+                component_path = os.path.join(path, name)
+            elif base_dir:
+                component_path = os.path.join(base_dir, "components", name)
+            else:
+                raise CommandError(
+                    "You must specify a path or set BASE_DIR in your django settings"
+                )
 
             if os.path.exists(component_path):
                 if force:
--- a/django_components/management/commands/startcomponent.py
+++ b/django_components/management/commands/startcomponent.py
@@ -1,0 +1,70 @@
+from django.core.management.base import BaseCommand, CommandError
+from django.conf import settings
+import os
+
+
+class Command(BaseCommand):
+    help = 'Creates a new component'
+
+    def add_arguments(self, parser):
+        parser.add_argument('name', type=str, help='The name of the component to create')
+
+    def handle(self, *args, **kwargs):
+        name = kwargs['name']
+        base_dir = settings.BASE_DIR
+
+        components_path = os.path.join(base_dir, f'components/{name}')
+
+        if os.path.exists(components_path):  # If component directory already exists
+            raise CommandError(f'The component "{name}" already exists at {components_path}.')
+
+        os.makedirs(components_path)
+
+        with open(components_path + '/index.js', 'w') as f:
+            script_content = f"""
+window.addEventListener('load', (event) => {{
+    console.log("{name} component is fully loaded");
+}});
+            """
+            f.write(script_content.strip())
+
+        with open(components_path + '/style.css', 'w') as f:
+            style_content = f"""
+.component-{name} {{
+    background: red;
+}}
+            """
+            f.write(style_content.strip())
+
+        with open(components_path + '/template.html', 'w') as f:
+            template_content = f"""
+<div class="component-{name}">
+    Hello from {name} component!
+</div>
+            """
+            f.write(template_content.strip())
+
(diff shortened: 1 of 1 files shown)


django_components/management/commands/startcomponent.py
```diff
@@ -0,0 +1,80 @@
...
+            )
+
+        os.makedirs(components_path)
+
+        with open(components_path + "/index.js", "w") as f:
+            script_content = f"""
+window.addEventListener('load', (event) => {{
+    console.log("{name} component is fully loaded");
+}});
+            """
```
EmilStenstrom, line 35: Maybe we can use textwrap.dedent here to make the indentation look better?
```suggestion
            script_content = textwrap.dedent(f"""
                window.addEventListener('load', (event) => {{
                    console.log("{name} component is fully loaded");
                }});
            """)
```

EmilStenstrom (changes requested):
Great work! This is very close to done I think. The only major issue is being dependent on the non-standard `settings.BASE_DIR`, and that all component dirs are in BASE_DIR/components.

EmilStenstrom:
Do we need tests for this? To make sure we don't inadvertently break it in the future?

django_components/management/commands/startcomponent.py
```diff
@@ -0,0 +1,80 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "name", type=str, help="The name of the component to create"
+        )
+
+    def handle(self, *args, **kwargs):
+        name = kwargs["name"]
+        base_dir = settings.BASE_DIR
+
+        components_path = os.path.join(base_dir, f"components/{name}")
```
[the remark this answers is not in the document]
hanifbirgani: We can set a global constant `COMPONENTS_DIR` in the main django settings.

django_components/management/commands/startcomponent.py
```diff
@@ -0,0 +1,80 @@
...
+        if os.path.exists(
+            components_path
+        ):  # If component directory already exists
+            raise CommandError(
+                f'The component "{name}" already exists at {components_path}.'
+            )
+
+        os.makedirs(components_path)
+
+        with open(components_path + "/index.js", "w") as f:
```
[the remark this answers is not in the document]
hanifbirgani: I guess it will make the command confusing for users. We can change it to `script.js` or another fixed parameter.

django_components/management/commands/startcomponent.py
hanifbirgani: Nice 👍

django_components/management/commands/startcomponent.py
```diff
@@ -0,0 +1,80 @@
...
+@component.register("{name}")
+class {name.capitalize()}(component.Component):
+    # you can override def get_template_name() instead of specifying the below variable.
+    template_name = "{name}/template.html"
+
+    # This component takes one parameter, a date string to show in the template
+    def get_context_data(self, date):
+        return {{
+            "date": date,
+        }}
```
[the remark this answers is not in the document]
hanifbirgani: They don't, I used the example in the readme. What's your suggestion?

django_components/management/commands/startcomponent.py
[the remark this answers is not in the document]
EmilStenstrom: If you look at the autodiscover code it allows multiple component directories, eg. one per app. Or you could collect all your components in one global directory. What do we do about the first case?

django_components/management/commands/startcomponent.py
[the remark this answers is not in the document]
EmilStenstrom: I was thinking we keep all the names you have specified now, and make sure they default to those values if you don't specify any value. But IF you want script.js instead, you can specify that as a paramerer?

django_components/management/commands/startcomponent.py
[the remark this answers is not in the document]
EmilStenstrom: Yeah, I thought about that. The readme walks you through how to built a specific date-component, but I think this should be more "a starting component you can build on". I think just changing "date" to "param" makes more sense?

EmilStenstrom:
> > 
> 
> It depends on the system file and the components' directory path. What's your idea to test it?

Just a simple smoke test. Set the Components dir, Call the command, and see that files are in that specified dir? I just want to make sure we don't mess stuff up inadvertedly.

I tend to use this context processor for file system dependent tests:

```python
@contextmanager
def temporary_directory():
    temp_dir = tempfile.mkdtemp()
    try:
        yield temp_dir
    finally:
        shutil.rmtree(temp_dir)
```

And the override_settings decorator to specify settings for a test:
```python
@override_settings(SPAMCHECK_ENABLED=True)
def test_only_link_in_text(self):
    ...
```

EmilStenstrom:
Here's an idea: the [unicorn project's management command](https://www.django-unicorn.com/docs/components/#create-a-component) requires that you specify were you want to put your component giving the app name. Should we do the same?

hanifbirgani:
> 

Ok, I will implement the idea in the upcoming commits.

EmilStenstrom:
> > 
> 
> Ok, I will implement the idea in the upcoming commits.

Only if you agree that this is a good idea. I want your opinion is a skilled developer too! :)

hanifbirgani:
> > > 
> > 
> > 
> > Ok, I will implement the idea in the upcoming commits.
> 
> Only if you agree that this is a good idea. I want your opinion is a skilled developer too! :)


I believe we should have a global constant `COMPONENTS_DIR` in our code that can be rewritten in the `settings.py`. We can then ask for a custom path in the `startcomponent` command, and if the user has provided a custom path, we can use it. Otherwise, we can use the default path from the `COMPONENTS_DIR` constant.

EmilStenstrom:
@hanifbirgani That's a good start, let's do it!

hanifbirgani:
@EmilStenstrom I've added the custom path option to the command without touching the core codes to prevent side effects.

django_components/management/commands/startcomponent.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def handle(self, *args, **kwargs):
+        name = kwargs["name"]
+
+        if name:
+            path = kwargs["path"]
+            js_filename = kwargs["js"]
+            css_filename = kwargs["css"]
+            template_filename = kwargs["template"]
+            base_dir = settings.BASE_DIR
```
EmilStenstrom, line 64: Isn't BASE_DIR a non-standard variable? Are we sure this one always exist, and what do we do if it doesn't exist? Is this a new requirement of django-components?

hanifbirgani: It exists on the Django default setup.

How should we calculate the base directory of the project? What's your suggestion?

EmilStenstrom: @hanifbirgani Sorry for the late reply. Let's check for BASE_DIR, then fall back to PROJECT_PATH (seems semi-standard too), and otherwise error out?

hanifbirgani: Do you mean `PROJECT_PATH` in the environment variables? like `os.environ['PROJECT_PATH']`?

EmilStenstrom: No, I've seen some project have a settings.PROJECT_PATH instead of BASE_DIR, and just thought this might be a good fallback if it exists?

hanifbirgani: IMO it's a rabbit hole. I prefer to check the standard `settings.BASE_DIR` and if it doesn't exist, ask the user to specify the full path to create the component. We already have the command line argument (`--path`) for this.

EmilStenstrom: Sure. But maybe then some error handling when BASE_DIR doesn't exist?

hanifbirgani: Sure, I'll add error handling in the next commit.

hanifbirgani: @EmilStenstrom I fixed this issue in one of the last commits. Please check the results.

django_components/management/commands/startcomponent.py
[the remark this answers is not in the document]
hanifbirgani: @EmilStenstrom I fixed this issue in one of the last commits. Please check the results.

EmilStenstrom (approved):

EmilStenstrom:
Thank you @hanifbirgani! Great job! :) 

This is now published as version 0.29 - https://pypi.org/project/django-components/0.29/

I think a good next step is to update the README to refer to the new management command: https://github.com/EmilStenstrom/django-components/issues/323

hipertracker:
It does not work
```
❯ python manage.py startcomponent  my_component
Unknown command: 'startcomponent'. Did you mean startproject?
```

EmilStenstrom:
@hipertracker What version of django_components do you use?

hipertracker:
I use django-components v0.29

poetry.lock:

```
[[package]]
name = "django-components"
version = "0.29"
description = "A way to create simple reusable template components in Django."
optional = false
python-versions = "*"
files = [
    {file = "django_components-0.29-py3-none-any.whl", hash = "sha256:42d98ebe72f9e7cab4b3a5a6d72a17f1065716e1967ef0dff0dfd9c59bfbbca0"},
    {file = "django_components-0.29.tar.gz", hash = "sha256:5a7220e65a75632eb6deb842f36ce690a115b6c4c5cd7ab5ffa62ee8fda46908"},
]
```

PR #311, opened 2023-08-26
fix: `endfill` instead of `endslot` in README.md

ar4s:
Hi! I found a small error in the `README.md`, nothing very important, but I think it will be helpful to newbies.


EmilStenstrom:
Thank you @ar4s! We really should add tests for our code examples too. Until then we rely on fixes like yours! :)

Issue #262, opened 2023-04-12
Component does not load js file, when "-" or "/" in register name


M4p4:
Given is following code:
```python
@component.register("foo-card")
class Foo(component.Component):
      template_name = "foo-card/foo-card.html"

      def get_context_data(self, url):
          return {
              "url": url,
          }

      class Media:
         js = "foo-card/foo-card.js"
```     

The component is registered with the name "foo-card". When this component is used the js file `foo-card.js` does not load!

When i change the register name of the component to:

```python
@component.register("foocard")
```
the component and the js file `foo-card.js` load!

I think the js should also load when there is a - or / in the component name.

EmilStenstrom:
@lemontheme I vaguely remember something about component names being valid identifiers?

@M4p4 Are you open to supplying a PR that fixes this?

lemontheme:
Ah, beat me to it @EmilStenstrom!

This is a bug. I'll explain why below. Fortunately, this one's easy to set right. I can easily patch this before the end of the day (CEST).

What's happened here is that, around the time we introduced fill tag aliases, we experimented with treating component names, fill names, and slot names as identifiers rather than as variables that need resolving. To illustrate the difference,
the names of Django's builtin 'block' tags are treated identifiers, e.g. `title` below, which has a constant value = 'title', regardless of the rendering context.

```django
{% block title %}{{ section.title }}{% endblock %}
```

By contrast, the argument of a builtin  'if' tag is treated as a variable, i.e. it is resolved against the rendering context. For example, the value of `athlete_list` below depends on the value associated with the corresponding key in the redering context.

```django
{% if athlete_list %} ...
```

In this setting, if the argument is quotes it is treated as a literal string. E.g. `"athlete_list"` resolves to `"athlete list"`.

```django
{% if "athlete_list" %} ...
```

In the end, we determined it made most sense to treat component names, fill names, and slot names as variables rather than as identifiers. Because variables only allow `_` as a separator, I (mistakenly) figured it would mean less confusion in the long run if we extended that naming constraint (i.e. that all names must be valid python identifiers) to component names – i.e. such that a component name like `foo-card` would be rejected, but `foo_card` would be permitted. Clever me forgot to add a check to the register() function, so that's why this didn't raise an error when registering the component.

[...]

PR #320, opened 2023-09-28
Add support for - and / in component names in middleware

imankulov:
- Fix the COMPONENT_COMMENT_REGEX to find component names with special symbols.
- Fix the test to catch the error.

Follow-up of e0c29e5 and ref: #262

### DIFF
--- a/django_components/middleware.py
+++ b/django_components/middleware.py
@@ -11,9 +11,9 @@ CSS_DEPENDENCY_PLACEHOLDER = '<link name="CSS_PLACEHOLDER">'
 JS_DEPENDENCY_PLACEHOLDER = '<script name="JS_PLACEHOLDER"></script>'
 
 SCRIPT_TAG_REGEX = re.compile("<script")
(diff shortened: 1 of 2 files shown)


EmilStenstrom:
@imankulov Thank you! Did i miss something with the tests?

The testing error is because Django adds type="text/css" in some versions, and some not. I simply remove that string and compare without it to fix it across django versions.

imankulov:
> The testing error is because Django adds type="text/css" in some versions, and some not. I simply remove that string and compare without it to fix it across django versions.

In the assertHTMLEqual() function, the type="text/css" attribute is removed from the first (left) argument. Initially, I passed a constant string in that argument. I have now changed the order of the arguments, and I hope that this resolves the issue.

EmilStenstrom:
Thank you! 

This is now released as django-components==0.28.3

(I also fixed it so that assertHTMLEquals removes those attributes from the right side, thanks for that feedback aswell).

imankulov:
Awesome! Thanks a lot for the quick turnaround, @EmilStenstrom!

Issue #323, opened 2023-10-09
Update README to use new management command


EmilStenstrom:
https://github.com/EmilStenstrom/django-components/pull/299 introduced a new management command that creates files for you. Let's update the recommended procedure to use that command to create the files.

PR #327, opened 2023-10-12
feat(docs): add `startcomponent` docs to the `README.md` file

hanifbirgani:
related #323

### DIFF
--- a/README.md
+++ b/README.md
@@ -535,6 +535,80 @@ COMPONENTS = {
 }
 ```
 
+## Management Command
+
+You can use the built-in management command `startcomponent` to create a django component. The command accepts the following arguments and options:
+
+- `name`: The name of the component to create. This is a required argument.
+
+- `--path`: The path to the components directory. This is an optional argument. If not provided, the command will use the `BASE_DIR` setting from your Django settings.
+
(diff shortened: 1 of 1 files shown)


EmilStenstrom:
This is great! The only thing I'm missing is changing the default usage in the topmost tutorial to use the new command instead of directing people to create files themselves. Separate PR och part of this one?

hanifbirgani:
> This is great! The only thing I'm missing is changing the default usage in the topmost tutorial to use the new command instead of directing people to create files themselves. Separate PR or part of this one?

So, do you mean moving the management command section above the "Create your first component" section?

EmilStenstrom:
No, I think the section you've written is great. But it's more like reference docs. The tutorial is "type this to get started", and currently that part is just assuming that people will create files in the right locations. Switching to your new command as the way to get started will make things so much easier!

hanifbirgani:
> No, I think the section you've written is great. But it's more like reference docs. The tutorial is "type this to get started", and currently that part is just assuming that people will create files in the right locations. Switching to your new command as the way to get started will make things so much easier!

I prefer a separate PR for that part.

PR #338, opened 2023-12-02
Clean up README.md

rwblokzijl:
- Remove trailing whitespace
- Use the same quotes as django defaults
- Use consisten name for css and js in tutorial

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ Start by creating empty files in the structure above.
 First you need a CSS file. Be sure to prefix all rules with a unique class so they don't clash with other rules.
 
 ```css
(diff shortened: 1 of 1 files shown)


EmilStenstrom:
@rwblokzijl Thank you! We are actually moving towards using style.css and script.js as the default names for components. Would you mind changing the README to use those instead of calendar.js/css instead?

https://github.com/EmilStenstrom/django-components/blob/master/django_components/management/commands/startcomponent.py

rwblokzijl:
@EmilStenstrom done

EmilStenstrom:
@rwblokzijl Thank you! You're very welcome to improve more things :D

Issue #345, opened 2023-12-13
Latest release (0.30) doesn't match the code in master


dylanjcastillo:
Hey @EmilStenstrom,

Awesome library! I've been using it alongside HTMX and have been loving it.

I noticed that the code, in at least the latest release, doesn't match the code in master. You are missing `managements/commands` and potentially other changes (?) since that was included 2 months ago. I noticed it because when I tried to use `startcomponent` it failed.

Here's a comparison of the contents of the release downloaded directly from pypi and the project:

From PyPi:
![image](https://github.com/EmilStenstrom/django-components/assets/7218756/b87e538f-c605-47b5-9a5d-8986bb8d0e66)

From GH:
<img width="1105" alt="image" src="https://github.com/EmilStenstrom/django-components/assets/7218756/0bfd8529-86a3-4dc5-a365-3f62f1095079">

I also noticed that you have a PR halfway through to include tests for `startcomponent`. I'd be happy to try to fix it if you're open to it.

EmilStenstrom:
Thank you for troubleshooting it this far.

I think I see what's going on. In setup.py we use find_packages to find the files to include. That one looks for only python packages and includes them. But management/commands lacks `__init__.py` files, so they are not packages, and are so not included.

Should be an easy fix. If you have the bandwidth to fix it it would of course be appreciated!

EmilStenstrom:
@dylanjcastillo Thank you for working on this. I've released 0.31 to PyPI, hoping that the startcomponent command is included this time? https://pypi.org/project/django-components/

dylanjcastillo:
Yes, it's there:

<img width="988" alt="image" src="https://github.com/EmilStenstrom/django-components/assets/7218756/b854a6ba-4dd0-48d5-b5e1-6ec37e3711c0">

Issue #324, opened 2023-10-10
Add test cases for `startcomponent` command


hanifbirgani:
I will add proper test cases for the newly introduced command.

PR #346, opened 2023-12-16
Add startcomponent test cases and setup

dylanjcastillo:
Hi @EmilStenstrom,

This PR addresses #324 and #345 let me know what you think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,3 +71,4 @@ pyproject.toml
 
 # Python environment
 .venv/
+.DS_Store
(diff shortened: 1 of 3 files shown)


EmilStenstrom:
Thanks for this!

Tests in CI seems to fail here:

```
           self.assertTrue(
              overwritten_date > created_date,
  >           "File was not overwritten with --force flag",
          )
  E       AssertionError: False is not true : File was not overwritten with --force flag
```

dylanjcastillo:
Not really sure why it was failing. Tests were working locally, so I think it might be something OS specific about `os.getmtime`.

I just updated the tests to use another method. They're working well locally.

Issue #351, opened 2023-12-25
Add setting for specifying "only" keyword as the default


EmilStenstrom:
From https://github.com/EmilStenstrom/django-components/issues/350 we can see that a use-case for django-components is:

```django
{% component_block "layout" ... only %}
  {% component_block "dialog" ... only %}
    {% fill "activator" %}
      {% component_block "button" ... only %}
        {% component "icon" icon="trash" ... only %}
        Delete {{ changes_count }} changes
      {% endcomponent_block %}
    {% endfill %}
  {% endcomponent_block %}
{% endcomponent_block %}
```

To simplify usage I would suggest we add a new setting that removes the need to specify "only" every time a new component is used, and instead that it picks the default from the setting.

PR #359, opened 2024-01-12
Use setting to limit context

dylanjcastillo:
Hi @EmilStenstrom,

I created a PR for this [issue](https://github.com/EmilStenstrom/django-components/issues/351), including tests.

Let me know if this is what you had in mind.

Best,
Dylan

### DIFF
--- a/django_components/app_settings.py
+++ b/django_components/app_settings.py
@@ -17,5 +17,9 @@ class AppSettings:
     def TEMPLATE_CACHE_SIZE(self):
         return self.settings.setdefault("template_cache_size", 128)
 
+    @property
+    def ISOLATE_COMPONENT_CONTEXT(self):
+        return self.settings.setdefault("isolate_component_context", False)
+
 
 app_settings = AppSettings()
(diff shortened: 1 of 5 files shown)


django_components/app_settings.py
```diff
@@ -17,5 +17,9 @@ def LIBRARIES(self):
     def TEMPLATE_CACHE_SIZE(self):
         return self.settings.setdefault("template_cache_size", 128)
 
+    @property
```
dylanjcastillo, line 20: @EmilStenstrom I added this to the app_settings, but I'm not sure what's the logic of picking an app vs. a global setting.

I see that in the Middleware you went for a global setting.

EmilStenstrom:
@dylanjcastillo What's the plan if we add another isolation behaviour as per the separate discussion? Here the value is just True/False... Should it be an enum from the getgo?

dylanjcastillo:
Hey @EmilStenstrom, you're right! I updated it to an enum

EmilStenstrom:
Awesome! I really appreciate these PRs @dylanjcastillo! Mind sending an e-mail to emil@emilstenstrom.se. There's some funding available for this project that I want to redirect to you!

dylanjcastillo:
Wow! That's very nice of you. Will reach out. Thank you!

Issue #183, opened 2022-10-17
Single file components


vb8448:
Have you ever thought to implement a single file component, similar to vuejs ?

Eg.:
```python
@component.register("calendar")
class Calendar(component.Component):
    # This component takes one parameter, a date string to show in the template
    def get_context_data(self, date):
        return {
            "date": date,
        }
   
    template = """
        <div class="calendar-component">Today's date is <span>{{ date }}</span></div>
    """
    css = """
        .calendar-component { width: 200px; background: pink; }
        .calendar-component span { font-weight: bold; }
    """
    js = """
        (function(){
            if (document.querySelector(".calendar-component")) {
                document.querySelector(".calendar-component").onclick = function(){ alert("Clicked calendar!"); };
            }
        })()
    """
```

vb8448:
> you lose all syntax highlighting in editors. Is there a way to get syntax highlighting some other way?

Pycharm seems to support [language injections](https://www.jetbrains.com/help/pycharm/using-language-injections.html#inject_language) and I guess also vscode does. Probably a specific extension will be needed.

EmilStenstrom:
I did some reading on this, and found a couple of crazy solutions such as https://github.com/twidi/mixt/ which have solved this in a mix of crazy and fantastic. But there's really nothing good out there. Language injections works in pycharm, but not in vscode. This also bypasses Django's template loading system, which loses caching. 

Overall, I think using strings like in your quick example would quickly break down for bigger components. So right now it's a "wontfix".

I do really like to have a solution for this, but I just don't see a solution I'd like to recommend to users. Losing syntax highlighting in editors seems like the major one right now.

aymericderbois:
I love the idea of having the template directly in a component variable. This is especially useful when you want to make small components!
Unfortunately for syntax highlighting, I think there is no solution. Even with Pycharm, the support of language injections is limited.

Nevertheless, as this is the kind of functionality I could use, I tried to make some tests and I made a quick development that allows to have "inline template".

Here is a "countdown" component, with the template directly in the component : 
```python
@component.register("countdown")
class Countdown(component.Component):
    template_string = """
        <span
          data-ended-at-timestamp="{{ until.timestamp }}"
          x-data="{
                interval: null,
                init() {
                    this.endTimestamp = parseInt($el.dataset.endedAtTimestamp);
                    this.processCountdown();
                    this.interval = setInterval(() => this.processCountdown(), 1000);
                },
                processCountdown() {
[...]

PR #361, opened 2024-01-13
Implement single file components

dylanjcastillo:
Hey @EmilStenstrom,

Sorry for so many PRs 😅. Just found some time and wanted to make progress on this.

Here's the PR for single-file components discussed in #183. The design is slightly different from your proposal, but hopefully still makes sense.

Let me know what you think.

Best,
Dylan

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,51 @@ The output from the above template will be:
 
 This makes it possible to organize your front-end around reusable components. Instead of relying on template tags and keeping your CSS and Javascript in the static directory.
 
+## Using single-file components
+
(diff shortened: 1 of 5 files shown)


EmilStenstrom:
Hi! I really appreciate you working on this, thanks! :) 

I've been going back and forth on the class variables. I like the simplicity, but at the same time I think it's hard to understand where the variables are coming from. Can they be used in the css and js templates too? Making the shortcuts methods and passing "context" to them would make that clear(er). At the same time, I think that use-case will be rare...

I say lets go and start with this. It's very useful for small components, and leaves lots of room for improvements down the line.

Issue #300, opened 2023-07-27
Not compatible with django-compressor


spapas:
When I try to run `python manage.py compress` on a project that uses django-compressor (https://github.com/django-compressor/django-compressor/) I get the following error:

```
  File "C:\progr\py3\ehcg\venv\lib\site-packages\compressor\offline\django.py", line 84, in remove_block_nodes
    setattr(node, attr, sub_nodelist)
AttributeError: can't set attribute 'nodelist'
```

After some debugging it seems that django-compressor can't handle the component_block for example like this: 

```
      {% component_block "card" "Εμφάνιση στοιχείων" color="" header_color="bg-info" header_text_color="text-light" %}
        {% fill "body" %}
          {% for p in ac.object_lists %}
            <a class='btn btn-info mt-1' href='{% url "object-list-detail" p.0 %}'>{{ p.1 }}</a>
          {% endfor %}
        {% endfill %}
      {% endcomponent_block %}
```

I opend that issue on django-compressor (https://github.com/django-compressor/django-compressor/issues/1201) but it is probably easier to fix django-components instead ?

EmilStenstrom:
@spapas We're open to reviewing PR:s if you write one up! I think supporting django-compressor is worthwhile.

spapas:
Hey @EmilStenstrom i'd be happy to provide a PR for this but unfortunately I'm not sure how this shouled be fixed :(

lemontheme:
Just looking at this now. I've yet to investigate it properly, but the exception seems to occur because django-compressor tries to reassign the nodelist attribute. (I've never used django-compressor, so I'll have to read up on why it does this.) Anyway, the reassignment fails on `ComponentNode.nodelist` this class's `nodelist` is implemented as a getter method using the \@property decorator.

Next steps:

1. Figure out what django-compressor does and what the implications would be of letting it make attribute reassignments to our internal Node objects.
2. Re-implement `ComponentNode.nodelist` as a bona fide public attribute, since this would appear to more in line with what other tooling expects.

Link to relevant LoC: https://github.com/EmilStenstrom/django-components/blob/3318bde9c36c237fcf88a965332b915340281780/django_components/templatetags/component_tags.py#L405C1-L407C1

dylanjcastillo:
@spapas how are you using `django-compress` with `django-components`? 

I tried it and it worked well, including the offline compression.

For this file:

```html
{% load compress %}
<!DOCTYPE html>
<html>
<head>
[...]

PR #364, opened 2024-01-19
Fix for issue with django-compressor

dylanjcastillo:
Hi @EmilStenstrom,

I was checking issue #300 and it looks like there's a quick fix.

There's a function called [handle_extendsnode](https://github.com/django-compressor/django-compressor/blob/939499ecca8886b3192da1343f20b30ab248baa4/compressor/offline/django.py#L20) in **django-compressor**, that creates a copy of the node tree replacing the block tags with the nodes of the children blocks.

During that process the `nodelist` is replaced [here](https://github.com/django-compressor/django-compressor/blob/939499ecca8886b3192da1343f20b30ab248baa4/compressor/offline/django.py#L84). But looks like those changes are compatible with **django-components**.

I did this small change, tested the results in an example provided by @spapas, and it looks like things are working well. Also, all tests are passing.

So I think it's safe to merge.

Best,
Dylan

### DIFF
--- a/django_components/templatetags/component_tags.py
+++ b/django_components/templatetags/component_tags.py
@@ -403,13 +403,13 @@ class ComponentNode(Node):
         self.context_kwargs = context_kwargs or {}
         self.isolated_context = isolated_context
         self.fill_nodes = fill_nodes
+        self.nodelist = self._create_nodelist(fill_nodes)
 
-    @property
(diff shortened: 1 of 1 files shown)


EmilStenstrom:
Wow, this is awesome! Great find! Fixes #300.

PR #366, opened 2024-01-22
Render components as views

dylanjcastillo:
Hi @EmilStenstrom,

I got a rough version of the changes and some basic testing, but would love to hear your thoughts about it. 

To avoid the registration duplicates issue, I thought it'd make sense to not raise an error if you're registering the same Component (by checking if the code is the same using `inspect.getsource`), and raise an error if you're trying to register a different component. This method should be useful when dealing with numerous components in an app, as opposed to not raising an error at all.

The way I'm handling filling the slots manually (`fill_slots`) method is a bit hacky, not sure what you think about it?


dylanjcastillo:
Would love to hear your thoughts as well, @CleitonDeLima

I understand that you don’t like some (or all) of the changes, but it’s hard to know why from just a 👎

django_components/component.py
```diff
@@ -59,7 +62,19 @@ def __new__(mcs, name, bases, attrs):
...
 
-class Component(metaclass=SimplifiedInterfaceMediaDefiningClass):
+class ComponentView(View):
+    http_method_names = [
+        "get",
+        "post",
+        "put",
+        "patch",
+        "delete",
+    ]
```
EmilStenstrom, line 72: I don't think this is needed really. My comment about sending an OPTIONS request is as valid with a PUT or PATCH request: "Why would you want...?". But if we are making it a view, lets go all the way. This just inheriting from View is fine here.

django_components/component_registry.py
```diff
@@ -11,11 +14,16 @@ def __init__(self):
         self._registry = {}  # component name -> component_class mapping
 
     def register(self, name=None, component=None):
-        if name in self._registry:
+        if (
+            name in self._registry
+            and name
+            and component
+            and inspect.getsource(component)
+            != inspect.getsource(self._registry[name])
```
EmilStenstrom, line 22: Are there other options than this? Feels a bit hacky...

sampleproject/calendarapp/urls.py
```diff
@@ -1,6 +1,8 @@
 from calendarapp.views import calendar
+from components.greeting import Greeting
 from django.urls import path
 
 urlpatterns = [
     path("", calendar, name="calendar"),
+    path("greeting/", Greeting.as_view(), name="calendar"),
```
EmilStenstrom, line 7: Should use another name for this path, instead of calendar twice!

tests/test_component.py
```diff
@@ -438,3 +439,52 @@ def test_instances_of_component_do_not_share_slots(self):
...
+            template_name = "simple_template.html"
+
+            def head(self, request, *args, **kwargs):
+                return HttpResponse(self.render({"variable": "test"}))
+
+        comp = MyComponent()
+        request = HttpRequest()
+        request.method = "HEAD"
+        response = comp.dispatch(request)
+        self.assertIsInstance(response, HttpResponseNotAllowed)
```
EmilStenstrom, line 490: This can be removed.

sampleproject/components/greeting.py
```diff
@@ -1,13 +1,29 @@
...
+        self.fill_slots(slots)
+        rendered_component = self.render(context)
+        return HttpResponse(rendered_component)
+
+    def get_context_data(self, request=None, **kwargs) -> Dict[str, Any]:
+        if request:
+            name = request.GET.get("name", "") if request else ""
+        else:
+            name = kwargs.get("name", "")
+        return {"name": name}
```
EmilStenstrom, line 22: I think I'd like to keep the get_context_data method free of the request object. I think it makes more sense to handle request related logic in the get method instead.

sampleproject/components/greeting.py
```diff
@@ -1,13 +1,29 @@
...
-class greeting(component.Component):
-    def get_context_data(self, greet, *args, **kwargs):
-        return {"greet": greet}
+class Greeting(component.Component):
+    def get(self, request, *args, **kwargs):
+        context = self.get_context_data(request, *args, **kwargs)
+        slots = [("message", "Hello, world!", None)]
+        self.fill_slots(slots)
+        rendered_component = self.render(context)
+        return HttpResponse(rendered_component)
```
EmilStenstrom, line 15: Hmm. This is the core new API right? I wonder if it would make more sense for self.render to get a new slots keyword arg, where you send in a dict with "name of slot" and "contents of slot". I think it might make it easier to understand what's going on... What do you think?

tests/test_component.py
```diff
@@ -438,3 +439,52 @@ def test_instances_of_component_do_not_share_slots(self):
...
+        self.assertIn(
+            "Variable: <strong>test</strong>", response.content.decode()
+        )
+
+    def test_component_view_post_request(self):
+        class MyComponent(component.Component):
+            template_name = "simple_template.html"
+
+            def post(self, request, *args, **kwargs):
+                return HttpResponse(self.render({"variable": "test"}))
```
EmilStenstrom, line 467: Maybe add some simple post data handling here to make the test make a little more sense?

tests/test_registry.py
```diff
@@ -37,13 +46,21 @@ def test_register_two_components(self):
...
             )
 
+    def test_allow_duplicated_registration_of_the_same_component(self):
+        self.registry.register(
+            name="testcomponent", component=MockComponentView
+        )
+        self.registry.register(
+            name="testcomponent", component=MockComponentView
+        )
+
```
EmilStenstrom, line 63: I think all tests should have an assertion somewhere. For this test I suggest this pattern:

```python
    try:
        myFunc()
    except ExceptionType:
        self.fail("myFunc() raised ExceptionType unexpectedly!")
```

EmilStenstrom (commented):
I think you're on the right track with this. I had some ideas of how the improve (in my mind) the main API. Let me know what you think.

(Again, I'm really appreciating your work here)

tests/test_component.py
```diff
@@ -447,7 +447,7 @@ class MyComponent(component.Component):
             template_name = "simple_template.html"
 
             def get(self, request, *args, **kwargs):
-                return self.render_to_response({"variable": "test"})
+                return HttpResponse(self.render({"variable": "test"}))
```
EmilStenstrom, line 450: Why is this better?

sampleproject/components/greeting.py
dylanjcastillo: Yes, definitely. There's no real reason for it to be here!

tests/test_component.py
dylanjcastillo: I wasn't sure about it. I ended up concluding that it wasn't worthwhile to add another method that was just a HttpResponse wrapper. What do you think?

EmilStenstrom: Hmm. I think I actually like it. It saves you from one import per component file, and it mimics the Django [shortcut render_to_response](https://docs.djangoproject.com/en/2.2/topics/http/shortcuts/#render-to-response). It's a small convenience I think.

django_components/component_registry.py
dylanjcastillo: I tried quite a few things, but none were as reliable. However, I found an alternative that might be a bit nicer.

The issue arises when adding components to `urlpatterns`. This re-registers the components and, in the case of Components used as views, they're treated as if they are different classes, due to how the packages are imported (or at least that's what I've been able to gather so far).

I tried using:

- `__name__`: But this will replace a component incorrectly if they share the same instance name.
- `__dict__`: For the same component `__module__`, and functions memory addresses aren't the same. So it raises an error about multiple components being registered with the same name. See example: https://pastebin.com/J6SH6eU9
- `__qualname__`: Has the same issue as `__dict_` as qualname (e.g., `greeting.Greeting` vs. `components.greeting.Greeting` are considered different).

I've found that the most reliable way to compare if two components are the same is by comparing the source codes, or using the location of the file + class name as follows:

1. Adding a `class_hash` to `Component`, based on the file the class is and its name:
```python
@classmethod
@property
class_hash(cls):
     return hash(str(inspect.getfile(cls)) + str(cls.__name__))
```

2. Doing the comparison in `component_registry.py` like this:
```python
    def register(self, name=None, component=None):
        if (
            existing_component
            and existing_component.class_hash != component.class_hash
        ):
            raise AlreadyRegistered(
                'The component "%s" has already been registered' % name
            )
        self._registry[name] = component
```

What do you think? Any thoughts on better/alternative approaches?

sampleproject/components/greeting.py
dylanjcastillo: I updated this and looks much better when passing a dictionary with the slots data!

tests/test_component.py
dylanjcastillo: For security purposes, I was thinking about escaping the content that goes into the slots by default:

```python
    def render(
        self,
        context_data: Dict[str, Any],
        slots_data: Optional[Dict[SlotName, str]] = None,
        escape_slots_content: Optional[bool] = True,
    ) -> str:
        context = Context(context_data)

        if slots_data:
            self._fill_slots(slots_data, escape_slots_content)

        updated_filled_slots_context: FilledSlotsContext = (
            self._process_template_and_update_filled_slot_context(
                context, template
            )
        )
        with context.update(
            {FILLED_SLOTS_CONTENT_CONTEXT_KEY: updated_filled_slots_context}
        ):
            return template.render(context)

    def _fill_slots(
        self,
        slots_data: Dict[SlotName, str],
        escape_content: bool = True,
    ):
        """Fill component slots outside of template rendering."""
        self.fill_content = [
            (
                slot_name,
                TextNode(escape(content) if escape_content else content),
                None,
            )
            for (slot_name, content) in slots_data.items()
        ]
```

What do you think?

django_components/component.py
```diff
@@ -84,6 +86,11 @@ def __init__(
         self.outer_context: Context = outer_context or Context()
         self.fill_content = fill_content
 
+    @classmethod
+    @property
+    def class_hash(cls):
+        return hash(str(inspect.getfile(cls)) + str(cls.__name__))
+
```
EmilStenstrom, line 93: Using inspect here bugs me a bit. I think it's because I've mostly used inspect when debugging or printing exceptions. I wonder if we could simply use `cls.__module__ + cls.__name__` here instead?

sampleproject/components/greeting.py
```diff
@@ -1,13 +1,26 @@
...
-    def get_context_data(self, greet, *args, **kwargs):
-        return {"greet": greet}
+class Greeting(component.Component):
+    def get(self, request, *args, **kwargs):
+        slots = {"message": "Hello, world!"}
+        context = {"name": request.GET.get("name", "")}
+        rendered_component = self.render(
+            context_data=context, slots_data=slots
+        )
+        return HttpResponse(rendered_component)
```
EmilStenstrom, line 16: This is a really nice API I think. Well done! :D

EmilStenstrom:
There's a small inspect issue that I'd like you to look at, but after that I think this is ready to merge. This is awesome!

(Also reminder to e-mail me at emil@emilstenstrom.se! :) )

django_components/component.py
EmilStenstrom: Well researched, and explained, thanks. Let's go with your solution!

tests/test_component.py
EmilStenstrom: Ugh, security. Yes, we really should think this through. My thoughts:

* Is rendering a component with render_to_response() protected already?
* **Not** escaping slots would be very dangerous, you are right. This definitely needs to be fixed before merging.
* Protential problem: wouldn't it be common to render a slot with HTML contents, the same contents that we wouldn't want a user to have access to? How should we handle this?
* A couple of tests that make sure we are not passing <script> tags unescaped by default would probably be good.

dylanjcastillo: > Is rendering a component with render_to_response() without slots protected already?

Yes, I added a test to verify that passing insecure parameters to the context is safe. From what I could gather, the context is processed with [this](https://github.com/django/django/blob/e412d85b4626bc56eb25206a40c3529162ce5dfc/django/template/context.py#L281) and that escapes the content by default. 

> Not escaping slots would be very dangerous, you are right. This definitely needs to be fixed before merging.
> Protential problem: wouldn't it be common to render a slot with HTML contents, the same contents that we wouldn't want a user to have access to? How should we handle this?

I made it the default to escape the content passed in slots but added a parameter so that users can control this. I also added a mention of this in the README, which hopefully is enough for users to understand the implications of this.

> A couple of tests that make sure we are not passing <script> tags unescaped by default would probably be good.

✅

EmilStenstrom:
Fantastic work on this! This is such a great addition to the library. I'll merge and release a new version tomorrow.

tests/test_component.py
EmilStenstrom: Excellent.

EmilStenstrom:
Sorry for the delay, I've now released version 0.34 which includes this fix: https://pypi.org/project/django-components/0.34/

Issue #368, opened 2024-01-27
django_components.component_registry.AlreadyRegistered error when using component as view.


Fingel:
Hello!
I found this library because I had the exact same use case in mind as outlined in this issue: https://github.com/EmilStenstrom/django-components/issues/349
I would like to use Django + HTMX, and being able to return a single component as a view fits perfectly. I believe this has the potential to become a very powerful paradigm.

It's mentioned in https://github.com/EmilStenstrom/django-components/issues/349#issuecomment-1891103065 that the author of the PR needed to explicitly register each component to avoid the `AlreadyRegistered` error. The pull request here by @dylanjcastillo  https://github.com/EmilStenstrom/django-components/pull/366 mentions that a modification had to be made to not raise an error if the same component is registered twice. 

I still get the error, so I'm wondering if this is not working as intended, or perhaps a different step needs to be taken? I haven't taken a look at the PR yet, but I'll be happy to debug.

Fingel:
OK, I think I've found the root cause. The class hash function here:
https://github.com/EmilStenstrom/django-components/blob/91b4accfebf74f3f7a8cf2ecab03789a1979d578/django_components/component.py#L93
will return 2 different hashes if a class is imported using a different path. 

Take for example, a Todo component with a .py file under `components/todo/todo.py`. When using the regular `@component.register("todo")` wrapper, the return value of `cls.__module__` is `todo`.

However, if you import the module in another file, such as `urls.py` like this:
```python
from components.todo.todo import TodoComponent

urlpatterns = [...]
```
the return value of `cls.__module__` will be `components.todo.todo`.

This is an issue because the documentation at
https://github.com/EmilStenstrom/django-components/blob/91b4accfebf74f3f7a8cf2ecab03789a1979d578/README.md?plain=1#L449 instructs you to place your components `urls.py` in the top level `components` folder, and import the example competent like so:

`from calendar import Calendar`

However this import _will not work_ if the urls.py module is in a directory above the calendar folder. You need to write it like this:

`from components.calendar.calendar import Calendar` 

or use a relative import. Either way, the hash function will not match because the module will not simply be `calendar`.

The only way to get this to work is if where you are importing the module (such as in a `urls.py`) is in the same module as where the component is defined. And indeed, in the sample/test project that the component views tests use:
https://github.com/EmilStenstrom/django-components/blob/91b4accfebf74f3f7a8cf2ecab03789a1979d578/sampleproject/components/urls.py

[...]

PR #369, opened 2024-01-27
Fix issue #368 and add tests

dylanjcastillo:
Hey @EmilStenstrom @Fingel,

I got a working version of a fix for issue #368. There was a quick fix (but @EmilStenstrom you might not be its biggest fan 😅). The hardest part was getting tests that replicated the issue, but I now have a version that seems to achieve that.

@Fingel your research was very useful to understand where the issue originated. Thank you very much.

When Django [initializes](https://docs.djangoproject.com/en/5.0/ref/applications/#initialization-process) the application registry, for each application in `INSTALLED_APPS` Django imports the root package of the application.

For `django-components`, this means the `ready` method is called which then calls `autodiscover`, as pointed by Fingel. `autodiscover` looks for all the locations of components and then executes those files.

The issue occurs during the execution of the component files found in `autodiscover`, specifically in `spec.loader.exec_module(module)`. Due to differences in the way the modules are imported I couldn't find a way to verify if two components were the same except for checking the file it was coming from + the name of the class.

I found two ways to do that:

1. `sys.modules[cls.__module__].__file__`: I think this should always work, but I'm not really sure, I have some gaps on how some of the `spec` and `module` stuff works. So, I'd rather go for a less standard approach, but one that I feel more confident about.
2. `inspect.getfile(cls)`: This is the method I proposed previously. Based on my tests it works well, and even though it might seem a bit hacky, other projects such as spaCy seem to use it for [similar purposes](https://github.com/explosion/spaCy/blob/a493981163002d0cd2409950512eeeccb6fa4690/spacy/util.py#L1143).

I propose we use `inspect.getfile()` fix for now. I think it's worth it, given the use cases it unblocks. OTOH, for future releases (including some of the stuff we were discussing in the HTMX-enabled components thread), it might be worthwhile to rethink the whole registration process, given that Components are becoming more complex now.

Also, I found an unrelated issue in `autodiscover` that, depending on the path resolution, resulted in duplicated directories when running `Loader.get_dirs`. This wasn't causing this issue, but I fixed it as well (see changes to `template_loader.py`. Maybe it'd speed up a bit the initialization process.

Finally, after some research, I realized that there's [no guarantee](https://docs.djangoproject.com/en/5.0/ref/applications/#django.apps.AppConfig.ready) that `ready` will be called only once when initializing the application. So, `autodiscover` and thus the registry of components must be robust to be running them multiple times. At first, I thought that we could prevent `autodiscover` from running twice, but I don't think that makes a lot of sense based on this new information.

TL;DR:

1. I propose we use `inspect.getfile()` as it fixes the issue for now.
2. I got tests that replicate the issue.
3. I also found a smaller bug I fixed in `template_loader.py`
4. For future releases it might be worthwhile to rethink the whole registration process

### DIFF
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -1,4 +1,5 @@
 import difflib
+import inspect
 from collections import ChainMap
 from typing import Any, ClassVar, Dict, Iterable, Optional, Set, Tuple, Union
 
@@ -90,7 +91,7 @@ class Component(View, metaclass=SimplifiedInterfaceMediaDefiningClass):
     @classmethod
     @property
     def class_hash(cls):
-        return hash(str(cls.__module__) + str(cls.__name__))
+        return hash(inspect.getfile(cls) + cls.__name__)
 
     def get_context_data(self, *args, **kwargs) -> Dict[str, Any]:
         return {}
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -88,10 +88,8 @@ class Component(View, metaclass=SimplifiedInterfaceMediaDefiningClass):
         self.outer_context: Context = outer_context or Context()
         self.fill_content = fill_content
 
-    @classmethod
-    @property
-    def class_hash(cls):
-        return hash(inspect.getfile(cls) + cls.__name__)
+    def __init_subclass__(cls, **kwargs):
+        cls.class_hash = hash(inspect.getfile(cls) + cls.__name__)
 
     def get_context_data(self, *args, **kwargs) -> Dict[str, Any]:
         return {}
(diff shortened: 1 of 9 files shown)


EmilStenstrom:
@dylanjcastillo Thanks again!

@Fingel Just released [0.34.1](https://pypi.org/project/django-components/0.34.1/), can you verify it works?

Fingel:
Nice work, I had also considered using `inspect.getfile()` in the hash function. Kinda crazy how hard it is to uniquely identify a class like that.

PR #372, opened 2024-01-29
Fix for #290

dylanjcastillo:
Hey @EmilStenstrom,

I came across issue #290 and made a quick fix for it, and added tests. Let me know what you think!

Best,
Dylan


CleitonDeLima:
This also resolves the discussion #332

PR #385, opened 2024-02-21
Don't rely on BASE_DIR being a Path.

### LINKED ISSUE #381: If settings.BASE_DIR is a string, django-components errors when loading the templates
In my settings file I had this as `BASE_DIR`
`BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))`
Which was later used for this in `TEMPLATES`:
`'DIRS': [os.path.join(BASE_DIR, 'templates')],`

When I added django-components to the project this error happened
```
  File "/foo/lib/python3.11/site-packages/django_components/template_loader.py", line 18, in get_dirs
    path = (settings.BASE_DIR / component_dir).resolve()
            ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
TypeError: unsupported operand type(s) for /: 'str' and 'str'
```

The solution is to change the `BASE_DIR`to `BASE_DIR = Path(__file__).resolve().parent.parent.parent`, which is what https://github.com/EmilStenstrom/django-components/issues/290 does in the example settings.

Since `BASE_DIR` is not an official Django setting, just something mentioned [in some places of the documentation](https://docs.djangoproject.com/en/5.0/search/?q=base_dir), the library documentation should explain what the setting is and what value is expected to have (a `pathlib.Path` instance)

EmilStenstrom:
Fixes #381

I went through all usages of BASE_DIR in the project and made sure all of them are safe for strings. Didn't touch sampleproject which still defined BASE_DIR as a Path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ Modify STATICFILES_DIRS (or add it if you don't have it) so django can find your
 ```python
 STATICFILES_DIRS = [
     ...,
-    BASE_DIR / "components",
+    os.path.join(BASE_DIR, "components"),
 ]
 ```
 
--- a/django_components/template_loader.py
+++ b/django_components/template_loader.py
@@ -15,7 +15,7 @@ class Loader(FilesystemLoader):
         directories = set(get_app_template_dirs(component_dir))
 
         if hasattr(settings, "BASE_DIR"):
-            path = (settings.BASE_DIR / component_dir).resolve()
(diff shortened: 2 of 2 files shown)


dylanjcastillo:
The fix looks good. And sorry, I introduced this bug in a previous fix.

EmilStenstrom:
@dylanjcastillo Thanks for reviewing! Don't worry, this project is littered with my little omissions like this. We just fix them and release a new version!

Issue #390, opened 2024-03-01
Adding HTML, JS, and CSS typing hints


dylanjcastillo:
To get syntax highlighting in VSCode for single-file components, I think we should include typing hints for HTML, CSS, and JS.

You can use [sourcetypes](https://pypi.org/project/sourcetypes/), but I think it'd be nice if we provided that directly, to avoid installing an additional library.

What do you think?

I'd be happy to open a PR for that.

EmilStenstrom:
Agreed, we should include it as a dependency!

EmilStenstrom:
This is now supported in django-components==0.51: https://pypi.org/project/django-components/0.51/

PR #391, opened 2024-03-02
Add type hints

dylanjcastillo:
Hi @EmilStenstrom, this is the PR for #390. Let me know what you think!


django_components/types.py
```diff
@@ -0,0 +1,9 @@
+try:
+    from typing import Annotated
+except ImportError:
+    from typing_extensions import Annotated
```
EmilStenstrom, line 4: This needs to be added as a dependency to avoid crashes I think?

django_components/types.py
```diff
@@ -0,0 +1,9 @@
...
+
+html = Annotated[str, "html"]
```
EmilStenstrom, line 6: This seems not to be in use?

EmilStenstrom (commented):
Looks straightforward! Added two comments, but almost ready to merge!

django_components/types.py
dylanjcastillo: I removed typing-extensions and instead created a simple Annotated class for the cases of Python <3.9. I checked and it still works well to provide the syntax highlighting.

I think it's better not to add additional dependencies if not really necessary, especially for this small use cases.

EmilStenstrom:
Looks great, thanks! :)

Issue #380, opened 2024-02-14
Feature proposal: HTML, JS and CSS file autodiscovery


JuroOravec:
## TL;DR
Instead of explicitly listing all HTML, JS and CSS dependencies, add the option to allow `django-components` take care of it at runtime by either:
  - a) Looking for HTML/JS/CSS files in the directly where the python component file is
  - b) Looking for HTML/JS/CSS files in a different directory

---

## Description

Normally, we have to explicitly set the template `template_name`, as well as `Media.js` or `Media.css` files, like so:

```py
@component.register("button")
class Button(Component):
    template_name = "button.html"
    class Media:
        js = ["..."]
        css = ["..."]
```

In my project, the file paths are a lot longer, so the file declaration gets repetitive:

```py
@component.register("button")
class Button(Component):
    template_name = "module_name/components/component_name/component_name.html"
    class Media:
        js = "module_name/components/component_name/component_name.js"
        css = "module_name/components/component_name/component_name.css"
```

We also keep files for each component for in their own directory, e.g.:

```txt
app
|-- templates
    |--module_a
    |--module_b
        |-- pages
        |-- components
            |-- component_c
            |-- component_d
                |-- component_d.py
                |-- component_d.html
                |-- component_d.js
                |-- component_d.css
```

Thus, it would nice developer experience if we could skip listing all the files, and instead just do:

```py
@component.register("button")
class Button(Component):
	file_autodiscovery = True
```

As added bonus, this would make it easier to move the whole component folders around.

### Changing searched directory

In our use case, it's sufficient to look only in the parent directory of the component file. But for completeness, I think the feature should allow to change the directory which is searched for files.

For example, instead of searching in `app/templates/components/button`, following component would search for HTML, JS and CSS files in `app/templates/static/button`:

```py
@component.register("button")
class Button(Component):
	file_autodiscovery = True
    file_autodiscovery_root = "../static/button"
```

NOTE: To give users full flexibility, ideally they could specify 1) multiple dirs to be searched, and 2) set different dirs for each file type - HTML/CSS/JS. However, that's out of scope of this proposal.

## Implementation

Implementation makes use of a Metaclass that extends `django_components.component.SimplifiedInterfaceMediaDefiningClass`, which tries to pre-populate the `Component.template_name` and `Component.Media.js`/`Component.Media.css` attributes at class definition.

<details>
<summary> Click to expand  </summary>

```py
import os
from glob import glob
from typing import NamedTuple

[...]

PR #395, opened 2024-03-09
feat: resolve media and template files relative to component class dir

JuroOravec:
Here is the first go at supporting relative imports for template and media files (See https://github.com/EmilStenstrom/django-components/issues/380). Relative import means that, when we define a component with a template and/or media files, then we can specify import path relative to the directory where the component class is defined.

So instead of specifying the import path relative to the Django's templates dirs, e.g.:
```py
@component.register("button")
class Button(Component):
    template_name = "module_name/components/component_name/component_name.html"
    class Media:
        js = "module_name/components/component_name/component_name.js"
        css = "module_name/components/component_name/component_name.css"
```

We can shorten it to:

```py
@component.register("button")
class Button(Component):
    template_name = "template.html"
    class Media:
        js = "script.js"
        css = "style.css"
```

For this to work, the directory where the component is defines MUST be reachable from one of the dirs set in `TEMPLATES.DIRS` Django settings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -681,6 +681,36 @@ COMPONENTS = {
 }
 ```
 
+## Logging and debugging
+
+Django components supports [logging with Django](https://docs.djangoproject.com/en/5.0/howto/logging/#logging-how-to). This can help with troubleshooting.
+
+To configure logging for Django components, set the `django_components` logger in `LOGGING` in `settings.py` (below).
+
+Also see the [`settings.py` file in sampleproject](./sampleproject/sampleproject/settings.py) for a real-life example.
+
+```py
+import logging
+import sys
+
+LOGGING = {
+    'version': 1,
+    'disable_existing_loggers': False,
+    "handlers": {
+        "console": {
+            'class': 'logging.StreamHandler',
+            'stream': sys.stdout,
+        },
+    },
+    "loggers": {
+        "django_components": {
+            "level": logging.DEBUG,
+            "handlers": ["console"],
+        },
+    },
+}
+```
+
 ## Management Command
 
 You can use the built-in management command `startcomponent` to create a django component. The command accepts the following arguments and options:
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ from django_components import component
 class Calendar(component.Component):
     # Templates inside `[your apps]/components` dir and `[project root]/components` dir will be automatically found. To customize which template to use based on context
     # you can override def get_template_name() instead of specifying the below variable.
-    template_name = "calendar/calendar.html"
+    template_name = "calendar.html"
 
     # This component takes one parameter, a date string to show in the template
     def get_context_data(self, date):
@@ -205,8 +205,8 @@ class Calendar(component.Component):
         }
 
     class Media:
-        css = "calendar/style.css"
-        js = "calendar/script.js"
+        css = "style.css"
+        js = "script.js"
 ```
 
 And voilá!! We've created our first component.
--- a/README.md
+++ b/README.md
@@ -761,7 +761,10 @@ One of our goals with `django-components` is to make it easy to share components
 
 - [django-htmx-components](https://github.com/iwanalabs/django-htmx-components): A set of components for use with [htmx](https://htmx.org/). Try out the [live demo](https://dhc.iwanalabs.com/).
 
-## Install locally and run the tests
+
+## Running django-components project locally
+
+### Install locally and run the tests
 
 Start by forking the project by clicking the **Fork button** up in the right corner in the GitHub . This makes a copy of the repository in your own name. Now you can clone this repository locally and start adding features:
 
@@ -794,3 +797,34 @@ pyenv install -s 3.12
 pyenv local 3.6 3.7 3.8 3.9 3.10 3.11 3.12
 tox -p
 ```
+
+### Developing against live Django app
+
+How do you check that your changes to django-components project will work in an actual Django project?
+
+Use the [sampleproject](./sampleproject/) demo project to validate the changes:
+
+1. Navigate to [sampleproject](./sampleproject/) directory:
+    ```sh
+    cd sampleproject
+    ```
+
+2. Install dependencies from the [requirements.txt](./sampleproject/requirements.txt) file:
+    ```sh
+    pip install -r requirements.txt
+    ```
+
+3. Link to your local version of django-components:
+    ```sh
+    pip install -e ..
+    ```
+    NOTE: The path (in this case `..`) must point to the directory that has the `setup.py` file.
+
+4. Start Django server
+    ```sh
+    python manage.py runserver
+    ```
+
+Once the server is up, it should be available at <http://127.0.0.1:8000>.
+
+To display individual components, add them to the `urls.py`, like in the case of <http://127.0.0.1:8000/greeting>
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -4,7 +4,6 @@ import os
 from collections import ChainMap
 from typing import Any, ClassVar, Dict, Iterable, List, Optional, Set, Tuple, Union
 
-from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.forms.widgets import Media, MediaDefiningClass
 from django.http import HttpResponse
@@ -27,6 +26,7 @@ from django_components.component_registry import (  # NOQA
     register,
     registry,
 )
+from django_components.utils.autodiscover import search
 from django_components.templatetags.component_tags import (
     FILLED_SLOTS_CONTENT_CONTEXT_KEY,
     DefaultFillContent,
@@ -76,22 +76,21 @@ def _resolve_component_relative_files(attrs: dict):
 
     # Prepare all possible directories we need to check when searching for
     # component's template and media files
-    candidate_dirs = settings.STATICFILES_DIRS or []
-    candidate_dirs.extend(
-        str(dir_path) for template_backend in settings.TEMPLATES or [] for dir_path in template_backend.get("DIRS", [])
-    )
+    components_dirs = search()
 
     # Get the directory where the component class is defined
     try:
-        abs_comp_dir, comp_dir = _get_dir_path_from_component_module_path(attrs["__module__"], candidate_dirs)
+        comp_dir_abs, comp_dir_rel = _get_dir_path_from_component_module_path(attrs["__module__"], components_dirs)
     except RuntimeError:
         # If no dir was found, we assume that the path is NOT relative to the component dir
         return
 
     # Check if filepath refers to a file that's in the same directory as the component class.
+    # If yes, modify the path to refer to the relative file.
+    # If not, don't modify anything.
     def resolve_file(filepath: str):
-        maybe_resolved_filepath = os.path.join(abs_comp_dir, filepath)
-        component_import_filepath = os.path.join(comp_dir, filepath)
+        maybe_resolved_filepath = os.path.join(comp_dir_abs, filepath)
+        component_import_filepath = os.path.join(comp_dir_rel, filepath)
 
         if os.path.isfile(maybe_resolved_filepath):
             return component_import_filepath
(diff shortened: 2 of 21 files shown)


django_components/component.py
```diff
@@ -60,9 +62,93 @@ def __new__(mcs, name, bases, attrs):
             if hasattr(media, "js") and isinstance(media.js, str):
                 media.js = [media.js]
 
+        _resolve_component_relative_files(attrs)
+
         return super().__new__(mcs, name, bases, attrs)
 
 
+def _resolve_component_relative_files(attrs: dict):
```
JuroOravec, line 70: I've defined the logic as standalone functions and prefixed with underscore. Let me know if there are any styling guidelines I should follow

django_components/component.py
```diff
@@ -60,9 +62,93 @@ def __new__(mcs, name, bases, attrs):
...
+    """
+    Check if component's HTML, JS and CSS files refer to files in the same directory
+    as the component class. If so, modify the attributes so the class Django's rendering
+    will pick up these files correctly.
+    """
+
+    # Prepare all possible directories we need to check when searching for
+    # component's template and media files
+    candidate_dirs = settings.STATICFILES_DIRS or []
```
JuroOravec, line 79: NOTE: So in my initial code, I relied only on `settings.TEMPLATES`. But I've seen in this codebase that in the `sampleproject/settings`, the components are defined via `STATICFILES_DIRS`. So I've included in `STATICFILES_DIRS` among the looked up directories too.

However, when we run `component.render(...)`, then Django's templating engine cannot find the files unless we specify them in the `TEMPLATES.DIRS` setting too.

So that means that adding `STATICFILES_DIRS`  to `candidate_dirs` doesn't really make sense, because the same dirs must be set also in `TEMPLATES.DIRS` for things to work.

So I plan to remove this line, but wanted to share this.

sampleproject/components/calendar/calendar.py
```diff
@@ -23,3 +23,27 @@ def get(self, request, *args, **kwargs):
     class Media:
         css = "calendar/calendar.css"
         js = "calendar/calendar.js"
+
+
+@component.register("calendar_relative")
+class CalendarRelative(component.Component):
```
JuroOravec, line 29: I've added a variant of the Calendar component to show how to use it with relative paths

tests/test_context.py
```diff
@@ -242,9 +242,12 @@ def test_one_context_call_with_simple_component(self):
...
         )
-        rendered = template.render(Context()).strip()
-
-        self.assertEqual(rendered, '<p class="incrementer">value=1;calls=1</p>', rendered)
+        rendered = template.render(Context()).strip().replace("\n", "")
+        self.assertEqual(
+            rendered,
+            '<link href="relative_file/relative_file.css" media="all" rel="stylesheet"><script src="relative_file/relative_file.js"></script><p class="incrementer">value=1;calls=1</p>',
+            rendered,
+        )
```
JuroOravec, line 250: This above also needs attention. For some reason, when I've added the `relative_file`, then I needed to update this piece of code. Do you know why?

JuroOravec:
As for docs, I'm thinking of either adding a comment & code example to the "Create your first component" section, or creating new section after that one. What do you think?

django_components/component.py
EmilStenstrom: Makes sense that this is a private function.

django_components/component.py
EmilStenstrom: I'm afraid lots of people have touched that code over the years, and that no one (including me) has a clear grasp of how it's supposed to work.

Some things to consider:
* The point of this library is to bundle django templates and static files, and this goes against the standard way to structure Django projects, namely putting templates in TEMPLATES.DIRS and static files in STATICFILES_DIRS. I'm not sure what the recommended way should even be, given that we ideally should keep supporting the collectstatic command, and the component.render doesn't work with STATICFILES_DIRS as you said.
* I was under the impression (could very well be wrong), that adding 'django_components.template_loader.Loader' as a template loaded made it possible to make it possible to load templates where ever we wanted. This seems not to be true?
* In the README, under [installation](https://github.com/EmilStenstrom/django-components/?tab=readme-ov-file#installation), we advise people to set the STATICFILES_DIRS settings, so I'm guessing that this is what people have set. It would be good to take backwards compatibility into account here.

As the person that last read the code, who likely has the best knowledge of how the code works, what would you recommend?

django_components/component.py
```diff
@@ -60,9 +62,93 @@ def __new__(mcs, name, bases, attrs):
...
+    # holds the component file.
+    root_dir_abs = None
+    for candidate_dir in candidate_dirs:
+        candidate_dir_abs = os.path.abspath(candidate_dir)
+        if comp_dir_abs_path.startswith(candidate_dir_abs):
+            # Get the common part of the path
+            root_dir_abs = comp_dir_abs_path[: len(candidate_dir_abs)]
+            break
+
+    if root_dir_abs is None:
```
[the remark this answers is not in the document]
EmilStenstrom: Good catch. I think it would make sense to log an error at the debug level, so that enabling debugging could allow people to see what paths are resolved when looking for the correct file. We get a lot of questions from people that are unsure of how the configure paths correctly.

sampleproject/components/calendar/calendar.py
EmilStenstrom: Excellent!

tests/test_context.py
EmilStenstrom: Not sure why. I don't see how the previous version of that test could ever have worked? 🤣 You version looks right, why didn't the component_dependencies render anything before?

EmilStenstrom:
> As for docs, I'm thinking of either adding a comment & code example to the "Create your first component" section, or creating new section after that one. What do you think?

I think relative imports should be the default that we recommend, so I think changing the first example at the top would suffice.

django_components/component.py
dylanjcastillo: For the `candidate_dirs` would it make sense to use the same directories as the ones found during in `autodiscover()`? https://github.com/EmilStenstrom/django-components/blob/7058f05e0415965ddd162d0ba663c5106e7ffb66/django_components/__init__.py#L27

Or is that leaving out some directories?

I think it might be more consistent. It could be something like those dirs + staticfiles_dirs if that's needed, but I guess based on what @JuroOravec is saying it might not be the case.

EmilStenstrom:
The current test error is because different django versions include type="text/css" and not when rendering style tags (and similar for js tags). They can be resolved by using testutils from: https://github.com/EmilStenstrom/django-components/blob/master/tests/testutils.py#L14

django_components/component.py
EmilStenstrom: Yes, using the same directories everywhere sounds like a good idea.

README.md
```diff
@@ -794,3 +797,34 @@ pyenv install -s 3.12
 pyenv local 3.6 3.7 3.8 3.9 3.10 3.11 3.12
 tox -p
 ```
+
+### Developing against live Django app
```
JuroOravec, line 801: Added section on how to run `sampleproject` locally

django_components/component.py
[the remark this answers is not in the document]
JuroOravec: I just got around to look into logging. I wanted to set the log level via settings, but not sure how to proceed. How are settings handled internally in this project?

From the README, I would assume that it's configured via the `COMPONENTS` Django settings field. There are a few places where it's used like so:
```py
getattr(settings, "COMPONENTS", {}).get("RENDER_DEPENDENCIES", False)
```

However, then there's also the `AppSettings`, which is used in more places, like so:
```py
app_settings.CONTEXT_BEHAVIOR == "isolated"
```

So I would probably use `AppSettings` to define and access the log level settings.

django_components/utils/autodiscover.py
```diff
@@ -0,0 +1,32 @@
+import glob
+from pathlib import Path
+from typing import List, Optional
+
+from django.template.engine import Engine
+
+from django_components.template_loader import Loader
+
+
+def search(search_glob: Optional[str] = None, engine: Optional[Engine] = None):
```
JuroOravec, line 10: I managed to use the autodiscovery logic, thanks @dylanjcastillo for suggestion! I moved the logic to separate file, and refactored it to make glob search optional, so it can used for both use cases.

sampleproject/components/nested/calendar/calendar.py
```diff
@@ -0,0 +1,25 @@
+from django_components import component
+
+
+@component.register("calendar_nested")
+class CalendarNested(component.Component):
```
JuroOravec, line 5: Added one more variant of the calendar, CalendarNested, to confirm that the logic works even for components that are not directly under the `components` directory.

Can confirm it works!

Screenshot 2024-03-17 at 12 04 39

django_components/component.py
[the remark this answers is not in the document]
JuroOravec: Ok turns out that I didn't need to touch the AppSettings, as logging can be configured [via Django](https://docs.djangoproject.com/en/5.0/topics/logging/).

I've added:
- Logging example to `sampleproject`
- Section on Logging to README
- A few logging statements

<details>
<summary>
Here you can see that the logs are printed only if log level is set to DEBUG for `django_components`
</summary>

Screenshot 2024-03-17 at 20 00 27

Screenshot 2024-03-17 at 20 00 46

</details>


<details>
<summary>
And here are some examples of what's logged when component uses files relative to the component vs not
</summary>

Screenshot 2024-03-17 at 20 10 11

Screenshot 2024-03-17 at 20 10 21

</details>

django_components/template_loader.py
```diff
@@ -2,34 +2,84 @@
...
+# is marked for deprecation.
+def path_is_relative_to(child_path: str, parent_path: str):
+    # If the relative path doesn't start with `..`, then child is descendant of parent
+    # See https://stackoverflow.com/a/7288073/9788634
+    rel_path = os.path.relpath(child_path, parent_path)
+    return not rel_path.startswith("..")
 
 
 class Loader(FilesystemLoader):
     def get_dirs(self):
```
JuroOravec, line 26: !IMPORTANT - PLEASE REVIEW!

There are couple of points related to `Loader.get_dirs` that need to be discussed:

1. I tried to document the flow a bit. However, I'm not sure if I got things right around `settings.SETTINGS_MODULE`, so please give that a look

2. I don't feel like I have a full grasp of the relationship between `STATICFILES_DIRS`, `TEMPLATES.DIRS`, and the `component_dir` variable:

 	Initially, I thought components are configured via `TEMPLATES.DIRS`, because in the project where I introduced django components, the `STATICFILES_DIRS` and `TEMPLATES.DIRS` were colocated.

 	When I started looking into this codebase, I saw that they are configured by `STATICFILES_DIRS`.

 	The `sampleproject` demo confirms that components are configured by `STATICFILES_DIRS`.

 	However, for some reason I had to add [the components dir to `TEMPLATES.DIR` in tests](https://github.com/EmilStenstrom/django-components/pull/395/files#diff-d75294aaaae87b0a95ac59c71cd89d1d288e3670cabb568a294ed2748def5e83R10). And I don't understand why. So this confuses me.

3. Continuing with the changes, I made a change so that the `component_dir` is not hardcoded, but rather used only as a default if `STATICFILES_DIRS` are not set. The reason for this is that, if we go back to point 2), then the components' location is derived from 2-3 variables.

	However, in README the instructions say to use `STATICFILES_DIRS` to configure the components dir, and the example uses a path which already includes the `".../components"` part:

	```python
	STATICFILES_DIRS = [
	    ...,
	    os.path.join(BASE_DIR, "components"),
	]
	```

	So docs make it look like it's define via `STATICFILES_DIRS`, but there are some extra vars behind the scenes.
	So IMO doing away with the hardcoded variable `component_dir` should simplify things, and make the behaviour more likely to be in line with what the docs say.

	So... That's at least the theory. As I said in point 2), don't know this codebase good enough yet to say.

	However, the tests are passing. And the `sampleproject` demo works. So it seems like things are working correctly.

	HOWEVER, this could possibly break for those who defined `STATICFILES_DIRS` but didn't include the components dir (e.g. `"{BASE_DIR}/components"`) among the entries. But then again, docs say that user SHOULD include it.

django_components/template_loader.py
```diff
@@ -2,34 +2,84 @@
...
-        component_dir = "components"
-        directories = set(get_app_template_dirs(component_dir))
```
JuroOravec, line 15: Note: I've replaced `get_app_template_dirs` with [the loop on line 46](https://github.com/EmilStenstrom/django-components/pull/395/files#diff-c6c9a460aa194de49542b7c8e241152e7727fd8d1ea49f77c3437b0373427d24R46), which is based on the implementation of `get_app_template_dirs`

JuroOravec:
Lastly, I've updated the first example in README to use the relative imports. And additionally, altho just locally, I tested the case when components are defined in `<app>/components`, instead of the root `components` dir.

With that, this MR is ready for re-review @dylanjcastillo @EmilStenstrom

EmilStenstrom:
It's very hard to verify that these changes won't break anything. People could have configured their projects in hundreds of different ways. My view is that you have gone to lengths to both improve the automatic testing of these changes and the relevant code, and done manual testing on top of that. I've read through the code, and all your comments, and they all make sense. 

I don't feel super confident that this won't break the library for someone, but at the same time I think you've done everything you could to test all the special cases.

So let's merge this, and do the final testing in the wild! :)

EmilStenstrom:
This is now live in django-components 0.60.

Issue #394, opened 2024-03-06
Add support for MyPy


JuroOravec:
I've started using MyPy recently, and for `django-components`, I get an error

```
Skipping analyzing "django_components": module is installed, but missing library stubs or py.typed marker
```

I haven't looked yet into how to make a library compatible with mypy, but I assume it shouldn't be too complicated. I can look into it if greenlit.

Let me know what you think @EmilStenstrom.

EmilStenstrom:
@JuroOravec Supporting mypy, and getting better typing support, is interesting!

JuroOravec:
Thanks! Please see https://github.com/EmilStenstrom/django-components/pull/396

PR #396, opened 2024-03-09
feat: add mypy support

JuroOravec:
As per https://github.com/EmilStenstrom/django-components/issues/394, this MR adds the `py.typed` marker that tells MyPy to work with the package's types.

Based on [this short blog post](https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package), this tiny change should be all that's required for this package to work nicely with MyPy.

For more info, see:
- [PEP 561 – Distributing and Packaging Type Information](https://peps.python.org/pep-0561/)
- [See more about the "Missing library stubs or py.typed marker" issue in MyPy docs in the section with the same name](https://readthedocs.org/projects/mypy/downloads/pdf/latest/#page=118)

### DIFF
--- a/django_components/__init__.py
+++ b/django_components/__init__.py
@@ -37,6 +37,8 @@ def import_file(path):
     MODULE_PATH = path
     MODULE_NAME = Path(path).stem
     spec = importlib.util.spec_from_file_location(MODULE_NAME, MODULE_PATH)
+    if spec is None:
+        raise ValueError(f"Cannot import file '{path}' - invalid path")
     module = importlib.util.module_from_spec(spec)
     sys.modules[spec.name] = module
-    spec.loader.exec_module(module)
+    spec.loader.exec_module(module)  # type: ignore
--- a/django_components/component.py
+++ b/django_components/component.py
@@ -1,7 +1,7 @@
 import difflib
 import inspect
 from collections import ChainMap
-from typing import Any, ClassVar, Dict, Iterable, Optional, Set, Tuple, Union
+from typing import Any, ClassVar, Dict, Iterable, List, Optional, Set, Tuple, Union
 
 from django.core.exceptions import ImproperlyConfigured
 from django.forms.widgets import Media, MediaDefiningClass
@@ -40,7 +40,7 @@ from django_components.templatetags.component_tags import (
 class SimplifiedInterfaceMediaDefiningClass(MediaDefiningClass):
     def __new__(mcs, name, bases, attrs):
         if "Media" in attrs:
-            media = attrs["Media"]
+            media: Component.Media = attrs["Media"]
 
             # Allow: class Media: css = "style.css"
             if hasattr(media, "css") and isinstance(media.css, str):
@@ -54,7 +54,7 @@ class SimplifiedInterfaceMediaDefiningClass(MediaDefiningClass):
             if hasattr(media, "css") and isinstance(media.css, dict):
                 for media_type, path_list in media.css.items():
                     if isinstance(path_list, str):
-                        media.css[media_type] = [path_list]
+                        media.css[media_type] = [path_list]  # type: ignore
 
             # Allow: class Media: js = "script.js"
             if hasattr(media, "js") and isinstance(media.js, str):
@@ -67,20 +67,20 @@ class Component(View, metaclass=SimplifiedInterfaceMediaDefiningClass):
     # Either template_name or template must be set on subclass OR subclass must implement get_template() with
     # non-null return.
     template_name: ClassVar[Optional[str]] = None
-    template: ClassVar[Optional[str]] = None
-    js: ClassVar[Optional[str]] = None
-    css: ClassVar[Optional[str]] = None
+    template: Optional[str] = None
+    js: Optional[str] = None
+    css: Optional[str] = None
     media: Media
 
     class Media:
-        css = {}
-        js = []
+        css: Optional[Union[str, List[str], Dict[str, str], Dict[str, List[str]]]] = None
+        js: Optional[Union[str, List[str]]] = None
 
     def __init__(
         self,
         registered_name: Optional[str] = None,
         outer_context: Optional[Context] = None,
-        fill_content: Union[DefaultFillContent, Iterable[NamedFillContent]] = (),
+        fill_content: Union[DefaultFillContent, Iterable[NamedFillContent]] = (),  # type: ignore
     ):
         self.registered_name: Optional[str] = registered_name
         self.outer_context: Context = outer_context or Context()
@@ -173,7 +173,7 @@ class Component(View, metaclass=SimplifiedInterfaceMediaDefiningClass):
     def _fill_slots(
         self,
         slots_data: Dict[SlotName, str],
-        escape_content: bool,
+        escape_content: Union[bool, None],
     ):
         """Fill component slots outside of template rendering."""
         self.fill_content = [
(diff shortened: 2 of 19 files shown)


EmilStenstrom:
Should we also add mypy as a linter? I'm not sure we pass the mypy checks yet though? 

@dylanjcastillo You are my equal, don't worry about what you can do and not :)

django_components/__init__.py
```diff
@@ -37,6 +37,8 @@ def import_file(path):
     MODULE_PATH = path
     MODULE_NAME = Path(path).stem
     spec = importlib.util.spec_from_file_location(MODULE_NAME, MODULE_PATH)
+    if spec is None:
+        raise ValueError(f"Cannot import file '{path}' - invalid path")
```
JuroOravec, line 41: Added assertion

django_components/component.py
```diff
@@ -67,20 +67,20 @@ class Component(View, metaclass=SimplifiedInterfaceMediaDefiningClass):
     # Either template_name or template must be set on subclass OR subclass must implement get_template() with
     # non-null return.
     template_name: ClassVar[Optional[str]] = None
-    template: ClassVar[Optional[str]] = None
-    js: ClassVar[Optional[str]] = None
-    css: ClassVar[Optional[str]] = None
+    template: Optional[str] = None
+    js: Optional[str] = None
+    css: Optional[str] = None
```
JuroOravec, line 72: Removed `ClassVar` annotation from component media & template.

This was causing MyPy errors when using the [inlined / single-file component method](https://github.com/EmilStenstrom/django-components#using-single-file-components) like in [this example](https://github.com/JuroOravec/django-components/blob/5c5d80b57dea6932fb3cb134e28829dd96159761/sampleproject/components/greeting.py#L32)

django_components/component.py
```diff
@@ -195,7 +195,7 @@ def _process_template_and_update_filled_slot_context(
             named_fills_content = {}
         else:
             default_fill_content = None
-            named_fills_content = {name: (nodelist, alias) for name, nodelist, alias in self.fill_content}
+            named_fills_content = {name: (nodelist, alias) for name, nodelist, alias in list(self.fill_content)}
```
JuroOravec, line 198: Not sure why, but MyPy was complaining that `self.fill_content` was not an iterable (although `self.fill_content` is a Union of two iterable types...). But the error resolved when I've cast it to list

django_components/templatetags/component_tags.py
```diff
@@ -426,14 +427,15 @@ def do_component(parser, token):
...
     if block_has_content(body):
         for parse_fn in (
             try_parse_as_default_fill,
             try_parse_as_named_fill_tag_set,
         ):
-            fill_nodes = parse_fn(body)
-            if fill_nodes:
+            curr_fill_nodes = parse_fn(body)
+            if curr_fill_nodes:
+                fill_nodes = curr_fill_nodes
```
JuroOravec, line 438: Slight change so that we assign to `fill_nodes` only if it's non-null

django_components/templatetags/component_tags.py
```diff
@@ -457,14 +459,15 @@ def try_parse_as_named_fill_tag_set(
...
-    seen_name_fexps = set()
+    seen_name_fexps: Set[FilterExpression] = set()
     for node in nodelist:
         if isinstance(node, NamedFillNode):
             if node.name_fexp in seen_name_fexps:
                 raise TemplateSyntaxError(
                     f"Multiple fill tags cannot target the same slot name: "
                     f"Detected duplicate fill tag name '{node.name_fexp}'."
                 )
+            seen_name_fexps.add(node.name_fexp)
```
JuroOravec, line 470: IMPORTANT: I've found that this check with `seen_name_fexps` wasn't working - `seen_name_fexps` was never populated, so the error would never raise.

sampleproject/components/greeting.py
```diff
@@ -11,7 +11,9 @@ def get(self, request, *args, **kwargs):
         context = {"name": request.GET.get("name", "")}
         return self.render_to_response(context, slots)
 
-    def get_context_data(self, name, *args, **kwargs) -> Dict[str, Any]:
+    def get_context_data(self, *args, **kwargs) -> Dict[str, Any]:
+        # NOTE: Because of MyPy in Python v3.6, arg `name` cannot be declared as separate variable
```
JuroOravec, line 15: In older Python versions (definitely 3.6, don't know about 3.7 and 3.8), MyPy complained when a subclassed method signature had extra arguments, even if followed by `*args`.

But this brings up a topic - Is it time to drop support for 3.6 and 3.7? After all, [they have reached end of life](https://devguide.python.org/versions/)

JuroOravec:
@dylanjcastillo @EmilStenstrom Thanks both for feedback!

@EmilStenstrom Good point with ensuring that the project itself passes Mypy checks. It would be annoying if users had to deal with errors from upstream dependencies. I've added mypy check to CI here https://github.com/EmilStenstrom/django-components/pull/396/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957f.

Because of that, I also had to fix couple of errors. Please see the new changes and comments above.

EmilStenstrom:
Wow, this PR is already full of nice small fixes so that we can pass mypy checks! Thanks for all this great work.

I'm not sure if this is best practice, but the other tools are implemented as environments on the tox.ini file: https://github.com/EmilStenstrom/django-components/blob/master/tox.ini#L13 - the nice part of this is that all tools are run when you run tox locally. I'm fine with fixing this after this PR is merged if you want to.

Separate question: You have already contributed some really nice PR:s. Would you like to become an official contributor to the project? Send me an email to emil@emilstenstrom.se and lets talk!

sampleproject/components/greeting.py
dylanjcastillo: I think it's a good idea to drop support eventually, but it might be worth keeping it for this PR because the required change is small enough.

I wonder if there's a way to estimate how many projects are using Python 3.6 or Python 3.7. I tried using some of the search features on GitHub but it didn't work.

django_components/component.py
```diff
@@ -173,7 +173,7 @@ def render_to_response(
     def _fill_slots(
         self,
         slots_data: Dict[SlotName, str],
-        escape_content: bool,
+        escape_content: Union[bool, None],
```
dylanjcastillo, line 176: I saw this and realized that I got the `escape_slots_content` type wrong in `render_to_response`. I think that one should be:

```
    def render_to_response(
        self,
        context_data: Dict[str, Any],
        slots_data: Optional[Dict[SlotName, str]] = None,
        escape_slots_content: bool = True,
        *args,
        **kwargs,
    ):
```

And this one should be only `bool`, not `None`.

dylanjcastillo:
Hey @JuroOravec,

The changes look good! Very useful. I added a comment to one of the changes, and also found that there seem to be an issue with some of the tests:

image

sampleproject/components/greeting.py
EmilStenstrom: The policy we're trying to follow is to support the same versions of python that django supports. So all officially supported Django versions -> All supported Python versions that they support: https://docs.djangoproject.com/en/5.0/faq/install/#what-python-version-can-i-use-with-django

Open to discussing other policies (in separate issue!) :)

django_components/component.py
EmilStenstrom: Fixed in separate commit.

EmilStenstrom:
I fixed the outstanding issues and published as 0.52.

Also made the executive decision that the tests need to pass on all versions, but that mypy checks only need to work in the latest python version. So the syntax can't break old versions, but it doesn't need to pass.

JuroOravec:
@EmilStenstrom @dylanjcastillo thanks for pushing this MR over the finishing line!

And thanks for fixing the setup/pipeline. I haven't worked with tox yet, been using poetry, so didn't know how that works 😄

EmilStenstrom:
@JuroOravec That was the easy part :)

For a quick guide to how I run the tests there's a short section at the end of the readme: https://github.com/EmilStenstrom/django-components?tab=readme-ov-file#install-locally-and-run-the-tests

Issue #403, opened 2024-03-24
v0.60.0 broken due to missing django_components.utils module


kaiku:
v0.60.0 refuses to load due to a missing `__init__.py` in the newly created `utils` dir introduced in https://github.com/EmilStenstrom/django-components/pull/395, cc @JuroOravec.

Relevant stack trace:
```
    from django_components.utils.autodiscover import search
ModuleNotFoundError: No module named 'django_components.utils'
```

I'm not able to test this now but I think the solution is easy:
1. Add an `__init__.py` to `utils`
2. (Preferred) Rename `utils/autodiscover.py` to `utils.py`. I think for simplicity's sake, everything can and should go in one file for this project.

JuroOravec:
@kaiku Thanks for making an issue for this, I've noticed it too last night, but it was already too late. Working on an MR for this atm.

EmilStenstrom:
@kaiku Thanks for reporting, a fix for this is released as 0.61 in pypi. Please update!

PR #404, opened 2024-03-24
Fix realeases v0.60 and v0.52

JuroOravec:
1) As per https://github.com/EmilStenstrom/django-components/issues/403, there was a missing `__init__.py` in the `utils` dir, because of which the directory was not included in the pypi release.

	This is a screenshot of files in the release when django_components was installed. You can see that the `utils` dir is missing:

	Screenshot 2024-03-24 at 00 04 00

	I went for @kaiku's suggestion and changed the utils dir to a file.

2) I also noticed that the mypy support wasn't working as expected. I looked into the guide I followed initially ([guide](https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/), [repo](https://github.com/whtsky/pixelmatch-py)), and realized that `py.typed` should be in the released package, NOT in project root.

Closes https://github.com/EmilStenstrom/django-components/issues/403


JuroOravec:
cc @EmilStenstrom  or @dylanjcastillo

EmilStenstrom:
Released as part of 0.61. Thanks @JuroOravec!

EmilStenstrom:
I wonder if we could add a CI step that could catch this in the future?

PR #406, opened 2024-03-24
fix: type component registry

GabDug:
Hey :wave:

I noticed in my project that mypy (in strict mode) was complaining that the `@component.register` decorator was untyped:
```
Call to untyped function "register" in typed context  [no-untyped-call]
```

I went ahead and typed the decorator, as well as the rest of the `component_registry` module.

I can scale back the changes to only type the decorator if needed.

Please let met know if you'd be interested in some help for the package, notably around type hints, CI or packaging, I'd be happy to contribute.

Thanks!


EmilStenstrom:
Great little addition, thanks!

Issue #414, opened 2024-03-29
Error with django-compressor: `ValueError: dictionary update sequence element #0 has length 5; 2 is required`


JuroOravec:
**Setup**

settings.py
```py
INSTALLED_APPS = [
	# ...
    'api',
    'compressor',
]

# Compressor settings
COMPRESS_ROOT = STATIC_BASE_DIR / 'static'
COMPRESS_ENABLED = True
STATICFILES_FINDERS = ('compressor.finders.CompressorFinder',"django.contrib.staticfiles.finders.FileSystemFinder","django.contrib.staticfiles.finders.AppDirectoriesFinder",)
```

template.html
```html
{% load compress %}
{# other... #}

<!DOCTYPE html>
<html lang="en">
  <head>
    {% compress css %}
      {% component_css_dependencies %}
    {% endcompress %} 
  </head>
  <body>
    ....
  </body>
</html>
```

Full trace
```txt
Traceback (most recent call last):
  File "/Users/presenter/repos/chk/appimage/baseapp/home/helpers/testutil_view.py", line 90, in test_view_cases
    self.do_test_view(case)
  File "/Users/presenter/repos/chk/appimage/baseapp/home/helpers/testutil_view.py", line 99, in do_test_view
    response = self.call_endpoint_for_view_case(case)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/presenter/repos/chk/appimage/baseapp/home/helpers/testutil_view.py", line 74, in call_endpoint_for_view_case
    response = self.do_get(endpoint, case)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/presenter/repos/chk/appimage/baseapp/home/helpers/testutil_view.py", line 68, in do_get
    response = case.client.get(url)
               ^^^^^^^^^^^^^^^^^^^^
  File "/Users/presenter/repos/chk/.venv/lib/python3.11/site-packages/django/test/client.py", line 927, in get
    response = super().get(path, data=data, secure=secure, headers=headers, **extra)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/presenter/repos/chk/.venv/lib/python3.11/site-packages/django/test/client.py", line 457, in get
    return self.generic(
           ^^^^^^^^^^^^^
[...]

PR #415, opened 2024-03-30
fix: avoid nested django contexts when rendering

JuroOravec:
Closes https://github.com/EmilStenstrom/django-components/issues/414

The issue was with Context, turns out that the context_data in component.Component.render may already be an instance of Context, so a simple check for that fixed it for me.

I also updated the input for `_process_template_and_update_filled_slot_context`, so we're not passing in Django's Context, since it wasn't necessary, so there's less touchpoints with it.


EmilStenstrom:
@JuroOravec Excellent, big thanks!

EmilStenstrom:
Released as version 0.63: https://pypi.org/project/django-components/0.63/

Issue #350, opened 2023-12-21
Use parent context in slots?


JuroOravec:
## Context

In frontend frameworks like Vue, it's normal to expect that when you define slots, e.g.:

```vue
<!-- Parent.vue -->
<MyComponent>
  Hello {{ firstName }}!
</MyComponent>
```

that the variables used INSIDE the slot belong to the parent context. That is, the `firstName` is taken from the file/component, where this markup was written (`Parent.vue`).

Now, if we want django-components to properly separate context of components, we need to use the `only` keyword.

However, as I found today, the problem is that using the `only` keyword also cuts off slot's access to the parent context.

So in the example above, if it was written in Django, the `firstName` would be undefined / empty.

This seem like a design shortcoming - I can either separate contexts, or cannot use parent's context in slots. But not both, like in Vue.

## Workaround

To allow us to use slots with the parent's context in our project, I subclassed django-component's `Component` to pre-render the slot fills (available via `Component.fill_content`), so that, at the point of render, the slot content is no longer dependent on the parent context.:

```py
from django.template.base import Node, NodeList, TextNode
import django_components.component

def prerender_component_slots(comp: django_components.component.Component):
    def prerender_node(node: Node):
        return TextNode(node.render(comp.outer_context))

    # Case: Single default slot
    if isinstance(comp.fill_content, NodeList):
        rendered = prerender_node(comp.fill_content)
        comp.fill_content = NodeList([rendered])

    # Case: Multiple named slots
    # Where each is NamedFillContent - tuple of (slot_name, Node, alias)
    else:
        comp.fill_content = [
            (
                slot_fill[0],
                NodeList([
                    prerender_node(slot_fill[1])
                ]),
                slot_fill[2],
            )
            for slot_fill in comp.fill_content
        ]
```

## Closing thoughts

Can we get something similar into django-components? Has it been considered before? (didn't find relevant issues)

EmilStenstrom:
Hi! Thanks for a well worded issue that both goes into the use-case and also shows some example code!

As an author of a components library there are many decisions like this, when one framework does it one way, and another does it another way. I have no intention of making django-components "like Vue", but if there's a good case for Vue to do it that way, I'm definitely open to doing the same.

So, let's start there. Why is to useful to have access to the component's parents context, but not the full context, when rendering a context?

My thinking around the only keyword is to enforce strict separation, so that global variables get much harder to abuse. It's still rather easy to include them though, you just send them in when calling the component:

```django
[...]

PR #416, opened 2024-03-30
refactor: move slots logic to own file

JuroOravec:
This is part of work towards https://github.com/EmilStenstrom/django-components/issues/350.

Initially I had hard time understanding how the slots work, so I wanted to make it easier by making it clearer where the slots logic lives by moving it to a separate file and documenting it a bit.

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -1,19 +1,7 @@
 import inspect
 import os
 from pathlib import Path
-from typing import (
-    Any,
-    ClassVar,
-    Dict,
-    Iterable,
-    List,
-    Mapping,
-    MutableMapping,
-    Optional,
-    Tuple,
-    Type,
-    Union,
-)
+from typing import Any, ClassVar, Dict, Iterable, List, Mapping, MutableMapping, Optional, Tuple, Type, Union
 
 from django.core.exceptions import ImproperlyConfigured
 from django.forms.widgets import Media, MediaDefiningClass
@@ -41,14 +29,13 @@ from django_components.logger import logger
 from django_components.middleware import is_dependency_middleware_active
 from django_components.slots import (
     DefaultFillContent,
-    NamedFillContent,
-    SlotName,
-    NamedFillNode,
     ImplicitFillNode,
+    NamedFillContent,
+    NamedFillNode,
+    SlotName,
     render_component_template_with_slots,
 )
 from django_components.utils import search
-
 
 RENDERED_COMMENT_TEMPLATE = "<!-- _RENDERED {name} -->"
 
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -17,9 +17,8 @@ from django.views import View
 # Defining them here made little sense, since 1) component_tags.py and component.py
 # rely on them equally, and 2) it made it difficult to avoid circularity in the
 # way the two modules depend on one another.
+from django_components.component_registry import registry  # NOQA
 from django_components.component_registry import AlreadyRegistered, ComponentRegistry, NotRegistered, register  # NOQA
-from django_components.component_registry import registry
-from django_components.component_registry import registry as component_registry
 from django_components.logger import logger
 from django_components.middleware import is_dependency_middleware_active
 from django_components.slots import (
@@ -317,7 +316,7 @@ class ComponentNode(Node):
 
     def render(self, context: Context) -> str:
         resolved_component_name = self.name_fexp.resolve(context)
-        component_cls: Type[Component] = component_registry.get(resolved_component_name)
+        component_cls: Type[Component] = registry.get(resolved_component_name)
 
         # Resolve FilterExpressions and Variables that were passed as args to the
         # component, then call component's context method
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -17,13 +17,8 @@ from django.views import View
 # Defining them here made little sense, since 1) component_tags.py and component.py
 # rely on them equally, and 2) it made it difficult to avoid circularity in the
 # way the two modules depend on one another.
-from django_components.component_registry import (  # NOQA
-    AlreadyRegistered,
-    ComponentRegistry,
-    NotRegistered,
-    register,
-    registry,
-)
+from django_components.component_registry import AlreadyRegistered, ComponentRegistry, NotRegistered, register  # NOQA
+from django_components.component_registry import registry
 from django_components.component_registry import registry as component_registry
 from django_components.logger import logger
 from django_components.middleware import is_dependency_middleware_active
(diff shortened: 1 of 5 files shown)


src/django_components/templatetags/component_tags.py
```diff
@@ -728,32 +375,10 @@ def parse_component_with_args(
...
-
 def is_wrapped_in_quotes(s: str) -> bool:
     return s.startswith(('"', "'")) and s[0] == s[-1]
 
 
-def is_dependency_middleware_active() -> bool:
-    return getattr(settings, "COMPONENTS", {}).get("RENDER_DEPENDENCIES", False)
-
-
-def norm_and_validate_name(name: str, tag: str, context: Optional[str] = None) -> str:
```
JuroOravec, line 745: `norm_and_validate_name` isn't being used anywhere and neither is mentioned in any discussion/issue, I so removed it.

src/django_components/templatetags/component_tags.py
```diff
@@ -728,32 +375,10 @@ def parse_component_with_args(
...
-    """Resolve FilterExpressions and Variables in context if possible.  Return other items unchanged."""
-
-    return context_item.resolve(context) if hasattr(context_item, "resolve") else context_item
-
-
 def is_wrapped_in_quotes(s: str) -> bool:
     return s.startswith(('"', "'")) and s[0] == s[-1]
 
 
-def is_dependency_middleware_active() -> bool:
```
JuroOravec, line 741: Moved `is_dependency_middleware_active` to the `middleware` file.

src/django_components/templatetags/component_tags.py
```diff
@@ -728,32 +375,10 @@ def parse_component_with_args(
     return component_name, context_args, context_kwargs
 
 
-def safe_resolve(context_item: FilterExpression, context: Context) -> Any:
```
JuroOravec, line 731: This was moved to the `component` file along with `ComponentNode`

src/django_components/templatetags/component_tags.py
```diff
@@ -123,95 +113,6 @@ def component_js_dependencies_tag(preload: str = "") -> SafeString:
         return mark_safe("\n".join(rendered_dependencies))
 
 
-class UserSlotVar:
```
JuroOravec, line 126: Slot and fill nodes were moved to `slots` file

src/django_components/templatetags/component_tags.py
```diff
@@ -27,24 +32,9 @@
 register = django.template.Library()
 
 
-RENDERED_COMMENT_TEMPLATE = "<!-- _RENDERED {name} -->"
-
 SLOT_REQUIRED_OPTION_KEYWORD = "required"
 SLOT_DEFAULT_OPTION_KEYWORD = "default"
 
-FILLED_SLOTS_CONTENT_CONTEXT_KEY = "_DJANGO_COMPONENTS_FILLED_SLOTS"
```
JuroOravec, line 35: These type aliases and constants were moved to the `slots` file

src/django_components/component.py
```diff
@@ -303,106 +289,78 @@ def _fill_slots(
...
+
+        if isinstance(self.fill_nodes, ImplicitFillNode):
+            fill_content = self.fill_nodes.nodelist
+        else:
+            fill_content = []
+            for fill_node in self.fill_nodes:
+                # Note that outer component context is used to resolve variables in
+                # fill tag.
+                resolved_name = fill_node.name_fexp.resolve(context)
+                resolved_fill_alias = fill_node.resolve_alias(context, resolved_component_name)
```
JuroOravec, line 341: The alias resolve logic from here https://github.com/EmilStenstrom/django-components/pull/416/files#diff-7345171ed38e1dd2e4b8199b65a2fd082323bdb331ff0ef0dc0dd3bd43f111aeL378 was moved to `NamedFillNode.resolve_alias`

src/django_components/slots.py
```diff
@@ -0,0 +1,467 @@
...
+        if isinstance(node, TextNode) and node.s.isspace():
+            pass
+        elif isinstance(node, CommentNode):
+            pass
+        else:
+            return True
+    return False
+
+
+def render_component_template_with_slots(
```
JuroOravec, line 340: NOTE: I've moved the last chunk of `Component.render` logic here into `render_component_template_with_slots`. The reason being that before, the flow was a bit unclear, because the template instance as owned/attached by `Component` class, but then was modified by the slot fills, and then `Component` rendered it. So there was hopping back and forth.

Now with `render_component_template_with_slots` here, `Component` just calls it and expects to get back a string. And so `Component` doesn't need to know how to slots logic is implemented.

src/django_components/slots.py
```diff
@@ -0,0 +1,467 @@
...
+from django.utils.safestring import SafeString, mark_safe
+
+FILLED_SLOTS_CONTENT_CONTEXT_KEY = "_DJANGO_COMPONENTS_FILLED_SLOTS"
+
+# Type aliases
+
+SlotName = str
+AliasName = str
+
+DefaultFillContent: TypeAlias = NodeList
```
JuroOravec, line 28: I've set this to TypeAlias because I was getting errors with Mypy

src/django_components/component.py
```diff
@@ -41,24 +24,29 @@
...
 
-_T = TypeVar("_T")
+RENDERED_COMMENT_TEMPLATE = "<!-- _RENDERED {name} -->"
 
 
 class SimplifiedInterfaceMediaDefiningClass(MediaDefiningClass):
     def __new__(mcs, name: str, bases: Tuple[Type, ...], attrs: Dict[str, Any]) -> Type:
+        # NOTE: Skip template/media file resolution when then Component class ITSELF
+        # is being created.
+        if "__module__" in attrs and attrs["__module__"] == "django_components.component":
```
JuroOravec, line 47: Genuine change - had to add this check to avoid circular imports. This check ensures that the template/media file search logic is NOT run when the `Component` class is defined below in this file.

This makes sense - we want to resolve media files only for SUBCLASSES of `Component`, not `Component` itself.

src/django_components/component.py
```diff
@@ -34,31 +17,30 @@
...
+    ImplicitFillNode,
     NamedFillContent,
+    NamedFillNode,
     SlotName,
-    SlotNode,
+    render_component_template_with_slots,
 )
 from django_components.utils import search
 
-_T = TypeVar("_T")
```
JuroOravec, line 57: `_T` wasn't used anywhere

EmilStenstrom:
This looks great! Good first step in making the project easier to work with. With these kinds of refactorings there's always a chance that someone wrote code that used the internal API that breaks, but there's no way around that. I wonder if there was a way to specify the public API for the library, so that people would have an easier way to know what the stable API is and not. Worth thinking about for 1.0.

PR #417, opened 2024-03-30
build: move to PEP517 and pyproject.toml, drop support for Py3.6 and Py3.7

GabDug:
Hey :wave: 

Following the recent packaging changes, I propose to go a step further and move to pyproject.toml, to use more modern Python packaging tooling.

This change will enable us to use any modern build-system, like recent `setuptools`, `pdm-backend` `hatchling`, which supports plugins. This could be useful to set the version automatically from the Git tag for example.

Unfortunately, Python 3.6 being no longer supported, I had to drop its support. I opened an issue #418 to discuss Python support policy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,6 @@ Django-components supports all <a href="https://docs.djangoproject.com/en/dev/fa
 
 | Python version | Django version           |
 |----------------|--------------------------|
-| 3.6            | 3.2                      |
 | 3.7            | 3.2                      |
 | 3.8            | 3.2, 4.0, 4.1, 4.2       |
 | 3.9            | 3.2, 4.0, 4.1, 4.2       |
@@ -255,7 +254,7 @@ Components can also be defined in a single file, which is useful for small compo
 ```python
 # In a file called [project root]/components/calendar.py
 from django_components import component
-from django_components import types as t 
+from django_components import types as t
 
 @component.register("calendar")
 class Calendar(component.Component):
@@ -263,16 +262,16 @@ class Calendar(component.Component):
         return {
             "date": date,
         }
-    
+
     template: t.django_html = """
         <div class="calendar-component">Today's date is <span>{{ date }}</span></div>
     """
-    
+
     css: t.css = """
         .calendar-component { width: 200px; background: pink; }
         .calendar-component span { font-weight: bold; }
     """
-    
+
     js: t.js = """
         (function(){
             if (document.querySelector(".calendar-component")) {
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -817,7 +815,6 @@ pytest
 The library is also tested across many versions of Python and Django. To run tests that way:
 
 ```sh
-pyenv install -s 3.6
 pyenv install -s 3.7
```
EmilStenstrom, line 818: 3.7 should be dropped too right?

pyproject.toml
```diff
@@ -1,3 +1,48 @@
...
+    "Framework :: Django :: 4.2",
+    "Framework :: Django :: 5.0",
+    "Operating System :: OS Independent",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
```
EmilStenstrom, line 29: Slightly strange ordering of trove classifiers?

setup.py
```diff
@@ -1,43 +1,6 @@
-# -*- coding: utf-8 -*-
-import os
+# Configuration has moved to pyproject.toml
+# This file is maintained for backward compatibility
```
EmilStenstrom, line 2: What kinds of setup is this supporting? Why can't we remove this file entirely?

scripts/supported_versions.py
```diff
@@ -169,7 +192,7 @@ def build_pyenv(python_to_django: VersionMapping):
 
 
 def build_ci_python_versions(python_to_django: Dict[str, str]):
-    # Outputs python-version: ['3.6', '3.7', '3.8', '3.9', '3.10', '3.11']
+    # Outputs python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
```
EmilStenstrom, line 195: 3.7 should be remove too right?

EmilStenstrom (commented):
Found a couple of simple fixes, but looks good! Happy to be moving towards modern python packaging. Does this change how to release the package?

pyproject.toml
GabDug: You're right, probably due to an automatic sorting on import. Thanks!

GabDug:
> Does this change how to release the package?

This does not, if you where using `python -m build` or any other PEP517 aware tool.

setup.py
GabDug: You're probably right, I think we could remove the file entirely. 

Only some old versions of pip/setuptools/etc might not support `pyproject.toml` projects correctly, but on 3.8+ users are advised to upgrade.

This should not be an issue, as we distribute universal wheels: almost no users would package from the sdist.

EmilStenstrom:
@GabDug The build process is fully automated, and is currently based on a GitHub Action that reacts of new tags being pushed. So to release I (or anyone else with commit permissions to this repo) just update the version number, tag a commit with that version number and push the tag to the repo. 

https://github.com/EmilStenstrom/django-components/blob/master/.github/workflows/publish-to-pypi.yml

GabDug:
> @GabDug The build process is fully automated, and is currently based on a GitHub Action that reacts of new tags being pushed. So to release I (or anyone else with commit permissions to this repo) just update the version number, tag a commit with that version number and push the tag to the repo.
> 
> https://github.com/EmilStenstrom/django-components/blob/master/.github/workflows/publish-to-pypi.yml

This workflow uses `python -m build` so should work as is. Only change is you now have to edit the version in `pyproject.toml`.

I haven't setup `setuptools.scm` to retrieve the version automatically from the Git tag, I think it can come in another PR.

PR #432, opened 2024-04-11
refactor: use sys.modules to get component file path

JuroOravec:
Fix the issue identified in https://github.com/EmilStenstrom/django-components/discussions/425 by using `sys.modules` to obtain the path to the file where a component class was created.

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -1,5 +1,6 @@
 import inspect
 import os
+import sys
 from pathlib import Path
 from typing import Any, ClassVar, Dict, Iterable, List, Mapping, MutableMapping, Optional, Tuple, Type, Union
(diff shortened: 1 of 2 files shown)


JuroOravec:
Ok I tried to write a test that's similar to the issue we've seen, so that there would be a component A whose template uses component B, whereas component B uses relative paths. But when I checked by undoing the changes of this MR, then the tests still passed. So the added tests doesn't really test against this specific issue. But I've included it anyway as it could catch some edge cases in the future.

JuroOravec:
And thanks for the review!

EmilStenstrom:
Merged and released in 0.65: https://pypi.org/project/django-components/0.65/

PR #434, opened 2024-04-12
fix: remove components from global scope module

### LINKED ISSUE #431: python file names colide with global scope variables/modules
### Discussed in https://github.com/EmilStenstrom/django-components/discussions/430

<div type='discussions-op-text'>

<sup>Originally posted by **franciscobmacedo** April 11, 2024</sup>
# The problem
If a component directory has python files in it that are named after some global scope variable name or module of another library, it overrides that variable. This causes a big problem: I can't name my python files after any global scope variable in my project.

# Example
I have a simple component called "select". the file structure is something like this:

```
 sampleproject/
    ├── components/             
    │   └── select/           
    │       ├── select.py     # <- the problem
    │       └── template.html 
    ├── sampleproject/
    ├── manage.py
    └── requirements.txt
```

However, I also use celery, which uses the [select I/O module from the standard library](https://docs.python.org/3/library/select.html). When launching a celery task, it was causing this error:

```shell
AttributeError: module 'select' has no attribute 'select'
```

Mostly because in my component `select.py` file I don't really have anything named `select`.  For testing purposes I added an empty "django.py" file in that directory, and sure enough the same problem happens:

```
ImportError: cannot import name 'VERSION' from 'django' (/path/to/project/components/select/django.py)
```

**PS:** The variable names inside the python file and the registered component name don't cause any issues. It's only the python file(s) name(s) inside a component directory.

# Current solution

If I set the `COMPONENTS` settings variable to disable autodiscover, this is all fixed:

```
COMPONENTS = {
    "libraries": [
        "components.select",
    ],
     "autodiscover": False,
}
```

# Proposed ideas
I understand the problem lies somewhere in the [import_file function](https://github.com/EmilStenstrom/django-components/blob/master/src/django_components/__init__.py#L39), where each component is registered as an independent module and therefore overriding already existing ones.

I'm not adding this problem as an issue because there's a solution for it. However, I believe this is a limitation: I think one should be able to name a component python file "select.py" and not worry about repercussions. Or, at least, maybe this could be more explicit in the README.md?

I don't want by all means complain, I love this project, so thank you very much for it!! Just thought this can be a problem for others as well.

Thoughts?</div>

franciscobmacedo:
fixes #431 

- removed module name registering from global scope
- changed the way components are hashed so it doesn't depend on the global scope module name

### DIFF
--- a/src/django_components/__init__.py
+++ b/src/django_components/__init__.py
@@ -1,6 +1,5 @@
 import importlib
 import importlib.util
-import sys
 from pathlib import Path
 from typing import Union
 
@@ -36,5 +35,4 @@ def import_file(path: Union[str, Path]) -> None:
     if spec is None:
         raise ValueError(f"Cannot import file '{path}' - invalid path")
     module = importlib.util.module_from_spec(spec)
-    sys.modules[spec.name] = module
     spec.loader.exec_module(module)  # type: ignore
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -1,4 +1,3 @@
-import inspect
 import os
 import sys
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


franciscobmacedo:
Still need to fix subcomponents and components as views

PR #435, opened 2024-04-13
refactor: fix clash with autodiscovery

JuroOravec:
Fixes https://github.com/EmilStenstrom/django-components/issues/431

This was a lot of trial and error, but it seems that by using `importlib.import_module(module_name)`, we don't have to interact with the `sys.modules`. Instead, we just pass it a python import like `"myapp.components.calendar"`. And so, there are no conflcts with 3rd party libraries anymore.

I wonder what would be a good way to test this:
- I tried adding a `django.py` to one of the `test_structures` (e.g. `tests/test_structures/test_structure_1/components/django/django.py`), but it didn't get picked up.
- I checked if things worked or not by having a `django.py` component in the sampleproject (`sampleproject/components/django/django.py`). That worked, but don't know if we want to use the demo project for testing.
- I was thinking that maybe we could have a full django project for end-to-end tests? But didn't want to tinker with that as part of this issue. Maybe we could add that to the milesone for v1? What do you think @EmilStenstrom?

### DIFF
--- a/src/django_components/__init__.py
+++ b/src/django_components/__init__.py
@@ -50,7 +50,6 @@ def _filepath_to_python_module(file_path: Path) -> str:
     # Get the root dir, see https://stackoverflow.com/a/16413955/9788634
     project_root = os.path.abspath(os.path.dirname(__name__))
 
-
     rel_path = os.path.relpath(file_path, start=project_root)
     rel_path_without_suffix = str(Path(rel_path).with_suffix(""))
 
--- a/src/django_components/__init__.py
+++ b/src/django_components/__init__.py
@@ -22,9 +22,7 @@ def autodiscover() -> None:
 
         # Autodetect a <component>.py file in a components dir
         component_filepaths = search(search_glob="**/*.py").matched_files
-        logger.debug(
-            f"Autodiscover found {len(component_filepaths)} files in component directories."
-        )
+        logger.debug(f"Autodiscover found {len(component_filepaths)} files in component directories.")
 
         for path in component_filepaths:
             # This imports the file and runs it's code. So if the file defines any
--- a/src/django_components/__init__.py
+++ b/src/django_components/__init__.py
@@ -41,7 +41,7 @@ def import_file(file_path: Path) -> None:
     rel_path = os.path.relpath(file_path, start=project_root)
     rel_path_without_suffix = str(Path(rel_path).with_suffix(""))
     module_name = rel_path_without_suffix.replace(os.sep, ".")
-    
+
     # This imports the file and runs it's code. So if the file defines any
     # django components, they will be registered.
     importlib.import_module(module_name)
--- a/src/django_components/__init__.py
+++ b/src/django_components/__init__.py
@@ -1,8 +1,7 @@
 import importlib
 import importlib.util
-import sys
+import os
 from pathlib import Path
-from typing import Union
 
 import django
 from django.utils.module_loading import autodiscover_modules
@@ -21,7 +20,7 @@ def autodiscover() -> None:
         autodiscover_modules("components")
 
         # Autodetect a <component>.py file in a components dir
-        component_filepaths = search(search_glob="**/*.py")
+        component_filepaths = search(search_glob="**/*.py").matched_files
         for path in component_filepaths:
             import_file(path)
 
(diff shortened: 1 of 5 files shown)


src/django_components/__init__.py
```diff
@@ -21,20 +20,28 @@ def autodiscover() -> None:
...
+
+    # Derive python import path from the filesystem path
+    # Example:
+    # If project root is `/path/to/project`
+    # And file_path is `/path/to/project/app/components/mycomp.py`
+    # Then the path relative to project root is `app/components/mycomp.py`
+    # Which we then turn into python import path `app.components.mycomp`
+    rel_path = os.path.relpath(file_path, start=project_root)
+    rel_path_without_suffix = str(Path(rel_path).with_suffix(""))
+    module_name = rel_path_without_suffix.replace(os.sep, ".")
```
EmilStenstrom, line 43: A couple of tests for this logic would be nice I think. You pretty much already have the testcases there in the file. I see support for non-linux OS:es, that could be in the unit test as well.

JuroOravec: Good point with testing the `import_file` logic. It wasn't as straightforward:

I omitted test non-linux paths, as I was getting "internal error" when I changed the `os.path.sep` to `\`. 

Also, altho I wrote the test for the file import, it works when ran alone, but some other tests are failing when it's all ran in a single run. So I've [marked it as skipped](https://github.com/EmilStenstrom/django-components/pull/435/files#diff-6faab92c5a5f55ed5c40230f2e1878cfe9603debd5af3d9b90c4ee875d5c2e93R129), and instead created a follow-up ticket (https://github.com/EmilStenstrom/django-components/issues/436) regarding tests isolation.

So the only test that actually runs is a test that raises error if the path is invalid 😅

EmilStenstrom:
```python
def autodiscover() -> None:
    from django_components.app_settings import app_settings
    if app_settings.AUTODISCOVER:
        autodiscover_modules("components")

        component_filepaths = search(search_glob="**/*.py").matched_files
        for path in component_filepaths:
            path_lib = path_to_path_lib(path)  # <- THIS
            importlib.import_module(path_lib)  # <- THIS

    for path_lib in app_settings.LIBRARIES:
        importlib.import_module(path_lib)
```

I had a even simpler idea. What about breaking those three lines out from import_file, and then add a couple of simple unit tests for `path_to_path_lib`?

I think you can mock os.path.sep like this:

```python
with mock.patch("os.path.sep", new="//"):
    print(os.path.sep)
```

JuroOravec: Thanks, that makes it clearer! Yup, changing the `import_file` into a simple transformation function made things much simpler. Plus I added logger calls.

Also I noticed that previously I was using `os.sep` instead of `os.path.sep`, not sure if that had any effect. Anyway, let me know what you think 🙌

tests/test_autodiscover.py
```diff
@@ -25,11 +26,17 @@ def tearDown(self) -> None:
...
+        all_components_before = component_registry.registry.all().copy()
+
         try:
             autodiscover()
         except component.AlreadyRegistered:
             self.fail("Autodiscover should not raise AlreadyRegistered exception")
 
+        all_components_after = component_registry.registry.all().copy()
+        imported_components_count = len(all_components_after) - len(all_components_before)
+        self.assertEqual(imported_components_count, 1)
```
EmilStenstrom, line 38: Haha, excellent workaround! :)

src/django_components/__init__.py
```diff
@@ -21,20 +21,39 @@ def autodiscover() -> None:
...
+    Derive python import path from the filesystem path.
+
+    Example:
+    - If project root is `/path/to/project`
+    - And file_path is `/path/to/project/app/components/mycomp.py`
+    - Then the path relative to project root is `app/components/mycomp.py`
+    - Which we then turn into python import path `app.components.mycomp`
+    """
+    # Get the root dir, see https://stackoverflow.com/a/16413955/9788634
+    project_root = os.path.abspath(os.path.dirname(__name__))
```
EmilStenstrom, line 49: In a couple of places in the project we allow the user to set the BASE_DIR variable. Is this what you are trying to find here? Maybe we should use BASE_DIR if it's defined and fall back to this otherwise? https://github.com/search?q=repo%3AEmilStenstrom%2Fdjango-components%20BASE_DIR&type=code

tests/test_autodiscover.py
```diff
@@ -116,3 +123,38 @@ def test_base_dir(self):
...
+class TestFilepathToPythonModule(SimpleTestCase):
+    def test_prepares_path(self):
+        self.assertEqual(
+            _filepath_to_python_module(Path("tests.py")),
+            "tests",
+        )
+        self.assertEqual(
+            _filepath_to_python_module(Path("tests/components/relative_file/relative_file.py")),
+            "tests.components.relative_file.relative_file",
+        )
```
EmilStenstrom, line 137: What happens with trickier file names? Can they be imported even though they are not valid modules?

* relative-file.py
* relative.file.py
* _file.py
* .python-version

Not sure how exhaustive we should be here, but just worth thinking about?

EmilStenstrom (commented):
Really like this! It solves a tricky problem in a nice way. Left a few thoughts as separate comments.

tests/test_autodiscover.py
JuroOravec: Very good question, IMO the paths should be valid python module names so as per [PEP-8](https://peps.python.org/pep-0008/#package-and-module-names):

> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

A few relevant discussions I've found:
- https://stackoverflow.com/questions/711884/python-naming-conventions-for-modules
- https://stackoverflow.com/questions/62878577/can-module-names-in-python-end-with-an-underscore
- https://stackoverflow.com/questions/8350853/how-to-import-module-when-module-name-has-a-dash-or-hyphen-in-it

So I'd go with the direction of documenting this restriction, rather than handling all possible variants.

src/django_components/__init__.py
JuroOravec: Yes, that's it :) Initially I avoided using BASE_DIR, because I was unfamiliar with it, and when I looked it up, it looked  like it was not a Django setting per se, but rather a convention, so I wasn't sure how widely it is used. Your suggestion with using BASE_DIR and falling back to the default above sounds like the best approach.

JuroOravec: Updated in https://github.com/EmilStenstrom/django-components/pull/435/commits/9104a439445ed074037049627bb8cd51e978d959 to use BASE_DIR

tests/test_autodiscover.py
JuroOravec: I've added a section on autodiscovery to README, along with a note on valid module names. See https://github.com/EmilStenstrom/django-components/pull/435/commits/e4950b79f78513877d4655bf410195f3e7e9ab73

EmilStenstrom:
Released as part of [0.66](https://pypi.org/project/django-components/0.66/), great work @JuroOravec!

PR #437, opened 2024-04-15
feat: slot context resolution + slot fill mgmt refactor

JuroOravec:
First stab at https://github.com/EmilStenstrom/django-components/issues/350.

Unlike our discussion, I made the slot resolution configurable via `"slot_context_behavior"` key, with the options below:
- `"allow_override"` - The behavior how it was until now. Context variables defined in deeper components overshadows outer variables.
- `"prefer_root"` - The new default we've discussed in https://github.com/EmilStenstrom/django-components/issues/350.  Variables defined at the "root" context ("root context" being the top-most render context) overshadow everything else, but otherwise it's the same as `"allow_override"`. So if a variable is defined in the root context, then it's taken from there, and otherwise it's taken from the deepest context.
- `"isolated"` - Context variables are taken solely from the root context, and other components CANNOT override these.

The reason why I went for configurable behavior is that
1) What we agreed on in https://github.com/EmilStenstrom/django-components/issues/350 was the `prefer_root` option, 2) I don't remember the exact link, but when at one point I was going through older discussions in this project, I got a feeling that there was 1-2 tickets for changing the slot behavior akin to `prefer_root` / `isolated`, and one for `allow_override`.
3) Since I was implementing this, and the idea of supporting multiple "modes" was on the table, I'd like to have the option to have complete isolation, hence the `isolated` option.

I've included description and examples of each type in [the settings](https://github.com/EmilStenstrom/django-components/pull/437/files#diff-34a53dba485852aed0d1fc1961917753118ecc45b88f25cae04b8a03e8770b22R12).

I also did a bit of internal refactoring. See https://github.com/EmilStenstrom/django-components/pull/437/commits/2a8005c4d4fb16ae92679ad6997d0363e3ba37fe and https://github.com/EmilStenstrom/django-components/pull/437/commits/696e5c48d1a24a5e526e282e7b28dea015cd440e.

I also refactored the `FillNode` classes, merging `ImplicitFillNode`, `NamedFillNode` and `BaseFillNode`. Because instead of `ImplicitFillNode`, we could simply have a `NamedFillNode` with `is_implicit: True` (See https://github.com/EmilStenstrom/django-components/pull/437/commits/c66437792225058407a48146d7747845303f9205)

TODO:
- [X] Check that each mode is working well
- [X] Tests
- [X] Update README settings section

### UPDATE 2024-04-16

Ooof, the refactor got out of hand 😅
1. I had troubles debugging the flow, so I refactored how we manage the slot fills
	- Previously, we had a `ChainMap` under the `"_DJANGO_COMPONENTS_FILLED_SLOTS"` key of the `Context`. I had 2 issues with this:
		- Django's `Context` object is already a stack, meaning we can pop/push layers. But also `ChainMap` was a stack. It wasn't easy for me to understand how to update the `ChainMap` in relation to the `Context` (AKA when to pop/push what).
			- So to simplify it, I removed the `ChainMap`, and instead all slot fills now live directly on the `Context` stack. So now there is only one object that manages the stacks (`Context`).
  		- Slot fills stored on the `ChainMap` were using the `Template` instance as dictionary keys. My problem with this was that  it was practically impossible to make sense of it using terminal. Also, my understanding is that the `Template` instances are cached/reused by Django. Yet, in the code we modified the Nodes in the `Template.nodelists`. So on one hand we used something that's supposed to be immutable (dict key), yet there were some parts of it that we DID change (underlying Nodes). So I wanted to simplify this, so it's clear when something IS or IS NOT immutable.
	  		- So instead of dict keys, we now:
			  	1. Create unique IDs for the Nodes we create (`ComponentNode`, `SlotNode`, `FillNode`)
			  	2. We use and pass through the component's ID (`component_id`) to tie together the components, fill nodes and slots.
			  	3. We use the `component_id` instead of the `Template` instance in the dicts.
2. During the refactor, at one point everything was working, except for looping logic. I had hard time troubleshooting this, so I've inserted log statements to almost all important parts of the flow, to be able to go through the logs step-by-step.
	- I thought it might be useful to keep it in, so toubleshooting is easier in the future. So I cleaned it up and added it in https://github.com/EmilStenstrom/django-components/pull/437/commits/27c996954d7de255a6f382a2613cb98d0fe026d2

~What's still remaining is to update the README. Otherwise, both tests and the `sampleproject` are passing on my end.~

Here's example how the logs looks like with the tracing logs and node IDs:

Screenshot 2024-04-16 at 15 28 43


sampleproject/sampleproject/settings.py
```diff
@@ -80,6 +80,14 @@
 
 WSGI_APPLICATION = "sampleproject.wsgi.application"
 
+# COMPONENTS = {
```
JuroOravec, line 83: Added example of the django-components settings into the sampleproject to help people get started

src/django_components/app_settings.py
```diff
@@ -9,25 +9,118 @@ class ContextBehavior(str, Enum):
...
+    Then the rendered `{{ my_var }}` will equal `123`.
+    """
+
+
 class AppSettings:
-    def __init__(self) -> None:
-        self.settings = getattr(settings, "COMPONENTS", {})
+    @property
+    def settings(self) -> Dict:
+        return getattr(settings, "COMPONENTS", {})
```
JuroOravec, line 107: Also, I noticed that changes to the Django config via `@override_settings` didn't work in tests. The reason was because `AppSettings` only worked with the settings as they were at startup. But for tests, this needed to account for changes in the settings over time.

So changed the settings so that the value is always taken from the "current" settings.

src/django_components/slots.py
```diff
@@ -96,56 +109,83 @@ def __repr__(self) -> str:
...
-    def __repr__(self) -> str:
-        raise NotImplementedError
+        used_ctx = self.resolve_slot_context(context)
+        with used_ctx.update(extra_context):
+            output = nodelist.render(used_ctx)
+
+        trace_msg("RENDR", "SLOT", self.name, self.node_id, component_id=component_id, msg="...Done!")
+        return output
+
+    def resolve_slot_context(self, context: Context) -> Context:
```
JuroOravec, line 141: This `resolve_slot_context` is where we configure the slot context resolution. Basically, this is what the original MR was supposed to be about.

tests/test_context.py
```diff
@@ -385,7 +386,21 @@ def test_component_excludes_variable_with_isolated_context_from_settings(
...
+            "{% load component_tags %}{% component_dependencies %}"
+            "{% component 'outer_context_component' only %}{% endcomponent %}"
+        )
+        rendered = template.render(Context({"variable": "outer_value"})).strip()
+        self.assertIn("outer_value", rendered, rendered)
+
+    @override_settings(
+        COMPONENTS={"context_behavior": "isolated"},
+    )
+    def test_outer_context_property_with_component_isolated(self):
```
JuroOravec, line 403: Added test to cover both options for `context_behavior`

tests/test_component.py
```diff
@@ -482,3 +536,153 @@ def test_instances_of_component_do_not_share_slots(self):
             </custom-template>
         """,
         )
+
+
+class SlotBehaviorTests(SimpleTestCase):
```
JuroOravec, line 541: Tests to cover all modes for the `slot_context_behavior`

tests/test_component.py
```diff
@@ -204,6 +204,60 @@ def test_component_with_relative_paths_as_subcomponent(
 
         self.assertIn('<input type="text" name="variable" value="hello">', rendered, rendered)
 
+    def test_component_inside_slot(self):
```
JuroOravec, line 207: While we do have components inside slot fills in some other tests, I didnt't find any test that'd test explicitly for that, so added this one.

tests/test_templatetags.py
```diff
@@ -1320,8 +1320,8 @@ def test_inner_slot_iteration_nested(self):
         component.registry.register("slot_in_a_loop", self.ComponentSimpleSlotInALoop)
 
         objects = [
-            {"inner": ["OBJECT1_ITER1", "OBJECT2_ITER1"]},
-            {"inner": ["OBJECT1_ITER2", "OBJECT2_ITER2"]},
+            {"inner": ["ITER1_OBJ1", "ITER1_OBJ2"]},
```
JuroOravec, line 1323: Just renamed these, so the loop part (`ITER`) is first, so it's easier to skim through.

src/django_components/slots.py
```diff
@@ -52,36 +55,46 @@ def default(self) -> str:
         return mark_safe(self._slot.nodelist.render(self._context))
 
 
-class TemplateAwareNodeMixin:
-    _template: Template
+class ComponentIdMixin:
```
JuroOravec, line 58: Still the same mixin, except we now use component ID instead of Template instance

src/django_components/slots.py
```diff
@@ -96,56 +109,83 @@ def __repr__(self) -> str:
...
+            raise ValueError(f"Unknown value for SLOT_CONTEXT_BEHAVIOR: '{app_settings.SLOT_CONTEXT_BEHAVIOR}'")
 
-    def render(self, context: Context) -> str:
-        raise TemplateSyntaxError(
-            "{% fill ... %} block cannot be rendered directly. "
-            "You are probably seeing this because you have used one outside "
-            "a {% component %} context."
-        )
 
+class FillNode(Node, ComponentIdMixin):
```
JuroOravec, line 161: Merged `BaseFillNode`, `NamedFillNode` and `ImplicitFillNode` into `FillNode`

src/django_components/slots.py
```diff
@@ -303,114 +328,119 @@ def _try_parse_as_named_fill_tag_set(
...
             # Stop searching here, as fill tags are permitted inside component blocks
             # embedded within a default fill node.
             continue
         for nodelist_attr_name in node.child_nodelists:
             nodes_stack.extend(getattr(node, nodelist_attr_name, []))
     else:
-        return ImplicitFillNode(nodelist=nodelist)
-
-
-def _block_has_content(nodelist: NodeList) -> bool:
```
JuroOravec, line 329: Moved `_block_has_content` to `node.py`, since there is now a couple of helper functions all related to working with `Node` or `Nodelists`

tests/test_context.py
dylanjcastillo: If a `context_behavior="isolated"` wouldn't that mean that in this case it shouldn't display `outer_value`?

src/django_components/app_settings.py
dylanjcastillo: I've had the same issue before. This is great!

tests/test_context.py
```diff
@@ -65,7 +66,7 @@ class OuterContextComponent(component.Component):
     template_name = "simple_template.html"
 
     def get_context_data(self):
-        return self.outer_context
+        return self.outer_context.flatten()
```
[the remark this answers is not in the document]
dylanjcastillo: Would it be because `get_context_data()` outpus a dictionary, and outer_contex is a `Context`? I think there was a similar issue you solved before.

src/django_components/component.py
```diff
@@ -255,10 +270,21 @@ def render(
         context = context_data if isinstance(context_data, Context) else Context(context_data)
         template = self.get_template(context)
 
+        # Associate the slots with this component for this context
```
dylanjcastillo, line 273: Not really related to this PR, but I noticed that the types on render and render_to_response are wrong.

I think `context_data` should be `Dict[str, Any] | Context`

tests/test_context.py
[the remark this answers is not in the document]
JuroOravec: Oh yeah, good point, it could be connected to how we [coerce to Context in render](https://github.com/JuroOravec/django-components/blob/4ffccdf5dd9ab7b0c5d34c29928ed45efe55db65/src/django_components/component.py#L270)

tests/test_context.py
JuroOravec: Good catch! So in this particular test, the outer context does not leak into `outer_context_component` component. But the component itself manually [accesses the outer context via `self.outer_context` in `get_context_data`](https://github.com/EmilStenstrom/django-components/pull/437/files/4ffccdf5dd9ab7b0c5d34c29928ed45efe55db65#diff-0739130667ad94af9f1ac9b3ab331c42b6e4c5b0fe7bc42a3dbb47f900b7cfcaR69).

My intuition says that the `outer_context` should be available even if the templates are isolated, since `outer_context` could have also other uses. But do correct me if that's not how to feature was intented to work.

tests/test_context.py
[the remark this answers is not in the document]
dylanjcastillo: Or maybe this [one](https://github.com/JuroOravec/django-components/blob/4ffccdf5dd9ab7b0c5d34c29928ed45efe55db65/src/django_components/component.py#L379) gets called somewhere in the process?

src/django_components/component.py
JuroOravec: Updated!

tests/test_context.py
dylanjcastillo: Ah you’re right! Makes sense. 

This is very thoroughly tested! 👍

src/django_components/logger.py
```diff
@@ -1,3 +1,35 @@
...
+        if not component_id:
+            raise ValueError("component_id must be set for the RENDER action")
+        msg_prefix = f"FOR COMP {component_id}"
+
+    msg_parts = [f"{action} {node_type} {node_name} ID {node_id}", *([msg_prefix] if msg_prefix else []), msg]
+    full_msg = " ".join(msg_parts)
+
+    # NOTE: When debugging tests during development, it may be easier to change
+    # this to `print()`
+    logger.debug(full_msg)
```
EmilStenstrom, line 35: Not sure about this. I can see how debuging paths could be useful for a developer, but this is for us right? Since this will likely output a lot of text, it might make the other debug statement less useful. Should be change the other ones to info instead? Info = For developers, Debug = For us?

src/django_components/context.py
```diff
@@ -0,0 +1,121 @@
+"""
+This file cetralizes various ways we use Django's Context class
```
EmilStenstrom, line 2: "cetralizes" -> "centralizes"

tests/test_component.py
```diff
@@ -208,7 +208,7 @@ def test_component_inside_slot(self):
         class SlottedComponent(component.Component):
             template_name = "slotted_template.html"
 
-            def get_context_data(self, name: str | None = None) -> component.Dict[str, component.Any]:
+            def get_context_data(self, name: component.Optional[str] = None) -> component.Dict[str, component.Any]:
```
EmilStenstrom, line 211: I think we should import Optional from types here instead of reusing it from another file.

src/django_components/slots.py
```diff
@@ -6,7 +6,7 @@
 if sys.version_info[:2] < (3, 10):
     from typing_extensions import TypeAlias
 else:
-    from typing import TypeAlias
+    from typing import TypeAlias  # noqa # TODO: Is this required?
```
EmilStenstrom, line 9: If it's not used in the same file, you can report it entirely.

README.md
```diff
@@ -704,6 +706,27 @@ COMPONENTS = {
...
+
+What variables should be available from inside a component slot?
+
+By default, variables inside component slots are preferentially taken from the root context.
+This is similar to [how Vue renders slots](https://vuejs.org/guide/components/slots.html#render-scope),
+except that, if variable is not found in the root, then the surrounding context is searched too.
+
+You can change this with the `slot_contet_behavior` setting. Options are:
+- `"prefer_root"` - Default - as described above
+- `"isolated"` - Same behavior as Vue - variable is taken ONLY from the root context
```
EmilStenstrom, line 719: "variable is" -> "variables are"

EmilStenstrom (commented):
This is awesome as always @JuroOravec! I've left a couple of small comments, but can't see any problems with this. Again, hard to know for sure if things will break when the change is this big. But lets go, and fix any problems that occur.

src/django_components/logger.py
JuroOravec: Yeah, this is meant for us. I wouldn't want to bump up current debug to info, since debug level currently includes the info about which template names were matched against which files, and that too can be fairly verbose if there's a lot of files.

The `logger.debug` would've ideally been a TRACE level, but didn't search before if it's possible.

But I did now!

Supposedly something like this should do the trick:
```py
import logging

TRACE_LEVEL_NUM = 5  # Setting it lower than DEBUG which is 10

def trace(self, message, *args, **kws):
    if self.isEnabledFor(TRACE_LEVEL_NUM):
        self._log(TRACE_LEVEL_NUM, message, args, **kws)

logging.Logger.trace = trace
logging.addLevelName(TRACE_LEVEL_NUM, "TRACE")
```

So what we could have is:
- `trace` (LOG LEVEL 5) - This would be for us. This is where we would output the `trace_msg`.
- `debug` (LOG LEVEL 10) - This is for low-level or verbose output for users. This would contain info about template name resolution, as it does now.
- `info` (LOG LEVEL 20) - Regular messages for users. We don't use it currently.
-  We don't currently use others like `warn`, `error`, `critical` 

Lastly, in the snippet above, the bit where it does
```py
logging.Logger.trace = trace
logging.addLevelName(TRACE_LEVEL_NUM, "TRACE")
```

I assume we should probably first check if the "TRACE" level was already defined, and define it only it has not been yet.

EmilStenstrom: Sounds like a good plan to me! :)

src/django_components/slots.py
```diff
@@ -286,7 +286,7 @@ def parse_slot_fill_nodes_from_component_nodelist(
 def _try_parse_as_named_fill_tag_set(
     nodelist: NodeList,
     ComponentNodeCls: Type[Node],
-) -> Optional[Iterable[NamedFillNode]]:
+) -> Sequence[NamedFillNode]:
```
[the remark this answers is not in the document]
JuroOravec: Updated in https://github.com/EmilStenstrom/django-components/pull/437/commits/3e75db59aefee6424ded444d656f0536d063c3fe!

src/django_components/component.py
```diff
@@ -186,7 +186,7 @@ def __init__(
         self,
         registered_name: Optional[str] = None,
         outer_context: Optional[Context] = None,
-        fill_content: Union[DefaultFillContent, Iterable[NamedFillContent]] = (),  # type: ignore
+        fill_content: dict[str, FillContent] = {},  # type: ignore
```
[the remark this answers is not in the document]
JuroOravec: Yup, updated in https://github.com/EmilStenstrom/django-components/pull/437/commits/091a692993692a601796c219b0028158d9506770

src/django_components/component.py
```diff
@@ -322,7 +322,7 @@ def render(self, context: Context) -> str:
         # so it can be used by nested Slots.
         root_ctx_already_defined = OUTER_CONTEXT_CONTEXT_KEY in context
         if not root_ctx_already_defined:
-            context.push({ OUTER_CONTEXT_CONTEXT_KEY: context.__copy__() })
+            context.push({OUTER_CONTEXT_CONTEXT_KEY: context.__copy__()})
```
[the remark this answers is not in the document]
JuroOravec: Agree, I wasn't sure how the `copy()` func is used, because at first I thought it was a global, but there was no global `copy`. Updated in https://github.com/EmilStenstrom/django-components/pull/437/commits/691b663ed5df929d686f346914bf5feee67f6ff0!

sampleproject/sampleproject/settings.py
```diff
@@ -80,6 +80,14 @@
...
+#    "autodiscover": True,
+#    "libraries": [],
+#    "template_cache_size": 128,
+#    "context_behavior": "isolated",  # "global" | "isolated"
```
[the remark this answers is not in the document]
JuroOravec: Yes, great catch!🙏 Updated in https://github.com/EmilStenstrom/django-components/pull/437/commits/1b6cc7c1926f988336443d8dae2e14d3fadf5144

src/django_components/context.py
JuroOravec: Thanks! Typos fixed in https://github.com/EmilStenstrom/django-components/pull/437/commits/8b8121bde2dd306e2eeed3e5d3ed159d0c666014

tests/test_component.py
JuroOravec: Thanks for spotting that! Updated in https://github.com/EmilStenstrom/django-components/pull/437/commits/bacd683c44e3e476097850be5c8f3967d25a275e

JuroOravec:
Thanks again, really appeciate the feedback from both of you so far @EmilStenstrom @dylanjcastillo!

I rebased, and fixed/resolved the small points. I'll come back to the point on trace logging tomorrow/Thu.

src/django_components/logger.py
JuroOravec: Implemented in https://github.com/EmilStenstrom/django-components/pull/437/commits/2c451693abd886cee180828513fead5726c51ad5 🎉 

Also in the end I omitted the following line as it felt a bit too invasive.
```py
logging.Logger.trace = trace
```

dylanjcastillo (approved):

dylanjcastillo:
I just approved it!

EmilStenstrom:
@JuroOravec Wanna merge and do a release? :)

JuroOravec:
Thanks both, merged!

@EmilStenstrom Here's MR for the release https://github.com/EmilStenstrom/django-components/pull/442. Not sure if I can push tags. Or can I use the github UI for that?

EmilStenstrom:
Awesome!

Issue #436, opened 2024-04-14
Tests: Component registry is not isolated across tests


JuroOravec:
I had difficulty creating tests for https://github.com/EmilStenstrom/django-components/pull/435. Things worked when I ran tests just for the `import_file` function, but I was seeing errors in different files and tests when I ran the whole suite of tests. [See this comment and test for context](https://github.com/EmilStenstrom/django-components/pull/435/files#diff-6faab92c5a5f55ed5c40230f2e1878cfe9603debd5af3d9b90c4ee875d5c2e93R129). It looks to me like the state of either/both `component_registry` and `sys.modules` (imported modules) persists across tests.

Ideally, both `sys.modules` and `component_registry` would be isolated between tests.

Don't have deep knowledge about tox / pytest, so raising a ticket to discuss this, hoping someone more knowledgeable answers 😄 

E.g., for `component_registry`, should we use a mock for the registry? Or run `registry.clear()` before each test? Or is there a simpler approach?

As for `sys.modules`, is it possible to restart/clear the modules that have been imported as part of the test run?

Also, component registrations [like these in test_context](https://github.com/EmilStenstrom/django-components/blob/master/tests/test_context.py#L71) should probably also be scoped not globally, but per test.

Lastly, related to tests, @dylanjcastillo, I like how the tests you wrote in https://github.com/EmilStenstrom/django-components/blob/edf3885632ff32406510693d63cfbddf1c71faa6/tests/test_component_as_view.py. But I can't wrap my head around how it works. How are the routes set there? Is it by [setting the ROOT_URLCONF](https://github.com/EmilStenstrom/django-components/blob/edf3885632ff32406510693d63cfbddf1c71faa6/tests/test_component_as_view.py#L108)?

EmilStenstrom:
@JuroOravec For #435 I think it's enough to unit test the code that translates a path to a dotted import path. 

More generally, tests should of course not affect each other. I think simply running registry.clear() in teardown for each test class is probably a good idea. Maybe a new testcase base class that we use everywhere?

dylanjcastillo:
@JuroOravec, yes, it uses `ROOT_URLCONF=__name__`, which essentially means that it uses the `urlpatterns` defined in the same file.

Issue #418, opened 2024-03-30
Python versions support policy (drop 3.6/3.7?)


GabDug:
Hey :wave:

To use pyproject.toml packaging in #417, I had to drop Python 3.6. Here's an issue to discuss that.

## Summary

I propose to change the support Policy from "all Python versions allowed by the Django versions we target against" to "all Python versions currently receiving active or security support that are allowed by the Django versions we target against":

| Python version | Django version           |
|----------------|--------------------------|
| ~~3.6~~           | ~~3.2~~                     |
| ~~3.7~~            | ~~3.2~~                      |
| 3.8            | 3.2, 4.0, 4.1, 4.2       |
| 3.9            | 3.2, 4.0, 4.1, 4.2       |
| 3.10           | 3.2, 4.0, 4.1, 4.2, 5.0  |
| 3.11           | 4.1, 4.2, 5.0            |
| 3.12           | 4.2, 5.0                 |

> Python 3.7 could still be a target, as tooling is still more up-to-date for it.

## Proposed Change

I see that django-components still supports Python 3.6 and 3.7, released in 2018 and 2020, [which no longer receive active or security support](https://endoflife.date/python).

While I agree that a broad support of Python version is nice to have to some users, I'm questioning whether it could make maintenance and evolution of the project harder.

While [Pypistats](https://pypistats.org/packages/django-components) data should be taken with a grain of salt, we see that the number of downloads for Python 3.6 is almost 0 (2 downloads in 120 days) and quite low for 3.7. Unfortunately, we can't know if those downloads are for a current of previous version of django-components.

Removing 3.6 and 3.7 would only impact users of Django 3.2: we could argue that users of Django 3.2 that still upgrade their packages regularly have already moved to Python 3.8, 3.9 or 3.10, as a lot of packages in the Python ecosystem only support active version of Python.

This change would allow to have more modern tools and with consistent versions in the CI, as many tools dropped support for those versions.

[...]

PR #443, opened 2024-04-18
chore: Drop support for Django 3.2, 4.0, 4.1

JuroOravec:
While I was hacking away at tests for https://github.com/EmilStenstrom/django-components/issues/436, I had some strange error with Django 3.2. That got me searching, and along the way, I stumbled across the [Django download page](https://www.djangoproject.com/download/), which says that Django 3.2, 4.0, and 4.1 are no longer supported.

Support policy was recently discussed in https://github.com/EmilStenstrom/django-components/issues/418#issuecomment-2028479587, based on which I think we can drop these versions.

One other thing I noticed was that the `supported_versions` script still included v3.2-4.2 in the generated text. This is because it took the data from the table at https://docs.djangoproject.com/en/dev/faq/install/, which lists even unsupported Django versions. So instead I updated the script to fetch the data from the table at https://www.djangoproject.com/download/, which includes only supported versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,10 +144,10 @@ Django-components supports all supported combinations versions of [Django](https
 
 | Python version | Django version           |
 |----------------|--------------------------|
-| 3.8            | 3.2, 4.0, 4.1, 4.2       |
-| 3.9            | 3.2, 4.0, 4.1, 4.2       |
-| 3.10           | 3.2, 4.0, 4.1, 4.2, 5.0  |
-| 3.11           | 4.1, 4.2, 5.0            |
+| 3.8            | 4.2                      |
+| 3.9            | 4.2                      |
+| 3.10           | 4.2, 5.0                 |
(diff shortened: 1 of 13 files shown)


EmilStenstrom:
Great find! Makes sense.

EmilStenstrom (commented):
Looks good! Might be worth removing Django30CompatibleSimpleTestCase too?

JuroOravec:
@EmilStenstrom Good idea! I'll be using some TestCase subclass for https://github.com/EmilStenstrom/django-components/issues/436, so I've kept an empty subclass there and renamed it to `BaseTestCase`. See https://github.com/EmilStenstrom/django-components/pull/443/commits/eb40afba7b55a5649a380cc80ba3b99bf1b6ba16

dylanjcastillo (approved):

PR #445, opened 2024-04-19
refactor: fix slot context behavior

JuroOravec:


EmilStenstrom:
Don't forget to add a tests that catches the broken behavior!

src/django_components/component.py
```diff
@@ -380,12 +372,7 @@ def render(self, context: Context) -> str:
 
         # Prevent outer context from leaking into the template of the component
         if self.isolated_context:
-            # Even if contexts are isolated, we still need to pass down the
```
JuroOravec, line 383: Moved to own function

src/django_components/context.py
```diff
@@ -17,10 +16,50 @@
...
+    # context layer.
+    #
+    # This is necessary, because otherwise if we have a nested loop with a same
+    # component used recursively, the inner slot -> comp mapping would leak into the outer.
+    #
+    # NOTE: If you ever need to debug this, insert a print/debug statement into
+    # `django.template.defaulttags.ForNode.render` to inspect the context object
+    # inside the for loop.
+    if "forloop" in context:
+        context.dicts[-1][_SLOT_COMPONENT_ASSOC_KEY] = context[_SLOT_COMPONENT_ASSOC_KEY].copy()
```
JuroOravec, line 46: One of the issues was with a for loop. Don't know exactly why, but when there was nested loops, then, after the inner loop finished, some state 
surrounding `_SLOT_COMPONENT_ASSOC_KEY` persisted after an inner loop finished.

So what was basically happening in the bug was that if we had nested loops with slots, then after the inner loops finished, then the next iteration of the outer loop rendered a slot of the INNER loop, instead of a slot of an OUTER loop.

src/django_components/context.py
```diff
@@ -17,10 +16,50 @@
...
+
+
+def make_isolated_context_copy(context: Context) -> Context:
+    # Even if contexts are isolated, we still need to pass down the
+    # outer context so variables in slots can be rendered using
+    # the outer context.
+    root_ctx = get_outer_root_context(context)
+    slot_assoc = context.get(_SLOT_COMPONENT_ASSOC_KEY, {})
```
JuroOravec, line 54: Another issue was that the slot->component associations weren't being passed. To fixed that I reverted to store the associations on a shared dict (instead of each key living directly on the `Context`), and passing this dict through even if the contexts are to be isolated.

src/django_components/context.py
```diff
@@ -38,77 +81,82 @@ def set_slot_fill(context: Context, component_id: str, slot_name: str, value: "F
...
     Note that we make use of the fact that Django's Context is a stack - we can push and pop
     extra contexts on top others.
-
-    For the slot fills to be pushed/popped wth stack layer, they need to have keys defined
-    directly on the Context object.
     """
     trace_msg("SET", "FILL", slot_name, component_id)
-    slot_key = (_FILLED_SLOTS_CONTENT_CONTEXT_KEY, component_id, slot_name)
-    context[slot_key] = value
+    slot_key = f"{component_id}__{slot_name}"
```
JuroOravec, line 86: Changed the keys from tuples to strings so the data can be serialized to JSON, which I used during debugging

src/django_components/context.py
```diff
@@ -5,31 +5,74 @@
...
+    # metadata so variables in slots can be rendered using the correct context.
+    root_ctx = get_outer_root_context(context)
+    slot_assoc = context.get(_SLOT_COMPONENT_ASSOC_KEY, {})
+    slot_fills = context.get(_FILLED_SLOTS_CONTENT_CONTEXT_KEY, {})
+
+    context_copy = context.new()
+    context_copy[_SLOT_COMPONENT_ASSOC_KEY] = slot_assoc
+    context_copy[_FILLED_SLOTS_CONTENT_CONTEXT_KEY] = slot_fills
+    set_outer_root_context(context_copy, root_ctx)
+    copy_forloop_context(context, context_copy)
```
JuroOravec, line 62: Also had to do pass down the "outer context" and for loop info for things to work

tests/testutils.py
```diff
@@ -32,3 +33,22 @@ def create_and_process_template_response(template, context=None, use_middleware=
     else:
         response.render()
     return response.content.decode("utf-8")
+
+
+def print_nodes(nodes: List[Node], indent=0) -> None:
```
JuroOravec, line 38: Added a convenience func to print out nodeslist, which I used a lot during debugging

src/django_components/context.py
```diff
@@ -38,77 +81,82 @@ def set_slot_fill(context: Context, component_id: str, slot_name: str, value: "F
...
+    else:
+        outer_root_context = Context()
+
+    # Include the mappings.
+    if _SLOT_COMPONENT_ASSOC_KEY in context:
+        outer_root_context[_SLOT_COMPONENT_ASSOC_KEY] = context[_SLOT_COMPONENT_ASSOC_KEY]
+    if _FILLED_SLOTS_CONTENT_CONTEXT_KEY in context:
+        outer_root_context[_FILLED_SLOTS_CONTENT_CONTEXT_KEY] = context[_FILLED_SLOTS_CONTENT_CONTEXT_KEY]
+
+    context[_OUTER_ROOT_CTX_CONTEXT_KEY] = outer_root_context
```
JuroOravec, line 131: Tbh even though I wrote it, the assignments in the `context.py` file feel a bit arcane to me - in other words I haven't wrapped my head comfortably around when to use `Context.push`, when to use `Context[key]`, etc.

...But as things are now it works, and trying to delete any of the parts resulted in errors.

JuroOravec:
@EmilStenstrom I've fix the issues and and tests :)

EmilStenstrom (commented):
I'm comfortable with merging this, as these are fairly low level changes, and no high level tests are breaking. Appreciate the new tests!

(I'm still not sure it's worth it to support all the different isolation levels at the same time. Maybe we should just go with the default, remove the config and be happy? :) )

JuroOravec:
Agree that ideally there should be only one way to do it. But since personally I'd prefer the `isolated` option as the way, in the other discussion we've suggested the `prefer_root`, and users may have been used to the option that's now called `allow_override`, then I'm not confident in setting the behavior to be only a single way, without providing users a way to go back.

Maybe we could decide on a single behavior and remove the setting as part of the v1.0 release? What do you think?

PR #452, opened 2024-04-23
refactor: clear component registry between tests

JuroOravec:
Closes https://github.com/EmilStenstrom/django-components/issues/436

### DIFF
--- a/tests/test_component_as_view.py
+++ b/tests/test_component_as_view.py
@@ -161,14 +161,6 @@ class TestComponentAsView(BaseTestCase):
             response.content,
         )
 
-    def test_replace_slot_in_view_with_insecure_content(self):
-        response = self.client.get("/test_slot_insecure/")
-        self.assertEqual(response.status_code, 200)
-        self.assertNotIn(
-            b"<script>",
-            response.content,
-        )
-
     def test_replace_context_in_view_with_insecure_content(self):
         response = self.client.get("/test_slot_insecure/")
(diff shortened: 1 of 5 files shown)


tests/test_component_as_view.py
```diff
@@ -150,14 +162,6 @@ def test_replace_slot_in_view(self):
             response.content,
         )
 
-    def test_replace_slot_in_view_with_insecure_content(self):
```
dylanjcastillo, line 153: Shouldn't we still test this?

JuroOravec: Ah, forgot to attach a comment here, but this is a duplicate, it has the same function body as the function below (`test_replace_slot_in_view_with_insecure_content`)

JuroOravec: But thanks for pointing out, as I'm looking at it now, I guess the first functions probavbly should have used `/test_context_insecure/` endpoint instead?

dylanjcastillo: Yes, you're right. I think I made a typo when I wrote that 😅

JuroOravec: Haha, happens :D Reverted the deletion and fixed the test in https://github.com/EmilStenstrom/django-components/pull/452/commits/8cc12ef4488500b87dbc0f411527bcc0872e35b8 :tada:

dylanjcastillo (approved):

Issue #454, opened 2024-04-24
Bug: Context vars missing in slots when `isolated` settings


JuroOravec:
Currently when both `context_behavior` and `slot_context_behavior` are set to "isolated", then in some cases the context variables in slots are not rendered.

I already got a fix for this (https://github.com/EmilStenstrom/django-components/pull/455), but writing this down for visibility.

Issue #453, opened 2024-04-24
Nested slot fills with same name (and same component?) override outer slot fills


JuroOravec:
See the error from https://github.com/EmilStenstrom/django-components/discussions/419#discussioncomment-9209027. It seems the slot fills of inner component override the OTHER slots of the outer component.

Using the template from `slotted_template.html`:

```django
{% load component_tags %}
{% component "test" name='Igor' %}
    {% fill "header" %}  <!-- This fill is rendered correctly -->
        {% component "test" name='Joe2' %}
            {% fill "header" %} <!-- This fill is rendered correctly -->
                Name2: {{ name }}
            {% endfill %}
            {% fill "main" %}  <!-- This is rendered correctly. BUT fill is applied also to the outer component -->
                Day2: {{ day }}
            {% endfill %}
            {% fill "footer" %}  <!-- This is rendered correctly. BUT fill is applied also to the outer component -->
                XYZ
            {% endfill %}
        {% endcomponent %}
    {% endfill %}

	<!-- ERROR HERE: explicit "footer" slot fill is ignored because of the inner slot fill -->
    {% fill "footer" %}
        WWW
    {% endfill %}

    <!-- ERROR HERE: the default "main" slot fill is ignored because of the inner slot fill -->
{% endcomponent %}
```

EmilStenstrom:
Good find!

PR #455, opened 2024-04-24
refactor: fix context vars missing in isolated slot

JuroOravec:
Closes https://github.com/EmilStenstrom/django-components/issues/454


src/django_components/context.py
```diff
@@ -19,11 +20,17 @@
 _FILLED_SLOTS_CONTENT_CONTEXT_KEY = "_DJANGO_COMPONENTS_FILLED_SLOTS"
 _OUTER_ROOT_CTX_CONTEXT_KEY = "_DJANGO_COMPONENTS_OUTER_ROOT_CTX"
 _SLOT_COMPONENT_ASSOC_KEY = "_DJANGO_COMPONENTS_SLOT_COMP_ASSOC"
+_PARENT_COMP_KEY = "_DJANGO_COMPONENTS_PARENT_COMP"
+_CURRENT_COMP_KEY = "_DJANGO_COMPONENTS_CURRENT_COMP"
```
JuroOravec, line 24: On the Context object we now store the ID of the current component node that we're rendering, as well as it's parent.

The parent info is used in this MR as an indicator of whether the rendered component node is a top-level node, or it is inside another component.

If I remember right, the info on current component ID is / will be used to fix https://github.com/EmilStenstrom/django-components/issues/453

src/django_components/context.py
```diff
@@ -110,15 +133,31 @@ def set_outer_root_context(context: Context, outer_ctx: Optional[Context]) -> No
...
-        # - All later indices (2, 3, ...) are extra layers added by the rendering
-        #   logic (each Node usually adds it's own context layer)
+    # Special case for handling outer context of top-level components when
+    # slots are isolated. In such case, the entire outer context is to be the
+    # outer root ctx.
+    if (
+        outer_ctx
+        and not context.get(_PARENT_COMP_KEY)
+        and app_settings.SLOT_CONTEXT_BEHAVIOR == SlotContextBehavior.ISOLATED
+        and _OUTER_ROOT_CTX_CONTEXT_KEY in context  # <-- Added to avoid breaking tests
```
JuroOravec, line 143: As the comment says, the last line was added to avoid breaking other tests (not quite sure why)

dylanjcastillo (approved):

PR #456, opened 2024-04-25
refactor: fix slot fills for components nested in themselves

JuroOravec:
Closes https://github.com/EmilStenstrom/django-components/issues/453


src/django_components/component.py
```diff
@@ -304,8 +304,10 @@ def _fill_slots(
         """Fill component slots outside of template rendering."""
         self.fill_content = {
             slot_name: FillContent(
-                TextNode(escape(content) if escape_content else content),
-                None,
+                nodes=NodeList([
```
JuroOravec: Wrapped this one in `NodeList`

EmilStenstrom (approved):
Looks good to me!

PR #457, opened 2024-04-25
feat: Allow multiple slots with the same name in the same template

JuroOravec:
With the current implementation it's safe to allow multiple slots with same name in the same template. See this thread for context https://github.com/EmilStenstrom/django-components/discussions/419#discussioncomment-9209027.

Enabling it was easy - just removing an error path. I added a few tests, also for tricky parts like duplicate slots in loops and nested in fills.

### DIFF
--- a/src/django_components/slots.py
+++ b/src/django_components/slots.py
@@ -417,13 +417,14 @@ def _collect_slot_fills_from_component_template(
             continue
 
         slot_name = node.name
+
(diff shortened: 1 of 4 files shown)


EmilStenstrom (commented):
Let's do this! Tests look good, and I agree that this is a good addition. I remember wanting one fill, multiple slots from django blocks before, and not finding a good way to do it. Happy we'll be fixing it here.

Issue #450, opened 2024-04-23
Remove options for slot context behavior for v1.0


JuroOravec:
This is a follow up from https://github.com/EmilStenstrom/django-components/pull/445#issuecomment-2073079048.

For context, slot context settings is about how the variables inside the slot are evaluated. For more info see https://github.com/EmilStenstrom/django-components/issues/350

@EmilStenstrom suggested to have a single behavior for slot context, instead of having settings that allow you to configure how it works. (See https://github.com/EmilStenstrom/django-components/pull/445#pullrequestreview-2017902791)

As for me, my motivation to introduce the feature was to be able to use the `isolated` setting for my project (I prefer the isolation as it mimics behavior of Vue or React). I can also imagine that some people may want to set the config to `allow_override` to keep the behavior as it was before v0.67. So IMO we should provide backward compatibility until v1.0.

Thoughts?

EmilStenstrom:
I'd love to hear other thoughts here. 

My thoughts: Removing options is never something that a community will vote for, it's something that you do to remove complexity, both from the code and for users. For users that means fewer ways to use the system, which means easier troubleshooting.

So long story short: I think removing the other options might make this project easier to maintain. It's not a strong opinion though.

dylanjcastillo:
I guess people would be more used to the isolated setting given that it’s the default option in Vue/React. I also wonder how many people are consciously aware of how allow_override works. IMO feels like a behavior prone to errors.

Maybe we could change the isolated to default, add a deprecation warning for the other two until 1.0 and, if there’s no push back, deprecate them?

JuroOravec:
@dylanjcastillo Yeah, I'm thinking the same.

---

Also, as I was fixing things around in the last few days, I also got a bit of experience from user's perspective, which leads me to think that, **if `context_behavior` settings was to remain, then**:
1. `context_behavior` and `slot_context_behavior` should be merged.
	
	Originally I added `slot_context_behavior` separately, as a standalone feature. But from user's perspective, configuring these two as separate values feels like unnecessarily too much detail.

2. The `prefer_root` option of `slot_context_behavior` should be dropped.

	After all, the `prefer_root` is very close in behavior to `isolated`.

3. The `context_behavior` options could be renamed to `isolated` and `django` - `django` meaning that the templates behave like in regular django templates.

EmilStenstrom:
@JuroOravec So one option context_behavior, with two options, isolated and django, with isolated being the default. That would work for me.

[...]

Issue #280, opened 2023-05-02
Using `if_filled` with the built in `if`


timothyis:
Is it possible to use `if_filled` along with the built-in `if` that Django templates provide? 

In some of our components we offer – for example – filling a title with either a prop passed in on the component definition, or a slot fill if we need some greater flexibility. This allows us to keep complexity down when using components. It would be great to render the title either if the prop is passed or the slot is used but not have to use duplicate elements in the component definition because we need to use two conditional types.

EmilStenstrom:
Could you give an example of how you would like the code to work?

timothyis:
Yes, here is a minimal example that is based on our note component:

```django
<div class="bg-slate-100 rounded-lg p-4 flex flex-col sm:flex-row">
	{% if iconName %}
		<div class="mr-3 mt-0.5">
			{% component "icon" name=iconName category=iconCategory set=iconSet only %}
		</div>
	{% endif %}

	{% if title %}
		<div class="flex flex-col">
			 <h4 class="font-medium">{% slot "title" %}{{ title }}{% endslot %}</h4>
 	{% endif %}
			<div>
				{% slot "message" %}{{ message }}{% endslot %}
		 	</div>
	{% if title %}
		</div>
	{% endif %}
</div>
```

In this example, I'm conditionally rendering a div if there is a title because otherwise it interacts with the icon (if there is one) incorrectly, but I also render the h4 that the title goes inside of so that the styling is consistent (I've removed a lot of classes here to make it easier to read).

I'd like to conditionally render all of these where `{% if title %}` but also taking into account if the slot is filled or not.

Currently we have to say "title=True". Here's an example of the current usage:
```django
{% component "note" title="This is a title" message="This is a note" %}
```
^ With a prop (clean version)

```django
{% component_block "note" title=True message="This is a note" %}
	{% fill "title" %}
		This is a title with something dynamic {{ pageContext.item }}
	{% endfill %}
{% endcomponent_block %}
```
^ With something dynamic inside the title to avoid doing messy stuff like `title="This is a title with something dynamic"|add:pageContext.item`, which can get messy with multiple variables and placing variables in the centre or using integers etc.

We also have a dialog component where you can either pass a title or define a custom header and currently we have to duplicate the header HTML depending on if there's a "header" fill _or_ if there's a title. 

```
{% if_filled "header" %}
	<header class="w-full min-h-10 pb-2">
		{% slot "header" %}{% endslot %}
	</header>
[...]

PR #465, opened 2024-04-30
feat: merge context settings, replace if_filled tag with var

JuroOravec:
This MR:

- Removes the nodes and template tags involved in the If/Elif/Else logic (`{% if_filled %}`, etc). Instead, the context is populated with a slots dict that tracks which slots have been filled, and can be used like `{% if component_vars.is_filled.<slot_name> %}`

- Merges `context_behavior` and `slot_context_behavior` into `context_behavior`. It now has two options - `"isolated"` and `"django"` (this was previously called `global`). `isolated` is the default.

- Revamps the slot resolution logic. It has now been moved "upstream", meaning that the slot resolution is now done in `Component.render()`. And so, by the time we reach `SlotNode.render()`, we only need to look up the corresponding slot fill.

- To make it easier to onboard people to the slot resolution flow, I've documented the flow in [docs/slot_rendering.md](https://github.com/EmilStenstrom/django-components/pull/465/files#diff-32aee5c5d3d2dd970fcf0d9f694027cc890de905f65dd493a8bfc2b6d9fa7bb8)

- Updates README

- See other comments for smaller changes

This is a big MR, but also about 2/3 of it are documentation and test changes.

Closes https://github.com/EmilStenstrom/django-components/issues/280
Closes https://github.com/EmilStenstrom/django-components/issues/450

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -306,9 +306,7 @@ class Component(View, metaclass=SimplifiedInterfaceMediaDefiningClass):
         # For users, we expose boolean variables that they may check
         # to see if given slot was filled, e.g.:
         # `{% if variable > 8 and component_vars.is_filled.header %}`
-        slot_bools = {
-            slot_fill.escaped_name: slot_fill.is_filled for slot_fill in resolved_fills.values()
-        }
+        slot_bools = {slot_fill.escaped_name: slot_fill.is_filled for slot_fill in resolved_fills.values()}
 
         with context.update(
             {
@@ -318,7 +316,7 @@ class Component(View, metaclass=SimplifiedInterfaceMediaDefiningClass):
                 # See https://github.com/EmilStenstrom/django-components/issues/280#issuecomment-2081180940
                 "component_vars": {
                     "is_filled": slot_bools,
-                }
+                },
             }
         ):
             return template.render(context)
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -20,16 +20,14 @@ from django.views import View
 # way the two modules depend on one another.
 from django_components.component_registry import registry  # NOQA
 from django_components.component_registry import AlreadyRegistered, ComponentRegistry, NotRegistered, register  # NOQA
-from django_components.context import make_isolated_context_copy, prepare_context, set_slot_component_association
+from django_components.context import make_isolated_context_copy, prepare_context
 from django_components.logger import logger, trace_msg
 from django_components.middleware import is_dependency_middleware_active
-from django_components.node import walk_nodelist
 from django_components.slots import (
     DEFAULT_SLOT_KEY,
     FillContent,
     FillNode,
     SlotName,
-    SlotNode,
     render_component_template_with_slots,
 )
 from django_components.utils import gen_id, search
@@ -189,11 +187,11 @@ class Component(View, metaclass=SimplifiedInterfaceMediaDefiningClass):
         registered_name: Optional[str] = None,
         component_id: Optional[str] = None,
         outer_context: Optional[Context] = None,
-        fill_content: Dict[str, FillContent] = {},
+        fill_content: Optional[Dict[str, FillContent]] = None,
     ):
         self.registered_name: Optional[str] = registered_name
         self.outer_context: Context = outer_context or Context()
-        self.fill_content = fill_content
+        self.fill_content = fill_content or {}
         self.component_id = component_id or gen_id()
 
     def __init_subclass__(cls, **kwargs: Any) -> None:
@@ -265,15 +263,6 @@ class Component(View, metaclass=SimplifiedInterfaceMediaDefiningClass):
         context = context_data if isinstance(context_data, Context) else Context(context_data)
         prepare_context(context, component_id=self.component_id, outer_context=self.outer_context or Context())
         template = self.get_template(context)
-
-        # Associate the slots with this component for this context
-        # This allows us to look up component-specific slot fills.
-        def on_node(node: Node) -> None:
-            if isinstance(node, SlotNode):
-                trace_msg("ASSOC", "SLOT", node.name, node.node_id, component_id=self.component_id)
-                set_slot_component_association(context, node.node_id, self.component_id)
-
-        walk_nodelist(template.nodelist, on_node)
 
         if slots_data:
             self._fill_slots(slots_data, escape_slots_content)
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -20,7 +20,13 @@ from django.views import View
 # way the two modules depend on one another.
 from django_components.component_registry import registry  # NOQA
 from django_components.component_registry import AlreadyRegistered, ComponentRegistry, NotRegistered, register  # NOQA
-from django_components.context import make_isolated_context_copy, prepare_context
+from django_components.context import (
+    make_isolated_context_copy,
+    _FILLED_SLOTS_CONTENT_CONTEXT_KEY,
+    _ROOT_CTX_CONTEXT_KEY,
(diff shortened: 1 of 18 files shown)


src/django_components/component.py
```diff
@@ -254,33 +258,76 @@ def get_template(self, context: Mapping) -> Template:
             f"Note: this attribute is not required if you are overriding the class's `get_template*()` methods."
         )
 
+    def render_from_input(self, context: Context, args: Union[List, Tuple], kwargs: Dict) -> str:
```
JuroOravec, line 261: Created this function so that the logic that manages / calls `get_context_data` lives on the `Component` class. The reason for this is also that, down the line, I'd like a feature to render a `Component` class directly, without needing to define a Template, but still being able to pass in args/kwargs like in a template. And so with this function, we can do so.

src/django_components/component.py
```diff
@@ -254,33 +258,76 @@ def get_template(self, context: Mapping) -> Template:
...
 
-        # Associate the slots with this component for this context
-        # This allows us to look up component-specific slot fills.
-        def on_node(node: Node) -> None:
-            if isinstance(node, SlotNode):
-                trace_msg("ASSOC", "SLOT", node.name, node.node_id, component_id=self.component_id)
-                set_slot_component_association(context, node.node_id, self.component_id)
+        # Support passing slots explicitly to `render` method
+        if slots_data:
+            fill_content = self._fills_from_slots_data(slots_data, escape_slots_content)
```
JuroOravec, line 289: Renamed `_fill_slots` to `_fills_from_slots_data`, and also changed it so it does not change the value of `self.fill_slots`. So calling `Component.render` DOES NOT change the state of `Component` instance.

src/django_components/component.py
```diff
@@ -254,33 +258,76 @@ def get_template(self, context: Mapping) -> Template:
...
+        else:
+            fill_content = self.fill_content
+
+        # If this is top-level component and it has no parent, use outer context instead
+        if not context[_PARENT_COMP_CONTEXT_KEY]:
+            context_data = self.outer_context.flatten()
+        if context_data is None:
+            context_data = {}
+
+        slots, resolved_fills = resolve_slots(
```
JuroOravec, line 299: All slot resolution / preparation is done here.

src/django_components/component.py
```diff
@@ -254,33 +258,76 @@ def get_template(self, context: Mapping) -> Template:
...
-            self.component_id, template, context, self.fill_content, self.registered_name
-        )
+        with context.update(
+            {
+                _ROOT_CTX_CONTEXT_KEY: self.outer_context,
+                _FILLED_SLOTS_CONTENT_CONTEXT_KEY: updated_slots,
+                # NOTE: Public API for variables accessible from within a component's template
+                # See https://github.com/EmilStenstrom/django-components/issues/280#issuecomment-2081180940
+                "component_vars": {
+                    "is_filled": slot_bools,
```
JuroOravec, line 326: As noted above, this is the public API, and this where we get the `component_vars.is_filled` variable

src/django_components/component.py
```diff
@@ -254,33 +258,76 @@ def get_template(self, context: Mapping) -> Template:
...
-                trace_msg("ASSOC", "SLOT", node.name, node.node_id, component_id=self.component_id)
-                set_slot_component_association(context, node.node_id, self.component_id)
+        # Support passing slots explicitly to `render` method
+        if slots_data:
+            fill_content = self._fills_from_slots_data(slots_data, escape_slots_content)
+        else:
+            fill_content = self.fill_content
+
+        # If this is top-level component and it has no parent, use outer context instead
+        if not context[_PARENT_COMP_CONTEXT_KEY]:
```
JuroOravec, line 294: Note that we need to keep track of the current/parent component for this if/else here, so we can properly access the data from the top-level context.

src/django_components/node.py
```diff
@@ -15,13 +15,20 @@ def nodelist_has_content(nodelist: NodeList) -> bool:
     return False
 
 
-def walk_nodelist(nodes: NodeList, callback: Callable[[Node], None]) -> None:
+class NodeTraverse(NamedTuple):
+    node: Node
+    parent: Optional["NodeTraverse"]
+
+
+def walk_nodelist(nodes: NodeList, callback: Callable[[Node], Optional[str]]) -> None:
```
JuroOravec, line 23: Added a feature to `walk_nodelist`, so that we can walk UP the node tree. We use this to detect nested slots, so that we can skip rendering of inner slots if the outer slot has been overriden by a fill.

src/django_components/slots.py
```diff
@@ -1,41 +1,84 @@
...
+        {% endfill %}
+    {% endcomponent %}
+    ```
+    """
 
     nodes: NodeList
     alias: Optional[AliasName]
 
 
+class Slot(NamedTuple):
```
JuroOravec, line 47: Now we don't mutate the `SlotNode` and `FillNode` instances. Instead, we create instances of `Slot` and `SlotFill` to tell SlotNodes how they should be rendered.

src/django_components/slots.py
```diff
@@ -99,70 +116,56 @@ def __init__(
...
         return output
 
-    def resolve_slot_context(self, context: Context) -> Context:
-        """
-        Prepare the context used in a slot fill based on the settings.
-
-        See SlotContextBehavior for the description of each option.
-        """
-        root_ctx = get_outer_root_context(context) or Context()
+    def _resolve_slot_context(self, context: Context, slot_fill: "SlotFill") -> Context:
```
JuroOravec, line 152: This function implements the `context_behavior` setting.

src/django_components/slots.py
```diff
@@ -207,68 +211,6 @@ def resolve_alias(self, context: Context, component_name: Optional[str] = None)
         return resolved_alias
 
 
-class _IfSlotFilledBranchNode(Node):
```
JuroOravec, line 210: Deleted If/Elif/Else/etc nodes.

src/django_components/slots.py
```diff
@@ -498,3 +512,20 @@ def _report_slot_errors(
         if fuzzy_slot_name_matches:
             msg += f"\nDid you mean '{fuzzy_slot_name_matches[0]}'?"
         raise TemplateSyntaxError(msg)
+
+
+name_escape_re = re.compile(r"[^\w]")
+def _escape_slot_name(name: str) -> str:
```
JuroOravec, line 518: This function implements a graceful slot name resolution, as described in https://github.com/EmilStenstrom/django-components/issues/280#issuecomment-2081180940, so that users may define slots with non-identifier characters like "my slot!", but it can still be access via `component_vars.is_filled.my_slot_`

tests/test_context.py
```diff
@@ -213,17 +213,63 @@ def setUpClass(cls):
...
-        template = Template(
-            "{% load component_tags %}{% component_dependencies %}"
-            "{% component 'parent_with_args' parent_value=parent_value %}"
-            "{% endcomponent %}"
+    @override_settings(
+        COMPONENTS={
+            "context_behavior": "django",
+        }
+    )
+    def test_parent_args_can_be_drawn_from_context__django(self):
```
JuroOravec, line 221: For tests that were failing after changing the default `context_behavior`, I created tests for both variants, suffixed test names with `__django` or `__isolated`.

dylanjcastillo:
This looks great. And the new documentation section about slot rendering is super useful. Thank you, Juro!

dylanjcastillo (approved):

JuroOravec:
@dylanjcastillo Thanks!

On README formatting - Yeah I think I would be good to have README linted too, do you know how to set it up?

Issue #464, opened 2024-04-30
Async middleware


TheSteveBurgess:
Hi,

Are there any plans to implement async in the current middleware? At the moment using the django-components middleware forces the entire request pipeline to be sync as the middleware doesn't support a async.

Cheers,

Steve.

JuroOravec:
Hey @SteveBurgessStantec, I don't see why not. From [the docs](https://docs.djangoproject.com/en/5.0/topics/http/middleware/#asynchronous-support), it looks like making an async copy of the [ComponentDependencyMiddleware middleware](https://github.com/EmilStenstrom/django-components/blob/master/src/django_components/middleware.py#L28) should do the trick.

Do you have a project that uses async middlewares? Could you hack together a PR and check if things work for you? Or alternatively some demo project that we could use for testing. I think demo project in the repo is synchronous, and so is the project where I use django-components, so I don't have a quick way to test if things work.

SteveBurgessStantec:
Hey @JuroOravec  I converted it earlier in dev and its looks like its working fine :)

Let me just check internally and see if i'm allowed to share the code. If i get the OK I'll create a PR.

TheSteveBurgess:
I've created the PR now (i think!). You'll have to excuse me, it's been a very long time since I've created pull requests on Github :)

Let me know if you need anything from me.

JuroOravec:
Merged in https://github.com/EmilStenstrom/django-components/pull/466

PR #466, opened 2024-05-01
feat: Support both sync and async in ComponentDependencyMiddleware

TheSteveBurgess:
Hi,

PR for converting the middleware to async as mentioned here (https://github.com/EmilStenstrom/django-components/issues/464)

This is working fine for me both sync and async via Daphne.

### DIFF
--- a/src/django_components/middleware.py
+++ b/src/django_components/middleware.py
@@ -54,7 +54,7 @@ class ComponentDependencyMiddleware:
         response = self.process_response(response)
         return response
 
-    def process_response(self, response):
+    def process_response(self, response: HttpResponse) -> HttpResponse:
         if (
             getattr(settings, "COMPONENTS", {}).get("RENDER_DEPENDENCIES", False)
             and not isinstance(response, StreamingHttpResponse)
--- a/src/django_components/middleware.py
+++ b/src/django_components/middleware.py
@@ -6,8 +6,9 @@ from django.conf import settings
 from django.forms import Media
 from django.http import HttpRequest, HttpResponse, StreamingHttpResponse
 from django.http.response import HttpResponseBase
-
+from django.utils.decorators import sync_and_async_middleware
 from django_components.component_registry import registry
+from asgiref.sync import iscoroutinefunction, markcoroutinefunction
 
 if TYPE_CHECKING:
     from django_components.component import Component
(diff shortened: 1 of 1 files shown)


TheSteveBurgess:
Well, that looks like I've broken something :) 

Let me take a look why the tests are failing.

src/django_components/middleware.py
```diff
@@ -33,14 +36,32 @@ class ComponentDependencyMiddleware:
...
+        if iscoroutinefunction(self.get_response):
+            markcoroutinefunction(self)
+
     def __call__(self, request: HttpRequest) -> HttpResponseBase:
+
+        if iscoroutinefunction(self):
+            return self.__acall__(request)
+
         response = self.get_response(request)
+        response = self.process_response(request)
```
JuroOravec, line 48: Ah, maybe the issue is that we pass `request` here instead of `response`?

JuroOravec:
@TheSteveBurgess thanks a lot for the PR! 🚀 It looks great! There's one typo, let's see if things work after that :)

src/django_components/middleware.py
TheSteveBurgess: That'd do it :) Hopefully it's working with that change

src/django_components/middleware.py
```diff
@@ -33,14 +36,32 @@ class ComponentDependencyMiddleware:
...
     def __call__(self, request: HttpRequest) -> HttpResponseBase:
+
+        if iscoroutinefunction(self):
+            return self.__acall__(request)
+
         response = self.get_response(request)
+        response = self.process_response(response)
+        return response
+
+    async def __acall__(self, request):
```
JuroOravec, line 51:
```suggestion
    async def __acall__(self, request: HttpRequest) -> HttpResponseBase:
```

src/django_components/middleware.py
```diff
@@ -33,14 +36,32 @@ class ComponentDependencyMiddleware:
...
+
+        response = await self.get_response(request)
+        response = self.process_response(response)
+        return response
+
+    def process_response(self, response):
```
JuroOravec, line 57:
```suggestion
    def process_response(self, response: HttpResponse) -> HttpResponse:
```

TheSteveBurgess:
Amazing! Thanks for that :)

JuroOravec:
Merged! @TheSteveBurgess thanks a lot again :) I plan to release this as part of v0.70 together with https://github.com/EmilStenstrom/django-components/pull/465 later today

Issue #134, opened 2022-02-05
Make `{% component %}` play well with `{% block %}`


EmilStenstrom:
It seems to me that `{% block %}` tags should be executed first, and only after blocks are filled should components parse and fill their slots. This would enable use-cases as:

```
{% block content %}
    {% component_block 'calendar' date="2021-01-01" %}
        {% slot 'header' %}Calendar for January{% endslot %}
    {% endcomponent %}
{% endblock %}
```

As well as:

```
{% component_block 'calendar' date="2021-01-01" %}
    {% slot 'header' %}{% block content %}{% endblock %}{% endslot %}
{% endcomponent %}
```

Currently only the first example works, and not the second one, because slots gets executed before blocks. 

I'm not sure if there's an easy way to solve this, but I think we should examine if there is. If it's not possible to solve, let's at least document the shortcoming to solve people some trouble.

housUnus:
Django principle when dealing with NESTED blocks is the same as THE LIBRARY, if you tried to use nested BLOCKS it won't work, I don't see it as an issue, once I'm using the BLOCK on the inheriting child, it means I'm redefining its structure including everything declared inside of it from the parent.

housUnus:
```
{% block test_parent %}
    parent from base
    {% block test_child %}
        child_from_bbase
    {% endblock test_child %}
{% endblock test_parent %}


{% extends "base.html" %}
{% block test_parent %}
    parent_from_skelton 
    {% block test_child %}
        child_from_skelton
    {% endblock test_child %}
{% endblock test_parent %}
```

This would print :
parent_from_skelton child_from_skelton just as SLOTS work

EmilStenstrom:
@JuroOravec I think the most important thing here is to have an opinion on how blocks and components should work together. The two are almost solving the same problem, where blocks work pretty much like slots, and I think people largely use them for "templating" rather than components.

I didn't think blocks could be dynamic at all, meaning you couldn't wrap them in an if tag, much less a component tag?

JuroOravec:
Thanks for info! I started playing around it yesterday - mostly just writing tests to understand how things interact. What I can say so far is that:
- Slots inside `extends` or `include` blocks do not get picked up (and should)
- Blocks cannot be "passed down" to components via `component` tag. I think this makes sense.

PR #468, opened 2024-05-01
fix: loader tags compatibility

JuroOravec:
Added tests to explore behavior of components with "loader tags" (`extends`, `include`, `block`). Tested out all combinations I could think of:
- Slots inside `include`
- Slots inside `extends`
- Slots inside `block`
- Block inside `slot`

Surprisingly, it all mostly worked as expected 😄, and I had to update only the traversal logic, so it could extract nodes from `extends` / `include` tags.

Closes https://github.com/EmilStenstrom/django-components/issues/134

### DIFF
--- a/src/django_components/node.py
+++ b/src/django_components/node.py
@@ -1,6 +1,8 @@
 from typing import Callable, List, NamedTuple, Optional
 
+from django.template import Context, Template
 from django.template.base import Node, NodeList, TextNode
+from django.template.loader_tags import ExtendsNode, IncludeNode, construct_relative_path
 from django.template.defaulttags import CommentNode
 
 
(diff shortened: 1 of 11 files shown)


src/django_components/node.py
```diff
@@ -20,26 +22,79 @@ class NodeTraverse(NamedTuple):
...
+    # Special case - {% extends %} tag - Load the template and go deeper
+    if isinstance(node, ExtendsNode):
+        # NOTE: When {% extends %} node is being parsed, it collects all remaining template
+        # under node.nodelist.
+        # Hence, when we come across ExtendsNode in the template, we:
+        # 1. Go over all nodes in the template using `node.nodelist`
+        # 2. Go over all nodes in the "parent" template, via `node.get_parent`
+        nodes = NodeList()
+        nodes.extend(node.nodelist)
+        template = node.get_parent(context)
```
JuroOravec, line 55: To resolve the parent Template referenced via `{% extends %}` tag, we needed to pass it a Context. That's why much of the non-test changes is adding a `context` argument to functions

src/django_components/node.py
```diff
@@ -20,26 +22,79 @@ class NodeTraverse(NamedTuple):
...
             nodes.extend(nodelist)
     return nodes
+
+
+def get_template_for_include_node(include_node: IncludeNode, context: Context) -> Template:
+    """
+    This snippet is taken directly from `IncludeNode.render()`. Unfortunately the
+    render logic doesn't separate out template loading logic from rendering, so we
+    have to copy the method.
+    """
```
JuroOravec, line 77: As it says above, this function is a copy from Django's codebase

tests/test_templatetags.py
```diff
@@ -1062,7 +1079,7 @@ def test_non_unique_fill_names_is_error(self):
             Template(
                 """
                 {% load component_tags %}
-                {% component "broken_component" %}
+                {% component "test" %}
```
JuroOravec, line 1082: This test was using a wrong component for testing duplicate fills

EmilStenstrom (commented):
I'm OK with releasing this, even though I don't understand it fully.

What would you say is the intuition on how block and slot works together? That the outermost tag executes first?

JuroOravec:
I'll write the behavior out, since it also took me some time to wrap my head around it:

1. First let's clarify how `include` and `extends` tags work inside components. So when component template includes `include` or `extends` tags, it's as if the "included" template was inlined. So if the "included" template contains `slot` tags, then the component uses those slots.

	So if I have a template `abc.html`:
	```django
	<div>
	  hello
	  {% slot "body" %}{% endslot %}
	</div>
	```
	
	And components that make use of `abc.html` via `include` or `extends`:
	```py
	@component.register("my_comp_extends")
	class MyCompWithExtends(component.Component):
	    template = """{% extends "abc.html" %}"""
	
	@component.register("my_comp_include")
	class MyCompWithInclude(component.Component):
	    template = """{% include "abc.html" %}"""
	```

	Then I can set slot fill for the slot imported via `include/extends`:

	```django
	{% component "my_comp_extends" %}
		{% fill "body" %}
			123
		{% endfill %}
	{% endcomponent %}
	```

	And it will render:
	```html
	<div>
	  hello
	  123
	</div>
	```

2. Slot and block

	So if I have a template `abc.html` like so:
	```django
	<div>
	  hello
	  {% block inner %}
        1
  	    {% slot "body" %}
          2
        {% endslot %}
	  {% endblock %}
	</div>
	```

	and component `my_comp`:
	```py
	@component.register("my_comp")
	class MyComp(component.Component):
		template_name = "abc.html"
	```

	Then:
	1. Since the `block` wasn't overriden, we can use the `body` slot:
		```django
		{% component "my_comp" %}
			{% fill "body" %}
				XYZ
			{% endfill %}
		{% endcomponent %}
		```
		And we get:
		```html
		<div>
		  hello
		  1
		    XYZ
		</div>
		```

	3. `blocks` CANNOT be overriden through the `component` tag, so something like this:
		```django
		{% component "my_comp" %}
			{% fill "body" %}
				XYZ
			{% endfill %}
		{% endcomponent %}
		{% block "inner" %}
			456
		{% endblock %}
		```
		Will still render the component content just the same:
		```html
		<div>
		  hello
		  1
		    XYZ
		</div>
		```
	
	4. I CAN override the `block` tags of `abc.html` if my component template uses `extends`. In that case, just as you would expect, the `block inner` inside `abc.html` will render `OVERRIDEN`:
		```py
		@component.register("my_comp")
		class MyComp(component.Component):
			template_name = """
				{% extends "abc.html" %}

				{% block inner %}
					OVERRIDEN
				{% endblock %}
			"""
		```

	5. This is where it gets interesting (but still intuitive).  I can insert even new `slots` inside these "overriding" blocks:
		```py
		@component.register("my_comp")
		class MyComp(component.Component):
			template_name = """
				{% extends "abc.html" %}

				{% load component_tags %}
				{% block "inner" %}
					OVERRIDEN
					{% slot "new_slot" %}
						hello
					{% endslot %}
				{% endblock %}
			"""
		```

		And I can then pass fill for this `new_slot` when rendering the component:
		```django
		{% component "my_comp" %}
			{% fill "new_slot" %}
				XYZ
			{% endfill %}
		{% endcomponent %}
		```

		NOTE: Currently I can supply fills for both `new_slot` and `body` slots, and I will not get an error for an invalid/unknown slot name. But since `body` slot is not rendered, it just won't do anything. So this renders the same as above:
		```django
		{% component "my_comp" %}
			{% fill "new_slot" %}
				XYZ
			{% endfill %}
			{% fill "body" %}
				www
			{% endfill %}
		{% endcomponent %}
		```

JuroOravec:
I've included the explainer in [docs/slots_and_blocks.md](https://github.com/EmilStenstrom/django-components/pull/468/files#diff-8ccb1410d9a7c0f8495682256e5221040b2c803bb962e243b0d23b44bbda692b)

PR #469, opened 2024-05-01
Update README.md

KPCOFGS:
Changed relative paths to absolute paths for PyPI compatibilities


JuroOravec:
@KPCOFGS Thanks for the fix!

PR #474, opened 2024-05-02
tests: add inline syntax highlight, inline template files

### LINKED ISSUE #470: Tests refactor: Move from template files to inline templates
It might be just me (please disagree in the comments), but I'm finding it hard to look at many of the tests, and understand what the tests are testing. To do it, you first look at the test, then at the component it's defining, then at the template file it's referencing. This is OK I guess, but I'm slowing finding it more and more annoying as we add more tests. 

Idea: Could we make each test function self-containing, with very few external references? I would be fine if we duplicate some code between tests here; sharing code between tests is likely not a good idea anyways.

Suggestion:
1. Move all the template code from external files to inline templates in the component definition
2. Move all components from classes defined outside of the test function to inline classes defined in the function. 

The goal with these changes would be to be able to look at a test, and immediately understand what it's testing, without having to look stuff up in several different places.

JuroOravec:
- Where possible move template files into python
 - Where it made sense, I moved the component classes inside tests or `TestCases` that used them, so there's fewer global component classes
 - Added line syntax highlight to most in-Python Django templates in tests

Closes #470


benchmarks/component_rendering.py
```diff
@@ -3,18 +3,27 @@
...
+from django_components import component, types
 from django_components.middleware import CSS_DEPENDENCY_PLACEHOLDER, JS_DEPENDENCY_PLACEHOLDER
 from tests.django_test_setup import *  # NOQA
 from tests.testutils import BaseTestCase, create_and_process_template_response
 
 
 class SlottedComponent(component.Component):
-    template_name = "slotted_template.html"
+    template: types.django_html = """
+        {% load component_tags %}
```
JuroOravec, line 14: I moved the template files into Python, and added syntax highlighting with `types.django_html` where possible.

src/django_components/types.py
```diff
@@ -19,6 +19,9 @@ def __getitem__(self, params: Any) -> "Annotated":  # type: ignore
                 params = (params,)
             return Annotated(self.type_, *params, **self.metadata[1])  # type: ignore
 
+        def __class_getitem__(self, *params: Any) -> "Annotated":  # type: ignore
+            return Annotated(*params)  # type: ignore
```
JuroOravec, line 23: Because I'm using the type annotations in test files, in Python 3.8 our shim is used. And there it turned out that the invocations like `Annotated[str, "css"]` were failing, because the class needed to define `__class_getitem__`

EmilStenstrom:
Great work again, let's merge this! :)

I think there might be some more work to inline components as well.

PR #475, opened 2024-05-02
Inline component view tests

EmilStenstrom:
This PR inlines all the tests in tests/test_component_as_view.py with the purpose of making each test be self-contained. This means the component, registration, and the views, are all defined in the same function. Let me know what you think of the structure.


JuroOravec (approved):
Awesome, thanks for cleaning up this one, I intentioanlly skipped this file in my MR because I wasn't sure how to break it down :)

PR #477, opened 2024-05-04
feat: support special chars `# @ - . :` in component kwargs

JuroOravec:
Part of https://github.com/EmilStenstrom/django-components/discussions/471

This MR allows us to use component argument keys with special characters `# @ - . :`.

The parsing logic override is based on [django-slippers implementation](https://github.com/mixxorz/slippers/blob/main/slippers/template.py)

### DIFF
--- a/README.md
+++ b/README.md
@@ -667,6 +667,44 @@ COMPONENTS = {
 }
 ```
 
+## Passing data to components
+
+As seen above, you can pass arguments to components like so:
+
+```django
+<body>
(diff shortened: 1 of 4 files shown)


dylanjcastillo (approved):
Looks good to me!

EmilStenstrom:
We are taking on a big maintenance burden by copying in a modified version of django's parsing like this. If we would go with the attrs="""x""" solutions this wouldn't be needed... :)

But it's looking like we will need it, but the latest messages, so maybe we're fine.

PR #489, opened 2024-05-07
fix: make template_loader consider tuples for STATICFILES_DIRS

### LINKED ISSUE #476: template_loader blocks on STATICFILES_DIRS if there are tuples
## How to reproduce

Simply run django with, in your _settings.py_, something like :
```
STATICFILES_DIRS = [
    # ...
    ("downloads", "/opt/webfiles/stats"),
]
```

(as made possible [natively in Django](https://docs.djangoproject.com/en/3.2/ref/settings/#std-setting-STATICFILES_DIRS))

## Current behaviour

Getting `TypeError: expected str, bytes or os.PathLike object, not tuple` given [this line] (https://github.com/EmilStenstrom/django-components/blob/e566d8ecbb7016cf3a8861a47a4a52e4a503b838/src/django_components/template_loader.py#L49) 

## Expected

`class Loader` from _src/django_components/template_loader.py_ should consider the second item of STATICFILES_DIRS list item when not a str.

## Furthermore

Thank you a lot, sir, for this django library. It helped us, a French news company, a lot to center our codebase almost fully on Django.
I might try to suggest a clean PR for this

alexandreMartinEcl:
fixes #476  

Hoping to fix issue [here](https://github.com/EmilStenstrom/django-components/issues/476)

I suggested to use `logger.warning` and simply not to consider the _STATICFILES_DIR_ value if the value is incorrect. But depends on your preference.

### DIFF
--- a/src/django_components/template_loader.py
+++ b/src/django_components/template_loader.py
@@ -37,7 +37,7 @@ class Loader(FilesystemLoader):
 
         directories: Set[Path] = set()
         for component_dir in component_dirs:
-            if isinstance(component_dir, tuple | list) and len(component_dir) == 2:
+            if isinstance(component_dir, (tuple, list)) and len(component_dir) == 2:
                 component_dir = component_dir[1]
             try:
                 Path(component_dir)
--- a/src/django_components/template_loader.py
+++ b/src/django_components/template_loader.py
@@ -37,6 +37,16 @@ class Loader(FilesystemLoader):
 
         directories: Set[Path] = set()
         for component_dir in component_dirs:
+            if isinstance(component_dir, tuple | list) and len(component_dir) == 2:
+                component_dir = component_dir[1]
+            try:
+                Path(component_dir)
+            except TypeError:
+                logger.warning(
+                    f"STATICFILES_DIRS expected str, bytes or os.PathLike object, or tuple/list of length 2. "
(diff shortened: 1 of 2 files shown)


JuroOravec:
@alexandreMartinEcl Looks good to me! I pushed one fix for a typing error.

One thing that remains - the tests fail for Python 3.8-3.10 somewhere around [where we've added `MagicMock` or `@patch()`](https://github.com/EmilStenstrom/django-components/pull/489/files#diff-6faab92c5a5f55ed5c40230f2e1878cfe9603debd5af3d9b90c4ee875d5c2e93R141)
 
See logs https://github.com/EmilStenstrom/django-components/actions/runs/8990294238/job/24695481406?pr=489

alexandreMartinEcl:
Indeed older version of python does not seem to appreciate my way to mock. It returns 
`'django_components.logger' is not a package` on `@patch("django_components.logger.logger.warning")`

I'm trying another way, but did not find a quick way to test on python3.8

HAve you maybe also experienced a difficulty with _mocking_ on old versions of python ?

JuroOravec:
@alexandreMartinEcl Thanks again, tests have been fixed, so I've merged it :) It is now avaiable in  https://github.com/EmilStenstrom/django-components/releases/tag/0.73

Issue #483, opened 2024-05-06
Feature parity with Juro's fork


JuroOravec:
Here I'm documenting ther remaining features/changes that I made in my fork that I'd want to eventually discuss.

Feature name | Description | Resolved
---|---|---
Fallthrough attributes | Allow to pass ad hoc key-value pairs to components as "attributes" to enable component users to configure styling and event handling. See [Vue's flalthrough attributes](https://vuejs.org/guide/components/attrs). | ✅ <br/>  (added in v0.74)
HTML attribute formatting | One thing is passing ad hoc attributes to components, but other is formatting them as HTML attributes. E.g. see how [django-web-components](https://github.com/Xzya/django-web-components#default--merged-attributes) does it. | ✅ <br/>  (added in v0.74)
Global rendering context | Allow to specify a dict of values that should be automatically made available inside all components in given rendering run. Similar to [Vue's inject/provide](https://vuejs.org/guide/components/provide-inject) or [React's Context/Provide](https://legacy.reactjs.org/docs/context.html) | ✅ <br/> (added in v0.80)
Inlined component render | Allow to render a component like a function, passing in the same args and kwargs as the `get_context_data` accepts. E.g. `content = MyComponent.render_from_input(args, kwargs)` | ✅ <br/> (added in v0.81)
Typed component inputs | When I render template with `django.shortcuts.render("mytemplate.html", data)`, there's not typing support for what should be the fields of `data`. Likewise, if the "Inlined component render" was implemented, the `args, kwargs` would likely not be typed either. Ideally, there'd be a way to define the typing of the component input, so I get a Mypy error if I pass invalid field/type. <br/><br/> NOTE: We will see once we get to it, but in my experience it was easier to add typing to component inputs after enforcing that components could have ONLY kwargs instead of args + kwargs. | [ ]

EmilStenstrom:
For the Global rendering context, maybe we could take some hints from Django's RequestContext? https://docs.djangoproject.com/en/5.0/ref/templates/api/#using-requestcontext

PR #491, opened 2024-05-07
feat: Support for HTML attributes and `html_attrs` tag

JuroOravec:
TODO:
- [x] Agree on interfaces/features
- [x] Fix tests
- [x] Update README

### DIFF
--- a/README.md
+++ b/README.md
@@ -739,16 +739,16 @@ But as you can see in the case above, the event handler `@click.stop` and stylin
 
 Luckily, there's a better way.
 
-When we want to pass a dictionary to a component, we can define individual key-value pairs as component kwargs, so we can keep all the relevant information in the template. For that, we prefix the key with the name of the dict and `::`. So key `class` of input `attrs` becomes `attrs::class`. And our example becomes:
+When we want to pass a dictionary to a component, we can define individual key-value pairs as component kwargs, so we can keep all the relevant information in the template. For that, we prefix the key with the name of the dict and `:`. So key `class` of input `attrs` becomes `attrs:class`. And our example becomes:
 
 ```py
 @component.register("my_comp")
 class MyComp(component.Component):
     template = """
         {% component "other"
-            attrs::class="pa-4 flex"
-            attrs::data-some-id=some_id
-            attrs::@click.stop="onClickHandler"
+            attrs:class="pa-4 flex"
+            attrs:data-some-id=some_id
+            attrs:@click.stop="onClickHandler"
         %}
         {% endcomponent %}
     """
@@ -761,18 +761,18 @@ Sweet! Now all the relevant HTML is inside the template, and we can move it to a
 
 ```django
 {% component "other"
-    attrs::class="pa-4 flex"
-    attrs::data-some-id=some_id
-    attrs::@click.stop="onClickHandler"
+    attrs:class="pa-4 flex"
+    attrs:data-some-id=some_id
+    attrs:@click.stop="onClickHandler"
 %}
 {% endcomponent %}
 ```
 
 > Note: It is NOT possible to define nested dictionaries, so
-`attrs::my_key::two=2` would be interpreted as:
+`attrs:my_key:two=2` would be interpreted as:
 >
 > ```py
-> {"attrs": {"my_key::two": 2}}
+> {"attrs": {"my_key:two": 2}}
 > ```
 
 ## Rendering HTML attributes
@@ -835,7 +835,7 @@ instead of overriding them. For example, if you're authoring a component, you ma
 want to ensure that the component will ALWAYS have a specific class. Yet, you may
 want to allow users of your component to supply their own classes.
 
-We can achieve this by prefixing the keys with `add::` to indicate that these values
+We can achieve this by prefixing the keys with `add:` to indicate that these values
 should be appended, instead of overwriting the previous value.
 
 So if we have a variable `attrs`:
@@ -845,10 +845,10 @@ attrs = {
 }
 ```
 
-And on `html_attrs` tag, we set the key `add::class`:
+And on `html_attrs` tag, we set the key `add:class`:
 
 ```django
-<div {% html_attrs attrs add::class="some-class" %}>
+<div {% html_attrs attrs add:class="some-class" %}>
 </div>
 ```
 
@@ -868,8 +868,8 @@ class MyComp(component.Component):
         <div
             {% html_attrs attrs
                 class="pa-4 text-red"
-                add::class="my-comp-date"
-                add::class=class_from_var
+                add:class="my-comp-date"
+                add:class=class_from_var
                 data-id="123"
             %}
         >
@@ -889,9 +889,9 @@ class Parent(component.Component):
     template: t.django_html = """
         {% component "my_comp"
             date=date
-            attrs::class="pa-0 border-solid border-red"
-            attrs::data-json=json_data
-            attrs::@click="(e) => onClick(e, 'from_parent')"
+            attrs:class="pa-0 border-solid border-red"
+            attrs:data-json=json_data
+            attrs:@click="(e) => onClick(e, 'from_parent')"
         %}
         {% endcomponent %}
     """
@@ -911,8 +911,8 @@ Inside `MyComp`, we define these default attributes:
 
 So if `attrs` includes these keys, the defaults above will be ignored.
 
-`MyComp` also defines `add::class` twice. It means that whether the `class`
-attribute is taken from `attrs` or the defaults, the two `add::class` values
+`MyComp` also defines `add:class` twice. It means that whether the `class`
+attribute is taken from `attrs` or the defaults, the two `add:class` values
 will be appended to it.
 
 So by default, `MyComp` renders:
@@ -931,12 +931,12 @@ contents of that dictionary are rendered as the HTML attributes.
 In `Parent`, we make use of passing dictionary key-value pairs as kwargs to define
 individual attributes as if they were regular kwargs.
 
-So all kwargs that start with `attrs::` will be collected into an `attrs` dict.
+So all kwargs that start with `attrs:` will be collected into an `attrs` dict.
 
 ```django
-    attrs::class="pa-0 border-solid border-red"
-    attrs::data-json=json_data
-    attrs::@click="(e) => onClick(e, 'from_parent')"
+    attrs:class="pa-0 border-solid border-red"
+    attrs:data-json=json_data
+    attrs:@click="(e) => onClick(e, 'from_parent')"
 ```
 
 And `MyComp` will receive `attrs` input with following keys:
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@ Read on to learn about the details!
 
 ## Release notes
 
+**Version 0.74** introduces `html_attrs` tag and `prefix:key=val` construct for passing dicts to components.
+
 🚨📢 **Version 0.70**
 
 - `{% if_filled "my_slot" %}` tags were replaced with `{{ component_vars.is_filled.my_slot }}` variables.
@@ -816,7 +818,7 @@ This feature is inspired by [`merge_attrs` tag of django-web-components](https:/
 
 ### Default attributes
 
-Sometimes you may want to specify default values for attributes. You can pass a second argument (or kwarg `default`) to set the defaults.
+Sometimes you may want to specify default values for attributes. You can pass a second argument (or kwarg `defaults`) to set the defaults.
 
 ```django
 <div {% html_attrs attrs defaults %}>
@@ -830,7 +832,7 @@ In the example above, if `attrs` contains e.g. the `class` key, `html_attrs` wil
 
 Otherwise, `html_attrs` will render:
 
-`class="{{ default.class }}"`
+`class="{{ defaults.class }}"`
 
 ### Appending attributes
 
--- a/README.md
+++ b/README.md
@@ -940,7 +940,7 @@ Then:
 	renders: <br/>
 	`class="from-attrs" type="submit"`
 
-- ONLY defaults (as kwarg) <br/>
+- Only defaults (as kwarg) <br/>
 	`{% html_attr defaults=defaults %}`
 
 	renders: <br/>
--- a/README.md
+++ b/README.md
@@ -869,14 +869,14 @@ To simplify merging of variables, you can supply the same key multiple times, an
 
 ```django
 {# my_var = "class-from-var text-red" #}
-<div {% html_attrs attrs class="some-class" class=my_var %}>
+<div {% html_attrs attrs class="some-class another-class" class=my_var %}>
 </div>
 ```
 
 Renders:
 
 ```html
-<div data-value="my-class pa-4 some-class class-from-var text-red">
+<div data-value="my-class pa-4 some-class another-class class-from-var text-red">
 </div>
 ```
 
--- a/README.md
+++ b/README.md
@@ -701,9 +701,75 @@ class Calendar(component.Component):
         return {
             "date": kwargs["my-date"],
             "id": kwargs["#some_id"],
-            "@click.native": kwargs["@click.native"]
+            "on_click": kwargs["@click.native"]
         }
 ```
+
+### Pass dictonary by its key-value pairs
+
+Sometimes, a component may expect a dictionary as one of its inputs.
+
+Most commonly, this happens when a component accepts a dictionary
+of HTML attributes (usually called `attrs`) to pass to the underlying template.
+
+In such cases, we may want to define some HTML attributes statically, and other dynamically. But for that, we need to define this dictionary on Python side:
+
+```py
+@component.register("my_comp")
+class MyComp(component.Component):
+    template = """
+        {% component "other" attrs=attrs %}
+        {% endcomponent %}
+    """
+
+    def get_context_data(self, some_id: str):
+        attrs = {
+            "class": "pa-4 flex",
+            "data-some-id": some_id,
+            "@click.stop": "onClickHandler",
+        }
+        return {"attrs": attrs}
+```
+
+But as you can see in the case above, the event handler `@click.stop` and styling `pa-4 flex` are disconnected from the template. If the component grew in size and we moved the HTML to a separate file, we would have hard time reasoning about the component's template.
+
+Luckily, there's a better way.
+
+When we want to pass a dictionary to a component, we can define individual key-value pairs as component kwargs, so we can keep all the relevant information in the template. For that, we prefix the key with the name of the dict and `::`. So key `class` of input `attrs` becomes `attrs::class`. And our example becomes:
+
+```py
+@component.register("my_comp")
+class MyComp(component.Component):
+    template = """
+        {% component "other"
+            attrs::class="pa-4 flex"
+            attrs::data-some-id=some_id
+            attrs::@click.stop="onClickHandler"
+        %}
+        {% endcomponent %}
+    """
+
+    def get_context_data(self, some_id: str):
+        return {"some_id": some_id}
+```
+
+Sweet! Now we can move the template to it's own file without losing the context:
+
+```django
+{% component "other"
+    attrs::class="pa-4 flex"
+    attrs::data-some-id=some_id
+    attrs::@click.stop="onClickHandler"
+%}
+{% endcomponent %}
+```
+
+> Note: It is NOT possible to pass nested dictionaries like this, so
+`attrs::my_key::two=2` would be interpreted as:
+>
+> ```py
+> {"attrs": {"my_key::two": 2}}
+> ```
 
 ## Component context and scope
 
--- a/README.md
+++ b/README.md
@@ -716,7 +716,8 @@ Sometimes, a component may expect a dictionary as one of its inputs.
 Most commonly, this happens when a component accepts a dictionary
 of HTML attributes (usually called `attrs`) to pass to the underlying template.
 
-In such cases, we may want to define some HTML attributes statically, and other dynamically. But for that, we need to define this dictionary on Python side:
+In such cases, we may want to define some HTML attributes statically, and other dynamically.
+But for that, we need to define this dictionary on Python side:
 
 ```py
 @component.register("my_comp")
@@ -735,11 +736,16 @@ class MyComp(component.Component):
         return {"attrs": attrs}
 ```
 
-But as you can see in the case above, the event handler `@click.stop` and styling `pa-4 flex` are disconnected from the template. If the component grew in size and we moved the HTML to a separate file, we would have hard time reasoning about the component's template.
+But as you can see in the case above, the event handler `@click.stop` and styling `pa-4 flex`
+are disconnected from the template. If the component grew in size and we moved the HTML
+to a separate file, we would have hard time reasoning about the component's template.
 
 Luckily, there's a better way.
 
-When we want to pass a dictionary to a component, we can define individual key-value pairs as component kwargs, so we can keep all the relevant information in the template. For that, we prefix the key with the name of the dict and `:`. So key `class` of input `attrs` becomes `attrs:class`. And our example becomes:
+When we want to pass a dictionary to a component, we can define individual key-value pairs
+as component kwargs, so we can keep all the relevant information in the template. For that,
+we prefix the key with the name of the dict and `:`. So key `class` of input `attrs` becomes
+`attrs:class`. And our example becomes:
 
 ```py
 @component.register("my_comp")
@@ -810,23 +816,21 @@ This feature is inspired by [`merge_attrs` tag of django-web-components](https:/
 
 ### Default attributes
 
-Sometimes you may want to specify default values for attributes. You can pass additional keyword arguments to set the defaults.
+Sometimes you may want to specify default values for attributes. You can pass a second argument (or kwarg `default`) to set the defaults.
 
 ```django
-<div {% html_attrs attrs class="alert" role="alert" %}>
+<div {% html_attrs attrs defaults %}>
     ...
 </div>
 ```
 
-In the example above, if `attrs` contains the `class` key, `html_attrs` will render:
+In the example above, if `attrs` contains e.g. the `class` key, `html_attrs` will render:
 
 `class="{{ attrs.class }}"`
 
 Otherwise, `html_attrs` will render:
 
-`class="alert"`
-
-Same applies to the `role` attribute.
+`class="{{ default.class }}"`
 
 ### Appending attributes
 
@@ -835,8 +839,8 @@ instead of overriding them. For example, if you're authoring a component, you ma
 want to ensure that the component will ALWAYS have a specific class. Yet, you may
 want to allow users of your component to supply their own classes.
 
-We can achieve this by prefixing the keys with `add:` to indicate that these values
-should be appended, instead of overwriting the previous value.
+We can achieve this by adding extra kwargs. These values
+will be appended, instead of overwriting the previous value.
 
 So if we have a variable `attrs`:
 ```py
@@ -845,10 +849,10 @@ attrs = {
 }
 ```
 
-And on `html_attrs` tag, we set the key `add:class`:
+And on `html_attrs` tag, we set the key `class`:
 
 ```django
-<div {% html_attrs attrs add:class="some-class" %}>
+<div {% html_attrs attrs class="some-class" %}>
 </div>
 ```
 
@@ -859,7 +863,118 @@ Then these will be merged and rendered as:
 </div>
 ```
 
-### Full example
+To simplify merging of variables, you can supply the same key multiple times, and these will be all joined together:
+
+```django
+{# my_var = "class-from-var text-red" #}
+<div {% html_attrs attrs class="some-class" class=my_var %}>
+</div>
+```
+
+Renders:
+
+```html
+<div data-value="my-class pa-4 some-class class-from-var text-red">
+</div>
+```
+
+### Rules for `html_attrs`
+
+1. Both `attrs` and `defaults` can be passed as positional args
+    
+    `{% html_attrs attrs defaults key=val %}`
+    
+    or as kwargs
+    
+    `{% html_attrs key=val defaults=defaults attrs=attrs %}`
+
+2. Both `attrs` and `defaults` are optional (can be omitted)
+
+3. Both `attrs` and `defaults` are dictionaries, and we can define them the same way [we define dictionaries for the `component` tag](#pass-dictonary-by-its-key-value-pairs). So either as `attrs=attrs` or `attrs:key=value`.
+
+4. All other kwargs are appended and can be repeated.
+
+### Examples for `html_attrs`
+
+Assuming that:
+
+```py
+class_from_var = "from-var"
+
+attrs = {
+	"class": "from-attrs",
+	"type": "submit",
+}
+
+defaults = {
+	"class": "from-defaults",
+	"role": "button",
+}
+```
+
+Then:
+
+- Empty tag <br/>
+	`{% html_attr %}`
+
+	renders (empty string): <br/>
+	` `
+
+- Only kwargs <br/>
+	`{% html_attr class="some-class" class=class_from_var data-id="123" %}`
+
+	renders: <br/>
+	`class="some-class from-var" data-id="123"`
+
+- Only attrs <br/>
+	`{% html_attr attrs %}`
+
+	renders: <br/>
+	`class="from-attrs" type="submit"`
+
+- Attrs as kwarg <br/>
+	`{% html_attr attrs=attrs %}`
+
+	renders: <br/>
+	`class="from-attrs" type="submit"`
+
+- ONLY defaults (as kwarg) <br/>
+	`{% html_attr defaults=defaults %}`
+
+	renders: <br/>
+	`class="from-defaults" role="button"`
+
+- Attrs using the `prefix:key=value` construct <br/>
+	`{% html_attr attrs:class="from-attrs" attrs:type="submit" %}`
+
+	renders: <br/>
+	`class="from-attrs" type="submit"`
+
+- Defaults using the `prefix:key=value` construct <br/>
+	`{% html_attr defaults:class="from-defaults" %}`
+
+	renders: <br/>
+	`class="from-defaults" role="button"`
+
+- All together (1) - attrs and defaults as positional args: <br/>
+	`{% html_attrs attrs defaults class="added_class" class=class_from_var data-id=123 %}`
+
+	renders: <br/>
+	`class="from-attrs added_class from-var" type="submit" role="button" data-id=123`
+
+- All together (2) - attrs and defaults as kwargs args: <br/>
+	`{% html_attrs class="added_class" class=class_from_var data-id=123 attrs=attrs defaults=defaults %}`
+
+	renders: <br/>
+	`class="from-attrs added_class from-var" type="submit" role="button" data-id=123`
+
+- All together (3) - mixed: <br/>
+	`{% html_attrs attrs defaults:class="default-class" class="added_class" class=class_from_var data-id=123 %}`
+
+	renders: <br/>
+	`class="from-attrs added_class from-var" type="submit" data-id=123`
+
+### Full example for `html_attrs`
 
 ```py
 @component.register("my_comp")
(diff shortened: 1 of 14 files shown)


tests/test_attributes.py
```diff
@@ -0,0 +1,257 @@
...
+        template = Template(template_str)
+        with self.assertRaisesMessage(TypeError, "cannot change object - object is immutable"):
+            template.render(Context({"class_var": "padding-top-8"}))
+
+    def test_attrs_tag(self):
+        @component.register("test")
+        class AttrsComponent(component.Component):
+            template: types.django_html = """
+                {% load component_tags %}
+                <div {% merge_attrs component_vars.attrs class+="added_class" data-id=123 %}>
```
JuroOravec, line 238: So there's 2 ways to render the attributes into the template. If user does not need to change the attributes, they can simply call `{{ component_vars.attrs }}` as seen here https://github.com/EmilStenstrom/django-components/pull/491/files#diff-a22fb83b06ea7d4165ec648dddac689cf529335fc2dc994c61eb1250ca26465cR216

But if they need to add extra fields along with the attrs, then, I used the same approach for merging/rendering the attributes as used in [`django-web-components`](https://github.com/Xzya/django-web-components/tree/master?tab=readme-ov-file#passing-data-to-components). That is, using a tag to render the dict and specify any additional fields as kwargs on the tag.

JuroOravec, line 238: Also note here that I used component_vars.attrs to access the attributes. It's a bit wordy, it would've been awesome if it could be something simple like $attrs, but 1) dollar sign is not a valid identifier character, and 2) succint names can be achieved later. Now it's imo important that things work.

tests/test_attributes.py
```diff
@@ -0,0 +1,257 @@
...
+)
+
+# isort: off
+from .django_test_setup import *  # NOQA
+from .testutils import BaseTestCase
+
+# isort: on
+
+
+class AttributesToStringTest(BaseTestCase):
```
JuroOravec, line 19: Tests on lines ~ 18-150 are copied from https://github.com/Xzya/django-web-components/blob/master/tests/test_attributes.py

src/django_components/component.py
```diff
@@ -318,6 +349,7 @@ def render(
                 # See https://github.com/EmilStenstrom/django-components/issues/280#issuecomment-2081180940
                 "component_vars": {
                     "is_filled": slot_bools,
+                    "attrs": attrs,
```
JuroOravec, line 352: I exposed the parsed attributes via `{{ component_vars.attrs }}`. Also, I made `attrs` immutable.

src/django_components/attributes.py
```diff
@@ -0,0 +1,182 @@
+# Initial implementation based on attributes.py from django-web-components
+# See https://github.com/Xzya/django-web-components/blob/b43eb0c832837db939a6f8c1980334b0adfdd6e4/django_web_components/templatetags/components.py
+# And https://github.com/Xzya/django-web-components/blob/b43eb0c832837db939a6f8c1980334b0adfdd6e4/django_web_components/attributes.py
```
JuroOravec, line 3: As it says here, this file is mostly copied from django-web-components, with added types, and with a tweak here and there.

src/django_components/component.py
```diff
@@ -253,11 +255,19 @@ def get_template(self, context: Mapping) -> Template:
...
         )
 
-    def render_from_input(self, context: Context, args: Union[List, Tuple], kwargs: Dict) -> str:
-        component_context: dict = self.get_context_data(*args, **kwargs)
+    def render_from_input(self, context: Context, args: Union[List, Tuple], kwargs: Dict[str, Any]) -> str:
+        processed_kwargs, attrs = extract_attrs_from_component_input(kwargs)
+        # TODO: Change `get_context_data` signature so we can distinguish
+        #   input named "attrs" from the fallthrough attrs:
+        #   `get_context_data(kwargs: Dict[Any, Any], attrs: Dict[str, Any])`
+        component_context: dict = self.get_context_data(*args, attrs=attrs, **processed_kwargs)
```
JuroOravec, line 263: Passing `attrs` to `get_context_data` is not idea, because practically it's not possible to distinguish between the attributes and a component input named `attrs`.

IMO a better approach would be to pass the component inputs as a dict (instead of spreading it with `**kwargs`). For this, we'd have to remove the option to pass positional arguments (`*args`). So this would be a breaking change. But IMO this would be the right way forward, because there's more features for which we'd need to make this change, like the "Typed component inputs" in https://github.com/EmilStenstrom/django-components/issues/483.

So instead of
```py
class MyClass(component.Component):
	def get_context_data(self, *args: Any, **kwargs: Any):
		return {...}
```

we'd have

```py
class MyClass(component.Component):
	def get_context_data(self, props: Dict[str, Any], attrs: Dict[str, Any]):
		return {...}
```

Furthermore, we could go a step further and future-proof the function signature in a similar manner how Vue does it, where the last argument would be an object that would serve as a container for any possible future fields.

```py
# in django_components/component.py
class ComponentContext(NamedTuple):
	attrs: Dict[str, Any]

# in my_component.py
class MyComponent(component.Component):
	def get_context_data(self, props, ctx)
		# Access attrs like this:
		ctx.attrs
```

tests/test_attributes.py
EmilStenstrom: I'm not a big fan of adding the "+=" operator here. My long term hope is to merge this into the django codebase, and each new thing like this increases the things you have to understand, and that need to be "sold" to the django maintainers.

For this, I'd suggest we drop the "+" sign, and just do "=" to mean "add".

EmilStenstrom:
Happy to see you are working on this! 👏👏👏

I have a couple of big questions about this approach:
* My intention with the `{% component attrs:x-data="..." %}` design was that attrs was user-selected, so `{% component my_array:x-data="..." %}` would be just as valid. The above means we wouldn't need any components_vars. What do you think of this idea?
* Copying hundreds of lines of code from django-web-components (impl + tests) and taking on the maintenance of that code feels like a big thing that should be done carefully. Looking at the changes needed, are we sure this is a good idea?

tests/test_attributes.py
JuroOravec: Maybe we could similar thing as we did with the `attrs`, and have something like `add:class="abc"`, so there's still a way to differentiate between "adding" and "setting" attributes, but without the unconventional `+=`?

JuroOravec:
@EmilStenstrom thanks for feedback! Yes, it's exactly these big questions that I wanted to discuss here.

For me, the steps I'm making/suggesting are with the end goal of having a similar developer experience (DevExp; ease of coding) for writing frontend code in Python/Django as I had with Vue/React. django-components is the closest thing I found in that regard. But it means that my goal may not be entirely aligned with your goal of pushing django-components to django.

IMO neither of the two is better, it's just different opinions, with different modes of operations, but IMO both roads could lead to stable maintenance of the django-components codebase.

### Pursuing DevExp vs integration into Django codebase

My feeling from Django so far is that while it's great at being a "batteries-included" starting point, I feel like almost every aspect of it could be replaced with another library that "fixes" what Django lacks:
- I use [`django-ninja`](https://github.com/vitalik/django-ninja) to have better validation of inputs for views/endpoints
- For ORM, I would rather use [Prisma's python client](https://prisma-client-py.readthedocs.io/en/stable/) over Django's ORM, to have a better control over database queries and to have queries validated with Mypy (so I get errors when I update the models).
- While Django Forms are OK for simple forms, it's hard to work with them when one needs add or customize frontend interactivity, as the forms feel opaque
- I use django-components for templating to have better control over which variables are redered where, and to group logic by components.

So from the above, it feels like Django is great at being the backbone that ties everything together, but spreads itself too thin, and then provides only basic support for each of the subcategories of web app/server development. Overall I don't feel like the Django landscape is going in the direction of merging successful projects into core Django. Instead, when I look at e.g. django-ninja, I get the feeling that while they maintain compatibility with Django, they are focusing on DevExp by porting popular features from FastAPI, rather than trying to maintain the codebase compatible with Django codebase (e.g. djang-ninja uses pydantic for validation).

So IMO if I could have a say, I'd rather for django-components to go in the similar direction as django-ninja. Still work within the Django framework, but provide the best possible experience for users.

I believe your point on pushing django-components upstream is coming from the point of concern for maintainability, and that's a very valid point (correct me if wrong). But in that case you have to educate me on this - have there been similar projects which have been integrated into code Django codebase like this? And has there already been any discussions on merging django-components into django? I haven't researched this so I'm unable to consider this side of the argument.

My concern is that I'm not entirely sure why Django would want to merge django-components, when they already have `include`, `block` and `extends` tags. Sure, django-components expands on this and does things a bit better. But by the same logic then I would expect also an effort from Django maintainers to integrate popular features from django-ninja. (Again, I don't know if there are or are not such efforts, as I simply haven't come across them).

Moreover, because of my feeling that Django spreads itself too thin, and because Django has much larger maintenance burden because of the framework's age and popularity, I feel like merging django-components into django effectively means an end to new features (maybe a few once a year). Which in itself is not necessarily a bad thing, but because I feel like there's still so much to do on django-components to improve dev experience, merging to django would hinder that.

Then, what could be a sustainable alternative to offloading django-components to django? Could be either minimizing support/development hours for django-components (declaring the project as feature-complete), or educating more people on the project so it can be community-driven, or getting sponsorship/donations to make maintenance sustainable. While first option is easily achievable, it would also mean an end to better user experience. Ofc, the latter two are easier said than done.   However, IMO this project still has a plenty of room to grow and community members to attract, which could make the latter two options feasible. Personally, in this space I'm still missing a UI component library like what [Vuetify](https://github.com/vuetifyjs/vuetify) does for Vue.

Also, important to point out, is that things don't have to be either-or. If it came down to actual merge of django-components into django, I would expect that there would still be plenty of API changes or deletions, as probably not all features would be accepted. So what would live in Django would not be 100% copy of this project. What means that this project can afford some deviances (altho, ofc still minimizing them if possible). Or, you can always take an older version with less features but better compatibility.

---

To understand where I'm coming from - I have deep experience with React/Vue/Vuetify, so I have a good sense for what I'm missing. I also for long wanted to work on open source project, and (for personal reasons) it's only now that I'm ticking things off of my bucket list. And, by chance, it seems landed on this project. I'm also currently at stage of life where I have very few family or other duties (still need to work part time tho). Hence the drive and the belief that this project could still be improved.

So with this in mind, yesterday I estimated that to get in many of the features that I miss (ofc, only if accepted), I could work on django-components for next ~ 3-4 months (assuming 1 feature per week). But as I said above, that would also mean going in the direction of e.g. django-ninja, prioritizing user experience over compatibility with Django codebase.

So I'd be interested to hear what you think about that.

JuroOravec:
@EmilStenstrom on the topic of `{% component attrs:x-data="..." %}` vs `{% component my_array:x-data="..." %}`

Have a look at https://vuejs.org/guide/components/attrs#attribute-inheritance, this was where I was coming from. There, you can see that attributes are collected into `$attrs`. Component user can specify any extra attrs like `x-data=...` or `@click="..."`, it's all collected into `$attrs`, and then it's the component _author_ that decides where the attributes are rendered. So it's as if the component author chooses the prefix name.

Also to me it feels like it makes sense to allow only the special `attrs:` prefix, because it has special meaning ("These inputs will be rendered as HTML attributes (most likely) of the top-level HTML tag"). Because if I have something else like `my_array:abc=def`, then it doesn't have the meaning of HTML attributes anymore, and it's effectively only collecting inputs. So it would be as if instead of:
```py
def get_context_data(self, abc):
	my_array = {"abc": abc}
	return {"my_array": my_array}
```
we do it for them:
```py
def get_context_data(self, my_array):
	return {"my_array": my_array}
```

To me, this mental models feels like a leaky abstraction. If `my_array:abc=def` doesn't have a special meaning, then why use `my_array:abc` if I could instead use something else like `my_array__abc=def` or simply `abc=def`. Why would the component user need/want to know what variable is used inside of the component to group the inputs?

Also, if we think of `my_array:some=value` as simply grouping inputs, then it also raises questions like:
- Can a component user apply the prefix multiple times to create a nested input? E.g. `my_data:my_array:abc=def`

EmilStenstrom:
@JuroOravec I agree with you that a great developer experience, definitely can go hand in hand with great maintainability. It requires that we are careful in taking on technical debt though.

### About merging django-components into Django
You are right that it's not at all clear that this is possible. There are a couple of instances where this has happened though: South, that became django's migration system, Channels, which is now maintained in parallel with Django, and Crispy forms which was the blueprint for improvements in Django's form handling.

The core of why I think being part of Django is interesting is that **Django needs this**. It needs a better story of how to build complex interfaces with it. Currently people are not able to use big parts of the library and instead fall back to just using DRF or django-ninja. Django can be much more than an API builder, and I think this project might be the solution to that. I actually held a presentation about this almost 10 years ago: https://www.youtube.com/watch?v=Niq-HoraNPo (slightly dated, but the same core idea).

Does this mean that we **have to** be part of Django? Not necessarily, but I think the prospect of helping 100x more people "see the light" is a worthy goal. We could put django-components in a state where it could be included, and see if we could get it merged. If not, then fine, no harm done, and we could continue like it has for years.

### What would it take to be included?
That what we build would match the other things that are already in Django. That we added as few new concepts as possible. That the code was maintainable and easy to understand. That we had a clear story around how you should use components, and were the scope of the project ended, and customization left off.

I think the above are all worthy goals, anyways, that would also make the project easier to explain and maintain over time.

### What would happen after an eventual merge?
This library could be a "components-extended" package, including all the parts that didn't get merged into django. It would take the core parts of the library and move that into django core. After a couple of years we could remove those parts from our code. Would things significantly change for this project? I think not. Would django benefit? I think it definitely would!

I'm not at all saying we should stop developing the library, put it into maintenance mode, or anything like that.

EmilStenstrom:
@JuroOravec I'm extremely happy that you are able to put some time into working on this project, and I think donating time like this should mean that you have an even greater say in where the project is going. And I share your vision of building a great developer experience for users. So I'll just share my image of what a great experience means, to see if we share a similar view. 

### How to build a great developer experience
I think the most important dev experience is **how quickly you can onboard new users**. I have done a couple of stints as a programming teacher, teaching beginners how Django works from scratch. I saw first hand how all the new concepts, sometimes introduced to make dev experience easier for advanced users, confused new users just because they had to learn something new. This is actually the reason why I proposed Django drop regexps from urls and switch to something easier, leading to the current path() function that is now the default: https://groups.google.com/g/django-developers/c/u6sQax3sjO4/m/l3nvZCuuAAAJ

Sometimes simplifying for new users colldes with adding new stuff for experienced users. I think this is very tricky, as maintainers of a library always are the most advanced users. That's why I really appreciate all the great stuff that you've done to simplify things, merging settings and possibly removing the middleware. Of course we should also build for more advanced users, but I feel very strongly about "first user experience", and will keep advocating for them! :) 

### My view of React
I have a little experience with React, but not enough that I have built anything substantial. I find it super confusing, with lots of magic going on, and way too many concepts crammed into a library that is supposed to be usable to beginners, while in reality being built for Facebook's needs. You even have to learn a new language, JSX, that has it's own kinks, to be productive. There's build steps and hydration, and fibers, and hooks, and, and and... All this in addition to understanding Django. 

This just can't be the best way to build a website in 2024!

That's why I really appreciate HTMX and Alpine.js. They are trying to simplify things for everyone. And building on a "heavy" backend framework like Django, it makes sense the the frontend is light. I have talked about this before, but in the ideal world, I think being able to build interactive sites fully in python, not caring about the js layer at all, could be a really good goal to reach. I've talked before about [Phoenix Liveview](https://www.youtube.com/watch?v=AsyGYpQFiIA), which I think is a really elegant way to build heavy backend fullstack apps.

EmilStenstrom:
> @EmilStenstrom on the topic of `{% component attrs:x-data="..." %}` vs `{% component my_array:x-data="..." %}`
> 
> Have a look at https://vuejs.org/guide/components/attrs#attribute-inheritance, this was where I was coming from. There, you can see that attributes are collected into `$attrs`. Component user can specify any extra attrs like `x-data=...` or `@click="..."`, it's all collected into `$attrs`, and then it's the component _author_ that decides where the attributes are rendered. So it's as if the component author chooses the prefix name.

I definitely see your use-case here. As with all these features, I think it makes sense to back up one step from their implementation to the problem they are solving, and try to figure out what makes the most sense for this project. I don't think it's important to pick a solution that is similar to another project. That

> Also to me it feels like it makes sense to allow only the special `attrs:` prefix, because it has special meaning ("These inputs will be rendered as HTML attributes (most likely) of the top-level HTML tag"). 

I though of the interface differently: We're building a way to send an array of parameters to the component, instead of just key=value pairs that are supported today. I know I talked about a HtmlDict, which I agree doesn't work with the mental model I talk about above. Maybe the merge_attrs component tag could handle all the html specific stuff? 

This would mean that {% component variable:key1=value1 variable:key2=value2 %} would be a way to send the dict variable={key1: value1, key2: value2} to the get_context_data method, so that it could do whatever it wanted with it. Maybe it passes it to the template, maybe it does something else, doesn't matter. The elegant part to me, is that the name you pick, in this case "variable", will be present as a get_context_data parameter, so you can easily track where stuff is ending up. A concept like Vue's "fallthrough" saves some characters, but adds another concept to the mix that you have to learn. I think it goes against Python's "explicit over implicit" ethos...

I'm open to changing my mind here, just trying to explain what I'm thinking.

> If `my_array:abc=def` doesn't have a special meaning, then why use `my_array:abc` if I could instead use something else like `my_array__abc=def` or simply `abc=def`. Why would the component user need/want to know what variable is used inside of the component to group the inputs?

my_array__abc=def would be an OK api for me too. 

> Also, if we think of `my_array:some=value` as simply grouping inputs, then it also raises questions like:
> 
>     * Can a component user apply the prefix multiple times to create a nested input? E.g. `my_data:my_array:abc=def`

I have a very hard time to see who would find that useful, so I think we can just specify that this is something we support for just one level of arrays.

JuroOravec:
@EmilStenstrom Btw, just want to say, whether these concept may get into Django codebase or not, I think it's still extremely useful to have that sort of "selective pressure" from your side (or as you said, advocating for "first user experience"), really appreciate it, and hope my message didn't come off as too strong.

And thanks for the examples of projects that became part of Django.

Agree with you on wanting people to see the light, altho I'm maybe on the advanced end of the spectrum - but I'm still annoying that with Vue+Vuetify I can build a whole webpage in a day, whereas with Django (+components), I spend a day just creating an autocomplete field. And then I see someone on the Django forums who shrugs at the idea of using JS because "it's a mess", and I'm just perplexed at their ignorance. (ofc, there may be a bias in Django community to be more "anti-other-programming-languages", because those who can/want use other languages for web apps may find better tooling elsewhere and not come back). Anyway...

Agree that having a clear story is definitely useful to have either way!

And also like the idea that this could potentially shift into "components-extended". Overall, it is encouraging, hearing from you that you too believe that two ideas (keep improving the project vs being open to pushing some concepts to upstream) don't have to be exclusive!

Personally, achieving "everything in Python" is not a goal for me, because I have experience with the whole stack, so I don't feel the need. But still happy to help if that becomes one of the goals. I just won't be the one to spearhead that, as I don't have the end goal vision like I have for the templating part of django-components.

### On React / first user experience

I keep mentioning Vue AND React so there's higher chance people can relate with their experience. But truth be told, my inspiration is 99% Vue. To Vue, React feels low-level / unpolished, requiring a lot of micromanagement, and being more chaotic because it has less prescribed structure*. In other words, I think Vue has better developer experience.

\* But also here there's more nuance - here I'm talking about variables / component state. When it comes purely to templating, they have different approaches, but not necessarily better/worse.

Sidenote: That's one thing that I find ironic, that in JS ecosystem, there is (or at least was) a sentiment that "React has less magic than Vue and is therefore more straightforward to understand". But then Vue has more guardrails, and it requires more mental power to maintain React apps than it does for Vue. So as always, there's no silver bullets, no absolutes, and there's always trade-offs.

Also for django-components, Vue is much more relevant, simply because the template is separated from the code. In React, you have JSX where you directly inline components as variables. In Vue, you render a component by "referencing" the component via it's registered name (Which is the same how django-components does it).

So basically for me, I want to bring the ease of use of Vue+Vuetify to Django via components + Alpine/HTMX.

JuroOravec:
@EmilStenstrom 

> This would mean that {% component variable:key1=value1 variable:key2=value2 %} would be a way to send the dict variable={key1: value1, key2: value2} to the get_context_data method

Ok, we can also go in this direction. So that means that `attrs:` would NOT be a special prefix. In that case, I'm thinking of renaming the `merge_attrs` tag to `html_attrs`, as it's about taking _any_ dict, and formatting it as a string of HTML attributes.

So in terms of documenting it, I'm thinking:
- We should show the example with `{% component variable:key1=value1 variable:key2=value2 %}`
- For `html_attrs`, I suggest we show an example where we call the variable `attrs:`, e.g. `{% component attrs:key1=value1 attrs:key2=value2 %}`, so it's at least the "conventional" variable name for HTML attributes.
- We should explain the caveat that `variable:` cannot be nested.

EmilStenstrom:
> @EmilStenstrom Btw, just want to say, whether these concept may get into Django codebase or not, I think it's still extremely useful to have that sort of "selective pressure" from your side (or as you said, advocating for "first user experience"), really appreciate it, and hope my message didn't come off as too strong.

Absolutely not! In open source the rarest resource is time, so all arguments should be weighted by the time you are willing to spend. See everything I say as opinions, and I hope I didn't come off as too strong either.

> And thanks for the examples of projects that became part of Django.
> 
> Agree with you on wanting people to see the light, altho I'm maybe on the advanced end of the spectrum - but I'm still annoying that with Vue+Vuetify I can build a whole webpage in a day, whereas with Django (+components), I spend a day just creating an autocomplete field. 

This is actually a really good use-case. Maybe a good subgoal could be to make creating an autocomplete field really straightforward? Something similar to https://dhc.iwanalabs.com/cascading_selects/ perhaps?

> And then I see someone on the Django forums who shrugs at the idea of using JS because "it's a mess", and I'm just perplexed at their ignorance. (ofc, there may be a bias in Django community to be more "anti-other-programming-languages", because those who can/want use other languages for web apps may find better tooling elsewhere and not come back). Anyway...

I think it's partly because of ignorance, as you say, but also partly because of the mental load of being in two development stacks at once. Knowing all the ins and outs, and that's not even including all the additional libraries you need to learn, and keep in your head. For advanced users like us, it's easy to forget the huge set of knowledge it takes to understand say Django + requirements.txt + Node + React + package.json. SOOO many concepts!

> Agree that having a clear story is definitely useful to have either way!
> 
> And also like the idea that this could potentially shift into "components-extended". Overall, it is encouraging, hearing from you that you too believe that two ideas (keep improving the project vs being open to pushing some concepts to upstream) don't have to be exclusive!
> 
> Personally, achieving "everything in Python" is not a goal for me, because I have experience with the whole stack, so I don't feel the need. But still happy to help if that becomes one of the goals. I just won't be the one to spearhead that, as I don't have the end goal vision like I have for the templating part of django-components.

Fully understand. Back to the open source ethos: it's "free" work, so everyone needs to find motivation for their work elsewhere, not doing what other people tell you to. That's work! :)

> ### On React / first user experience
> 
> I keep mentioning Vue AND React so there's higher chance people can relate with their experience. But truth be told, my inspiration is 99% Vue. To Vue, React feels low-level / unpolished, requiring a lot of micromanagement, and being more chaotic because it has less prescribed structure*. In other words, I think Vue has better developer experience.

I should look into Vue again. Read up on it many years ago, but didn't get to building anything.

> So basically for me, I want to bring the ease of use of Vue+Vuetify to Django via components + Alpine/HTMX.

That's a really good goal I think. We want to make Alpine/HTML + Django a much better developer experience than Django + DRF/django-ninja + Vue/React (the most common development stack right now).

EmilStenstrom:
> @EmilStenstrom
> 
> > This would mean that {% component variable:key1=value1 variable:key2=value2 %} would be a way to send the dict variable={key1: value1, key2: value2} to the get_context_data method
> 
> Ok, we can also go in this direction. So that means that `attrs:` would NOT be a special prefix. In that case, I'm thinking of renaming the `merge_attrs` tag to `html_attrs`, so it's more declarative, as it's about taking _any_ dict, and formatting it as a string of HTML attributes.

Makes sense to have html_attrs be the name instead.

> So in terms of documenting it, I'm thinking:
> 
>     * We should show the example with `{% component variable:key1=value1 variable:key2=value2 %}`

Not sure. The most common use-case will be setting HTML attributes, so maybe it still makes sense to have those as examples? I think the alpine params that you used shows the reason behind the feature well. It's just that now you can have two attrs and wrapper_attrs at the same time.

>     * For `html_attrs`, I suggest we show an example where we call the variable `attrs:`, e.g. `{% component attrs:key1=value1 attrs:key2=value2 %}`, so it's at least the "conventional" variable name for HTML attributes.

Yeah, or we could do a tailwind example: `attrs:class="m-0" attrs:class="flex-wrap-reverse"`? You decide.

>     * We should explain the caveat that `variable:` cannot be nested.

Sure!

JuroOravec:
@EmilStenstrom I've refactored this so that the construct `prefix:key=val` can be used for any props.  I also changed it from single colon `:` to two colons `::`, so it becomes `prefix::key=val`

The reason for using `::` is that I realized that I don't know if there are any frameworks or libraries that make use of colon in HTML attribute name. But if there is, then we wouldn't be able to distinguish it from our prefix. So it would be safer to use `::`. Coincidentally, I feel like this makes it also easier to read, and it's similar to [scoped variables in some programming languages](https://en.wikipedia.org/wiki/Scope_resolution_operator), so hopefully it'd be more intuitive at a glance.

I kept the ability of the `html_attrs` tag to "append" values, so we can join `class` attribute that is coming from component user with `class` defined by component author. But instead of `key+=val`, I reused the `prefix::key=val` construct. So user's can set `add::class="extra-class"` to indicate that the value `"extra-class"` should be appended to whatever is set to attribute `class`.

I've also updated README, trying to document this from user's perspective, introducing one concept at a time.

And since we don't treat `class` as a special attribute (like the code from django-web-components originally did), I was able to shave off about half of the code + tests copied from django-web-components.

JuroOravec:
@EmilStenstrom 

> This is actually a really good use-case. Maybe a good subgoal could be to make creating an autocomplete field really straightforward? Something similar to [dhc.iwanalabs.com/cascading_selects](https://dhc.iwanalabs.com/cascading_selects/) perhaps?

Yeah, that's a good goal. Creating pre-made components will be a whole new can of worms, but it would be IMO very valuable. However, I already feel like there might be a desire for 2 different libraries, one for AlpineJS and one for HTMX (unless we figure out how to merge them). And there's also the question of supporting Jinja2 vs Django templates. So definitely a lot of things to consider before we get to making components.

> For advanced users like us, it's easy to forget the huge set of knowledge it takes to understand say Django + requirements.txt + Node + React + package.json. SOOO many concepts!

Very true!

EmilStenstrom:
> @EmilStenstrom I've refactored this so that the construct `prefix:key=val` can be used for any props. I also changed it from single colon `:` to two colons `::`, so it becomes `prefix::key=val`
> 
> The reason for using `::` is that I realized that I don't know if there are any frameworks or libraries that make use of colon in HTML attribute name. But if there is, then we wouldn't be able to distinguish it from our prefix. So it would be safer to use `::`. 

`:` is a character reserved for XML namespaces, so it can't be used for other things in HTML. https://stackoverflow.com/questions/16021123/is-colon-valid-in-attribute-names-for-html5 - so it would be safe to use for this.

> Coincidentally, I feel like this makes it also easier to read, and it's similar to [scoped variables in some programming languages](https://en.wikipedia.org/wiki/Scope_resolution_operator), so hopefully it'd be more intuitive at a glance.

Double colon feels a bit "foreign" I think, but so does the colon, so you decide :)

> I kept the ability of the `html_attrs` tag to "append" values, so we can join `class` attribute that is coming from component user with `class` defined by component author. But instead of `key+=val`, I reused the `prefix::key=val` construct. So user's can set `add::class="extra-class"` to indicate that the value `"extra-class"` should be appended to whatever is set to attribute `class`.

It feels strange to have reserved prefixes like that. Could we instead just say that if you use the same variable that means you add an attribute? `{% merge_attrs attrs attrs::class="some-class" %}`?

> I've also updated README, trying to document this from user's perspective, introducing one concept at a time and finishing it with a full example.

Awesome! I see a couple of mentions of merge_attrs in the docs, I think they should be html_attrs.

> And since we don't treat `class` as a special attribute (like the code from django-web-components originally did), I was able to shave off about half of the code + tests copied from django-web-components.

Awesome! I think this is shaping up to be a really nice feature. Thanks for being patient with all my side-thoughts :)

JuroOravec:
> `:` is a character reserved for XML namespaces, so it can't be used for other things in HTML. [stackoverflow.com/questions/16021123/is-colon-valid-in-attribute-names-for-html5](https://stackoverflow.com/questions/16021123/is-colon-valid-in-attribute-names-for-html5) - so it would be safe to use for this.

Ok, that's great, that means we can use just single colon!

> It feels strange to have reserved prefixes like that. Could we instead just say that if you use the same variable that means you add an attribute? `{% merge_attrs attrs attrs::class="some-class" %}`?

I don't like using a same variable `attrs::`, because then we're breaking expectations on how `prefix::` should work, as seen in `component` tag. Sure enough, the suggested `add::` also doesn't work 100% the same as in `component` tag (we can pass `add::class` multiple times to `html_attrs`, but we cannot do that in `component`), but IMO then it at least feels like the input to `html_attrs` tag would be `{% html_attrs attrs add key1=val1 key2=val2 ... %}`, where both `attrs` and `add` are dicts.

Since it's safe to use single colon, then one more options comes to mind - we could simply use `:class="abc"` instead of `add:class="abc"`. Similarly how in [django-web-components](https://github.com/Xzya/django-web-components/tree/master?tab=readme-ov-file#slot-attributes), they use colon-prefixed `:let`. Altho then it might be harder to distinguish between the "appended" value and "default" values, e.g. `{% html_attrs attrs key1=val1 :key1=val1 %}`,

JuroOravec:
One more idea on this:

> It feels strange to have reserved prefixes like that. Could we instead just say that if you use the same variable that means you add an attribute? `{% merge_attrs attrs attrs::class="some-class" %}`?

Above, I suggested interface like

`{% html_attrs attrs add key1=val1 key2=val2 ... %}`

But actually, we could avoid the problem with trying to distinguish between "defaults" and "appends" by switching it up:

`{% html_attrs attrs defaults key1=val1 key2=val2 ... %}`

So all `key1=val1` key2=val2`, etc, they would be appended.

And the `attrs defaults` would be parsed the same way we parse components, so component author could do:
```django
{% html_attrs attrs
	defaults:class="some-class"
	defaults:data-id="123"
	class="extra-class"
	class=class_var
%}
```

JuroOravec:
Ok, I like this last approach, with `attrs` and `defaults`, where we're treating them _almost_ the same as component kwargs (The only difference is that we allow them to be unset).

So the ideas are:
- Both `attrs` and `defaults` can be passed as positional args, or as kwargs (`attrs=attrs`, `defaults=defaults`), same as we do for component name
- Both `attrs` and `defaults` are optional (can be omitted)
- Both `attrs` and `defaults` are dicts, so they can be set as `attrs=attrs` or as `attrs:key=value ...`
- All other kwargs are appended

Examples:

Assuming that:
```py
class_from_var="from-var"
attrs = {
	"class": "from-attrs",
	"type": "submit",
}
defaults = {
	"class": "from-defaults",
	"role": "button",
}
```

- Empty tag
	`{% html_attr %}`
	renders:
	` `

- ONLY kwargs - these would all be appended
	`{% html_attr class="some-class" class=class_from_var data-id="123" %}`
	renders:
	`class="some-class from-var" data-id="123"`

- ONLY attrs
	`{% html_attr attrs %}`
	renders:
	`class="from-attrs" type="submit"`

- attrs as kwarg
	`{% html_attr attrs=attrs %}`
	renders:
	`class="from-attrs" type="submit"`

- ONLY defaults (as kwarg)
	`{% html_attr defaults=defaults %}`
	renders:
	`class="from-defaults" role="button"`

- attrs using the `prefix:key=value` construct
	`{% html_attr attrs:class="from-attrs" attrs:type="submit" %}`
	renders:
	`class="from-attrs" type="submit"`

- defaults using the `prefix:key=value` construct
	`{% html_attr defaults:class="from-defaults" %}`
	renders:
	`class="from-defaults" role="button"`

- All together (1) - attrs and defaults as positional args:
	`{% html_attrs attrs defaults class="added_class" class=class_from_var data-id=123 %}`
	renders:
	`class="from-attrs added_class from-var" type="submit" role="button" data-id=123`

- All together (2) - attrs and defaults as kwargs args:
	`{% html_attrs class="added_class" class=class_from_var data-id=123 attrs=attrs defaults=defaults %}`
	renders:
	`class="from-attrs added_class from-var" type="submit" role="button" data-id=123`

- All together (3) - mixed:
	`{% html_attrs attrs defaults:class="default-class" class="added_class" class=class_from_var data-id=123 %}`
	renders:
	`class="from-attrs added_class from-var" type="submit" data-id=123`

JuroOravec:
Updated the code, tests, and README to the API described in my last 2 comments.

@EmilStenstrom @dylanjcastillo Would you give this one a review?

EmilStenstrom (commented):
I think this is really good API design. Very happy for this feature Juro!

~~In the review I found one question: prefixes are fallthrough, which I'm not sure is the right choice. Wouldn't be sane to require all params used to be available in get_context_data?~~

Found one place with old docs too, an easy fix.

Overall, this is a great PR ⭐

README.md
```diff
@@ -701,10 +705,381 @@ class Calendar(component.Component):
...
+        {% component "other"
+            attrs:class="pa-4 flex"
+            attrs:data-some-id=some_id
+            attrs:@click.stop="onClickHandler"
+        %}
+        {% endcomponent %}
+    """
+
+    def get_context_data(self, some_id: str):
+        return {"some_id": some_id}
```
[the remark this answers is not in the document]
JuroOravec: You are right, and the example does work that way. The `get_context_data` is for component `my_comp`, where we don't need to specify anything, because it's the parent. If we've shown also code for component `other`, then `other`'s `get_context_data` must include `attrs` among the inputs.

I can add a dummy code for the `other` component to avoid the confusion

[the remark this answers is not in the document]
EmilStenstrom: Oh, my fault! No need to add anything else then.

README.md
```diff
@@ -701,10 +705,381 @@ class Calendar(component.Component):
...
+<div data-value="my-class pa-4 some-class">
+</div>
+```
+
+To simplify merging of variables, you can supply the same key multiple times, and these will be all joined together:
+
+```django
+{# my_var = "class-from-var text-red" #}
+<div {% html_attrs attrs class="some-class" class=my_var %}>
+</div>
```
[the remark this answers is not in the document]
JuroOravec: Yup! :) I will update the example to answer this question

tests/test_attributes.py
```diff
@@ -0,0 +1,364 @@
...
+        self.assertEqual(
+            attributes_to_string({"required": False}),
+            "",
+        )
+
+    def test_attribute_with_true_value(self):
+        self.assertEqual(
+            attributes_to_string({"required": True}),
+            "required",
+        )
```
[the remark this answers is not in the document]
JuroOravec: Do you mean the `attributes_to_string` function? Didn't think of that, but it does make sense to document it, so people can use the feature also outside of templates.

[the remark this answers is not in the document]
EmilStenstrom: @JuroOravec No, that attributes with True gets translated to a "bare" attribute.

src/django_components/templatetags/component_tags.py
```diff
@@ -248,6 +249,29 @@ def do_component(parser: Parser, token: Token) -> ComponentNode:
...
+    ```
+
+    Normal kwargs (`key=value`) are treated as defaults, meaning that if the same
+    key is set on the given dictionary, the dict takes precedence.
+
+    Kwargs that start with `add:` will be concatenated to existing keys. So if e.g.
+    key "class" is supplied from the dict or another kwarg, with value "my-class",
+    then adding `add:"extra-class"` will result in value `"my-class extra-class"
+    for key "class".
+    """
```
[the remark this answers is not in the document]
JuroOravec: Updated in https://github.com/EmilStenstrom/django-components/pull/491/commits/39ba5fe1c579d4f1b87423086525b06bbaa3302e

tests/test_attributes.py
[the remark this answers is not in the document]
JuroOravec: Updated in https://github.com/EmilStenstrom/django-components/pull/491/commits/4cbafd55576408dc9ccd258116a076cf87358a45.

I also documented the `attributes_to_string`, because I can imagine that otherwise there might be someone who would do the following just to get that behavior:
```py
Template("{% html_attrs attrs ... %}").render(Context({ "attrs": attrs, ... }))
```

JuroOravec:
Thanks @EmilStenstrom, I'm also happy with how it turned out! And thanks again for all the feedback, we wouldn't have arrived at this without it 😄

And thanks for the review, I've updated the docs.

EmilStenstrom (approved):

Issue #494, opened 2024-05-12
Scoped slots - Passing data to slots and accessing them in fills


JuroOravec:
After https://github.com/EmilStenstrom/django-components/pull/491 I wanted to do a small one now.

Another feature that I'm missing as a component author is what [Vue calls "scoped slots"](https://vuejs.org/guide/components/slots.html#scoped-slots). Basically, it means that **you can pass data to slots and then access them in fills**. 

### Why is it useful?

Consider a component with slot(s). This component may do some processing on the inputs, and then use the processed variable in the slot's default template:

```py
@component.register("my_comp")
class MyComp(component.Component):
	template = """
		<div>
			{% slot "content" default %}
				input: {{ input }}
			{% endslot %}
		</div>
	"""

	def get_context_data(self, input):
		processed_input = do_something(input)
		return {"input": processed_input}
```

Currently, if I decide to override the "content" slot, then I cannot make use of the `input` variable (if I'm using the `isolated` option for `context_behavior`). I would either have to recompute `input` myself, or pass the slot content as a function to MyComp. Recomputing is not ideal if the computation is expensive. And passing slot fills as functions means we cannot separate Python and HTML logic.

Vue's approach is that instead, the component with the slot can decide to pass some variables to the slot. And these variables can then be _optionally_ used by the slot fill:

```vue
<!-- Child component -->
<div>
	<!-- Note: "name" identifies the slot, same as we do {% slot "content" %} -->
	<slot name="content" :text="greetingMessage" :count="1" />
</div>

<!-- Parent component -->
<Child>
	<template #content="slotProps">
		text: {{ slotProps.text }}
		count: {{ slotProps.count }}
	</template>
</Child>
```

This has also been implemented in [django-web-component (DWC)](https://github.com/Xzya/django-web-components/tree/master?tab=readme-ov-file#scoped-slots). DWC allows to pass only one variable to the slot. The name under which it is then accessed in the slot fill is defined by `:let`. Also notice that the example below uses a default slot, so `:let` is defined on the component itself.

```django
{# unordered_list component #}
<li>
    <!-- We are passing the `entry` as the second argument to render_slot -->
    {% render_slot slots.inner_block entry %}
</li>

{# parent component #}
{% unordered_list :let="fruit" entries=entries %}
    I like {{ fruit }}!
{% endunordered_list %}
```

---

### Implementation notes:

- API for passing data to slots - I'm thinking about two approaches:
	1. All extra slot kwargs are passed through:
[...]

Issue #478, opened 2024-05-04
Allow to render component dependencies without middleware


JuroOravec:
Continuation from this comment https://github.com/EmilStenstrom/django-components/issues/277#issuecomment-2092358246

To refine the previous idea a bit more, I suggest following API for users to render dependencies:

1. Preferably, use `{% component_dependencies %}` (or the JS/CSS variants) in the top-level component BEFORE any other tags that could render components.

2. Alternatively, instead of `{% component_dependencies %}` (or the JS/CSS variants) users may also use `track_dependencies` for the same effect:
	```py
	track_dependencies(
		context,
		lambda ctx: Template("""
			{% component 'hello' %}{% endcomponent %}
		""").render(ctx),
	) -> str:
	```

3. If users need to place `{% component_dependencies %}` (or the JS/CSS variants) somewhere else than at the beginning of the top-level component, then we will need to the replacement strategy. I suggest to still use the `track_dependencies`. Basically, if the template given to `track_dependencies` contains `{% component_dependencies %}` tag, then we do the replacement strategy. If `{% component_dependencies %}` is NOT present, then we first render the text and collect all dependencies, and prepend them to the rendered content.

	For users, the API would still be the same as in 2.:
	```py
	track_dependencies(
		context,
		lambda ctx: Template("""
			{% component 'hello' %}{% endcomponent %}
			{% component_dependencies %} <-- comp_deps tag found, so replacement strategy used
		""").render(ctx),
	) -> str:
	```

4. And if users do not want to call `track_dependencies` for each template render, they can use the Middleware (as it is currently), with the caveat that it works only for non-streaming responses and of content type `text/html`.

EmilStenstrom:
To complicate things, I think many will put component_dependencies_js in the footer of the page (where it doesn’t block rendering), and the css-version in the head.

I’m wondering if we’re mixing two concepts here? “Where do I want my link tag” vs “I want to dynamically load my dependencies”. 

Is adding a new tag another option?

about the track_dependencies-function, where would the user put it?

JuroOravec:
> To complicate things, I think many will put component_dependencies_js in the footer of the page (where it doesn’t block rendering), and the css-version in the head.

Hm, ok, so in that case the point 1. won't work very well.

> I’m wondering if we’re mixing two concepts here? “Where do I want my link tag” vs “I want to dynamically load my dependencies”.

Yeah, 

> Is adding a new tag another option?

[...]

PR #495, opened 2024-05-13
feat: Scoped slots + Updated docs

JuroOravec:
~Implementation may still change based on feedback.~ This MR is now ready for review, it adds "scoped slot" feature as per https://github.com/EmilStenstrom/django-components/issues/494#issuecomment-2113363070.

I also moved the README sections around a bit, and added a view missing sections. See  the screenshot in https://github.com/EmilStenstrom/django-components/pull/495#discussion_r1598567122 and other comments below.

Closes https://github.com/EmilStenstrom/django-components/issues/494

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Read on to learn about the details!
 
 This change is done to simplify the API in anticipation of a 1.0 release of django_components. After 1.0 we intend to be stricter with big changes like this in point releases.
 
-**Version 0.34** adds components as views, which allows you to handle requests and render responses from within a component. See the [documentation](#components-as-views) for more details.
+**Version 0.34** adds components as views, which allows you to handle requests and render responses from within a component. See the [documentation](#use-components-as-views) for more details.
 
 **Version 0.28** introduces 'implicit' slot filling and the `default` option for `slot` tags.
 
@@ -237,19 +237,43 @@ class Calendar(component.Component):
 
 And voilá!! We've created our first component.
 
-## Autodiscovery
+## Using single-file components
 
-By default, the Python files in the `components` app are auto-imported in order to auto-register the components (e.g. `components/button/button.py`).
+Components can also be defined in a single file, which is useful for small components. To do this, you can use the `template`, `js`, and `css` class attributes instead of the `template_name` and `Media`. For example, here's the calendar component from above, defined in a single file:
 
-Autodiscovery occurs when Django is loaded, during the `ready` hook of the `apps.py` file.
+```python
+# In a file called [project root]/components/calendar.py
+from django_components import component
+from django_components import types as t
 
-If you are using autodiscovery, keep a few points in mind:
+@component.register("calendar")
+class Calendar(component.Component):
+    def get_context_data(self, date):
+        return {
+            "date": date,
+        }
 
-- Avoid defining any logic on the module-level inside the `components` dir, that you would not want to run anyway.
-- Components inside the auto-imported files still need to be registered with `@component.register()`
-- Auto-imported component files must be valid Python modules, they must use suffix `.py`, and module name should follow [PEP-8](https://peps.python.org/pep-0008/#package-and-module-names).
+    template: t.django_html = """
+        <div class="calendar-component">Today's date is <span>{{ date }}</span></div>
+    """
 
-Autodiscovery can be disabled via in the [settings](#disable-autodiscovery).
+    css: t.css = """
+        .calendar-component { width: 200px; background: pink; }
+        .calendar-component span { font-weight: bold; }
+    """
+
+    js: t.js = """
+        (function(){
+            if (document.querySelector(".calendar-component")) {
+                document.querySelector(".calendar-component").onclick = function(){ alert("Clicked calendar!"); };
+            }
+        })()
+    """
+```
+
+This makes it easy to create small components without having to create a separate template, CSS, and JS file.
+
+Note that the `t.django_html`, `t.css`, and `t.js` types are used to specify the type of the template, CSS, and JS files, respectively. This is not necessary, but if you're using VSCode with the [Python Inline Source Syntax Highlighting](https://marketplace.visualstudio.com/items?itemName=samwillis.python-inline-source) extension, it will give you syntax highlighting for the template, CSS, and JS.
 
 ## Use the component in a template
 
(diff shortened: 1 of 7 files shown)


src/django_components/component.py
```diff
@@ -433,20 +442,3 @@ def render(self, context: Context) -> str:
 
         trace_msg("RENDR", "COMP", self.name_fexp, self.component_id, "...Done!")
         return output
-
-
-def safe_resolve_list(args: List[FilterExpression], context: Context) -> List:
```
JuroOravec, line 438: Moved function that work with FilterExpression into file `expression.py`

src/django_components/templatetags/component_tags.py
```diff
@@ -278,23 +242,11 @@ def do_html_attrs(parser: Parser, token: Token) -> HtmlAttrsNode:
     ```
     """
     bits = token.split_contents()
-    attributes, default_attrs, append_attrs = parse_html_attrs_args(parser, bits, "html_attrs")
+    attributes, default_attrs, append_attrs = _parse_html_attrs_args(parser, bits, "html_attrs")
     return HtmlAttrsNode(attributes, default_attrs, append_attrs)
 
 
-def is_whitespace_node(node: Node) -> bool:
```
JuroOravec, line 285: These three `is_whitespace_node`,  `is_whitespace_token` and `is_block_tag_token` weren't being used. Actually, even in the MR that introduced them didn't use them. Since it's not documented, I assume it's not part of the public API.

src/django_components/templatetags/component_tags.py
```diff
@@ -393,7 +469,3 @@ def _get_positional_param(
 
 def is_wrapped_in_quotes(s: str) -> bool:
     return s.startswith(('"', "'")) and s[0] == s[-1]
-
-
-def strip_quotes(s: str) -> str:
```
JuroOravec, line 398: Instead of stripping quotes outselves, I used `FilterExpression.resolve()` to do that for us here https://github.com/EmilStenstrom/django-components/pull/495/files#diff-7345171ed38e1dd2e4b8199b65a2fd082323bdb331ff0ef0dc0dd3bd43f111aeR345

README.md
```diff
@@ -237,19 +237,43 @@ class Calendar(component.Component):
 
 And voilá!! We've created our first component.
 
-## Autodiscovery
+## Using single-file components
```
JuroOravec, line 240: Moved the section on "single-file components" and "components as Views" higher up. Might not be easy to see in this diff, and just my personal opinion, but it feels to me that now README flows even a bit better:
1. installation
2. creating components (and the different variants thereof)
3. registering componnt with autodiscovery
4. using slots and passing data to components (which requires the other component to be registered)
5. and them some more niche stuff

Overall the ordering is:

Screenshot 2024-05-13 at 16 19 07

README.md
```diff
@@ -460,86 +532,88 @@ This is fine too:
 {% endcomponent %}
 ```
 
-### Components as views
+### Render fill in multiple places
```
JuroOravec, line 535: Added this section since it wasn't documented yet.

README.md
```diff
@@ -638,35 +712,131 @@ To negate the meaning of `component_vars.is_filled`, simply treat it as boolean
...
+#### Accessing `is_filled` of slot names with special characters
 
 To be able to access a slot name via `component_vars.is_filled`, the slot name needs to be composed of only alphanumeric characters and underscores (e.g. `this__isvalid_123`).
 
 However, you can still define slots with other special characters. In such case, the slot name in `component_vars.is_filled` is modified to replace all invalid characters into `_`.
 
 So a slot named `"my super-slot :)"` will be available as `component_vars.is_filled.my_super_slot___`.
 
-### Setting Up `ComponentDependencyMiddleware`
+### Scoped slots
```
JuroOravec, line 723: Added section on scoped slots and moved `ComponentDependencyMiddleware` further down to its own section on rendering dependencies.

README.md
```diff
@@ -1175,6 +1349,40 @@ NOTE: `{% csrf_token %}` tags need access to the top-level context, and they wil
 
 Components can also access the outer context in their context methods by accessing the property `outer_context`.
 
+## Rendering JS and CSS dependencies
```
JuroOravec, line 1352: Moved the section on middleware here, as I expect that this section will be expanded when we tackle https://github.com/EmilStenstrom/django-components/issues/478

src/django_components/slots.py
```diff
@@ -77,6 +81,7 @@ class SlotFill(NamedTuple):
     nodelist: NodeList
     context_data: Dict
     alias: Optional[AliasName]
+    scope: Optional[ScopeName]
```
JuroOravec, line 84: How the feature is implemented is that when we parse `slot` tags, we note down the kwargs parsed to it, and pass it down. Similarly, when we parse `fill` tags, we note down the `slot_data` variable name, and pass it down. These two then meet when we are rendering a slot, and we assign the slot kwargs to the context under a key set in `fill` tag.

`slot_data` value is required to be string literal, e.g. `"quoted_string"`, same as we do for fill name or the `as var` syntax.

tests/test_templatetags.py
```diff
@@ -2392,3 +2392,277 @@ def test_inner_slot_iteration_nested_with_slot_default_and_outer_scope_variable_
...
+
+
+class ScopedSlotTest(BaseTestCase):
+    def test_slot_data_as_kwarg(self):
+        @component.register("test")
+        class TestComponent(component.Component):
+            template: types.django_html = """
+                {% load component_tags %}
+                <div>
+                    {% slot "my_slot" data=slot_data %}Default text{% endslot %}
```
EmilStenstrom: Oh? I expected the API to be `abc="def" 123="456"`?

EmilStenstrom (commented):
The code looks great. I thought the API was that you set the names of your parameters in the slot tag, but now it takes a data param with a dict instead?

tests/test_templatetags.py
JuroOravec: Hm, maybe you had old version of the MR loaded in the browser. The initial implementation used the `data` object, but then I changed it a couple of days back in https://github.com/EmilStenstrom/django-components/pull/495/commits/3f90d51994ed3143a5c402d680b8bdd71cecfd18. But there was still an old error message, I've fixed that just now!

JuroOravec:
I also added a table of contents to README

tests/test_templatetags.py
```diff
@@ -2392,3 +2392,215 @@ def test_inner_slot_iteration_nested_with_slot_default_and_outer_scope_variable_
...
+
+
+class ScopedSlotTest(BaseTestCase):
+    def test_slot_data(self):
+        @component.register("test")
+        class TestComponent(component.Component):
+            template: types.django_html = """
+                {% load component_tags %}
+                <div>
+                    {% slot "my_slot" abc=abc 123=var123 %}Default text{% endslot %}
```
EmilStenstrom, line 2404: Using 123 as an attribute here is super strange, as neither HTML or python allows them. But I guess OK that it still works? :)

tests/test_templatetags.py
```diff
@@ -2392,3 +2392,215 @@ def test_inner_slot_iteration_nested_with_slot_default_and_outer_scope_variable_
...
+            def get_context_data(self):
+                return {
+                    "abc": "def",
+                    "var123": 456,
+                }
+
+        template: types.django_html = """
+            {% load component_tags %}
+            {% component "test" %}
+                {% fill "my_slot" slot_data="slot_data_in_fill" as "slot_var" %}
```
EmilStenstrom, line 2485: I think we should change the API (separately) to a keyword arg instead of "as var".

tests/test_templatetags.py
```diff
@@ -2392,3 +2392,215 @@ def test_inner_slot_iteration_nested_with_slot_default_and_outer_scope_variable_
...
+        expected = """
+            <div>
+                Default text
+                def
+                456
+            </div>
+        """
+        self.assertHTMLEqual(rendered, expected)
+
+    def test_slot_data_raises_on_slot_data_and_as_same_var(self):
```
EmilStenstrom, line 2502: Nice catch!

tests/test_templatetags.py
```diff
@@ -2392,3 +2392,215 @@ def test_inner_slot_iteration_nested_with_slot_default_and_outer_scope_variable_
...
+                    "abc": "def",
+                    "var123": 456,
+                }
+
+        template: types.django_html = """
+            {% load component_tags %}
+            {% component "test" %}
+                {% fill "my_slot" %}
+                    overriden
+                    {{ slot_data_in_fill }}
```
EmilStenstrom, line 2553: Not sure what this test is testing, as this is just an undefined variable?

tests/test_templatetags.py
```diff
@@ -2392,3 +2392,215 @@ def test_inner_slot_iteration_nested_with_slot_default_and_outer_scope_variable_
...
+        template: types.django_html = """
+            {% load component_tags %}
+            {% component "test" %}
+                {% fill "my_slot" slot_data="data" %}
+                    {{ data|safe }}
+                {% endfill %}
+            {% endcomponent %}
+        """
+        rendered = Template(template).render(Context())
+        expected = "<div> {} </div>"
```
EmilStenstrom, line 2580: Minor: Where do these brackets come from? Are they empty dict -> str? Shouldn't the default be empty string instead?

EmilStenstrom (commented):
Really good work on this! 

I see no tests with context_behavior="isolated", so I guess that all of them are testing "django"?

tests/test_templatetags.py
JuroOravec: I was thinking the same. I've moved the discussion to https://github.com/EmilStenstrom/django-components/issues/502

tests/test_templatetags.py
JuroOravec: The undefined variable is not necessary, but the point of this test is that things still work even if user doesn't provide the `slot_data` kwarg

tests/test_templatetags.py
JuroOravec: Yes, it's from doing `{{ data|safe }}`.

The dict is not intended to be printed like I did with `{{ data }}`, I did so only for the sake of tests. So I don't like the idea of coercing empty dict to a string. Instead, it's more likely that the `slot_data` will be used as an object, accessing keys like `{{ data.some_key }}`. So we should keep it as a dict, so users don't get surprised when they try to access a key, and instead they accidentally access a string property/method.

tests/test_templatetags.py
JuroOravec: I removed the variable in https://github.com/EmilStenstrom/django-components/pull/495/commits/40c2b1af9b6878a922cd2456b76dbd69ba1e0fef

tests/test_templatetags.py
JuroOravec: Changed the key name in https://github.com/EmilStenstrom/django-components/pull/495/commits/40c2b1af9b6878a922cd2456b76dbd69ba1e0fef

JuroOravec:
@EmilStenstrom 
> I see no tests with context_behavior="isolated", so I guess that all of them are testing "django"?

No, `isolated` is still the default. Changing the default involves also changing some tests, as well as documentation. That's why I moved it to a separate issue in https://github.com/EmilStenstrom/django-components/issues/498. And I want to first clean up this one to avoid needlessly complex rebases/merges.

JuroOravec:
Btw, @EmilStenstrom, what do you think of renaming the kwarg through which we access the data from `slot_data` to simply `data`?

When we take also https://github.com/EmilStenstrom/django-components/issues/502 into consideration, then IMO the shorter names would make for a cleaner API. 

Consider:

```django
{% fill "my_slot" data="slot_data" default="slot_default" %}
```

vs

```django
{% fill "my_slot" slot_data="slot_data" slot_default="slot_default" %}
```

EmilStenstrom:
@JuroOravec Definitely like the data variable better.

EmilStenstrom (approved):

JuroOravec:
@EmilStenstrom Thanks again for your time and for all the feedback! I've changed `slot_data` to just `data`, and I'll go ahead and merge this.

Issue #458, opened 2024-04-29
"django_components.component" does not explicitly export attribute "register"


thacoon:
I have a component like this

```python
from typing import Dict, Any

from django_components import component

from server.example.models import Example


@component.register('example')
class ExampleComponent(component.Component):
    template_name = 'example/template.html'

    def get_context_data(self, instance: Example) -> Dict[str, Any]:
        return {
            'instance': instance,
        }
```

When running mypy I get the following error: `error: Module "django_components.component" does not explicitly export attribute "register"  [attr-defined]`

EmilStenstrom:
@thacoon How would we fix this, and are you open to writing a PR?

JuroOravec:
Looking at [this thread](https://github.com/konradhalas/dacite/issues/133), I guess a temporary workaround could be to either:
- Use
	```py
	from django_components.component_registry import registry
	```
- Set in mypy.ini
	```toml
	implicit_reexport = True
	```

What would be a proper fix? @thacoon can you share your mypy config?

thacoon:
@JuroOravec my latest configuration is:

```
[mypy]
# Mypy configuration:
# https://mypy.readthedocs.io/en/latest/config_file.html
enable_error_code =
    truthy-bool,
    truthy-iterable,
    redundant-expr,
    unused-awaitable,
    ignore-without-code,
    possibly-undefined,
    redundant-self,

explicit_package_bases = true
ignore_missing_imports = true
strict = true
warn_unreachable = true

plugins =
    mypy_django_plugin.main

[mypy-server.apps.*.migrations.*]
# Django migrations should not produce any errors:
ignore_errors = true

[mypy.plugins.django-stubs]
# Docs: https://github.com/typeddjango/django-stubs
django_settings_module = config.settings
strict_settings = false
```

EmilStenstrom:
I think the problem this mypy error is trying to point us at, is that we rely on the component module importing register as an API, but it's all implicit. So if we would refactor the module, and accidentally remove that import, we wouldn't notice that the API broke.

A simple way to explicityly define the public API is specifying __all__, and list all the symbols we export. Isn't that the simplest solution here?

JuroOravec:
@EmilStenstrom So we want to define `__all__` inside `src/__init__.py`, is that correct?

JuroOravec:
@thacoon I wasn't able to get the same error in my project, but it might be because there I'm on django_components 0.66
and mypy 1.7.0. What versions are you on?

[...]

PR #499, opened 2024-05-19
feat: alias component_registry imports for mypy

mikucz:
### Resolving issue:
- https://github.com/EmilStenstrom/django-components/issues/458

As asked by @EmilStenstrom I'm making this small contribution

#### Full mypy error message:
```
"register" is not exported from module "django_components.component" [reportPrivateImportUsage] 
```

--- 
Tested also output in `pyright` lsp and it is happy too:
Before:
image

After:
image

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -19,8 +19,11 @@ from django.views import View
 # Defining them here made little sense, since 1) component_tags.py and component.py
 # rely on them equally, and 2) it made it difficult to avoid circularity in the
 # way the two modules depend on one another.
+from django_components.component_registry import AlreadyRegistered as AlreadyRegistered  # NOQA
+from django_components.component_registry import ComponentRegistry as ComponentRegistry  # NOQA
+from django_components.component_registry import NotRegistered as NotRegistered  # NOQA
+from django_components.component_registry import register as register  # NOQA
 from django_components.component_registry import registry  # NOQA
-from django_components.component_registry import AlreadyRegistered, ComponentRegistry, NotRegistered, register  # NOQA
 from django_components.context import (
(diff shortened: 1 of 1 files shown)


JuroOravec:
Thanks @mikucz, looks great, really appreciate the clear description! I'll release this soon as part of v0.75.

Off topic, but do you prefer mypy or pyright? Did you notice any differences between them? I haven't heard of pyright before (up until this year my primary language was Javascript/Typescript). IMO Typescript and the type checker provide great user experience, so exciting to see that Microsoft also supports similar effort for Python. A quick googling suggested that currently it might be better to use Mypy for open source projects due to Mypy having the largest userbase. So likely not relevant in terms of django-components, but I might try it in own projects.

mikucz:
> Thanks @mikucz, looks great, really appreciate the clear description! I'll release this soon as part of v0.75.
> 
> Off topic, but do you prefer mypy or pyright? Did you notice any differences between them? I haven't heard of pyright before (up until this year my primary language was Javascript/Typescript). IMO Typescript and the type checker provide great user experience, so exciting to see that Microsoft also supports similar effort for Python. A quick googling suggested that currently it might be better to use Mypy for open source projects due to Mypy having the largest userbase. So likely not relevant in terms of django-components, but I might try it in own projects.

@JuroOravec here is great article with comparison between those two:
https://github.com/microsoft/pyright/blob/main/docs/mypy-comparison.md

I mainly use `pyright` as lsp in my editor because I like it more than `pylsp`. Never had any annoying conflicts between pure `mypy` and `pyright`. It has great futures for auto imports and docs preview. But last time when I tried `pylsp` was couple years ago so many things could have changed :slightly_smiling_face: 

At the end I still have `pre-commit` attached that is triggering pure `mypy` to check everything.

PR #504, opened 2024-05-23
refactor: change fill alias from "as var" to default=var

### LINKED ISSUE #502: Refactor accessing of default slot from `as var` to kwarg
See https://github.com/EmilStenstrom/django-components/pull/495#discussion_r1607117487

JuroOravec:
Change the API for accessing slot's default content from

```django
{% fill "my_slot" as "alias" %}
	{{ alias.default }}
{% endfill %}
```

to

```django
{% fill "my_slot" default="slot_default" %}
	{{ slot_default }}
{% endfill %}
```

Closes #502

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -358,8 +358,8 @@ class Component(View, metaclass=SimplifiedInterfaceMediaDefiningClass):
         slot_fills = {
             slot_name: FillContent(
                 nodes=NodeList([TextNode(escape(content) if escape_content else content)]),
-                alias=None,
-                scope=None,
+                slot_default_var=None,
+                slot_data_var=None,
             )
             for (slot_name, content) in slots_data.items()
         }
@@ -422,12 +422,12 @@ class ComponentNode(Node):
                         f"Detected duplicate fill tag name '{resolved_name}'."
                     )
 
-                resolved_fill_alias = fill_node.resolve_alias(context, resolved_component_name)
-                resolved_scope_var = fill_node.resolve_scope(context, resolved_component_name)
+                resolved_slot_default_var = fill_node.resolve_slot_default(context, resolved_component_name)
+                resolved_slot_data_var = fill_node.resolve_slot_data(context, resolved_component_name)
                 fill_content[resolved_name] = FillContent(
                     nodes=fill_node.nodelist,
-                    alias=resolved_fill_alias,
-                    scope=resolved_scope_var,
+                    slot_default_var=resolved_slot_default_var,
+                    slot_data_var=resolved_slot_data_var,
                 )
 
         component: Component = component_cls(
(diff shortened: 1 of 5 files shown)


src/django_components/component.py
```diff
@@ -422,12 +422,12 @@ def render(self, context: Context) -> str:
                         f"Detected duplicate fill tag name '{resolved_name}'."
                     )
 
-                resolved_fill_alias = fill_node.resolve_alias(context, resolved_component_name)
```
JuroOravec, line 425: I renamed the internal variables from `alias` to `slot_default_var` and from `scope` to `slot_data_var`, so it's clear what they refer to.

src/django_components/slots.py
```diff
@@ -80,27 +80,26 @@ class SlotFill(NamedTuple):
...
     nodelist: NodeList
     context_data: Dict
-    alias: Optional[AliasName]
-    scope: Optional[ScopeName]
+    slot_default_var: Optional[SlotDefaultName]
+    slot_data_var: Optional[SlotDataName]
 
 
-class UserSlotVar:
+class LazySlot:
```
JuroOravec, line 87: Renamed and changed so that the bound slot + context is rendered as `{{ slot }}` instead of `{{ slot.default }}`.

tests/test_templatetags.py
```diff
@@ -912,7 +912,7 @@ def test_both_slots_overridden(self):
         self.assertHTMLEqual(rendered, '<p id="a">Override A</p><p id="b">Override B</p>')
 
 
-class SlotSuperTests(BaseTestCase):
```
JuroOravec, line 915: Renamed the `slot super` to `slot default`

tests/test_templatetags.py
```diff
@@ -998,6 +1000,52 @@ def test_super_under_if_node(self):
             """,
         )
 
+    def test_nested_fills(self):
```
JuroOravec, line 1003: Added a test to ensure the variable assigned via `default` is accessible also in nested components. And did the same for the `data` attribute too.

src/django_components/templatetags/component_tags.py
```diff
@@ -382,19 +383,13 @@ def _parse_fill_args(
...
 
     slot_name = bits.pop(0)
     slot_name_fexp = parser.compile_filter(slot_name)
 
-    alias_fexp: Optional[FilterExpression] = None
-    # e.g. {% fill <name> as <alias> %}
-    if len(bits) >= 2 and bits[-2].lower() == "as":
-        alias = bits.pop()
-        bits.pop()  # Remove the "as" keyword
-        alias_fexp = parser.compile_filter(alias)
```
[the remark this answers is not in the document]
EmilStenstrom: There are many other Django tags that use valueless attributes, so I'm not that bothered about this complexity personally.

EmilStenstrom (approved):
Looks great! Since this is a breaking change, I think we should have a warning at the top of the README about this changed behavior.

JuroOravec:
@EmilStenstrom Thanks for the review! Good catch with the breaking change, I've added a note.

Issue #498, opened 2024-05-16
Revert back to "django" as the default setting for "context_behavior"


JuroOravec:
See the thread in https://github.com/EmilStenstrom/django-components/issues/494

TL;DR:

https://github.com/EmilStenstrom/django-components/issues/494#issuecomment-2111041694

> I wonder if we should switch back to "django" as the default context_behavior, and have isolated be a mode the we reserve for everyone that want the kind of detailed scope control you are after? I think that might be the best path here, that both gives you the control you want, but does not complicate things for beginners?

and https://github.com/EmilStenstrom/django-components/issues/494#issuecomment-2111091032

> I think we should [...] be very specific when in isolated mode, but at the same time switch back to "django" as the default. I fear that the kinds of issues you bring up are for later in a project lifetime, are great for maintainability, but might hinder some people that are getting started and don't (yet) need strict isolation. They can always add it later on.

Notes:
- Ideally we'd update tests so that those tests that render templates would be would be duplicated to be tested for both the "djagno" and "isolated" settings.

Issue #508, opened 2024-05-26
Bug: Cannot override `block` tag inside `component` tag inside a template imported via `extends` in "django" mode


JuroOravec:

JuroOravec:
## The problem

The original problem was that, for some reason, the `context.render_context` (instance of `RenderContext`) inside Django's `BlockNode.render` had one extra layer pushed to the stack in the "django" mode.

This `RenderContext` is what Django uses to keep track of the `{% block %}` tags used in a template. So this way Django knows where to insert one block into another.

The extra layer in the stack was problematic, because unlike `Context`, `RenderContext` checks only the latest layer for variables. Since there was an extra item in the stack, the variables to keep track of `{% block %} tags were defined in the second-last item, and so were UNDEFINED from the perspective of `RenderContext`.
    
See https://github.com/django/django/blob/4.2/django/template/loader_tags.py

## Solution

At first I thought that the culprit would be in the difference of how we handle the "django" and "isolated" modes. The only difference being is that for the latter we make a standalone copy of the `Context` instance and keep only a subset of keys.

But, the "django" mode was still failing even when the context was copied too (but with all keys kept).

I checked the state of the `RenderContext` dicts, and they were identical, except for the extra last item in the stack.

So I looked into all places where we add to the RenderContext stack, to see where the extra layer is coming from.

After some trial and error, I came across [`Template.render`](https://github.com/django/django/blob/4.2/django/template/base.py#L171). 

It calls `render_context.push_state` without the `isolated_context` kwarg, which defaults to `True`. When `True`, empty dict is pushed to the stack of `RenderContext`.

```py
    def render(self, context):
        "Display stage -- can be called many times"
        with context.render_context.push_state(self):
            if context.template is None:
                with context.bind_template(self):
                    context.template_name = self.name
                    return self._render(context)
            else:
                return self._render(context)
```

At this point I realized that the error doesn't happen in "isolated" mode, because it makes a standalone copy of `Context`. While the extra stack layer is pushed to one copy of `Context`, `BlockNode.render` receives the OTHER copy of `Context`, which never had the extra layer pushed. So the error was still there, but in "isolated" mode it was bypassed by coincidence.

So next I tried to modify the `Template.render` method to ALWAYS set `isolated_context=True`. But while this fixed the original error, it created new ones.

```py
    def render(self, context):
        "Display stage -- can be called many times"
        with context.render_context.push_state(self, isolated_context=True):
            if context.template is None:
[...]

PR #509, opened 2024-05-26
tests: Split test files and run template tests under both context behavior modes

JuroOravec:
Part of https://github.com/EmilStenstrom/django-components/issues/498.

Changes:
- Add a decorator that we can use in tests to be able to run tests under both `context_behavior="django"` and `context_behavior="isolated"` settings.
- Added this decorator to many tests to ensure that, whichever mode is the default for `context_behavior`, that we test both modes.
- Move some tests around, so the test files are (hopefully) more self-explanatory.
	- E.g. now the `test_component.py` focuses on testing the `Component` class.
	- And template tags are split across:
		1. `test_templatetags_component.py` for `{% component %}` tag,
		2. `test_templatetags_slot_fill.py` for `{% slot %}` and `{% fill %}` tags,
		3. `test_templatetags_templating.py` for nesting, and using multiple different tags together,
		4. `test_templatetags.py` as a catch-all.


tests/test_command_startcomponent.py
JuroOravec, line 1: Renamed, so the file name starts with `command_`, so the files will be grouped together if we ever have more command test files.

tests/test_component.py
```diff
@@ -97,107 +99,7 @@ class Media:
             """,
         )
 
-    def test_css_only_component(self):
```
JuroOravec, line 100: Mostly moved stuff out of here, and to `test_templatetags_component.py`

tests/test_component_as_view.py
```diff
@@ -162,6 +164,28 @@ def get(self, request, *args, **kwargs) -> HttpResponse:
             response.content,
         )
 
+    @parametrize_context_behavior(["django", "isolated"])
+    def test_replace_context_in_view(self):
```
JuroOravec, line 168: Added test since we already have `test_replace_slot_in_view` and `test_replace_slot_in_view_with_insecure_content`. And there is `test_replace_context_in_view_with_insecure_content`, but no corresponding `test_replace_context_in_view`.

tests/test_context.py
```diff
@@ -243,42 +246,8 @@ def setUpClass(cls):
         component.registry.register(name="parent_with_args", component=cls.ParentComponentWithArgs)
         component.registry.register(name="variable_display", component=VariableDisplay)
 
-    @override_settings(
-        COMPONENTS={
-            "context_behavior": "django",
-        }
-    )
-    def test_parent_args_can_be_drawn_from_context__django(self):
```
JuroOravec, line 251: Thanks to the `parametrize_context_behavior` decorator, we could remove duplicate tests that differed only in the `context_behavior` mode.

tests/test_context.py
```diff
@@ -471,27 +417,12 @@ def setUpClass(cls):
...
-    @override_settings(
-        COMPONENTS={"context_behavior": "isolated"},
-    )
-    def test_simple_component_cannot_use_outer_context__isolated(self):
+    # NOTE: Second arg in tuple is expected value.
+    @parametrize_context_behavior([
+        ("django", "outer_value"),
+        ("isolated", ""),
+    ])
+    def test_simple_component_can_use_outer_context(self, context_behavior_data):
```
JuroOravec, line 429: Here's an example of how we can pass extra data along with the modes in `parametrize_context_behavior`.

tests/test_context.py
```diff
@@ -537,17 +470,7 @@ def setUpClass(cls):
         super().setUpClass()
         component.registry.register(name="simple_component", component=SimpleComponent)
 
-    def setUp(self):
-        self.patcher = patch(
```
JuroOravec, line 541: This way of patching is no longer needed as we achieve the same with the decorator.

tests/test_templatetags.py
```diff
@@ -1,208 +1,47 @@
-import re
-import textwrap
-from typing import Any, Callable, Dict, Optional
+"""Catch-all for tests that use template tags and don't fit other files"""
```
JuroOravec, line 1: Tests were taken out of this file into other test files

tests/test_templatetags.py
```diff
@@ -216,2484 +55,408 @@ class Media:
...
-        template = Template(template_str)
-        with self.assertRaises(TemplateSyntaxError):
-            template.render(Context({}))
-
-    def test_fill_tag_can_occur_within_component_nested_in_implicit_fill(
-        self,
-    ):
-        component.registry.register("test_comp", self.ComponentWithDefaultSlot)
-        component.registry.register("slotted", SlottedComponent)
+    # TODO: https://github.com/EmilStenstrom/django-components/issues/508
```
JuroOravec, line 209: I found out that there's an issue with using `block` tags with `slot` tags (e.g. `slot` tag inside `block` tag), for the "django" mode. Created an issue for this bug at https://github.com/EmilStenstrom/django-components/issues/508

EmilStenstrom (approved):

PR #521, opened 2024-06-02
feat: refactor render fn and allow slots as functions

### LINKED ISSUE #520: Allow to render a component, passing args and kwargs, from Python
This is what I called the "Inlined component render" in https://github.com/EmilStenstrom/django-components/issues/483.

## Description

The idea is that I want to be able to render a component in Python, but using the same inputs (args and kwargs) like I would if I rendered the component inside a Django template via `{% component %}` tag.

So if component's template interface is:
```django
{% component "my_comp" data height=123 width=456 readonly=True %}
{% endcomponent %}
```

Then I'd like to be able to call the component from inside Python like so:
```py
content = MyComp.render_from_input(data, height=123, width=456, readonly=True)
# or
content = MyComp.render_from_input(args=[data], kwargs{"height": 123, "width": 456, "readonly": True})
```
and get back the rendered content as a string.

## Why?

I mostly do this when I want to nest rendered data into other data structures.

E.g. currently I have a `table` component, which accepts a list of `TableRow`s, whereas `TableRow` is roughly:
```py
@dataclass
class TableRow:
	cols: Dict[str, TableCell]
	attrs: str|None = None
	klass: str|None = None

@dataclass
class TableCell:
	content: str
	attrs: str|None = None
	klass: str|None = None
	href: str|None = None
```

So with this, I can construct tables, but I can tweak styling / HTML attributes on the level of rows AND cells. Or make specific cells to act as links.

But you see above that `TableCell` accepts a `content` as string. I have some elaborate content inside the table cells, e.g. an icon that acts as a button. But the icon is defined as an `icon` component. So I first need to render the `icon` to string, and then pass the result into `TableCell` to achieve that.

```py
row = TableRow(
	cols={
		"save": TableCell(
			content=IconComp.render(...),
			...
		)
	}	
)
```

## Current state

Currently, `Component` class has 3 rendering methods - `render`, `render_from_input`, and `render_to_response`.
- `render_to_response`
	- Part of [`django.views.View` API](https://github.com/django/django/blob/main/django/views/generic/base.py#L190), this method allows the components to behave as Views.
	- Calls `render`.
	- Allows to pass slot fills in Python
	- **Doesn't** handle args and kwargs
- `render_from_input`
	- Used by `ComponentNode`, so used when we render the component inside a Django template as `{% component ... %}`.
	- Calls 'render`
 	- **Doesn't** handle slot fills
	- Accepts args and kwargs.
- `render`
	- The common path for both `render_to_response` and `render_from_input`.
	- 	*Doesn't* accept args and kwargs, and instead expects the args/kwargs to be pre-processed into a Context instance.

NOTE: Initially I thought that `Component.render` needs follow the `Node.render` API, but this is not true. `ComponentNode.render` has to follow `Node.render` API, but `Component.render` can be whatever!

So the problem is that currently, we don't have a unified interface for rendering a component from inside Python
- `render_from_input` accepts args and kwargs but doesn't handle slots
- `render_to_response` handles slots but doesn't accept args and kwargs
- `render` requires users to pre-process the args and kwargs themselves.

## Future state

So instead I suggest to merge these to have singular interface that can handle both args & kwargs, AND slots, AND can accept a Context:

```py
def render(
    self,
    context: Union[Dict[str, Any], Context] = None,
    args: Optional[Union[List, Tuple]] = None,
    kwargs: Optional[Dict[str, Any]] = None,
    slots: Optional[Dict[SlotName, str]] = None,
    escape_slots_content: bool = True,
) -> str:
```

And because we still need `render_to_response`, then `render_to_response` would be just a thin wrapper around `render`, that wraps the content in an HtmlResponse class.

## API

While the new `render` is fairly flexible, then if we wanted to call it in Python, we'd still need to instanciate the class:
```py
MyComp("").render(args=args, kwargs=kwargs, ...)
```

A nicer API would be if we could call `render` as a class method:
```py
MyComp.render(args=args, kwargs=kwargs, ...)
```

I think it's safe to use `render` as class method, because all the data that's passed to Component at instantiation are related component's behavior when it's nested in the template:
```py
component: Component = component_cls(
    registered_name=resolved_component_name,
    outer_context=context,
    fill_content=fill_content,
    component_id=self.component_id,
)
```
- `registered_name` - Not relevant when component rendered directly
- `outer_context` - None when called directly, because component is not nested
- `fill_content` - Handled by the `slots` input to `Component.render`, so no need to set that at instantiation
- `component_id` - No need to set it, as component will define one internally automatically if not set.
---

What do you think?

JuroOravec:
Initial implementation for https://github.com/EmilStenstrom/django-components/issues/520.

See individual comments for details.

What still remains is to do:
- [x] Document the changes in README
- [x] Update docstrings of `render` and `render_to_response`

Closes #520

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -348,21 +348,45 @@ class Component(View, metaclass=SimplifiedInterfaceMediaDefiningClass):
         **response_kwargs: Any,
     ) -> HttpResponse:
         """
+        Render the component and wrap the content in the response class.
+
+        The response class is taken from `Component.response_class`. Defaults to `django.http.HttpResponse`.
+
         This is the interface for the `django.views.View` class which allows us to
         use components as Django views with `component.as_view()`.
+
+        Inputs:
+        - `args` - Positional args for the component. This is the same as calling the component
+          as `{% component "my_comp" arg1 arg2 ... %}`
+        - `kwargs` - Kwargs for the component. This is the same as calling the component
+          as `{% component "my_comp" key1=val1 key2=val2 ... %}`
+        - `slots` - Component slot fills. This is the same as pasing `{% fill %}` tags to the component.
+            Accepts a dictionary of `{ slot_name: slot_content }` where `slot_content` can be a string or render function.
+        - `escape_slots_content` - Whether the content from `slots` should be escaped.
+        - `context` - A context (dictionary or Django's Context) within which the component
+          is rendered. The keys on the context can be accessed from within the template.
+            - NOTE: In "isolated" mode, context is NOT accessible, and data MUST be passed via
+              component's args and kwargs.
+
+        Any additional args and kwargs are passed to the `response_class`.
 
         Example:
         ```py
         MyComponent.render_to_response(
             args=[1, "two", {}],
             kwargs={
-                "key"=123,
+                "key": 123,
             },
             slots={
-                header='STATIC TEXT HERE',
+                "header": 'STATIC TEXT HERE',
+                "footer": lambda ctx, slot_kwargs, slot_ref: f'CTX: {ctx['hello']} SLOT_DATA: {slot_kwargs['abc']}',
             },
             escape_slots_content=False,
+            # HttpResponse input
+            status=201,
+            headers={...},
         )
+        # HttpResponse(content=..., status=201, headers=...)
         ```
         """
         content = cls.render(
@@ -384,15 +408,31 @@ class Component(View, metaclass=SimplifiedInterfaceMediaDefiningClass):
         escape_slots_content: bool = True,
     ) -> str:
         """
+        Render the component into a string.
+
+        Inputs:
+        - `args` - Positional args for the component. This is the same as calling the component
+          as `{% component "my_comp" arg1 arg2 ... %}`
+        - `kwargs` - Kwargs for the component. This is the same as calling the component
+          as `{% component "my_comp" key1=val1 key2=val2 ... %}`
+        - `slots` - Component slot fills. This is the same as pasing `{% fill %}` tags to the component.
+            Accepts a dictionary of `{ slot_name: slot_content }` where `slot_content` can be a string or render function.
+        - `escape_slots_content` - Whether the content from `slots` should be escaped.
+        - `context` - A context (dictionary or Django's Context) within which the component
+          is rendered. The keys on the context can be accessed from within the template.
+            - NOTE: In "isolated" mode, context is NOT accessible, and data MUST be passed via
+              component's args and kwargs.
+
         Example:
         ```py
         MyComponent.render(
             args=[1, "two", {}],
             kwargs={
-                "key"=123,
+                "key": 123,
             },
             slots={
-                header='STATIC TEXT HERE',
+                "header": 'STATIC TEXT HERE',
+                "footer": lambda ctx, slot_kwargs, slot_ref: f'CTX: {ctx['hello']} SLOT_DATA: {slot_kwargs['abc']}',
             },
             escape_slots_content=False,
         )
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -37,8 +37,17 @@ from django_components.context import (
 from django_components.expression import safe_resolve_dict, safe_resolve_list
 from django_components.logger import logger, trace_msg
 from django_components.middleware import is_dependency_middleware_active
-from django_components.node import RenderedContent, nodelist_to_render_func
-from django_components.slots import DEFAULT_SLOT_KEY, FillContent, FillNode, SlotContent, SlotName, resolve_slots
+from django_components.slots import (
+    DEFAULT_SLOT_KEY,
+    FillContent,
+    FillNode,
+    SlotContent,
+    SlotName,
+    SlotRef,
+    SlotRenderedContent,
+    _nodelist_to_slot_render_func,
+    resolve_slots,
+)
 from django_components.template_parser import process_aggregate_kwargs
 from django_components.utils import gen_id, search
 
@@ -495,13 +504,13 @@ class Component(View, metaclass=SimplifiedInterfaceMediaDefiningClass):
         slot_fills = {}
         for slot_name, content in slots_data.items():
             if isinstance(content, (str, SafeString)):
-                content_func = nodelist_to_render_func(
+                content_func = _nodelist_to_slot_render_func(
                     NodeList([TextNode(escape(content) if escape_content else content)])
                 )
             else:
 
-                def content_func(ctx: Context) -> RenderedContent:
-                    rendered = content(ctx)
+                def content_func(ctx: Context, kwargs: Dict[str, Any], slot_ref: SlotRef) -> SlotRenderedContent:
+                    rendered = content(ctx, kwargs, slot_ref)
                     return escape(rendered) if escape_content else rendered
 
             slot_fills[slot_name] = FillContent(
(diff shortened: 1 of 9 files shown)


tests/test_component.py
```diff
@@ -81,26 +82,64 @@ class Media:
...
+        rendered = SimpleComponent.render(kwargs={"variable": "test"})
+        self.assertHTMLEqual(
+            rendered,
+            """
+            Variable: <strong>test</strong>
+            """,
+        )
+
+    @parametrize_context_behavior(["django", "isolated"])
+    def test_template_string_dynamic(self):
```
JuroOravec, line 94: Added missing tests for `Component.get_template_string`, `template_name`,  and `get_template`.

src/django_components/component_registry.py
```diff
@@ -20,7 +20,7 @@ def __init__(self) -> None:
 
     def register(self, name: str, component: Type["component.Component"]) -> None:
         existing_component = self._registry.get(name)
-        if existing_component and existing_component.class_hash != component.class_hash:
+        if existing_component and existing_component._class_hash != component._class_hash:
```
JuroOravec, line 23: Prefixed `_class_hash` to hint that it's being used internally.

tests/test_component.py
```diff
@@ -113,16 +152,297 @@ def get_context_data(self, name, css_class="", title="", **attrs):
...
+            kwargs: {}
+            ---
+            from_context:
+            ---
+            slot_second: SLOT_SECOND_DEFAULT
+            """,
+        )
+
+    @parametrize_context_behavior(["django", "isolated"])
+    def test_render_full(self):
```
JuroOravec, line 231: See this test for a full example on how we to use `Component.render`

src/django_components/component.py
```diff
@@ -180,14 +181,28 @@ def _get_dir_path_from_component_path(
...
+    """Inlined JS associated with this component."""
     css: Optional[str] = None
+    """Inlined CSS associated with this component."""
     media: Media
+    """
+    Normalized definition of JS and CSS media files associated with this component.
+
+    NOTE: This field is generated from Component.Media class.
+    """
+    response_class = HttpResponse
```
JuroOravec, line 200: Before, the `render_to_response` always used the `HttpResponse` class. Now users can override this using `response_class`. This was inspired by a mixin class in the same file as `django.views.View`

src/django_components/component.py
```diff
@@ -205,17 +220,39 @@ def __init__(
...
+
+    @property
+    def name(self) -> str:
+        return self.registered_name or self.__class__.__name__
 
     def get_context_data(self, *args: Any, **kwargs: Any) -> Dict[str, Any]:
         return {}
 
-    def get_template_name(self, context: Mapping) -> Optional[str]:
+    def get_template_name(self, context: Context) -> Optional[str]:
```
JuroOravec, line 232: Changed to `Context` since that's what it actually is

src/django_components/component.py
```diff
@@ -205,17 +220,39 @@ def __init__(
         self._context: Optional[Context] = None
 
     def __init_subclass__(cls, **kwargs: Any) -> None:
-        cls.class_hash = hash(inspect.getfile(cls) + cls.__name__)
+        cls._class_hash = hash(inspect.getfile(cls) + cls.__name__)
+
+    @property
+    def name(self) -> str:
+        return self.registered_name or self.__class__.__name__
```
JuroOravec, line 227: Moved component name to own property, as I used in a couple of places.

src/django_components/component.py
```diff
@@ -205,17 +220,39 @@ def __init__(
...
 
-    def get_template_string(self, context: Mapping) -> Optional[str]:
+    def get_template_string(self, context: Context) -> Optional[str]:
         return self.template
 
+    # NOTE: When the template is taken from a file (AKA specified via `template_name`),
+    # then we leverage Django's template caching. This means that the same instance
+    # of Template is reused. This is important to keep in mind, because the implication
+    # is that we should treat Templates AND their nodelists as IMMUTABLE.
+    def get_template(self, context: Context) -> Template:
```
JuroOravec, line 242: Github doesn't show it, but this `get_template` method was just moved up a bit, so all the methods related to the template are together.

src/django_components/component.py
```diff
@@ -303,135 +320,195 @@ def get_context_data(self):
...
             raise RuntimeError(
-                f"Method 'inject()' of component '{comp_name}' was called outside of 'get_context_data()'"
+                f"Method 'inject()' of component '{self.name}' was called outside of 'get_context_data()'"
             )
 
-        return get_injected_context_var(comp_name, self._context, key, default)
+        return get_injected_context_var(self.name, self._context, key, default)
+
+    @classmethod
+    def render_to_response(
```
JuroOravec, line 331: `render_to_response` is now also a class method

src/django_components/component.py
```diff
@@ -303,135 +320,195 @@ def get_context_data(self):
...
+            },
+            escape_slots_content=False,
+        )
+        ```
+        """
+        comp = cls()
+        return comp._render(context, args, kwargs, slots, escape_slots_content)
 
-    def render_from_input(
+    def _render(
```
JuroOravec, line 395: Merged `render` and `render_from_input` into `_render`. 

Note the difference:
- `render` is public method and class method
- `_render` is private method and instance method.

`render` (public class method) internally instantiates Component class and calls `_render`.

tests/test_component.py
```diff
@@ -113,16 +152,315 @@ def get_context_data(self, name, css_class="", title="", **attrs):
...
+
+        with self.assertRaises(TemplateSyntaxError):
+            SimpleComponent.render()
+
+        SimpleComponent.render(
+            slots={"first": "FIRST_SLOT"},
+        )
+
+    @parametrize_context_behavior(["django", "isolated"])
+    def test_render_with_include(self):
```
JuroOravec, line 421: In my fork my implementation had issue when I used `Component.render()` and the template included an `{% include %}` tag. This is why I associated the template with the context [here](https://github.com/EmilStenstrom/django-components/pull/521/files#diff-35c84ebceae50f7149f03bbcaf2b4f1289bc2984bf7859ca521b5dbe7960c396R431) when context was internally created inside `render` method.

src/django_components/component.py
EmilStenstrom: This breaks backwards compatibility if people have self.render_to_response I think? Or can class methods be called on instances?

EmilStenstrom (commented):
Nice changes. Auto-approve when you've done the two tasks you specify in this PR.

JuroOravec:
Made two more changes:
- When rendering a component with `Component.render()`, it has a `slots` kwarg to define the slot fills. Before this MR, we could define slots like this only as string. Then, in previous version of this MR, we could define it also as a function `(ctx: Context) -> str`. Now, in the newer version, I've passed also slot data and slot default to the slot render function. So the function signature is `(ctx: Context, slot_data: Dict[str, Any], slot_default: SlotRef) -> str` (where `SlotRef` is the "lazy-eval" slot content.

- In previous version of the MR, when we called `MyComponent.render()`, then this class method created an instance internally. But since class methods can be called also from instances, it felt strange and wasteful that `MyComponent().render()` would still create another instance inside `render()`. So in https://github.com/EmilStenstrom/django-components/pull/521/commits/36cd6e3f063e69488fbd20260f98e0c37e6a8b86 I made is so that the `render` and `render_to_response` methods are changed from class methods to instance methods at instantiation.

@EmilStenstrom Given these changes, could you have another look at this MR? Thanks!

EmilStenstrom:
Oups, accidentaly merged when I wanted to approve the PR, sorry!

Issue #522, opened 2024-06-04
Support for Paths as objects is missing


ldurey:
Hello everyone,
First of all, thanks for you great work, this project seems like really useful for our team.
This is my first contribution to an open source project, so apologies if i'm not sufficiently clear. 

I wanted to report that the support for [Paths as objects](https://docs.djangoproject.com/en/5.0/topics/forms/media/#paths-as-objects) declaration is not possible with the current state of the `django_components` project.

## Context
- We are using a [ManifestStaticFilesStorage](https://docs.djangoproject.com/en/5.0/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage) to manage our static files.
- We want to be able to use `Javascript module` declaration for rather complex components.

So we did a declaration of js asset like this
```
from django.templatetags.static import static
from django.utils.html import format_html

js = [format_html(f'<script type="module" src="{static(static_path)}"></script>')]
```
But in our case, this is clashing with `ManifestStaticFilesStorage` because the `static` call is evaluated at initialisation of the project. And the static files manifest may miss some keys for newly added files (not registered yet). So we end up with `ValueError` raised at init.
Because of that evaluation at initialisation we can't call `collectstatic` to populate the manifest which is actually needed to populate itself. (DEBUG = False must be used to raise the Exception, so this is a production only Bug.)

To solve this problem we tried to use `Paths as objects`, and this is the issue I report.

## Issue 
```
from django.templatetags.static import static
from django.utils.html import format_html, html_safe

@html_safe
class JSModulePath:
	def __init__(self, static_path):
		self.static_path = static_path 

	def __str__(self):
		return format_html(f'<script type="module" src="{static(self.static_path)}"></script>')

# Component
	js = [JSModulePath(static_path=static_path)]
```

With this asset declaration, we end up with a `TypeError` raised by `os.path.join()` at this [line of code](https://github.com/EmilStenstrom/django-components/blob/3a7d5355cfa5d37641a4c80ee9fd2423cea10c1b/src/django_components/component.py#L120).

## How to solve
Reading the code, it seems like all the problem come from support for relative path in assets declaration which is handled by this method [_resolve_component_relative_files](https://github.com/EmilStenstrom/django-components/blob/3a7d5355cfa5d37641a4c80ee9fd2423cea10c1b/src/django_components/component.py#L80C5-L80C38).
If I just comment this line, everything is all right and support for `Paths as objects` is back.

[...]

PR #526, opened 2024-06-12
feat: paths as objects + user-provided Media cls + handle static

JuroOravec:
Changes:
- Allow to specify template/js/css files as str, bytes, FileLike, SafeString, or a function that returns one of the other types.
- Support Django's "path as objects" feature - when an object is a "safe string" (defines `__html__` method), then we do NOT try to resolve it.
- Allow to override how JS/CSS paths are rendered into `<script>`/`<link>` tags by subclassing Django's `Media` class and supplying that to `Component.media_class`
- ~Validate that keys are valid CSS media types when `Media.css` is defined as a dictionary~

Documents the above plus also how component's template/css/js paths are resolved relative to component's dir if possible.

While we should wait to see what the guys from https://github.com/EmilStenstrom/django-components/issues/522 say, the tests suggest that this should fix their issue.

Closes https://github.com/EmilStenstrom/django-components/issues/522


src/django_components/component.py
```diff
@@ -54,37 +55,113 @@
...
+        # `MediaDefiningClass` defines `media` as a getter (defined in djago.forms.widgets.media_property).
+        # So we reused that and convert it to user-defined Media class
+        media_prop: property = cls.media
+        media_getter = media_prop.fget
+        def media_wrapper(self: "Component") -> Any:
+            if not media_getter:
+                return None
+            media: Media = media_getter(self)
+            return self.media_class(js=media._js, css=media._css)
+        cls.media = property(media_wrapper)
```
JuroOravec, line 86: The bit above makes it possible for users to set `media_class` on the component class to override the Media class. For examples see https://github.com/EmilStenstrom/django-components/pull/526/files#diff-7667a514b72a24c770553fe3421223dcc03252c458518310f25a5ef7746d4669R473 and https://github.com/EmilStenstrom/django-components/pull/526/files#diff-7667a514b72a24c770553fe3421223dcc03252c458518310f25a5ef7746d4669R496

src/django_components/component.py
```diff
@@ -54,37 +55,113 @@
...
             return super().__new__(mcs, name, bases, attrs)
 
         if "Media" in attrs:
-            media: Component.Media = attrs["Media"]
+            media_data: Component.Media = attrs["Media"]
+            _normalize_media(media_data)
+            _normalize_media_filepaths(media_data)
 
-            # Allow: class Media: css = "style.css"
-            if hasattr(media, "css") and isinstance(media.css, str):
```
JuroOravec, line 68: Moved the normalization to a function for clarity

src/django_components/component.py
```diff
@@ -54,37 +55,113 @@
...
         # NOTE: Skip template/media file resolution when then Component class ITSELF
         # is being created.
         if "__module__" in attrs and attrs["__module__"] == "django_components.component":
             return super().__new__(mcs, name, bases, attrs)
 
         if "Media" in attrs:
-            media: Component.Media = attrs["Media"]
+            media_data: Component.Media = attrs["Media"]
+            _normalize_media(media_data)
+            _normalize_media_filepaths(media_data)
```
JuroOravec, line 68: `_normalize_media_filepaths` what allows users to define file paths as str/bytes/PathLike/Callable

src/django_components/component.py
```diff
@@ -54,37 +55,113 @@
...
+        return True
+    
+    return False
+
+
+def _normalize_media_filepath(filepath: Any) -> str | SafeData:
+    print("_normalize_media_filepath: ", filepath)
+
+    if isinstance(filepath, SafeData) or hasattr(filepath, "__html__"):
+        return filepath
```
JuroOravec, line 158: This bit above ensures that when users creates an "object path" with `@ html_safe`, then we don't try to convert it to a string.

src/django_components/component.py
```diff
@@ -126,15 +203,17 @@ def _resolve_component_relative_files(attrs: MutableMapping) -> None:
...
-        maybe_resolved_filepath = os.path.join(comp_dir_abs, filepath)
-        component_import_filepath = os.path.join(comp_dir_rel, filepath)
-
-        if os.path.isfile(maybe_resolved_filepath):
-            logger.debug(
-                f"Interpreting template '{filepath}' of component '{module_name}' relatively to component file"
-            )
-            return component_import_filepath
+    def resolve_file(filepath: str | SafeData) -> str | SafeData:
+        if isinstance(filepath, str):
```
JuroOravec, line 212: Path resolution logic is applied only to paths that resolved to plain string (so paths defined as str, bytes, PathLike, functions).

src/django_components/component.py
```diff
@@ -126,15 +203,17 @@ def _resolve_component_relative_files(attrs: MutableMapping) -> None:
...
+            maybe_resolved_filepath = os.path.join(comp_dir_abs, filepath)
+            component_import_filepath = os.path.join(comp_dir_rel, filepath)
+
+            if os.path.isfile(maybe_resolved_filepath):
+                logger.debug(
+                    f"Interpreting template '{filepath}' of component '{module_name}' relatively to component file"
+                )
+                return static(component_import_filepath)
+            return static(filepath)
```
JuroOravec, line 221: And here we apply the `static()` call to the string path which has been possibly resolved as a relative path,

tests/test_component_media.py
```diff
@@ -282,6 +286,250 @@ class Media:
             """,
         )
 
+    # TODO - ALLOW CALLABLE!
+    # TODO - DOCUMENT IT ALL!
+    # TODO - Document how when we pass a safe string to css/js, then
+    #        Media.render_js/css DOES NOT format it (same as Django's
+    #        see https://docs.djangoproject.com/en/5.0/topics/forms/media/#paths-as-objects)
```
EmilStenstrom, line 293: TODO

tests/test_component_media.py
```diff
@@ -282,6 +286,250 @@ class Media:
...
+            <script src="path/to/script2.js"></script>
+            """,
+        )
+
+    def test_media_custom_render_js(self):
+        class MyMedia(Media):
+            def render_js(self):
+                tags: list[str] = []
+                for path in self._js:
+                    tags.append(f'<my_script_tag src="{self.absolute_path(path)}"></my_script_tag>')
```
EmilStenstrom, line 484: `absolute_path(path)` doesn't seem to do anything?

EmilStenstrom (commented):
Really like the API here! Looking forward to seeing the final PR!

tests/test_component_media.py
JuroOravec: In that tests, `absolute_path(path)` did nothing, because neither `staticfiles` app was given, nor was `STATIC_URL` set. I've added a separate test here https://github.com/EmilStenstrom/django-components/pull/526/files#diff-7667a514b72a24c770553fe3421223dcc03252c458518310f25a5ef7746d4669R530 to the use of `static(path)` / `absolute_path(path)` when `STATIC_URL` is given.

src/django_components/component.py
```diff
@@ -49,145 +48,22 @@
...
-            # Allow: class Media: js = "script.js"
-            if hasattr(media, "js") and isinstance(media.js, str):
-                media.js = [media.js]
-
-        _resolve_component_relative_files(attrs)
-
         return super().__new__(mcs, name, bases, attrs)
 
 
-def _resolve_component_relative_files(attrs: MutableMapping) -> None:
```
JuroOravec, line 90: I've moved the logic that prepares the Media class to `component_media`

README.md
```diff
@@ -1645,7 +1650,211 @@ If you find yourself using the `only` modifier often, you can set the [context_b
 
 Components can also access the outer context in their context methods like `get_context_data` by accessing the property `self.outer_context`.
 
-## Rendering JS and CSS dependencies
+## Defining HTML/JS/CSS files
```
JuroOravec, line 1653: Documented how to work with the Media class, the inputs, how to override it, etc. So conceptually there's 2 separate steps:
1. "Defining HTML/JS/CSS files" describes how to define/link files to a components
2. "Rendering JS and CSS dependencies" describes how, given that I have components that have linked files, how I render out the `<script>`/`<link>` tags.

src/django_components/component_media.py
```diff
@@ -0,0 +1,375 @@
...
+from django_components.utils import search
+
+if TYPE_CHECKING:
+    from django_components.component import Component
+
+
+# For full list, see https://docs.djangoproject.com/en/5.0/topics/forms/media/#css
+# Note that many of these are marked as deprecated. For the recommended
+# media types, see https://developer.mozilla.org/en-US/docs/Web/CSS/@media#media_types.
+CSS_MEDIA_TYPES = (
```
JuroOravec, line 19: Normally Django's Media class accepts a dictionary for the `css` attribute. There, the dict keys represent [CSS media types](https://developer.mozilla.org/en-US/docs/Web/CSS/@media#media_types). There is only a small set of all the media types, so I added validation

src/django_components/component_media.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+def _resolve_component_relative_files(attrs: MutableMapping) -> None:
+    """
+    Check if component's HTML, JS and CSS files refer to files in the same directory
+    as the component class. If so, modify the attributes so the class Django's rendering
+    will pick up these files correctly.
+    """
+    # First check if we even need to resolve anything. If the class doesn't define any
+    # JS/CSS files, just skip.
```
JuroOravec, line 277: `_resolve_component_relative_files` and `_get_dir_path_from_component_path` didn't change, just added the check here to skip the relative path resolution logic when the component class doesn't define any files.

tests/test_component_media.py
```diff
@@ -282,6 +287,437 @@ class Media:
...
+
+            <script hi src="calendar/script.js"></script>
+            <script src="calendar/script1.js"></script>
+            <script src="calendar/script2.js"></script>
+            <script src="calendar/script3.js"></script>
+            """,
+        )
+
+    @override_settings(STATIC_URL="static/")
+    def test_works_with_static(self):
```
JuroOravec, line 579: Added a separate test here to check that the paths defined ins `Media.css` / `Media.js` work with staticfiles.

tests/test_component_media.py
```diff
@@ -282,6 +287,437 @@ class Media:
...
+            <script hi src="/static/calendar/script.js"></script>
+            <script src="/static/calendar/script1.js"></script>
+            <script src="/static/calendar/script2.js"></script>
+            <script src="/static/calendar/script3.js"></script>
+            <script src="/static/calendar/script4.js"></script>
+            """,
+        )
+
+
+class MediaStaticfilesTests(BaseTestCase):
```
JuroOravec, line 618: These following 2 tests use a similar setup as described in https://github.com/EmilStenstrom/django-components/issues/522. The first test uses a default Storage class, while second uses `ManifestStaticFilesStorage`.

JuroOravec:
@EmilStenstrom This one is now ready for review :)

src/django_components/component_media.py
```diff
@@ -0,0 +1,378 @@
...
+
+def _normalize_media(media: ComponentMediaInput) -> None:
+    if hasattr(media, "css") and media.css:
+        # Allow: class Media: css = "style.css"
+        if _is_media_filepath(media.css):
+            media.css = [media.css]  # type: ignore[list-item]
+
+        # Allow: class Media: css = ["style.css"]
+        if isinstance(media.css, (list, tuple)):
+            media.css = {"all": media.css}
```
EmilStenstrom, line 191: Are we changing the media type from "None" to "all" here? media="all" is implicit in CSS so I don't think we should add that to everyone's CSS when it's not needed.

tests/components/relative_file/relative_file_pathobj.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def post(self, request, *args, **kwargs) -> HttpResponse:
+        variable = request.POST.get("variable")
+        return self.render_to_response({"variable": variable})
+
+    def get(self, request, *args, **kwargs) -> HttpResponse:
+        return self.render_to_response({"variable": "GET"})
+
+    def get_context_data(self, variable, *args, **kwargs) -> Dict[str, Any]:
+        return {"variable": variable}
```
EmilStenstrom, line 41: Not a super fan of having shared fixtures used for testing like this, because it makes tests hard to follow (you have to open several files to understand them), and can introduce cross-test dependencies inadvertently. Not super important, because I think we're already doing this in other places in the tests.

tests/test_autodiscover.py
```diff
@@ -36,7 +36,7 @@ def test_autodiscover_with_components_as_views(self):
 
         all_components_after = component_registry.registry.all().copy()
         imported_components_count = len(all_components_after) - len(all_components_before)
-        self.assertEqual(imported_components_count, 1)
+        self.assertEqual(imported_components_count, 2)
```
EmilStenstrom, line 39: This is a very strange test. Checks that we didn't add any new autodiscoverable components?

tests/test_component_media.py
```diff
@@ -282,6 +287,448 @@ class Media:
...
+        # Configure static files. The dummy files are set up in the `./static_root` dir.
+        # The URL should have path prefix /static/.
+        # NOTE: We don't need STATICFILES_DIRS, because we don't run collectstatic
+        #       See https://docs.djangoproject.com/en/5.0/ref/settings/#std-setting-STATICFILES_DIRS
+        STATIC_URL="static/",
+        STATIC_ROOT=os.path.join(Path(__file__).resolve().parent, "static_root"),
+        # NOTE: STATICFILES_STORAGE is deprecated since 5.1, use STORAGES instead
+        #       See https://docs.djangoproject.com/en/5.0/ref/settings/#staticfiles-storage
+        STORAGES={
+            # This was NOT changes
```
EmilStenstrom, line 686:
```suggestion
            # This was NOT changed
```

tests/test_component_media.py
```diff
@@ -282,6 +287,448 @@ class Media:
...
+
+        comp = SimpleComponent()
+
+        # NOTE: Since we're using ManifestStaticFilesStorage, we expect the rendered media to link
+        # to the files as defined in staticfiles.json
+        self.assertHTMLEqual(
+            comp.render_dependencies(),
+            """
+            <link href="/static/calendar/style.0eeb72042b59.css" media="all" rel="stylesheet">
+            <my_script_tag src="/static/calendar/script.e1815e23e0ec.js"></my_script_tag>
```
EmilStenstrom, line 728: Really nice that we can support this now!

EmilStenstrom (approved):
Really nice work as always @JuroOravec!

Added some small nitpicks, and a question about if we can render media="all" in fewer cases. In general I approve! :)

src/django_components/component_media.py
JuroOravec: I didn't implement it, so I might be missing something, but I'm not aware of the "None" option, was that maybe something in older Django versions? Because in v5, when you [go to the docs](https://docs.djangoproject.com/en/5.0/topics/forms/media/#css) and see the examples there, it looks like `"all"` is the default / catch-all option.

tests/components/relative_file/relative_file_pathobj.py
JuroOravec: That's a good point, I've moved this file to it's own dir `relative_file_pathobj`

tests/test_autodiscover.py
JuroOravec: Hm, looking at the test, I think the test would make more sense if we checked that `autoimport()` imports the components from the `tests/components` dir.

For now I've put a TODO there. I will be reviewing these tests when I'll be making an MR for autoimport, as mentioned here https://github.com/EmilStenstrom/django-components/issues/527#issuecomment-2183091056

src/django_components/component_media.py
JuroOravec: Aha, I made a mistake. In Django docs, they use also this example, with two values separated by comma:

```py
class Media:
    css = {
        "tv,projector": ["lo_res.css"],
    }
```
Which corresponds to:
```html
<link href="http://static.example.com/lo_res.css" media="tv,projector" rel="stylesheet">
```

So I assume that this means that they keys in the dictionary can use also the [logical operators](https://developer.mozilla.org/en-US/docs/Web/CSS/@media#logical_operators) (comma, `not`, `only`, etc...).

So in that case IMO validating the CSS keys is not worth it, as there could be many different combinations.

src/django_components/component_media.py
JuroOravec: Btw @EmilStenstrom I'll merge this MR so we can close the original issue. But if there's a way to not generate the `media="all"`, e.g. with the "None" option, we can open a new issue for that or I can just include it in on of my other MRs

EmilStenstrom: I was thinking that `css = "style.css"` (our version) would generate a link tag without media set. But let's keep it the way to wrote it.

PR #530, opened 2024-06-18
Test: trying to setup a detection for collectstatic error with github action

ldurey:


ldurey:
This is not needed anymore, see https://github.com/EmilStenstrom/django-components/issues/522#issuecomment-2176064872

PR #531, opened 2024-06-20
add support for jsx notation

Bokitza:
a simple translator for React-ish developers to be able to write in JSX.
makes the experience of the signal file view-component feels like Remix and Python got married and had a child.

Usage:
```
template = jsx2django_components(
    <body>
      <Head title="this is a prop for the Header component title" />
         <body class="text-center">
         <h1> A beautiful Web App</h1>
         <h2 class="text-2xl "> HomePage is here </h2>
      <Footer logo="this is a prop for the logo in the Footer component" />
    </body>
```)

### DIFF
--- a/src/django_components/utils.py
+++ b/src/django_components/utils.py
@@ -1,7 +1,9 @@
 import glob
 from pathlib import Path
 from typing import Any, Callable, List, NamedTuple, Optional
+
 from django.template.engine import Engine
+
 from django_components import types as t
 from django_components.template_loader import Loader
 
--- a/src/django_components/utils.py
+++ b/src/django_components/utils.py
@@ -1,9 +1,8 @@
 import glob
 from pathlib import Path
 from typing import Any, Callable, List, NamedTuple, Optional
-
 from django.template.engine import Engine
-
+from django_components import types as t
 from django_components.template_loader import Loader
 
 
@@ -56,3 +55,31 @@ def find_last_index(lst: List, predicate: Callable[[Any], bool]) -> Any:
         if predicate(elem):
             return len(lst) - 1 - r_idx
     return -1
+
+
+def jsx2django_components(jsx: t.django_html):
+    html = jsx
+    more_components = True
+
+    while more_components:
+        tag = ""
+        try:
+            tag = jsx.split("<")[1].split("/>")[0]
+        except:
+            more_components = False
+            continue
+        if not tag:
+            more_components = False
+            continue
+
+        is_compoment = not str(tag)[0].isupper()
+        if is_compoment:
+            jsx = jsx.split(tag)[1]
+            continue
+        component_name, attributes = tag.split(" ", 1)
+
+        parsed_component = "{% component '" + component_name + "' " + attributes + "%} {% endcomponent %}"
+
+        html = html.replace("<" + tag + "/>", parsed_component)
+        jsx = jsx.split(tag)[1]
+    return html

### CODE REVIEW

**Comment by EmilStenstrom** [comment]
Hi @Bokitza! Interesting approach. Could you give us a little more context on how this would be used?

**Comment by Bokitza** [comment]
It's syntactic sugar for Js-frameworks oriented developers.

one of the pro's is the ability to support React tooling in VS code (like component highlighting), and I'm my own opinion a bit more clean than the {% notation

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

**Comment by JuroOravec** [comment]
Hey @Bokitza looks interesting. Do I understand correctly that you're still using Django template language underneath? That you just write the templates with `<` `>` instead of `{%` `%}`, then pre-process it, and then load it normally onto Django?

I'm not sure if adding the `jsx2django_components` function is the preferable approach here. It feels like a simple custom script that anyone could copy-paste into their project, so it could live outside of django_components, or as a standalone package.

However, I could imagine this working if it was implemented as a proper [Django templating backend](https://docs.djangoproject.com/en/5.0/howto/custom-template-backend/). The reason being, and this is still months away, but in my head I'm playing with the idea of making django_compontents work also with Jinja2, since Django supports Jinja2 out of the box. If we got to support both Jinja2 and Django templates at the same time, it means that django_components would have a pluggable interface that would allow any other template backends to integrate with it too.

So then there could be a standalone Django template backend that parses JSX syntax, and it could be used hand-in-hand with django_components.

**Comment by Bokitza** [comment]
Cool (:

Just pushed a PR for the thought of it. I know it will help me write 'cleaner'. No worries. Keep up the good work

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

**Comment by JuroOravec** [comment]
I looked more into this out of curiosity about how Django handles templates from different engines (Django templates vs Jinja.

From Django docs, it says that custom backend must implement `get_template()`, and optionally `from_string()`. The question then is, what does `get_template()` return. Which can be confusing, because there's at least 2 concepts in Django that are  named "Template" (one coming from `django.template`, the other being the return type of `get_template()`, the "backend" Template.

We're interested in the "backend Template" ([Django](https://github.com/django/django/blob/main/django/template/backends/django.py#L93), [Jinja2](https://github.com/django/django/blob/7ba2a0db20c37a5b1500434ca4ed48022311c171/django/template/backends/jinja2.py#L55)), which has this minimal interface:
```py
class Template:
	template: Renderable
	backend: Backend
	origin: Origin

    def __init__(self, template: Renderable, backend: Backend) -> None:

    def render(self, context: Mapping = None, request = None) -> str:
```

Where, for the sake of clarity, I've called the inner template `Renderable` which has following interface:
```py
class Renderable:
	def render(ctx: Mapping) -> str
```

And other interfaces are:

```py
class Origin:
	name: str
    template_name: str

class Backend:
	def get_template(self, template_name: str) -> Template:
```

---

Equipped with this knowledge, if I were to use `jsx2django_components` in a custom Django template backend, it could possibly be something along these lines:

```py
# See https://github.com/django/django/blob/main/django/template/backends/django.py#L77
from django.template.backends.django import DjangoTemplates

class MyJsxBackend(DjangoTemplates):
    def from_string(self, template_code):
		transformed = jsx2django_components(template_code)
        return Template(self.engine.from_string(transformed), self)

    def get_template(self, template_name):
        try:
			template = self.engine.get_template(template_name), self
			transformed_template = transform_template_jsx2django_components(template)
            return Template(transformed_template, self)
        except TemplateDoesNotExist as exc:
            reraise(exc, self)
```

Issue #527, opened 2024-06-17
Allow users to customize component tags + reintroduce component "inline" block


JuroOravec:
Another week, another feature! @EmilStenstrom Let me know what you think about this one.

## Description

When I was considering libraries like django_components, I felt that the way how components are declared
in the template is too lengthy:

```django
{% component "my_comp" ... %}
{% endcomponent %}
```

and the appeal of django-slippers or django-web-components was that they had a terser syntax, e.g.:

```django
{% my_comp %}
{% endmy_comp %}

OR 

{% #my_comp %}
{% /my_comp %}
```

This has, I think, two important considerations:
1. For an library like django_components, it doesn't add anything useful if we changed how the component tags
   are rendered, from one opinionated approach to another.
2. However, on the level of the broader community, there are some people who prefer shorter syntax. And they may
   choose other libraries like django-slippers/django-web-components, but those libraries may not have many of
   our features. So IMO it's not a great "offering" if one has to choose between e.g. longer syntax + autodiscovery (django_components) vs shorter syntax but with no autodiscovery (django-slippers).

In django-web-components, they did a really interesting solution to this, which is that they allow users to
decide what tags they want to use for the components. [See Component tag formatter](https://github.com/Xzya/django-web-components#component-tag-formatter).

In their case, one can configure the used template tags as so:

```py
class ComponentTagFormatter:
    def format_block_start_tag(self, name):
        return f"#{name}"

    def format_block_end_tag(self, name):
        return f"/{name}"

    def format_inline_tag(self, name):
        return name

# inside your settings
WEB_COMPONENTS = {
    "DEFAULT_COMPONENT_TAG_FORMATTER": "path.to.your.ComponentTagFormatter",
}
```

Which then allows them to call components inside the template as so:

```django
{% #my_comp %}
{% /my_comp %}
```

### TagFormatter class

I gave this a try. For django_components, this had to be a bit modified, because we use the `component` tag,
which is shared by all components. So in our case it's not sufficient to know only about the tag, but we need
to also get a second piece of information (the component name) to know how to render it. Conversely, if we used
tags like `#my_comp`, then we'd have the component name already in the tag.

For this reason, I had to add also `parse_block_start_tag` and `parse_inline_tag`, where the user would have the chance to parse tag inputs, and extract info that specifies what component it is.

<details>
<summary> See TagFormatter API </summary>

```py
class TagFormatter():
    def format_block_start_tag(self, name: str) -> str:
[...]

PR #533, opened 2024-06-22
refactor: Prepare autodiscover and template loader for v1

JuroOravec:
MR for moving autodiscover to it's own file + documenting it, as mentioned in https://github.com/EmilStenstrom/django-components/issues/527#issuecomment-2183091056.


src/django_components/utils.py
```diff
@@ -1,41 +1,4 @@
...
-
-from django_components.template_loader import Loader
-
-
-class SearchResult(NamedTuple):
-    searched_dirs: List[Path]
-    matched_files: List[Path]
-
-
-def search(search_glob: Optional[str] = None, engine: Optional[Engine] = None) -> SearchResult:
```
JuroOravec, line 15: I've split this function to `get_dirs` and `search_dirs` and moved them to `autodiscover.py`

src/django_components/template_loader.py
```diff
@@ -56,18 +78,19 @@ def get_dirs(self) -> List[Path]:
             for app_config in apps.get_app_configs():
                 if not app_config.path:
                     continue
-                if not Path(component_dir).is_dir():
-                    continue
 
-                if path_is_relative_to(component_dir, app_config.path):
+                # NOTE: This applies only to absolute paths
+                if path_is_relative_to(component_dir, app_config.path) and Path(component_dir).is_dir():
                     curr_directories.add(Path(component_dir).resolve())
```
JuroOravec, line 84: Thinking out loud as I'm trying to make sense of this - If we get here to line 84, that means that the current dir from `STATICFILES_DIRS` points to a place WITHIN the current app.

So if the Django project root path is `/path/to/project`, and app's path is `/path/to/project/my_app/`, then `component_dir` also starts with `/path/to/project/my_app/`. So e.g. `/path/to/project/my_app/my_comps/`, in which case `path_is_relative_to()` would return `./my_comps/`.

If all this is true, then we know that the current path from STATICFILES_DIRS is an absolute path, and we can end the loop early, as the rest of the code makes sense only if the path is relative.

src/django_components/template_loader.py
```diff
@@ -56,18 +78,19 @@ def get_dirs(self) -> List[Path]:
...
 
+            # NOTE: This applies only to relative paths
             if hasattr(settings, "BASE_DIR"):
-                path = (Path(settings.BASE_DIR) / component_dir).resolve()
+                path: Path = (Path(settings.BASE_DIR) / component_dir).resolve()
```
JuroOravec, line 88: This branch allows us to treat a relative path in `STATICFILES_DIRS` as relative to the project root. The question is whether we want to support this, since in docs we ask users to set the dirs as:

```py
STATICFILES_DIRS = [
    ...,
    os.path.join(BASE_DIR, "components"),
]
```

At first I was in favour as it could be nicer experience. However, looking at Django's docs on [STATICFILES_DIRS](https://docs.djangoproject.com/en/5.0/ref/settings/#staticfiles-dirs), they ask for full paths. So if we on the other hand ask users to put relative paths into `STATICFILES_DIRS`, it could lead to unexpected behaviour. Instead, we should expect `STATICFILES_DIRS` to have full paths only.

Which means we'd remove the feature of supplying relative paths to `STATICFILES_DIRS`. I think this is fine, as resolving the paths relative to `BASE_DIR` is simple:
```py
Path(settings.BASE_DIR) / component_dir).resolve()
```

src/django_components/template_loader.py
```diff
@@ -56,18 +78,19 @@ def get_dirs(self) -> List[Path]:
...
                 if path.is_dir():
                     curr_directories.add(path)
 
             # Add the directory that holds the settings file
+            # NOTE: This applies only to relative paths
             if settings.SETTINGS_MODULE:
```
JuroOravec, line 94: Same issue is here as I mention in https://github.com/EmilStenstrom/django-components/pull/533/files#r1649603172. Basically we treat the dirs in `STATICFILES_DIRS` as relative altho Django docs ask for absolute paths.

At first I didn't understand why we resolve relative to `SETTINGS_MODULE`, but turns out that was the initial implementation in https://github.com/EmilStenstrom/django-components/blob/58c6e476f1da7a227512f360946fc8f4f88796eb/django_components/template_loader.py. And then, couple of months ago from now, it was branched out in https://github.com/EmilStenstrom/django-components/blob/567c8ccc3d5747bc01a0a4b83c6d54f3880c4c7a/django_components/template_loader.py to fix https://github.com/EmilStenstrom/django-components/issues/290.

Correct me if wrong, but I think the code branch can be summarized as "Look for `component_dir` as relative subpath in settings dir (e.g. `my_app/settings/`), the app where settings are set (e.g. `my_app`), and the root". 

If that's right, we're practically trying to get to the same place as we do in the [code branch above](https://github.com/EmilStenstrom/django-components/pull/533/files#diff-756de4c35ec676b6ad61e551979238b310a7806c7c612add84aaec6f376c1b16R87) with `settings.BASE_DIR`. So we could just use `BASE_DIR` instead of relying on `SETTINGS_MODULE`.

And as I mention in [the comment above](https://github.com/EmilStenstrom/django-components/pull/533/files#r1649603172) on `BASE_DIR`, I think that it's simple enough to use to let users specify the full path via `BASE_DIR` as:
```py
Path(settings.BASE_DIR) / component_dir).resolve()
```

So personally I would remove this code block for `settings.SETTINGS_MODULE`, and tell users to replace that with something like:
```py
Path(settings.BASE_DIR) / "app_with_settings" / "settings").resolve()
```

Is there maybe something I'm missing about `settings.SETTINGS_MODULE`?

src/django_components/template_loader.py
```diff
@@ -56,18 +78,19 @@ def get_dirs(self) -> List[Path]:
             for app_config in apps.get_app_configs():
                 if not app_config.path:
                     continue
-                if not Path(component_dir).is_dir():
```
JuroOravec, line 59: Merged the these two if statements on lines 59 and 62 into a single line

src/django_components/template_loader.py
```diff
@@ -37,6 +57,7 @@ def get_dirs(self) -> List[Path]:
 
         directories: Set[Path] = set()
         for component_dir in component_dirs:
+            # Consider tuples for STATICFILES_DIRS (See #489)
```
JuroOravec, line 60: TODO - Add link to https://docs.djangoproject.com/en/5.0/ref/settings/#prefixes-optional

src/django_components/template_loader.py
```diff
@@ -23,7 +23,27 @@ def path_is_relative_to(child_path: str, parent_path: str) -> bool:
...
+        for a "components" dir. These dirs are tested as both relative and absolute paths.
+
+        Absolute paths are accepted only if they resolve to a directory and point to within one
+        of the apps. E.g. `/path/to/django_project/[app_name]/nested/components/`.
+
+        Relative paths are resolved relative to `BASE_DIR`, e.g. `[BASE_DIR]/components/`.
+
+        Relative paths are also tested against the settings dir. Settings dir is the parent
+        directory of where `SETTINGS_MODULE` points, if set.
+        """
```
JuroOravec, line 46: This docstring needs to be updated.

src/django_components/template_loader.py
```diff
@@ -23,7 +23,27 @@ def path_is_relative_to(child_path: str, parent_path: str) -> bool:
 
 
 class Loader(FilesystemLoader):
+    # TODO - Look deeper into how this function works
+    #      -> Are STATICFILES_DIRS absolute paths or relative to project root? README suggest absolute,
+    #         but the default suggests relative, and code suggests both. Document all the supported cases.
+    #      -> Why do we add settings directory?
+    #      -> Could we merge this with `autodiscover_modules()` in `autodiscover.py`, since there we also
+    #         look for `[app_name]/components`, or do these two accomplish different things?
```
JuroOravec, line 31: Questions:

1. Are STATICFILES_DIRS absolute paths or relative to project root? README suggest absolute, but the default suggests relative, and code suggests both. Document all the supported cases.
	- I've already touched on this in other comments . But [Django asks for full paths](https://docs.djangoproject.com/en/5.0/ref/settings/#staticfiles-dirs).

2. Why do we add settings directory?
	- Addressed in https://github.com/EmilStenstrom/django-components/pull/533/files#r1649615698

3. Could we merge this with `autodiscover_modules()` in `autodiscover.py`, since there we also look for `[app_name]/components`, or do these two accomplish different things?
	- No. `autodiscover_modules` imports individual modules (AKA files), whereas in `STATICFILES_DIRS`, we set DIRS as inputs.

src/django_components/apps.py
```diff
@@ -5,4 +5,9 @@ class ComponentsConfig(AppConfig):
     name = "django_components"
 
     def ready(self) -> None:
+        # This is the code that gets run when user adds django_components to Django's INSTALLED_APPS
+        #
+        # TODO - Allow only if autodiscovery is true?
+        #        Because right now it's kinda weird that we import the COMPONENTS.libraries
+        #        Even if autodiscovery is off.
         self.module.autodiscover()
```
JuroOravec, line 13: I think it makes sense to make the `autodiscover` function as part of the public API, so that people can possibly run it when they need to. However, that invites a discussion on how `autodiscover` should behave:

So autodiscover currently actually imports from 3 places:
1. Component modules - `[app_name]/components.py` for each app in the project
2. Component dirs - As defined by `Loader.get_dirs`, which takes the `STATICFILES_DIRS`, and for each one tries to look in a couple of places
3. Libraries - Modules listed in `COMPONENTS.libraries`.

How things worked up to now is that:
- 1. and 2., so `[app_name]/components.py` and `STATICFILES_DIRS`, were loaded only if `AUTODISCOVER` setting was ON.
- 3., the `COMPONENTS.libraries`, were ALWAYS imported.

This feels a bit random, and raises a few questions:

1. Should we split the logic or rename the function? The "auto*discovery*" is really only the part where we search `STATICFILES_DIRS` for python files. The importing of `COMPONENTS.libraries` and `[app_name]/components.py` is more like "import_libraries" or "import_modules"

2. Should we drop support for automatically searching for `[app_name]/components.py`, and let users set that manually via `COMPONENTS.libraries`? Or we can externalize it as a helper function that generates a list of paths like `[app_name]/components.py`, so users can then pass the results to `COMPONENTS.libraries`.

	Note that the `[app_name]/components.py` feature is NOT documented currently.

	```py
	from django_components.utils import gen_app_paths
	COMPONENTS = {
		"libraries": [
			# Generates `[app_name]/components.py` for each app
			*gen_app_paths("components"),

			# Other 
			"my_app.some.path",
			...
		],
	}
	```

5. How should we handle when `AUTODISCOVER` is ON / OFF?
	First of all, I think we should remove `[app_name]/components.py` (1. from the beginning of this comment) and let users set the same thing via `COMPONENTS.libraries`
	
	Then, there would be only imports via `COMPONENTS.libraries` (3.) and `STATICFILES_DIRS` (2.).

	- If user wanted to turn off imports via `COMPONENTS.libraries`, they could just set it to an empty list
	- And if user wanted to turn off imports from `STATICFILES_DIRS`, they would set `AUTODISCOVER` to `False`.

	So that means that `AUTODISCOVER` would only decide whether to search for python files in `STATICFILES_DIRS`.

---

So all-in-all, my approach would be:
- The `autodiscover.py` file would export two functions: `autodiscover` and `import_libraries`

-  `autodiscover` function would have signature
	```py
	def autodiscover(
	    map_module_names: Optional[Callable[[str], str]] = None,
	) -> List[str]:
	```
	Where:
	- `map_module_names` is renamed from `map_components_modules`, and it is kept as we use that in some of the tests.
	- `autodiscover` function returns a list of module names that were imported
	- `autodiscover` function only searches `STATICFILES_DIRS` for python files, and does NOT import from `COMPONENTS.libraries`, nor from `[app_name]/components.py`.
	- `autodiscover` function would run inside the [`ready` hook](https://github.com/EmilStenstrom/django-components/pull/533/files#diff-2e6e554ee041581d50482b56d52761e5d70e4d1c58dbe75352ac2e72d49b3151R13) only if `AUTODISCOVER` setting is ON. However, user could import the function and run it themselves whenever.

-  `import_libraries` function would have signature
	```py
	def import_libraries(
	    map_module_names: Optional[Callable[[str], str]] = None,
	) -> List[str]:
	```
	Where:
	- `map_module_names` is added to have the same functionality as `autodiscover`
	- `import_libraries` function returns a list of module names that were imported
	- `import_libraries` only imports from `COMPONENTS.libraries`, nothing else.
	- `import_libraries` would run inside the [`ready` hook](https://github.com/EmilStenstrom/django-components/pull/533/files#diff-2e6e554ee041581d50482b56d52761e5d70e4d1c58dbe75352ac2e72d49b3151R13) ALWAYS. However, user could import the function and run it themselves whenever.

So [src/django_components/apps.py](https://github.com/EmilStenstrom/django-components/pull/533/files#diff-2e6e554ee041581d50482b56d52761e5d70e4d1c58dbe75352ac2e72d49b3151) would look like this:
```py
from django.apps import AppConfig

class ComponentsConfig(AppConfig):
    name = "django_components"

    def ready(self) -> None:
        # This is the code that gets run when user adds django_components to Django's INSTALLED_APPS
	from django_components.app_settings import app_settings
	from django_components.autodiscover import autodiscover, import_libraries	

	import_libraries()

	if app_settings.AUTODISCOVER:
	        autodiscover()
```

src/django_components/autodiscover.py
```diff
@@ -0,0 +1,151 @@
...
+            if map_components_modules:
+                module_name = map_components_modules(module_name)
+
+            # This imports the file and runs it's code. So if the file defines any
+            # django components, they will be registered.
+            logger.debug(f'Importing module "{module_name}" (derived from path "{path}")')
+            importlib.import_module(module_name)
+            imported_modules.append(module_name)
+
+    if not skip_libraries:
```
JuroOravec, line 94: It doesn't show up in this diff because I moved the file, but in this WIP I was also playing with the idea of allowing to skip some parts of the `autodiscover`. The reason being that `autodiscover` searches in 3 distinct places. But I prefer more a different approach, the one I've described in https://github.com/EmilStenstrom/django-components/pull/533/files#r1649626042

JuroOravec:
@EmilStenstrom This is a WIP, but I'm eager to hear your thoughts on this, especially on https://github.com/EmilStenstrom/django-components/pull/533#discussion_r1649626042

src/django_components/template_loader.py
JuroOravec: UPDATE: I've refactored the Loader to expect only full paths from `STATICFILES_DIRS`.

src/django_components/template_loader.py
JuroOravec: UPDATE: I've refactored the Loader to use `BASE_DIR` instead of `SETTINGS_MODULE`

src/django_components/template_loader.py
JuroOravec: TODO: Use this for updating README

src/django_components/apps.py
JuroOravec: UPDATE: Updated as described above

src/django_components/__init__.py
```diff
@@ -1,76 +1,7 @@
...
-from django_components.logger import logger
-from django_components.utils import search
+from django_components.autodiscover import autodiscover as autodiscover  # NOQA
+from django_components.autodiscover import import_libraries as import_libraries  # NOQA
 
 if django.VERSION < (3, 2):
     default_app_config = "django_components.apps.ComponentsConfig"
-
-
-def autodiscover(map_import_paths: Optional[Callable[[str], str]] = None) -> List[str]:
```
JuroOravec, line 18: Moved autodiscovery to its own file

src/django_components/template_loader.py
```diff
@@ -2,33 +2,34 @@
...
-# Same as `Path.is_relative_to`, defined as standalone function because `Path.is_relative_to`
-# is marked for deprecation.
-def path_is_relative_to(child_path: str, parent_path: str) -> bool:
-    # If the relative path doesn't start with `..`, then child is descendant of parent
-    # See https://stackoverflow.com/a/7288073/9788634
-    rel_path = os.path.relpath(child_path, parent_path)
-    return not rel_path.startswith("..")
-
-
+# TODO - Document that `STATICFILES_DIRS` MUST have absolute paths
```
JuroOravec, line 14: TODO

src/django_components/template_loader.py
```diff
@@ -2,33 +2,34 @@
...
+
+        If `STATICFILES_DIRS` is not set or empty, then `BASE_DIR` is required.
+        """
         # Allow to configure from settings which dirs should be checked for components
-        if hasattr(settings, "STATICFILES_DIRS") and len(settings.STATICFILES_DIRS):
+        if hasattr(settings, "STATICFILES_DIRS") and settings.STATICFILES_DIRS:
             component_dirs = settings.STATICFILES_DIRS
         else:
-            component_dirs = ["components"]
+            component_dirs = [settings.BASE_DIR / "components"]
```
JuroOravec, line 32: This has to be now full path instead of relative path, because also `settings.STATICFILES_DIRS` must be absolute paths.

src/django_components/template_loader.py
```diff
@@ -47,47 +50,11 @@ def get_dirs(self) -> List[Path]:
                     f"See Django documentation. Got {type(component_dir)} : {component_dir}"
                 )
                 continue
-            curr_directories: Set[Path] = set()
-
-            # For each dir in `settings.STATICFILES_DIRS`, we go over all Django apps
-            # and, for each app, check if the STATICFILES_DIRS dir is within that app dir.
-            # If so, we add the dir as a valid source.
-            # The for loop is based on Django's `get_app_template_dirs`.
-            for app_config in apps.get_app_configs():
```
JuroOravec, line 56: Removed the IF logic on this line (https://github.com/EmilStenstrom/django-components/pull/533/files#diff-756de4c35ec676b6ad61e551979238b310a7806c7c612add84aaec6f376c1b16L65):
```py
if hasattr(settings, "BASE_DIR"):
```

So that we treat STATICFILES_DIRS as full paths. This means that if people previously specified relative paths in STATICFILES_DIRS, they will now have to set them to full paths themselves.

Also removed the IF logic on this line https://github.com/EmilStenstrom/django-components/pull/533/files#diff-756de4c35ec676b6ad61e551979238b310a7806c7c612add84aaec6f376c1b16L71
```py
if settings.SETTINGS_MODULE:
```

Instead of relying on `SETTINGS_MODULE`, we now use only STATICFILES_DIRS and BASE_DIR. So if people used `SETTINGS_MODULE` before, they will have to set `BASE_DIR` instead.

tests/django_test_setup.py
```diff
@@ -1,8 +1,16 @@
+from pathlib import Path
+from typing import Dict, Optional
+
 import django
 from django.conf import settings
 
-if not settings.configured:
+
+def setup_test_config(components: Optional[Dict] = None):
```
JuroOravec, line 8: Instead of configuring Django setting when we import `django_test_setup`, it's now wrapped in a function, so we can configure the settings.

tests/test_template_loader.py
```diff
@@ -0,0 +1,76 @@
+from pathlib import Path
```
JuroOravec, line 1: I've split the `test_autodiscover.py` file and move the tests for template loader into `test_template_loader.py`.

tests/testutils.py
```diff
@@ -89,7 +89,7 @@ def autodiscover_with_cleanup(*args, **kwargs):
 ContextBehParam = Union[ContextBehStr, Tuple[ContextBehStr, Any]]
 
 
-def parametrize_context_behavior(cases: List[ContextBehParam]):
+def parametrize_context_behavior(cases: List[ContextBehParam], settings: Optional[Dict] = None):
```
JuroOravec, line 92: Updated this test helper so that we can pass also other settings to `override_settings`.

JuroOravec:
As of now, I still need to update the README. And, before merging, I want to test these changes on 1. sampleproject, and 2. my own Django project. Because while the tests are passing, I want to confirm that everything's working in live projects too.

JuroOravec:
Happy to report that everything works well both in sampleproject, and in my own Django project! So this MR is now ready for review @EmilStenstrom

README.md
```diff
@@ -46,6 +46,10 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
 
 ## Release notes
 
+🚨📢 **Version 0.83** Autodiscovery module resolution changed. Following undocumented behavior was removed:
+- Previously, autodiscovery also imported any `[app]/components.py` files, and used `SETTINGS_MODULE` to search for component dirs.
```
EmilStenstrom, line 50: If you have your components in a components.py file, what do you need to do to keep using the project? I think that's a good question to answer here.

README.md
```diff
@@ -591,7 +636,19 @@ If you are using autodiscovery, keep a few points in mind:
...
+
+Autodiscovery can be also triggered manually as a function call. This is useful if you want to run autodiscovery at a custom point of the lifecycle:
+
+```py
+from django_components import autodiscover
+
+autodiscover()
+```
+
+Note, though, that if a module was previously imported, it's body will NOT be executed upon calling `autodiscover`.
```
EmilStenstrom, line 651: This line is needlessly complicating things IMHO. Can we remove it?

README.md
```diff
@@ -1917,6 +1974,36 @@ COMPONENTS = {
...
+
+In the rare cases when you need to manually trigger the import of libraries, you can use the `import_libraries` function:
+
+```py
+from django_components import import_libraries
+
+import_libraries()
+```
+
+Note, though, that if a module was previously imported, it's body will NOT be executed upon calling `import_libraries`.
```
EmilStenstrom, line 2005: This line is needlessly complicating things IMHO. Can we remove it?

tests/test_autodiscover.py
```diff
@@ -1,186 +1,168 @@
...
+class _TestCase(TestCase):
+    @classmethod
+    def setUpClass(self) -> None:
+        component.registry.clear()
+        return super().setUpClass()
 
+    @classmethod
+    def tearDownClass(cls) -> None:
+        super().tearDownClass()
+        component.registry.clear()
```
EmilStenstrom, line 22: Why is this on the class level, and not on the test level? The component registry shouldn't be persisted between tests right? Also, I think just clearing on teardown should be enough if things are as they should...

EmilStenstrom (approved):
Excellent additions and clearups as usual. I've added a couple of comments and thoughts, but all in all looks good.

README.md
JuroOravec: Updated!

tests/test_autodiscover.py
JuroOravec: Good point, I used the class methods as I've seen them used before, but didn't question them. I've updated this throughout the tests

README.md
JuroOravec: Updated!

JuroOravec:
Thanks again for all the feedback!

PR #541, opened 2024-07-08
fix: various fixes for inject/provide and html_attrs

JuroOravec:
I came across these errors when upgrading from django-components v0.73 to v0.82:

- fix: Do not escape already escaped slots when slots are given to `slots` kwarg in `Component.render`

- fix: html_attrs allow `attrs` and `defaults` to be `None`

- fix: slots inside `{% provide %}` should have access to provided values

Also modified the raised errors from inside the components' `render` function, so it shows info on which component (name) raised the error.

### DIFF
--- a/tests/test_templatetags_slot_fill.py
+++ b/tests/test_templatetags_slot_fill.py
@@ -210,8 +210,13 @@ class ComponentSlottedTemplateTagTest(BaseTestCase):
             {% endcomponent %}
         """
         template = Template(template_str)
-        with self.assertRaises(TemplateSyntaxError):
+
+        try:
             template.render(Context({}))
+        except RuntimeError as err:
+            self.assertIsInstance(err.__cause__, TemplateSyntaxError)
+        else:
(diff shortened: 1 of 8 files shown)


tests/test_templatetags_slot_fill.py
```diff
@@ -417,20 +420,16 @@ def test_slot_name_fill_typo_gives_helpful_error_message(self):
...
-                self.assertEqual(
-                    textwrap.dedent(
-                        """\
-                        Component 'test1' passed fill that refers to undefined slot: 'haeder'.
-                        Unfilled slot names are: ['footer', 'header'].
-                        Did you mean 'header'?"""
-                    ),
-                    str(e),
-                )
-                raise e
```
JuroOravec, line 433: Simplified this assertion, as it needlessly re-raised the error

JuroOravec:
@EmilStenstrom Tagging you for awareness. I'll go ahead and merge these fixes as I need it for my project.

PR #553, opened 2024-07-20
Add documentation for jetbrains IDEs

daniboygg:
Add documentation about code highlight on single-file-components as discussed on https://github.com/EmilStenstrom/django-components/discussions/548


EmilStenstrom:
Thank you @daniboygg!

PR #561, opened 2024-07-29
fix: do not aggregate kwargs that start with colon

JuroOravec:
We have the feature of "aggregating kwargs", meaning that if I pass following kwarg to a component, `prop:key=value`, then this gets interpreted as "passing kwarg 'prop', which is a dict `{"key": value}`.

As per https://github.com/EmilStenstrom/django-components/discussions/556#discussioncomment-10182903, currently we "aggregate" even component kwargs that start with colon (`:`). These should be ignored.


JuroOravec:
@EmilStenstrom Tagging for visibility, I'll go ahead and merge this tiny fix so it's released as part of 0.85.

PR #562, opened 2024-07-29
refactor: Use top-level exports as public API

JuroOravec:
This MR makes 2 changes:

1. Export public API from `__init__.py` (see https://github.com/EmilStenstrom/django-components/pull/562/files#diff-1aa73be263a2679bafabb9bcc26cabafb8d4bfcf77e243fb9e56f44e0835b7be), so users can import them from the top-level of the package, like so:
	```py
	from django_components import ...
	```

2. Update the canonical way of importing the library, so that users don't have to import submodules. This also means that we will have the freedom to update files within the package without breaking the public contract.

	Before:

	```py
	from django_components import component, types

	types.django_html
	types.css
	types.js
	component.Component
	component.registry
	component.register
	component.AlreadyRegistered
	component.NotRegistered
	```
	
	After:

	```py
	import django_components as dc

    dc.django_html
    dc.css
    dc.js
    dc.component
    dc.registry
    dc.register
    dc.AlreadyRegistered
    dc.NotRegistered
	```

### DIFF
--- a/src/django_components/__init__.py
+++ b/src/django_components/__init__.py
@@ -2,16 +2,16 @@
 import django
 
 # Public API
-from django_components.autodiscover import autodiscover as autodiscover  # NOQA
-from django_components.autodiscover import import_libraries as import_libraries  # NOQA
-from django_components.component import Component as Component  # NOQA
-from django_components.component_registry import AlreadyRegistered as AlreadyRegistered  # NOQA
-from django_components.component_registry import ComponentRegistry as ComponentRegistry  # NOQA
-from django_components.component_registry import NotRegistered as NotRegistered  # NOQA
-from django_components.component_registry import register as register  # NOQA
-from django_components.component_registry import registry as registry  # NOQA
-from django_components.types import css as css  # NOQA
-from django_components.types import django_html as django_html  # NOQA
+from django_components.autodiscover import autodiscover as autodiscover
+from django_components.autodiscover import import_libraries as import_libraries
+from django_components.component import Component as Component
+from django_components.component_registry import AlreadyRegistered as AlreadyRegistered
+from django_components.component_registry import ComponentRegistry as ComponentRegistry
+from django_components.component_registry import NotRegistered as NotRegistered
+from django_components.component_registry import register as register
+from django_components.component_registry import registry as registry
+from django_components.types import css as css
+from django_components.types import django_html as django_html
 from django_components.types import js as js
 
 if django.VERSION < (3, 2):
--- a/src/django_components/__init__.py
+++ b/src/django_components/__init__.py
@@ -1,7 +1,17 @@
 import django
 
-from django_components.autodiscover import autodiscover as autodiscover  # NOQA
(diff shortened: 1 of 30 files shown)


src/django_components/component.py
```diff
@@ -50,6 +41,16 @@
 from django_components.template_parser import process_aggregate_kwargs
 from django_components.utils import gen_id
 
+# TODO_DEPRECATE_V1 - REMOVE IN V1, users should use top-level import instead
```
JuroOravec, line 44: To avoid needless breaking changes, files from component registry are re-exported in `component.py` until v1.

src/django_components/__init__.py
```diff
@@ -1,7 +1,18 @@
+# flake8: noqa F401
 import django
 
-from django_components.autodiscover import autodiscover as autodiscover  # NOQA
-from django_components.autodiscover import import_libraries as import_libraries  # NOQA
+# Public API
```
JuroOravec, line 4: This is where we define the public API

JuroOravec:
Beside the comments above, the rest of the file is just renaming the imports

JuroOravec:
@EmilStenstrom Sorry for the spam today, I had catching up to do 😄 This is another step towards https://github.com/EmilStenstrom/django-components/issues/527, let me know what you think 🙏

EmilStenstrom:
Happy to hear that you're continuing to improve the library with more PRs! I'm in vacation mode as you've noticed, and answer here more slowly. Sorry about that!

1) I'm not fond of with this change is the "dc" name, because it's not apparent what it means. I think `components` would be a better default name, giving some hint as to where things are coming from. Yes, this is a longer name to write, but I think that often being clearer is more important than being short. 

2) I'm not sure I think `dc.css` or `components.css` is as clear as `types.css`? Many are still unused to types in python (remember, Django isn't untyped), so just making clear that "the strange thing after the colon is a type" makes sense to me.

All in all, this is what I'm suggesting:

```python
import django_components as components, types

types.django_html
types.css
types.js
components.Component
components.registry
components.register
components.AlreadyRegistered
components.NotRegistered
```

What do you think?

JuroOravec:
@EmilStenstrom Don't be sorry, I was also away :) And while I do keep making more MRs when the previous ones are merged, there's no rush, enjoy the summer, and reply/review when you feel like it! There's also plenty of work for me on the UI lib.

As for the import I was mainly for `django_components as dc` because I exported `js/css/django_html` at top level, and it felt it would've been weird if people imported the types as

```py
from django_components import js
```

But if we keep the `types` as a separate module (which also makes sense because it's something auxiliary), then IMO we can skip the aliasing, and recommend people to do

```py
from django_components import Component, registry, types, ...
```

EmilStenstrom:
That looks like a good solution to me too!

JuroOravec:
@EmilStenstrom Updated!

EmilStenstrom (commented):
@JuroOravec Looks good to me! Really nice reduction in boilerplate, without sacrificing reability.

PR #563, opened 2024-07-30
Fix upgradecomponent

ekaj2:
Just tried to run this on my production codebase and ran into this easily reproducible issue:

```
    with open(file_path, "r+", encoding="utf-8") as f:
              ^^^^^^^^^
UnboundLocalError: cannot access local variable 'file_path' where it is not associated with a value
```

Also clarified the verbiage in the README for anyone else updating from an archaic version (maybe I'm the only one), but it took me a bit to realize I point this at my templates, not components dir.

Note that this still didn't get me 100% of the way there, but it went from 0-50 and saved me a ton of time. Still missing lots of opening component_block tags and plenty more {% endcomponent %}.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
 
 **Version 0.67** CHANGED the default way how context variables are resolved in slots. See the [documentation](https://github.com/EmilStenstrom/django-components/tree/0.67#isolate-components-slots) for more details.
 
-🚨📢 **Version 0.5** CHANGES THE SYNTAX for components. `component_block` is now `component`, and `component` blocks need an ending `endcomponent` tag. The new `python manage.py upgradecomponent` command can be used to upgrade a directory (use --path argument to point to each dir) of components to the new syntax automatically.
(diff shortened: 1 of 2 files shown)


JuroOravec:
@ekaj2 Thanks for the fix! Which version are you migrating from? Do you have other suggestions on how the migration could be madde easier?

ekaj2:
@JuroOravec Was migrating from 0.37! Really as it currently stands, there's no way around doing either a manual check of everything in your UI or a project-wide grep for the word component and the word component_block (wrapped in <> for the better regex match).

I imagine you could tune the regexes in this upgradecomponent to fix more of these, but perhaps wasted work since likely not many people were running this anyways? Maybe point people here from the README or add a warning there?

I didn't try to debug what exactly missed because I only had ~100 refs, most of which were ok. Mostly guessing, but `{% component a='a' b=b c="c" %}` is probably plenty of complexity to break the regex checker? Several that I manually fixed like that

Appreciate what you and Emil are doing with this, really one of my favorite django packages!

JuroOravec:
Thanks for the kind words!

Sidenote, but it's funny that 0.37 was actually just half a year ago. But since then there's been on average a new release every week, so it totally feels ancient!

Anyway, agree with you. Looking at the upgrade script, it's practically the same as searching for `component_block` throughout the project, so it feels like getting the library towards v1 is a time better spent.

JuroOravec:
Cheers, the fix was now released as part of v0.86!

PR #566, opened 2024-08-03
refactor: prepare registry for custom template tags and docs

JuroOravec:
Changes:
- Document how `ComponentRegistry` works
- `ComponentRegistry()` accepts an optional `library` kwarg
- `@register` decorator accepts an optional `registry` kwarg
- Preparation for custom component template tags:
	- There can be multiple template tags used for rendering the same component (for the "block" and "inline" forms)
	- When user registers a component with not-seen-before template tags, the tags are automatically registered
  into to `library`. And when the last component that uses those tags is unregistered,
  the tag is removed from `library`.

This is the last step before https://github.com/EmilStenstrom/django-components/issues/527

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,7 +64,6 @@ target/
 # lock file is not needed for development
 # as project supports variety of Django versions
 poetry.lock
-pyproject.toml
 
 # PyCharm
 .idea/
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
 - [Using single-file components](#using-single-file-components)
 - [Use components in templates](#use-components-in-templates)
 - [Use components outside of templates](#use-components-outside-of-templates)
+- [Registering components](#registering-components)
 - [Use components as views](#use-components-as-views)
 - [Autodiscovery](#autodiscovery)
 - [Using slots in templates](#using-slots-in-templates)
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -64,7 +64,6 @@ target/
 # lock file is not needed for development
 # as project supports variety of Django versions
 poetry.lock
-pyproject.toml
```
JuroOravec, line 67: Noticed that this was in the gitignore. Strangely, I don't think it made the top-level `pyproject.toml` ignored. Since the project now uses `pyproject.toml`, we should remove this one.

src/django_components/component_registry.py
```diff
@@ -14,49 +16,274 @@ class NotRegistered(Exception):
...
+        """
         existing_component = self._registry.get(name)
-        if existing_component and existing_component._class_hash != component._class_hash:
+        if existing_component and existing_component.cls._class_hash != component._class_hash:
             raise AlreadyRegistered('The component "%s" has already been registered' % name)
-        self._registry[name] = component
+
+        used_tags = ["component"]
+
+        # Keep track of which components use which tags, because multiple components may
```
JuroOravec, line 114: What this part does is that when a user defines [custom component tags](https://github.com/EmilStenstrom/django-components/issues/527) , e.g. `my_comp` and `#my_comp`, then we'd add these template tags to the underlying `django.template.Library`.

This way, so it's suficient for users to load only our template tag library (e.g. with `{% load component_tags %}`), and they can still use their custom tags.

And then, bit lower down (https://github.com/EmilStenstrom/django-components/pull/566/files#diff-a5dcf86b1b1bd690092e8ed60318835a1ce9ed2c62befcc05d43dfd4459393b6R150), we do the reverse, and remove any registered component-specific template tags, when the last component using said tag(s) is unregistered.

JuroOravec:
@EmilStenstrom Would you give this one a look? Thanks!

src/django_components/component_registry.py
```diff
@@ -14,49 +16,274 @@ class NotRegistered(Exception):
...
+        for tag in entry.tags:
+            # Unlink component from tag
+            self._tags[tag].remove(name)
+
+            # Cleanup
+            is_tag_empty = not len(self._tags[tag])
+            if is_tag_empty:
+                del self._tags[tag]
+
+            is_protected = tag in {"component", "#component"}
```
EmilStenstrom, line 159: Hmm... So we are bringing back #component as an inline tag available by default? I'm open to discussing this, but it doesn't match Django's defaults having tags prefixed with "#". Thoughts?

EmilStenstrom: is_protected should probably be global, and checked for in register as well, don't you think? Otherwise you could mess up the component tag as well...

tests/testutils.py
```diff
@@ -20,6 +20,13 @@
 
 
 class BaseTestCase(SimpleTestCase):
+    def setUp(self) -> None:
+        super().setUp()
+        # Ensure that the module that defines the template tags is executed again after import
+        if "django_components.templatetags.component_tags" in sys.modules:
+            del sys.modules["django_components.templatetags.component_tags"]
+        registry._library = None
```
EmilStenstrom, line 28: Could you explain this, I'm not really following...

EmilStenstrom (commented):
Great changes again, including the simplification of tests!

README.md
```diff
@@ -643,6 +644,126 @@ Note: slots content are automatically escaped by default to prevent XSS attacks.
...
+### Registering components to custom ComponentRegistry
+
+In rare cases, you may want to manage your own instance of `ComponentRegistry`,
+or register components onto a different `Library` instance than the default one.
+
+The `Library` instance can be set at instantiation of `ComponentRegistry`. If omitted,
+then the default Library instance from django_components is used.
+
+```py
+from django.template import Library
```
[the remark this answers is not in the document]
JuroOravec: Updated in https://github.com/EmilStenstrom/django-components/pull/566/commits/03fd922607890763d82bdea4034a31b9828c241a!

src/django_components/component_registry.py
JuroOravec: As for this

> Hmm... So we are bringing back #component as an inline tag available by default? I'm open to discussing this, but it doesn't match Django's defaults having tags prefixed with "#". Thoughts?

Let's discuss that on the next MR, which will include the tag formatter. In my first implementation of the tag formatter, I assumed that there will ALWAYS be both an "inline" and "block" forms. But since you're bringing this up, I can refactor the tag formatter such that it can define only inline or only block forms. That way, the default implementation could be Django-compatible, and users would opt-in into `#component` format.

EmilStenstrom (approved):
Happy to merge and discuss more in the next PR.

PR #572, opened 2024-08-05
feat: TagFormatter - Allow users to customize component template tags

JuroOravec:
~Changes:~ UPDATE  - See latest comments (from 9th Aug)
- ~Add inlined slot tag `#slot` (https://github.com/EmilStenstrom/django-components/pull/572/commits/c97bf87dba8a02ed3229c5bb4daf635ea1aa8ecc)~
- ~Add inlined fill tag `#fill` (https://github.com/EmilStenstrom/django-components/pull/572/commits/159cc64c8a650c9200038a6957da78696a5eb98d)~
	- UPDATE: All block tags can now be written as inline, if they end with a slash - `{% slot "xxx" / %}`
	- Also, all these block tags also support the `kwarg:key=val` syntax, and are processed the same way
- Prepare component tag for custom tags - Instead of registering the `component` tag upfront, the tags will be registered at the time of component registration (https://github.com/EmilStenstrom/django-components/pull/572/commits/fbb0f54799eed023dcd7693c967e312db701db67)
- Add TagFormatter (https://github.com/EmilStenstrom/django-components/pull/572/commits/e23ddf7603ad16a9b1e4b720cfb520bf4c98580b)
- Allow user to set TagFormatter via settings (https://github.com/EmilStenstrom/django-components/pull/572/commits/63d1fd4a0960c940c17290e7004c3eaf9f8efaf0)
- Add predefined TagFormatters to public API (https://github.com/EmilStenstrom/django-components/pull/572/commits/4606bedefdcc7ea5b913db16306222f5b4ddc3de)
- Hook up TagFormatter with ComponentRegistry (https://github.com/EmilStenstrom/django-components/pull/572/commits/4770a2376ea39f078a102a0f990ce75e93465279) 
- ~Add inline tags to protected tags (https://github.com/EmilStenstrom/django-components/pull/572/commits/ec598569c6b525190a853cbd86f67ffdde3f05c6)~
- Refactor: Pass around component name as string instead of template expression (https://github.com/EmilStenstrom/django-components/pull/572/commits/591c29d1f0914f7cc31b89d432cd796ffa9d75ce)

TODO:
- [x] Agree on the API
- [x] Tests
- [x] Docs

Closes https://github.com/EmilStenstrom/django-components/issues/527


src/django_components/templatetags/component_tags.py
```diff
@@ -217,6 +243,15 @@ def do_component(parser: Parser, token: Token) -> ComponentNode:
     return component_node
 
 
+def do_component_tag_block(parser: Parser, token: Token) -> ComponentNode:
```
JuroOravec, line 246: These two component tag functions intentionally don't have `@register.tag("component")` decorator. Instead, we call these functions at component registration from within `ComponentRegistry`, after we have obtained start and end tags from `TagFormatter`.

src/django_components/templatetags/component_tags.py
```diff
@@ -286,13 +321,32 @@ def _check_for_isolated_context_keyword(bits: List[str]) -> Tuple[List[str], boo
...
+def _parse_tag_body(parser: Parser, end_tag: str, inline: bool) -> NodeList:
+    if inline:
+        body = NodeList()
+    else:
+        body = parser.parse(parse_until=[end_tag])
+        parser.delete_first_token()
+    return body
+
+
 def _parse_component_with_args(
```
JuroOravec, line 333: In this function is logic to parse args/kwargs of a component tag. I removed the part that finds the component name. Instead, this logic now lives in `TagFormatter`. So, in a way, `TagFormatter` now pre-processes the args/kwargs, extracting the component name out. And what remains, we parse as regular args/kwargs.

src/django_components/tag_formatter.py
```diff
@@ -0,0 +1,269 @@
...
+
+from django_components.app_settings import app_settings
+from django_components.expression import resolve_string
+from django_components.template_parser import VAR_CHARS
+from django_components.utils import is_str_wrapped_in_quotes
+
+TAG_RE = re.compile(r"^[{chars}]+$".format(chars=VAR_CHARS))
+
+
+class TagFormatterABC(abc.ABC):
```
JuroOravec, line 16: This is the API of the `TagFormatter`, inspired by https://github.com/Xzya/django-web-components#component-tag-formatter. To define a custom `TagFormatter`, one must define following functions:

- `format_block_start_tag`
- `format_block_end_tag`
- `format_inline_tag`
- `parse_block_start_tag`
- `parse_inline_tag`

Unlike `django-web-components`, our `TagFormatter` requires that we define also how to extract (parse) the component name from the inputs. This was required so we could transform our current way of defining components into a TagFormatter.
```django
{% component "my_comp" %}
```
Because `django-web-components` assumed that the component name can be extracted from the tag name. But in our case, the tag name is `"component"`, so on its own it's not enough information. So instead we pass all the args/kwargs to the parser, which would extract the name `my_comp` from the example above, and return other args/kwargs unchanged.

src/django_components/tag_formatter.py
```diff
@@ -0,0 +1,269 @@
...
+    def safe_format_inline_tag(self, name: str) -> str:
+        """Formats the start tag of an inline component. Raises ValueError if the tag is invalid."""
+        tag = self.format_inline_tag(name)
+        self.validate_tag(tag, "inline_tag")
+        return tag
+
+
+# TODO - Update docs that these are names of positional args
+# TODO - Add tests
+class ComponentTagFormatter(TagFormatterABC):
```
JuroOravec, line 121: This is the default tag formatter, which has the same behavior as current. Except this also adds an "inlined" tag `#component`.

src/django_components/tag_formatter.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def _parse_start_or_inline_tag(self, tokens: List[str]) -> Tuple[str, List[str]]:
+        _, *comp_args = tokens
+
+        if not comp_args:
+            raise TemplateSyntaxError("ComponentTagFormatter: Component tag did not receive tag name")
+
+        # First arg is a kwarg, not a positional arg. Hence look for the "name" kwarg
+        # for component name.
+        if "=" in comp_args[0]:
```
JuroOravec, line 174: This is where the logic from https://github.com/EmilStenstrom/django-components/commit/fbb0f54799eed023dcd7693c967e312db701db67#r145067467 went.

src/django_components/tag_formatter.py
```diff
@@ -0,0 +1,269 @@
...
+        comp_name = resolve_string(comp_name)
+
+        return comp_name, final_args
+
+
+# TODO - Add tests
+# TODO - UPDATE DOCS FOR EVERYTHING - ShorthandTagFormatter, ComponentTagFormatter, TagFormatterABC, tag_formatter settings
+# TODO - DOCUMENT IN README HOW TAG_FORMATTER WORKS
+# TODO - DOCUMENT IN README HOW INLINE VS "BLOCK" components work
+class ShorthandTagFormatter(TagFormatterABC):
```
JuroOravec, line 211: This is the syntax similar to django-web-components or django-slippers, where the tag name is derived from the component name, e.g.:
```django
{% mycomp abc=123 %}
    {% fill "myfill" %}
        ...
    {% endfill %}
{% endmycomp %}
```

I included this as it seems as a recurring theme that some users want shorter syntax for components.

src/django_components/component_registry.py
```diff
@@ -268,6 +270,39 @@ def clear(self) -> None:
...
+        component: Type["Component"],
+    ) -> ComponentRegistryEntry:
+        # Lazily import to avoid circular dependencies
+        from django_components.templatetags.component_tags import do_component_tag_block, do_component_tag_inline
+
+        formatter = get_component_tag_formatter()
+
+        # TODO - Allow None here?
+        # register the inline tag
+        inline_tag = formatter.safe_format_inline_tag(comp_name)
```
JuroOravec, line 285: Currently we expect that `TagFormater` WILL define tags for both block and inline variants. If we want to allow `TagFormatters` that would for example implement only the block variant (like discussed here https://github.com/EmilStenstrom/django-components/pull/566#discussion_r1703654170), we could refactor this code to either allow :
- `format_inline_tag` not to be defined,
- or (better) allow `format_inline_tag` to return `None`, which would mean "this TagFormatter does not implement inline tags"

src/django_components/component_registry.py
```diff
@@ -268,6 +270,39 @@ def clear(self) -> None:
...
+
+        if is_tag_protected(self.library, inline_tag):
+            raise TagProtectedError('Cannot register inline tag "%s", this tag name is protected' % inline_tag)
```
JuroOravec, line 288: Here we throw if user tried to register a component tag with one of the protected names.

JuroOravec:
@EmilStenstrom This is the rest of the code I have on customizing the component tags + adding inline variants of our tags. Ofc, everything is up for discussion. I recommend to go commit by commit as listed in the main description, to get a better feeling for the flow.

One note - turns out that when you write a comment on the code in one of the commits, it doesn't show up in this feed. So I wrote all comments twice - once in the commits code, and once of the code for this MR (https://github.com/EmilStenstrom/django-components/pull/572/files). So watch out that, when writing/replying to comments, that you're on the "Files" tab of this MR.

src/django_components/app_settings.py
```diff
@@ -115,5 +120,10 @@ def _validate_context_behavior(self, raw_value: ContextBehavior) -> ContextBehav
             valid_values = [behavior.value for behavior in ContextBehavior]
             raise ValueError(f"Invalid context behavior: {raw_value}. Valid options are {valid_values}")
 
+    # TODO - ADD TEST
+    @property
+    def TAG_FORMATTER(self) -> Union[str, Type["TagFormatterABC"]]:
```
JuroOravec, line 125: When it comes to using the `TagFormatter` by users, the flow could look like this:
1. User defines their own `TagFormatter`, which inherits from the `TagFormatterABC`, and defines the `format_*` and `parse_*` methods, e.g. like the [ShorthandTagFormatter](https://github.com/EmilStenstrom/django-components/pull/572/files#diff-d49ecbc3623a4ed5408729717a2ce020b9f336330a24a90cc7e75da2ea7416e1R211) does.
2. They set their tag formatter via settings, either as import path:
	```py
	COMPONENTS = {
		"tag_formatter": "path.to.MyTagFormatter",
	}
	```
	Or by reference
	```py
	from path.to import MyTagFormatter

	COMPONENTS = {
		"tag_formatter": MyTagFormatter,
	}
	```
3. Profit

JuroOravec:
Lastly, on the topic of "inlined" tags - I'd like to have that feature, as I feel like 50% of the time, I use `{% component %}` tag with no body, and about 90% of the time, I use `{% slot %}` tag with no body. 

I added also inline variant for `{% fill %}` for completeness. While I didn't need it yet, I can imagine that it could be useful to have empty `fill` tag when someone just wants to overwrite the default fill.

On the other hand, the `{% provide %}` tag doesn't make sense without the fill.

For all of the "inline" tags, I went with using hashtag `{% #tag %}`, inspired by [django-web-components](https://github.com/Xzya/django-web-components#rendering-components), but we could also go with something else.

EmilStenstrom:
I'm not sure I like the look of `{% #component date="2024-08-08" %}`. If prefixing things with "#" was a standard in the django world, then maybe that would work, but I find no indication that it is. It doesn't feel intuitive what you see it for the first time either. My first thought was "is it inspired by CSS ids?" which is obviously not the right thought.

Another idea is inspired by HTML `<component></component>` -> `<component />` 
* `{% component date="2024-08-08" / %}` 
* or `{% component date="2024-08-08" end %}`
* or `{% component date="2024-08-08" empty %}`

Wouldn't that we a better default?

JuroOravec:
Agree, my initial go to was to have something similar to HMTL opening/closing tags `< />`, e.g. `{% /%}` (without the space between the slash and percent). But that would require making changes to Django's templating logic, because we'd be adding a custom closing symbol.

But I didn't think of having the slash separated with a space! I really like that.

`{% component date="2024-08-08" / %}`

While also a keyword could work, like `end` or `empty`, I think it's less clear that it's not part of the input.

EmilStenstrom:
Happy to hear you like it too!

Then I guess the next step is figuring out how this affects customization. Saying we add support for self-closing tags with a slash, what's the next most probable customization people might want?

* `{% component XYZ %}` -> `{% XYZ %}` - "I don't want to have to write component over and over"
* `{% component XYZ / %}` -> `{% #component XYZ %}` - "I want to be compatible with django-web-components"

What else?

If we support / even in custom tags, would there even be a need to customize that?

JuroOravec:
> `{% component XYZ %} `-> `{% XYZ %}` - "I don't want to have to write component over and over"

This is already possible with the TagFormatter.

> `{% component XYZ / %}` -> `{% #component XYZ %}` - "I want to be compatible with django-web-components"

My first instinct was not to allow users changing of the syntax for a self-closing tag. However, the way the `TagFormatter` is written, we can very easily extend it to give users the power to decide whether a component is interpreted as an inlined or block. 🤯 

In this MR we already have the [`parse_block/inline_tag` hooks](https://github.com/EmilStenstrom/django-components/pull/572/files#diff-d49ecbc3623a4ed5408729717a2ce020b9f336330a24a90cc7e75da2ea7416e1R150), where we ask the users to extract name from the tag and forward the rest of the args/kwargs. We could also allow them to return the additional parameter from those functions. A boolean or enum, indicating whether the parsed tag is inline or block.

And then, at the point of parsing the component inputs from the `{% component %}` tag ([here](https://github.com/EmilStenstrom/django-components/pull/572/files#diff-7345171ed38e1dd2e4b8199b65a2fd082323bdb331ff0ef0dc0dd3bd43f111aeR215) and [here](https://github.com/EmilStenstrom/django-components/pull/572/files#diff-7345171ed38e1dd2e4b8199b65a2fd082323bdb331ff0ef0dc0dd3bd43f111aeR339)), we'd receive the `is_inline` variable from `formatter.parse_tag`, instead of passing it from outside.

So, in terms of API, that means that:
- The slash at the end would be only our convention, but users could override that if needed.
- The TagFormatter would have only 2 formatting functions `format_start_tag` and `format_end_tag` (instead of 3), and 1 parsing function `parse_start_tag` (instead of 2). So TagFormatter wouldn't have specific methods for block and inline tags. Instead, `parse_start_tag` would tell us whether it's block or inline.

---

However, if we allow users to decide if `{% component %}` is block or inline, then IMO we should allow them that also for `{% slot %}` and `{% fill %}` tags - so users could potentially rename those, or change how the self-closing tags are written.

So, in terms of API:

- TagFormatter would have different methods for `{% component %}`, `{% slot %}`, and `{% fill %}` tags:
	```py
	class TagFormatter:
		# Component tag
	    def format_component_start_tag(self, name: str) -> str:
	        ...
	    def format_component_end_tag(self, name: str) -> str:
	        ...
	    def parse_component_start_tag(self, tokens: List[str]) -> Tuple[str, List[str], bool]:
	        ...

		# Slot tag
	    def format_slot_start_tag(self, name: str) -> str:
	        ...
	    def format_slot_end_tag(self, name: str) -> str:
	        ...
	    def parse_slot_start_tag(self, tokens: List[str]) -> Tuple[str, List[str], bool]:
	        ...

		# Fill tag
	    def format_fill_start_tag(self, name: str) -> str:
	        ...
	    def format_fill_end_tag(self, name: str) -> str:
	        ...
	    def parse_fill_start_tag(self, tokens: List[str]) -> Tuple[str, List[str], bool]:
	        ...
	```

	Or, for a snappier interface, we could break it down, and users would configure individual tags  via `COMPONENTS.tag_formatters`:
	```py
	class TagFormatter:
	    def start_tag(self, name: str) -> str:
	        ...
	    def end_tag(self, name: str) -> str:
	        ...
	    def parse(self, tokens: List[str]) -> Tuple[str, List[str], bool]:
	        ...
	```
	
	```py
	# settings.py

	COMPONENTS = {
		"tag_formatters": {
			"component": MyComponentFormatter,
			"slot": MySlotFormatter,
			"fill": MyFillFormatter,
		},
	}
	```

---

And I think a bonus point would be if we made it easy to rename the these tags, so there'd be 3 levels of usage:
1. No customization - uses default `{% component %}`, `{% slot %}`, `{% fill %}`
2. Light customization - Rename e.g. `{% component %}` to `{% comp %}` just by changing a variable or subclassing the default TagFormatter:
	```py
	# settings.py

	class MyTagFormatter(ComponentTagFormatter):
		component_tag = "comp"

	COMPONENTS = {
		"tag_formatter": MyTagFormatter,
	}
	```
3. Advanced customization - This involves writing their own TagFormatter, e.g. as we did here https://github.com/EmilStenstrom/django-components/pull/572/files#diff-d49ecbc3623a4ed5408729717a2ce020b9f336330a24a90cc7e75da2ea7416e1R121

JuroOravec:
Noice, got it working locally!

1. Default `{% component "name" ... %}{% endcomponent %}`, but with tag changed to `comp`:
	
	Screenshot 2024-08-08 at 21 12 50
	
	<br/>
	
	Screenshot 2024-08-08 at 21 12 57

2. "Shorthand" syntax `{% <name> ... %}{% end<name> %}`

	Screenshot 2024-08-08 at 21 15 17
	<br/>
	Screenshot 2024-08-08 at 21 15 35

And I also used the slash to indicate the "inline" component.

EmilStenstrom:
I think keeping things simple here is probably a good idea. We don't know how people will want to customize, so it's easy to over-engineer. 

* I think allowing customization of slot/fill tags is just overkill, and something we can skip
* Customizing if something is inline or block isn't needed either, as long as custom tags supports ending with "/" instead
* The combo start_tag/end_tag/parse_tag looks good to me.

src/django_components/__init__.py
```diff
@@ -3,14 +3,27 @@
...
+    registry as registry,
+)
+from django_components.library import TagProtectedError as TagProtectedError
+from django_components.tag_formatter import (
+    ComponentFormatter as ComponentFormatter,
+    ShorthandComponentFormatter as ShorthandComponentFormatter,
+    TagFormatterABC as TagFormatterABC,
+    TagResult as TagResult,
+    component_formatter as component_formatter,
+    component_shorthand_formatter as component_shorthand_formatter,
```
JuroOravec, line 25: And here, lines 18-25, are new public exports

src/django_components/app_settings.py
```diff
@@ -115,5 +118,9 @@ def _validate_context_behavior(self, raw_value: ContextBehavior) -> ContextBehav
             valid_values = [behavior.value for behavior in ContextBehavior]
             raise ValueError(f"Invalid context behavior: {raw_value}. Valid options are {valid_values}")
 
+    @property
+    def TAG_FORMATTER(self) -> Union["TagFormatterABC", str]:
```
JuroOravec, line 122: Updated as per feedback, the settings now allow to set only a single TagFormatter, which is then used for component tags.

src/django_components/attributes.py
```diff
@@ -5,69 +5,42 @@
...
 
 HTML_ATTRS_DEFAULTS_KEY = "defaults"
 HTML_ATTRS_ATTRS_KEY = "attrs"
 
 
-def _default(val: Any, default_val: Any) -> Any:
-    return val if val is not None else default_val
-
-
 class HtmlAttrsNode(Node):
```
JuroOravec, line 17: Previously, when we passed aggregate kwargs (those `attrs:class="border-red"`) to `slot` or `html_attrs` tags, we processed these keys only at the render time.

I did an update to tag parsing, so that all our tags use the same common logic for pasring the inputs. And as part of that, also the aggregate kwargs are pre-processed up front. So a lot of logic from here could be removed.

src/django_components/component_registry.py
```diff
@@ -1,25 +1,16 @@
...
+from django_components.library import Tags, is_tag_protected, mark_protected_tags, register_tags_from_formatter
+from django_components.tag_formatter import get_tag_formatter
+
 if TYPE_CHECKING:
     from django_components.component import Component
 
 _TComp = TypeVar("_TComp", bound=Type["Component"])
 
 
-PROTECTED_TAGS = [
```
JuroOravec, line 11: Moved the logic that's concerned with working with Django's Library object into `library.py`. Also, added a test that checks that the tag protection works correctly.

https://github.com/EmilStenstrom/django-components/pull/572/files#diff-0abc741bfcebe9debcf5eb70d01e66db1ccb9036877f195680d25bf73e9e9083R144

src/django_components/template_parser.py
```diff
@@ -41,7 +44,7 @@
     num=r"[-+\.]?\d[\d\.e]*",
     # The following is the only difference from the original FilterExpression. We allow variable names to have extra
     # special characters: - : . @ #
-    var_chars=r"\w\-\:\@\.\#",
```
JuroOravec, line 44: Refactored the regex out, so we can re-use it to validate the user-generated tag names from TagFormatter

src/django_components/tag_formatter.py
```diff
@@ -0,0 +1,256 @@
...
+        as that's the input to the component.
+
+        So in the end, we return a tuple:
+
+        `('my_comp', ['key=val', 'key2=val2'])`
+        """
+        ...
+
+
+class InternalTagFormatter:
```
JuroOravec, line 72: TagFormatter classes in this file:
- `TagResult` - The named tuple that users are expected to return from `TagFormatter.parse()`
- `TagFormatterABC` - The TagFormatter interface.
- `InternalTagFormatter` - When calling the user-provided methods, we validate that the generated tag names are valid identifiers. `InternalTagFormatter` encapsulates this validation.
- `SimpleTagFormatter` - Generic implementation of the `tag/endtag` style tags
- `ComponentFormatter` - The default behaviour with `component/endcomponent` tags
- `ShorthandComponentFormatter` - The `comp_name/endcomp_name` style tags.
	- Note the difference: While `SimpleTagFormatter` and `ComponentFormatter` define a single tag for all components, `ShorthandComponentFormatter` defines a different tag for each component

src/django_components/templatetags/component_tags.py
```diff
@@ -62,83 +65,89 @@ def get_components_from_preload_str(preload_str: str) -> List["Component"]:
...
 @register.tag("slot")
-def do_slot(parser: Parser, token: Token) -> SlotNode:
-    # e.g. {% slot <name> ... %}
-    tag_name, *args = token.split_contents()
-    slot_name, is_default, is_required, slot_kwargs = _parse_slot_args(parser, args, tag_name)
-    # Use a unique ID to be able to tie the fill nodes with components and slots
-    # NOTE: MUST be called BEFORE `parser.parse()` to ensure predictable numbering
-    slot_id = gen_id()
-    trace_msg("PARSE", "SLOT", slot_name, slot_id)
+def slot(parser: Parser, token: Token) -> SlotNode:
```
JuroOravec, line 119: For the review, I suggest to open this file separately, to get a better feel for the changes. Primarily, I cleaned up the tag functions. Instead of doing custom parsing in each tag, there's now [`_parse_tag`](https://github.com/EmilStenstrom/django-components/pull/572/files#diff-7345171ed38e1dd2e4b8199b65a2fd082323bdb331ff0ef0dc0dd3bd43f111aeR335), which handles it all.

So all tags (`component`, `slot`, `fill`, `provide`, ...) now offer the same behavior - using `/` for inline tags, handling the `kwarg:key=val` syntax, etc.

This doesn't apply to the template tags that deal with the JS/CSS dependencies, as those might yet change when we work on that.

---

Also, I updated the tag names to match the tag names.

JuroOravec:
@EmilStenstrom I retracked some of those changes, so only the component tags can are configurable. See the other comments for the rest of the changes. Still need to write up documentation, but the code is ready for review!

JuroOravec:
I've just updated the README, and this is now ready for review :)

README.md
```diff
@@ -414,12 +443,16 @@ First load the `component_tags` tag library, then use the `component_[js/css]_de
     {% component_css_dependencies %}
 </head>
 <body>
-    {% component "calendar" date="2015-06-19" %}{% endcomponent %}
+    {% component "calendar" date="2015-06-19" / %}{% endcomponent %}
```
EmilStenstrom, line 446: Remove slash here, alternative is shown below.

README.md
```diff
@@ -1885,6 +1910,122 @@ If you find yourself using the `only` modifier often, you can set the [context_b
...
+{# or #}
+
+{% button href="..." disabled / %}
+```
+
+### Available TagFormatters
+
+django_components provides following predefined TagFormatters:
+
+- **`ComponentFormatter` (`django_components.component_formatter`)**
```
EmilStenstrom, line 1947: Add short description and say it's default.

README.md
```diff
@@ -1885,6 +1910,122 @@ If you find yourself using the `only` modifier often, you can set the [context_b
...
+
+- **`ShorthandComponentFormatter` (`django_components.shorthand_component_formatter`)**
```
EmilStenstrom, line 1949: Short description?

src/django_components/tag_formatter.py
```diff
@@ -0,0 +1,256 @@
...
+
+class TagFormatterABC(abc.ABC):
+    def start_block_tag(self, name: str) -> str:
+        """Formats the start tag of a block component."""
+        return self.start_tag(name)
+
+    def start_inline_tag(self, name: str) -> str:
+        """Formats the start tag of an inline component."""
+        return self.start_tag(name)
+
```
EmilStenstrom, line 33: Do we still need these?

src/django_components/tag_formatter.py
EmilStenstrom: In the docstring for the code below it looks like SimpleTagFormatter and ShorthandComponentFormatter does the exact same thing.

Also, the README omitts SimpleTagFormatter as the main defined tags.

Can you explain?

EmilStenstrom (commented):
Great work as always! :)

Added some small comments. But a bigger one: Does we really need to keep the possibility to have separate inline and block tags? I think it looks like that adds some unnessesary complexity to the code....

README.md
JuroOravec: Fixed in https://github.com/EmilStenstrom/django-components/pull/572/commits/9061c3226ae70e2a23788063c3127d4136059752

README.md
JuroOravec: Added in https://github.com/EmilStenstrom/django-components/pull/572/commits/9061c3226ae70e2a23788063c3127d4136059752

src/django_components/tag_formatter.py
JuroOravec: Yeah, no. When I was implementing this, my point of reference was a feature that would support syntaxes as found in django_components, django-web-components, and django-slippers. And django-slippers technically uses a different tag for inline and block tags. But since we're using the slash `/` to indicate inline component, this is not relevant anymore.

Update in https://github.com/EmilStenstrom/django-components/pull/572/commits/d9593ffdff0cd8dc087a2a04fff919e82680c432

src/django_components/tag_formatter.py
JuroOravec: Not the same thing, `SimpleTagFormatter` takes uses the TAG name, while `ShorthandComponentFormatter` uses the COMPONENT name.

I defined `SimpleTagFormatter` initially for the version that supported TagFormatters also for slots and fills. But kept it afterwards, because `ComponentFormatter` technically uses the same logic. But `SimpleTagFormatter` was not meant to be public.

I removed it in https://github.com/EmilStenstrom/django-components/pull/572/commits/73a748194c829fdf7c0865d009adff035ff2ff0b to avoid confusion

JuroOravec:
@EmilStenstrom Thanks for the review, really appreciate it! I've addressed the comments.

EmilStenstrom (approved):
Fantastic work! I think this API ended up really slick! :)

Issue #580, opened 2024-08-13
Global context not available when rendering via view methods with RequestContext


larrybotha:
When rendering components dynamically via a view method, i.e `get`, `post`, etc. the global context is not available to render included templates when `RequestContext` is provided as context

## steps to reproduce

- install 0.88 (earlier versions fail, too - last known working version was 0.63, but I haven't determined where the issue was introduced)
- create a component that is rendered both directly and via a GET request
- use `{% include ... %}` to render another template from within the component
- make a GET request to render the component
- provide request context in the `get` method using Django's `RequestContext`

## expected

- included template renders, `request` object is available inside component

## actual

- component fails to render

## demos

- working, 0.63: https://replit.com/@larrybotha1/Django-Components-063
    - note that passing `RequestContext` via `render_to_response` provides access to `request` within the component
- failing, 0.88:  https://replit.com/@larrybotha1/Django-Components-088
    - passing `RequestContext` to `render_to_response` results in the component failing to render. Evaluating the context at this point shows that the context is empty, apart from the top-level boolean values that Django makes available in context

I may very well be misunderstanding how to pass context through, however, so any pointers would be appreciated!

EmilStenstrom:
@JuroOravec You have any idea of what's going on here?

JuroOravec:
@larrybotha thanks for the detailed report, I'll have a look later in the evening

JuroOravec:
Found the culprit. So it wasn't related to `RequestContext`, but to how we search the `{% include %}` tags within component's template for slots, [here](https://github.com/EmilStenstrom/django-components/blob/master/src/django_components/node.py#L72).

That piece of logic relies on the `Template` being "bound" to the `Context`. In other words, for `context.template` to be defined.

To be frank, I don't know much about Django's flow in this regard, e.g. at which point Django sets it. But setting it manually seems to work.

I came across this issue before, so just needed to adjust an if statement.

MR for the fix: https://github.com/EmilStenstrom/django-components/pull/582

JuroOravec:
Fix for this is now released in [v0.89](https://github.com/EmilStenstrom/django-components/releases/tag/0.89) 🎉 

Btw, @larrybotha, thanks again for including those demos, it helped tremendously!

larrybotha:
[...]

PR #582, opened 2024-08-16
fix: Set `Context.template` for components to work with include

JuroOravec:
Fixes https://github.com/EmilStenstrom/django-components/issues/580


src/django_components/component.py
```diff
@@ -387,9 +383,11 @@ def _render_impl(
         with context.update(context_data):
             template = self.get_template(context)
             _monkeypatch_template(template)
-            if not context_was_given:
+
+            if context.template is None:
```
JuroOravec, line 387: The fix itself was just adjusting this condition

PR #585, opened 2024-08-19
feat: Typing for component inputs and access inputs during render

JuroOravec:
Changes:
- Add typing for `Component.render` / `Component.render_to_response`, and allow to pass generics to the `Component` class to specify args, kwargs, slots, and data returned from `get_context_data`.

	```py
    TestCompArgs = Tuple[int, str]

    class TestCompKwargs(TypedDict):
        variable: str
        another: int

    class TestCompData(TypedDict):
        abc: int

    class TestCompSlots(TypedDict):
        my_slot: str

	class TestComponent(Component[TestCompArgs, TestCompKwargs, TestCompData, TestCompSlots]):
		...
	```
	- Args must be a `Tuple`
	- Kwargs, slots, and data can be any mapping, but most practical is to define them as `TypedDict`, so one gets an error when an invalid key is set.
	
	Screenshot 2024-08-19 at 10 11 53
	
	Screenshot 2024-08-19 at 10 15 20

- Allow to access inputs passed to `Component.render` via `self.input` ([see example](https://github.com/EmilStenstrom/django-components/pull/585/files#diff-b35cc8d48a3ea2ddde0c3bc9de0b0b768c567f9e8f532dc4ca1b260fe06e4ac8R234))

- BREAKING CHANGE: Do not inherit from `django.views.View`. Instead, make a View a nested class.
	- This one was necessary, because, for an unknown reason, inheriting from `View` was breaking the generics. Possibly there was some circular reference, because I didn't get the error when I copied the `View` class to local file and used that.

		So, instead of doing
		```py
		class MyComponent(Component):
		    template = """..."""
		
		    def post(self, request, *args, **kwargs) -> HttpResponse:
		        variable = request.POST.get("variable")
		        return self.render_to_response(kwargs={"variable": variable})
		```

		One should define the `get/post/...` methods inside a nested `View` class, subclassing `ComponentView`, and use `self.component` to access the component instance:
		
		```py
		class MyComponent(Component):
		    template = """..."""
		    
		    class View(ComponentView):
		        def post(self, request, *args, **kwargs) -> HttpResponse:
		            variable = request.POST.get("variable")
					# This line changed:
		            return self.component.render_to_response(kwargs={"variable": variable})
		```
	- I did manage to keep `Component.as_view`, so that part of the API didn't have to change.

Closes https://github.com/EmilStenstrom/django-components/issues/483, closes https://github.com/EmilStenstrom/django-components/discussions/516 (see https://github.com/EmilStenstrom/django-components/discussions/516#discussioncomment-10187403), addresses "Access slots as variables in template" in https://github.com/EmilStenstrom/django-components/discussions/486

### DIFF
--- a/tests/components/multi_file/multi_file.py
+++ b/tests/components/multi_file/multi_file.py
@@ -2,19 +2,20 @@ from typing import Any, Dict
 
 from django.http import HttpResponse
 
-from django_components import Component, register
+from django_components import Component, ComponentView, register
 
 
 @register("multi_file_component")
 class MultFileComponent(Component):
     template_name = "multi_file/multi_file.html"
 
-    def post(self, request, *args, **kwargs) -> HttpResponse:
-        variable = request.POST.get("variable")
-        return self.render_to_response({"variable": variable})
-
-    def get(self, request, *args, **kwargs) -> HttpResponse:
-        return self.render_to_response({"variable": "GET"})
-
     def get_context_data(self, variable, *args, **kwargs) -> Dict[str, Any]:
         return {"variable": variable}
+
+    class View(ComponentView):
+        def get(self, request, *args, **kwargs):
+            return self.component.render_to_response({"variable": "GET"})
+        
+        def post(self, request, *args, **kwargs) -> HttpResponse:
+            variable = request.POST.get("variable")
+            return self.component.render_to_response({"variable": variable})
--- a/tests/components/multi_file/multi_file.py
+++ b/tests/components/multi_file/multi_file.py
@@ -13,7 +13,7 @@ class MultFileComponent(Component):
         variable = request.POST.get("variable")
         return self.render_to_response({"variable": variable})
 
-    def get(self, request, *args, **kwargs):
+    def get(self, request, *args, **kwargs) -> HttpResponse:
         return self.render_to_response({"variable": "GET"})
 
     def get_context_data(self, variable, *args, **kwargs) -> Dict[str, Any]:
--- a/tests/components/multi_file/multi_file.py
+++ b/tests/components/multi_file/multi_file.py
@@ -15,7 +15,7 @@ class MultFileComponent(Component):
     class View(ComponentView):
         def get(self, request, *args, **kwargs):
             return self.component.render_to_response({"variable": "GET"})
-        
+
         def post(self, request, *args, **kwargs) -> HttpResponse:
             variable = request.POST.get("variable")
             return self.component.render_to_response({"variable": variable})
--- a/tests/components/multi_file/multi_file.py
+++ b/tests/components/multi_file/multi_file.py
@@ -2,20 +2,19 @@ from typing import Any, Dict
 
 from django.http import HttpResponse
 
-from django_components import Component, ComponentView, register
+from django_components import Component, register
 
 
 @register("multi_file_component")
 class MultFileComponent(Component):
     template_name = "multi_file/multi_file.html"
 
+    def post(self, request, *args, **kwargs) -> HttpResponse:
+        variable = request.POST.get("variable")
+        return self.render_to_response({"variable": variable})
+
+    def get(self, request, *args, **kwargs):
+        return self.render_to_response({"variable": "GET"})
+
     def get_context_data(self, variable, *args, **kwargs) -> Dict[str, Any]:
         return {"variable": variable}
-
-    class View(ComponentView):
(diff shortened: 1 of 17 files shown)


JuroOravec:
@EmilStenstrom Would you have a look at this one?

EmilStenstrom:
@JuroOravec I really like the typing bit. It's one thing that has always been hard to do well in Django when you're passing around dicts randomly all over the place. Good looks fine!

But I don't like the nested class View at all here. Seems it's complicating an API that were working fine before and breaking existing code for people. Is there really no way to make this work without changing the API?!

JuroOravec:
> But I don't like the nested class View at all here. Seems it's complicating an API that were working fine before and breaking existing code for people. Is there really no way to make this work without changing the API?!

@EmilStenstrom IMO there's two parts to this:
1. I don't know how else to solve the error, there's not much to grasp at. See below, I get the error when `Component` subclasses `django.views.View`. It doesn't happen when I copy the code for `View`, but I don't want to do that.

	Screenshot 2024-08-21 at 08 32 18

2. IMO in the long run it's better not to subclass directly from `View` as it can lead to unexpected conflicts. E.g. few months back, in my project I wanted to add a `setup` method to `Component`, and could figure out what was wrong, until I found `View` also has a method called `setup`. So from that perspective it's good to have the actual `View` class nested (or otherwise separated).

When it comes to preserving the API, what we can do is to support the `get/post/...` methods on the top-level, while `ComponentView` would, by default, take the `get/post/...` from the Component instance, e.g.:

```py
class ComponentView(View):
    component = cast("Component", None)

    def __init__(self, component: "Component", **kwargs: Any) -> None:
        super().__init__(**kwargs)
        self.component = component

	def get(self, *args, **kwargs):
		self.component.get(*args, **kwargs)

	def post(self, *args, **kwargs):
		self.component.post(*args, **kwargs)
```

That way, people can still define get/post/... on the Component, and will need to set `Component.View` only if they want to subclass or otherwise change the View class.

JuroOravec:
Nice, IMO this looks better. So `get/post/...` are still defined on the Component class. And users define `Component.View` only if they want to override the View class itself.
- [I added a section on modifying the View class](https://github.com/EmilStenstrom/django-components/pull/585/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R748)
- The `Component.View` class is by default `ComponentView`, which implements all the HTTP method handlers [here](https://github.com/EmilStenstrom/django-components/pull/585/files#diff-35c84ebceae50f7149f03bbcaf2b4f1289bc2984bf7859ca521b5dbe7960c396R106) to point to `Component.get`, `Component.post`, etc.

tests/components/multi_file/multi_file.py
```diff
@@ -13,7 +13,7 @@ def post(self, request, *args, **kwargs) -> HttpResponse:
         variable = request.POST.get("variable")
         return self.render_to_response({"variable": variable})
 
-    def get(self, request, *args, **kwargs) -> HttpResponse:
+    def get(self, request, *args, **kwargs):
```
EmilStenstrom, line 16: Why is this no longer typed?

tests/components/relative_file/relative_file.py
```diff
@@ -17,7 +17,7 @@ def post(self, request, *args, **kwargs) -> HttpResponse:
         variable = request.POST.get("variable")
         return self.render_to_response({"variable": variable})
 
-    def get(self, request, *args, **kwargs) -> HttpResponse:
+    def get(self, request, *args, **kwargs):
```
EmilStenstrom, line 20: Why is this no longer typed?

tests/components/single_file.py
```diff
@@ -19,7 +19,7 @@ def post(self, request, *args, **kwargs) -> HttpResponse:
         variable = request.POST.get("variable")
         return self.render_to_response({"variable": variable})
 
-    def get(self, request, *args, **kwargs) -> HttpResponse:
+    def get(self, request, *args, **kwargs):
```
EmilStenstrom, line 22: Why is this no longer typed?

EmilStenstrom (approved):
Looks really good! Some small docs issues, but go ahead!

JuroOravec:
@EmilStenstrom Thanks again for your time and help! I've put the types back and added links to the relevant sections to the release notes. With that resolved, I'm going to merge this 🚀 

catjam-vibin

Issue #587, opened 2024-08-19
Allow dynamic slots and fills


JuroOravec:
This is for the UI component library, which is a port of Vuetify. Vue allows to dynamically define and populate slots, and Vuetify uses this for their data tables, [allowing to define a slot for only a specific column / header](https://vuetifyjs.com/en/components/data-tables/basics/#slots).

With this in mind, it would be useful to allow dynamic names for the `slot` and `fill` tags.

```django
{% slot "content" %}            {# <--- literal #}
{% slot my_var %}               {# <--- dynamic #}
{% slot "{{ my_var }}" %}       {# <--- dynamic expression #}

{% fill "content" %}            {# <--- literal #}
{% fill my_var %}               {# <--- dynamic #}
{% fill "{{ my_var }}" %}       {# <--- dynamic expression #}
```

For this to work, the slot resolution would have to be moved from "compile-time" (parsing of the Template) to "render-time" (rendering of the Template with given Context).

EmilStenstrom:
Exciting! :)

Issue #539, opened 2024-07-07
Proposals to make it possible to keep HTML attributes inside the HTML file


JuroOravec:
## Problem

I'm updating one of my projects from v0.73 to v0.82. Features like the `html_attrs` tag, `prop:key=value` syntax and provide/inject cleaned up the code really nicely.

However, there are still a few use cases where I need to define HTML attributes in Python (instead of keeping them in the HTML), to work around the limitations of the library.  This adds unnecesary mental overhead as I need to trace where the HTML attributes belong.

Consider the example below, where:

- I am passing `attrs` (dict with HTML attrs) to child component `tab`. I want to add some extra HTML attrs to it (`@click=...`). Ideally the `@click=...` HTML attribute could live in the template, directly on the `tab` component. But because I needed to merge my HTML attribute with the user-given `attrs`, I had to define it in the Python section.

- I'm passing in a list of TabItem objects. These have `disabled` attribute. However, `input_form` component accepts `editable`, which in this case is the negation of `disabled`. I cannot do `editable=not disabled` inside the template, so I had to convert the list of `TabItems` to a list of dictionaries, so I could define an extra key `readonly`.

```py
class TabItem(NamedTuple):
	content: str
	disabled: bool = False

@component.register("tabs")
class Tabs(Component):
    template = """
		<div>
			{% for tab in tabs %}
				{% component "tab"
					content=tab.content
					disabled=tab.disabled
					attrs=attrs
				%}{% endcomponent %}

				{% component "input_form"
					editable=tab.editable
				%}{% endcomponent %}
			{% endfor %}
		</div>
	"""

    def get_context_data(
        self,
        *_args,
		tabs: list[TabItem],
        attrs: dict | None = None,
    ):
        final_attrs = {
			**(attrs or {}),
			"@click": "(evt) => alert(evt.target.data)"
		}

		final_tabs = []
		for tab in tabs:
			final_tabs.append({
				"content": tab.content,
				"disabled": tab.disabled,
				"editable": not tab.disabled,
			})

        return {
			"tabs": final_tabs,
            "attrs": final_attrs,
        }
```

### ❌ Custom filters

While custom filters could be enough for the negation:

```django
editable=tab.disabled|not
```

It's already insufficient for ternary (if/else). Django has the built-in [`yesno` filter](https://docs.djangoproject.com/en/5.0/ref/templates/builtins/#yesno), but the filter works only with strings. I cannot use `yesno` to decide between two objects.

This works:
```django
[...]

PR #594, opened 2024-08-23
refactor: move kwargs resolution to render-time + cleanup

JuroOravec:
Refactor that moves the processing of kwargs passed to template tags from compile- to render-time. 

This is a precursor for the dynamic expressions (https://github.com/EmilStenstrom/django-components/discussions/552), spread operator (https://github.com/EmilStenstrom/django-components/issues/539), and dynamic slots and fills (https://github.com/EmilStenstrom/django-components/issues/587).

### DIFF
--- a/src/django_components/__init__.py
+++ b/src/django_components/__init__.py
@@ -19,6 +19,10 @@ from django_components.component_registry import (
     registry as registry,
 )
 from django_components.library import TagProtectedError as TagProtectedError
+from django_components.slots import (
+    SlotContent as SlotContent,
+    SlotFunc as SlotFunc,
+)
 from django_components.tag_formatter import (
     ComponentFormatter as ComponentFormatter,
     ShorthandComponentFormatter as ShorthandComponentFormatter,
--- a/src/django_components/attributes.py
+++ b/src/django_components/attributes.py
@@ -8,7 +8,7 @@ from django.template import Context
 from django.utils.html import conditional_escape, format_html
 from django.utils.safestring import SafeString, mark_safe
 
-from django_components.expression import RuntimeKwargs, RuntimeKwargPairs
+from django_components.expression import RuntimeKwargPairs, RuntimeKwargs
 from django_components.node import BaseNode
 
 HTML_ATTRS_DEFAULTS_KEY = "defaults"
--- a/src/django_components/attributes.py
+++ b/src/django_components/attributes.py
@@ -4,37 +4,46 @@
 
 from typing import Any, Dict, List, Mapping, Optional, Tuple
 
-from django.template import Context, Node
+from django.template import Context
 from django.utils.html import conditional_escape, format_html
 from django.utils.safestring import SafeString, mark_safe
 
-from django_components.expression import Expression, safe_resolve
+from django_components.expression import RuntimeKwargs, RuntimeKwargPairs
+from django_components.node import BaseNode
 
 HTML_ATTRS_DEFAULTS_KEY = "defaults"
 HTML_ATTRS_ATTRS_KEY = "attrs"
 
 
-class HtmlAttrsNode(Node):
+class HtmlAttrsNode(BaseNode):
     def __init__(
         self,
-        attributes: Optional[Expression],
-        defaults: Optional[Expression],
-        kwargs: List[Tuple[str, Expression]],
+        kwargs: RuntimeKwargs,
+        kwarg_pairs: RuntimeKwargPairs,
+        node_id: Optional[str] = None,
     ):
-        self.attributes = attributes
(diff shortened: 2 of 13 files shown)


src/django_components/__init__.py
```diff
@@ -19,6 +19,10 @@
     registry as registry,
 )
 from django_components.library import TagProtectedError as TagProtectedError
+from django_components.slots import (
+    SlotContent as SlotContent,
+    SlotFunc as SlotFunc,
```
JuroOravec, line 24: Renamed `SlotRenderContent` and `SlotRenderFunc` and made them public. [See how SlotFunc is used here](https://github.com/EmilStenstrom/django-components/pull/594/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R628)

src/django_components/attributes.py
```diff
@@ -4,37 +4,46 @@
...
-from django_components.expression import Expression, safe_resolve
+from django_components.expression import RuntimeKwargPairs, RuntimeKwargs
+from django_components.node import BaseNode
 
 HTML_ATTRS_DEFAULTS_KEY = "defaults"
 HTML_ATTRS_ATTRS_KEY = "attrs"
 
 
-class HtmlAttrsNode(Node):
+class HtmlAttrsNode(BaseNode):
```
JuroOravec, line 18: Added a common base class for all `Node` subclasses

src/django_components/expression.py
```diff
@@ -57,7 +62,90 @@ def resolve_string(
...
+    return "=" in key
+
+
 def is_aggregate_key(key: str) -> bool:
     # NOTE: If we get a key that starts with `:`, like `:class`, we do not split it.
     # This syntax is used by Vue and AlpineJS.
     return ":" in key and not key.startswith(":")
+
+
+def process_aggregate_kwargs(kwargs: Mapping[str, Any]) -> Dict[str, Any]:
```
JuroOravec, line 75: Moved this here from `template_parser`, so `template_parser` contains only the template-parsing logic that we overrode from Django

src/django_components/templatetags/component_tags.py
```diff
@@ -436,31 +450,13 @@ def mark_kwarg_key(key: str, is_agg_key: bool) -> None:
...
     # Parse args/kwargs that will be passed to the fill
-    args, raw_kwarg_pairs = parse_bits(
+    args, kwarg_pairs = parse_bits(
         parser=parser,
         bits=bits,
         params=[] if isinstance(params, bool) else params,
         name=tag_name,
     )
 
-    # Post-process args/kwargs - Mark special cases like aggregate dicts
```
JuroOravec, line 446: Removed because we now process the `prop:key=val` syntax at render-time.

src/django_components/expression.py
```diff
@@ -1,50 +1,55 @@
...
 
+Expression = Union[FilterExpression]
+RuntimeKwargsInput = Dict[str, Expression]
+RuntimeKwargPairsInput = List[Tuple[str, Expression]]
 
-class AggregateFilterExpression:
-    def __init__(self, dict: Dict[str, FilterExpression]) -> None:
-        self.dict = dict
 
+class RuntimeKwargs:
```
JuroOravec, line 11: Made the kwargs object into it's own class, so we can call `.resolve()` on it. This is where the processing of the `prop:key=val` syntax happens now.

src/django_components/expression.py
```diff
@@ -1,50 +1,55 @@
...
+    if not value.isidentifier():
+        return False
+    return True
+
+
+def safe_resolve_list(context: Context, args: List[Expression]) -> List:
+    return [arg.resolve(context) for arg in args]
 
 
-def safe_resolve(context_item: Expression, context: Context) -> Any:
```
JuroOravec, line 42: Turns out `safe_resolve()` wasn't really needed, and instead we can call `.resolve()` directly on the objects that support it.

src/django_components/expression.py
```diff
@@ -1,50 +1,55 @@
-from typing import Any, Dict, List, Mapping, Optional, Union
+from typing import Any, Dict, List, Mapping, Optional, Tuple, Union
 
-from django.template import Context
+from django.template import Context, TemplateSyntaxError
 from django.template.base import FilterExpression, Parser
 
+Expression = Union[FilterExpression]
```
JuroOravec, line 6: It may look strange that there's a `Union` with a single element. But this is a preparation for the dynamic expressions and spread operator syntax, which will be part of the union.

EmilStenstrom (approved):
Looks good to me!

PR #596, opened 2024-08-23
feat: add spread operator

JuroOravec:
This MR adds a spread operator to the template tags:

```django
{% component 'test'
    var_a
    ...my_dict
    x=123
    ...item
/ %}
```

Closes https://github.com/EmilStenstrom/django-components/issues/539


README.md
```diff
@@ -1431,105 +1432,6 @@ As seen above, you can pass arguments to components like so:
 </body>
 ```
 
-### Special characters
```
JuroOravec, line 1434: Moved these two sections onto the chapter on template tag syntax.

JuroOravec:
@EmilStenstrom Another relatively small (tests are about 400 lines). Also I did this one first, as I might play a bit more with the tags/expressions. Thanks for your time!

src/django_components/expression.py
```diff
@@ -72,6 +109,24 @@ def is_aggregate_key(key: str) -> bool:
     return ":" in key and not key.startswith(":")
 
 
+def is_spread_operator(value: Any) -> bool:
+    if not isinstance(value, str) or not value:
+        return False
+
+    return value.startswith("...")
```
EmilStenstrom, line 116: There needs to be something after the spread operator too right?

src/django_components/expression.py
```diff
@@ -72,6 +109,24 @@ def is_aggregate_key(key: str) -> bool:
...
+
+# A string that starts with `...1=`, `...29=`, etc
+ESCAPED_SPREAD_OPERATOR_RE = re.compile(r"^\.\.\.\d+=")
+
+
+def is_escaped_spread_operator(value: Any) -> bool:
+    if not isinstance(value, str) or not value:
+        return False
+
+    return bool(ESCAPED_SPREAD_OPERATOR_RE.match(value))
```
EmilStenstrom, line 127: I don't get this. Why would you ever write "...123"?

EmilStenstrom (commented):
Great feature (as we've talked about before)!

I'm missing some negative testcase, like if you write just "..." in a component, and maybe ... followed by a non-dict?

src/django_components/expression.py
JuroOravec: I didn't check before, but added a check now, [here](https://github.com/EmilStenstrom/django-components/pull/596/files#diff-7345171ed38e1dd2e4b8199b65a2fd082323bdb331ff0ef0dc0dd3bd43f111aeR422). So if it's just `...` without a value, it raises an error

src/django_components/expression.py
JuroOravec: So the "escaped_spread_operator" is meant to be internal, users are not expected to write `...1=`.

We convert the spread syntax `...dict` to the "escaped/internal" format `...1=dict`, so Django still parses it as a kwarg, and it remains in the original position.

And there's the number, since there can be multiple spread operators, so they would then be `...1=`, `...2=`, etc.

This way it's trivial to merge the kwargs after the spread operator is replaced with actual values, because we don't have to guess the position where it originally was.

I've added this comment and renamed the vars to `internal_spread_operator` in https://github.com/EmilStenstrom/django-components/pull/596/commits/9ba0c8404b329dc6ed30fae538ab17de1e5a54c7

JuroOravec:
@EmilStenstrom Good idea with the negative tests, I've added them here https://github.com/EmilStenstrom/django-components/pull/596/commits/4f881703091551899b75d7f8d60e4f2dca0b41fe

src/django_components/expression.py
EmilStenstrom: Very clever! Thanks for explaining :)

EmilStenstrom (approved):

EmilStenstrom:
Great work, let's ship it! :)

PR #605, opened 2024-08-24
feat: add dynamic expressions

JuroOravec:
Good news, unlike I [said in the discussion](https://github.com/EmilStenstrom/django-components/discussions/552#discussioncomment-10394843) previously, I actually managed to pull off supporting both `{{ }}` and `{% %}` tags (and also `{# #}`)!

Also, after going over how [django-cotton designed it](https://github.com/wrabit/django-cotton#template-expressions-in-attributes), I feel like their approach is more useful than what I proposed initially.

So initially, I was saying that we should allow only a single tag, and no text.

So while this would work:

```django
{% my_component model="{{ model_name }}" %}
```

This would not

```django
{% my_component model="todos.{{ index }}.name" %}
```

But now I think that the latter is more flexible and more can be achieved with it. So the text is practically evaluated as a regular django template.

So @EmilStenstrom this is the same as [you suggested](https://github.com/EmilStenstrom/django-components/discussions/552#discussioncomment-10403531) with:

```django
{% component 'test' title="{{ person.first_name }} {{ person.last_name }}" / %}
```

But to still make it possible to use the outputs of those nested tags as inputs to component, I added following logic:
- If the string consists of a single tag, and no other text, then the output does not get stringified.

	So, below, the result from `random_int` would be passed to `test` component as `int`
	```django
	{% component 'test' page="{% random_int 10 20 %}" / %}
	```
- However, if there IS extra text, or multiple tags, then the result gets stringified.

	Because there are extra spaces around the tag, this would now pass a string.
	```django
	{% component 'test' page=" {% random_int 10 20 %} " / %}
	```

TODO:
- [x] Update README

Closes https://github.com/EmilStenstrom/django-components/discussions/552

### DIFF
--- a/README.md
+++ b/README.md
@@ -1972,8 +1972,6 @@ In the example above:
 
 This is inspired by [django-cotton](https://github.com/wrabit/django-cotton#template-expressions-in-attributes).
 
-> Note: Never use this feature to mix business logic and template logic. Business logic should still be in the template!
-
 #### Passing data as string vs original values
 
 Sometimes you may want to use the template tags to transform
--- a/README.md
+++ b/README.md
@@ -1972,6 +1972,8 @@ In the example above:
 
 This is inspired by [django-cotton](https://github.com/wrabit/django-cotton#template-expressions-in-attributes).
 
+> Note: Never use this feature to mix business logic and template logic. Business logic should still be in the template!
+
 #### Passing data as string vs original values
 
 Sometimes you may want to use the template tags to transform
@@ -2027,6 +2029,8 @@ Similar is possible with [`django-expr`](https://pypi.org/project/django-expr/),
   value="{% expr 'input_value if is_enabled else None' %}"
 / %}
 ```
+
+> Note: Never use this feature to mix business logic and template logic. Business logic should still be in the template!
 
 ### Pass dictonary by its key-value pairs
 
--- a/README.md
+++ b/README.md
@@ -65,6 +65,10 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
 
 ## Release notes
 
+**Version 0.93**
+- Spread operator `...dict` inside template tags. See [Spread operator](#spread-operator))
+- Use template tags inside string literals in component inputs. See [Use template tags inside component inputs](#use-template-tags-inside-component-inputs))
+
 🚨📢 **Version 0.92**
 - BREAKING CHANGE: `Component` class is no longer a subclass of `View`. To configure the `View` class, set the `Component.View` nested class. HTTP methods like `get` or `post` can still be defined directly on `Component` class, and `Component.as_view()` internally calls `Component.View.as_view()`. (See [Modifying the View class](#modifying-the-view-class))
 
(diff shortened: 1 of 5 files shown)


EmilStenstrom:
@JuroOravec really nice, I like this design!

JuroOravec:
@EmilStenstrom Great to hear! I've updated the README, and so this is now ready for review

README.md
```diff
@@ -1927,6 +1931,103 @@ Other than that, you can use spread operators multiple times, and even put keywo
...
+```
+
+Similar is possible with [`django-expr`](https://pypi.org/project/django-expr/), which adds an `expr` tag and filter that you can use to evaluate Python expressions from within the template:
+
+```django
+{% component "my_form"
+  value="{% expr 'input_value if is_enabled else None' %}"
+/ %}
+```
+
```
EmilStenstrom, line 2030: Maybe we should add a disclaimer here? Something like: "Note: Never use this feature to mix business logic and template logic. Business logic should still be in the template!"

(I really like this feature, component is an extremely powerful tag now!)

src/django_components/expression.py
```diff
@@ -109,6 +176,37 @@ def is_aggregate_key(key: str) -> bool:
...
+# A string that must start and end with quotes, and somewhere inside includes
+# at least one tag. Tag may be variable (`{{ }}`), block (`{% %}`), or comment (`{# #}`).
+DYNAMIC_EXPR_RE = re.compile(
+    r"""^{quote}.*?(?:{var_tag}|{block_tag}|{comment_tag}).*?{quote}$""".format(
+        var_tag=r"(?:\{\{.*?\}\})",
+        block_tag=r"(?:\{%.*?%\})",
+        comment_tag=r"(?:\{#.*?#\})",
+        quote=r"['\"]",
+    )
+)
```
EmilStenstrom, line 188: This matches strings starting with one quote type, and ending with another: "hallo'

ChatGPT suggests this fix:

```python
DYNAMIC_EXPR_RE = re.compile(
    r"""^(?P<quote>['"]).*?(?:{var_tag}|{block_tag}|{comment_tag}).*?(?P=quote)$""".format(
        var_tag=r"(?:\{\{.*?\}\})",
        block_tag=r"(?:\{%.*?%\})",
        comment_tag=r"(?:\{#.*?#\})",
    )
)

```

src/django_components/templatetags/component_tags.py
```diff
@@ -480,13 +481,27 @@ def mark_kwarg_key(key: str, is_agg_key: bool) -> None:
...
+            args.append(val)
+
     kwarg_pairs: RuntimeKwargPairsInput = []
     for key, val in raw_kwarg_pairs:
         is_spread_op = is_internal_spread_operator(key + "=")
 
         if is_spread_op:
-            expr = parser.compile_filter(val.token)
+            # Allow to use dynamic expressions with spread operator, e.g.
+            # `..."{{ }}"`
```
EmilStenstrom, line 497: This is just crazy, I not sure I think we should include this. Will confuse everyone that sees and expression like this!

src/django_components/templatetags/component_tags.py
```diff
@@ -560,6 +575,88 @@ def _parse_tag_body(parser: Parser, end_tag: str, inline: bool) -> NodeList:
     return body
 
 
+def _fix_nested_tags(parser: Parser, token: Token) -> None:
+    block_token = token
```
EmilStenstrom, line 579: Is the intent here to just rename the variable? Can't this be done in the def right above?

src/django_components/templatetags/component_tags.py
```diff
@@ -560,6 +575,88 @@ def _parse_tag_body(parser: Parser, end_tag: str, inline: bool) -> NodeList:
...
+    # When our template tag contains a nested tag, e.g.:
+    # `{% component 'test' "{% lorem var_a w %}"`
+    #
+    # Django parses this into:
+    # `TokenType.BLOCK: 'component 'test'     "{% lorem var_a w'`
+    #
+    # Above you can see that the token ends at the end of the NESTED tag,
+    # and includes `{%`. So that's what we use to identify if we need to fix
+    # nested tags or not.
+    has_unclosed_tag = block_token.contents.count("{%") > block_token.contents.count("%}")
```
EmilStenstrom, line 590: This wouldn't work for this crazy usage I think: `{% component 'test' "{% lorem var_a="{%}" %}`? Do we care? :)

tests/test_expression.py
```diff
@@ -61,6 +78,316 @@ def test_safe_resolve_dict(self):
...
+        self.assertEqual(captured["list_var"], "  ")
+
+        self.assertHTMLEqual(
+            rendered,
+            """
+            <div></div>
+            <div> </div>
+            <div></div>
+            <div> </div>
+            """,
```
EmilStenstrom, line 307: Maybe not use assertHTMLEqual here since this is testing stringification and handling of spaces?

tests/test_expression.py
```diff
@@ -61,6 +78,316 @@ def test_safe_resolve_dict(self):
...
+        self.assertEqual(captured["list_var"], " [{'a': 1}, {'a': 2}] ")
+
+        self.assertHTMLEqual(
+            rendered,
+            """
+            <div>lorem ipsum dolor</div>
+            <div>lorem ipsum dolor [{'a': 1}]</div>
+            <div>True</div>
+            <div>[{'a': 1}, {'a': 2}]</div>
+            <div>{'a': 3}</div>
```
EmilStenstrom, line 386: Maybe not use assertHTMLEqual here since this is testing stringification and handling of spaces?

tests/test_expression.py
```diff
@@ -96,7 +423,7 @@ def get_context_data(
             {% component 'test'
                 var_a
                 ...my_dict
-                ...item
+                ..."{{ list|first }}"
```
EmilStenstrom, line 426: Are we reaaaaaallly sure we want see this in people's code?! :) 

The steps to understand this code:
1) "Oh, I didn't know the spread operator worked on strings! Oh wait, that's not a string."
2) "It's a list! And first returns the first item. So we're spreading an item in a list, what's that?"
3) "Oh, the list contains little dicts, and this dict doesn't get stringified because there isn't spaces around it."
4) "So this is actually just "a=1", I get it!"

EmilStenstrom (commented):
This looks very good! I'm a little worried that the parsing here will break when people start doing crazy things like nesting tags several levels deep, quoting django template tags in strings, and expecting them to work.

src/django_components/templatetags/component_tags.py
JuroOravec: Of course we do! Fixed it :)

src/django_components/expression.py
JuroOravec: Very nice! I did check for the quotes mismatch, but separately in the `is_dynamic_expression`. I've updated it to check the quotes inside the regex

JuroOravec:
@EmilStenstrom Really good feedback, love how thorough it is!

I addressed the feedback [here](https://github.com/EmilStenstrom/django-components/pull/605/commits/0a473a86ec02262594f57bd2e209653dd176c2b3):
- Replaced `assertHtmlEqual` with `assertEqual`
- Covered the `"{%}"` edge case by also checking for unfinished quotes
- Updated the regex to check for quotes in it
- Added the disclaimer for `"Note: Never use this feature to mix business logic and template logic. Business logic should still be in the template!"`
- Added a test to check behavior of [nested tags in nested tags](https://github.com/EmilStenstrom/django-components/pull/605/files#diff-8200e7404420069f578b2bbd98b34348e75285cb3f9a273b85e488ba6eaeadb1R415)

When it comes to points on the complex code like `"'{{ '{{nested}}' }}'"` or `..."{{ list|first }}"`, I don't mind changing the changing the syntax, but I'd really like to have these features.

The reason is, that this comes really handy for the UI component library, where there are cases like this:

```vue
<MyComponent>
  <template #content="contentProps">

    <OtherComp v-bind="{ ...contentProps, ...attrs }" />

  </template>
</MyComponent>
```

In other words, they take data given by a slot (`contentProps`), and combine it with user-provided data (`attrs`), and then spread it onto `OtherComp` using the spread operator (`v-bind`).

So in Django, the equivalent would look like this:

```django
{% component "my_component" %}
	{% fill "content" data="content_props" %}

		{% component "other_comp" ..."{% expr '{ **content_props, **attrs }' %}" %}

	{% endfill %}
{% endcomponent %}
```

Now, the alternative could be again that I'd define the value to spread on a separate line

```django
{% component "my_component" %}
	{% fill "content" data="content_props" %}

		{% expr '{ **content_props, **attrs }'  as merged_props %}
		{% component "other_comp" ...merged_props %}

	{% endfill %}
{% endcomponent %}
```

But I don't see harm in allowing both at the same time. They are complementary concepts. I also didn't document the combined usage (correct me if wrong), so it's more up to the users to discover that it works. And I now also [updated the error message](https://github.com/EmilStenstrom/django-components/pull/605/commits/4beccb63353d68390e067cf814adc8a106f5726d) for when the spread operator receives a non-dict, to make things clear.

So IMO there are guardrails so that this can be used by power users like me, without promoting bad coding habits on the rest.

EmilStenstrom (approved):
Looks great, and and understand the use-case you are proposing. Lets go with your suggestion!

JuroOravec:
Awesome, thanks!

Issue #401, opened 2024-03-18
Move documentation to GitHub Pages


EmilStenstrom:
The README is super-long, so I think it's time to move it out to a separate site, and give it some structure. 

I suggest GitHub Pages because it's built in to GitHub already.

JuroOravec:
Just a cheeky thought. I reckon some sort of documentation builder would be used for this. But if the website had a custom landing page, or something like that, it'd be cool to build it with Django and Django components. https://github.com/meeb/django-distill can be used to build static sites with Django.

GabDug:
As the README is already using Markdown, I would say [`mkdocs-material`](https://squidfunk.github.io/mkdocs-material/) may be a good fit. I've been working with mkdocs for a while now, so I can contribute to the setup around Mkdocs setup and its CI if help is wanted :)

EmilStenstrom:
@JuroOravec While I like the idea of using the library for it's own documentation, I also like the idea of publishing the docs on GitHub Pages and not have to worry about a separate service/server. 

@GabDug That looks really good. Would love to get that structure in place. Though I think a lot of the work will be to split the current readme into logical chunks with good section names.

EmilStenstrom:
@GabDug Did you have any chance to make any more progress on this? I think it would make sense to split this work up into several parts, så that just moving what we have could be one step. During the last weeks the README has changed a lot, so I'm not sure this PR will be easy to rebase. If we start with just a split we can keep iterating on the new version while other work in ongoing?

dalito:
Maybe add an additional link to the docs in the "About" section of the gh project page?

EmilStenstrom:
@dalito Yes, when we get this working (this is my first time fiddling with this) we'll make sure it's replacing the current mega-README :)

PR #606, opened 2024-08-24
Fix docs deployment to gh-pages.

dalito:
Related to https://github.com/EmilStenstrom/django-components/issues/401#issuecomment-2308379192: I think these changes are at least needed to fix docs deployment to gh-pages.

@EmilStenstrom did you already activate gh-pages in settings of the repo?

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -14,8 +14,11 @@ on:
 
 jobs:
   docs:
+    # Grant GITHUB_TOKEN the permissions required to make a Pages deployment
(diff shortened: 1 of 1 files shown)


EmilStenstrom:
Thanks @dalito, I've merged that and now it seems to work for the version-specific pages https://emilstenstrom.github.io/django-components/dev/ https://emilstenstrom.github.io/django-components/latest/

Seems mike has a command to set the default, and that that might redirect us to latest. Any idea how to trigger that? :)

PR #607, opened 2024-08-25
Increase minimum coverage to 97%

EmilStenstrom:
Juro has done a great job increasing the test coverage for new code. 

This pull request increases the minimum coverage requirement to 97%. The `coverage report` command has been updated to fail if the coverage falls below this threshold.


JuroOravec (approved):

JuroOravec:
Good idea!

PR #608, opened 2024-08-25
Fix redirect to latest for mike 2.0

dalito:
@EmilStenstrom - At the root of gh-pages there should be an index.html file with the redirect. It is missing here in this repo.

Maybe issue https://github.com/jimporter/mike/issues/188 for mike2.0 was the reason for the missing redirects. This PR applies the suggested change.


JuroOravec (approved):

EmilStenstrom:
Thanks again @dalito!

PR #609, opened 2024-08-25
feat: dynamic slots, fills, and provides

JuroOravec:
Allow to specify slot and fill names as variables. ...And also for the `provide` tag, since it works the same way.

It would be nice to also have dynamic components, bit I'm leaving that for later, as that one's more complicated because of the TagFormatter.

Closes https://github.com/EmilStenstrom/django-components/issues/587

### DIFF
--- a/README.md
+++ b/README.md
@@ -2106,7 +2106,7 @@ Similar is possible with [`django-expr`](https://pypi.org/project/django-expr/),
 / %}
 ```
 
-> Note: Never use this feature to mix business logic and template logic. Business logic should still be in the component!
+> Note: Never use this feature to mix business logic and template logic. Business logic should still be in the view!
 
 ### Pass dictonary by its key-value pairs
 
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
 - [Use components as views](#use-components-as-views)
 - [Autodiscovery](#autodiscovery)
 - [Using slots in templates](#using-slots-in-templates)
-- [Passing data to components](#passing-data-to-components)
+- [Accessing data passed to the component](#accessing-data-passed-to-the-component)
 - [Rendering HTML attributes](#rendering-html-attributes)
 - [Template tag syntax](#template-tag-syntax)
 - [Prop drilling and dependency injection (provide / inject)](#prop-drilling-and-dependency-injection-provide--inject)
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -812,7 +813,7 @@ class ComponentView(View):
 
 If you want to define your own `View` class, you need to:
 1. Set the class as `Component.View`
-2. Subclass from `ComponentView`, so the View instance has access to the component class.
+2. Subclass from `ComponentView`, so the View instance has access to the component instance.
```
JuroOravec, line 816: Typo

README.md
```diff
@@ -1440,18 +1441,77 @@ to the same name. This raises an error:
 {% endcomponent %}
 ```
 
-## Passing data to components
```
JuroOravec, line 1443: It's not clear from the diff, but I didn't remove this section, just renamed `Passing data to components` to `Accessing data passed to the component`, because that was the sole sub-section.

README.md
```diff
@@ -1862,7 +1922,23 @@ attributes_to_string(attrs)
 ## Template tag syntax
 
 All template tags in django_component, like `{% component %}` or `{% slot %}`, and so on,
-support extra syntax that makes it possible to write components like in Vue or React.
+support extra syntax that makes it possible to write components like in Vue or React (JSX).
+
+### Self-closing tags
```
JuroOravec, line 1927: I realized that it would be good to document the self-closing tags in this section too.

README.md
```diff
@@ -1960,6 +2036,7 @@ Instead, template tags in django_components (`{% component %}`, `{% slot %}`, `{
   "As positional arg {# yay #}"
   title="{{ person.first_name }} {{ person.last_name }}"
   id="{% random_int 10 20 %}"
+  readonly="{{ editable|not }}"
```
JuroOravec, line 2039: I realized I didn't include an example with a filter here

src/django_components/component.py
```diff
@@ -634,23 +633,7 @@ def render(self, context: Context) -> str:
                 ),
             }
         else:
-            fill_content = {}
```
JuroOravec, line 637: Moved this logic into it's own function in `slots.py`

src/django_components/provide.py
```diff
@@ -19,34 +21,43 @@ class ProvideNode(BaseNode):
 
     def __init__(
         self,
-        name: str,
         nodelist: NodeList,
+        trace_id: str,
```
JuroOravec, line 25: Previously, for the tracing logging, we used slot/fill names. But we can no longer rely on those, as they are now resolved only at render time.

So here and there, I replaced it with `trace_id`s, which include the ID, and the raw value of `slot`, if given.

src/django_components/templatetags/component_tags.py
```diff
@@ -674,43 +685,3 @@ def _fix_nested_tags(parser: Parser, block_token: Token) -> None:
 
             expects_text = True
             continue
-
-
-class ParsedSlotTag(NamedTuple):
-    name: str
-
-
-def _parse_slot_args(
```
JuroOravec, line 683: `_parse_slot_args` and `_parse_provide_args` were moved to `SlotNode.resolveKwargs` and `ProvideNode.resolveKwargs` respectively.

EmilStenstrom (approved):
Looks good to me! :)

README.md
```diff
@@ -2028,7 +2106,7 @@ Similar is possible with [`django-expr`](https://pypi.org/project/django-expr/),
 / %}
 ```
 
-> Note: Never use this feature to mix business logic and template logic. Business logic should still be in the template!
+> Note: Never use this feature to mix business logic and template logic. Business logic should still be in the component!
```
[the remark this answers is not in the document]
JuroOravec: Yup, view makes more sense, thanks! I just noticed that template didn't sound right 😄

Issue #590, opened 2024-08-21
Allow different apps to have different settings


JuroOravec:
For the UI component library, I came across an issue that inside the library I want to use certain settings:

```py
COMPONENTS = {
    "context_behavior": "isolated",
    "tag_formatter": "django_components.component_shorthand_formatter"	
}
```

But I want to allow users to use different settings.

Since the components within the library render each other, it would break if the end user changed the `tag_formatter` setting.

E.g. if, within the UI library I referenced a component in another using the "ShorthandComponentFormatter":

```django
{% button %}
{% endbutton %}
```

But user's settings would require to use different format:

```django
{% component "button" %}
{% endcomponent %}
```

---

Currently, there's 5 settings: `autodiscover`, `libraries`, `template_cache_size`, `context_behavior`, `tag_formatter`
- Of these, `autodiscover` and `libraries` kinda are "global" settings. They are not used in the code, just to import things at the startup
- `template_cache_size`, `context_behavior`, `tag_formatter`, however, are used in template tags, `ComponentRegistry`, `Component`.
- I'm thinking that, the 3 settings above are directly or indirectly associated with `ComponentRegistry` instance that holds the components. So maybe we could add a `settings` field which would hold the 3 settings. And the settings set in `settings.py` would just pass the values to the default subclassed `Library`.
- Getting the settings info to slots will have to be through the Context, similarly as we do it [here](https://github.com/EmilStenstrom/django-components/blob/03af25aad6841b6edb56c925d2599817e08ceb44/src/django_components/component.py#L432).

So to allow different settings, one would create their own instance of `ComponentRegistry` (optionally also their own `Library`)

```py
my_lib = django.template.Library()
my_reg = ComponentRegistry(
    library=my_lib,
	settings={
		"context_behavior": "isolated",
		"tag_formatter": "django_components.component_shorthand_formatter",
		"template_cache_size": 256,
	},
)
```

And then, to use the correct registry, one would register components like so:

```py
@component.register("adivider", my_reg)
class ADivider(component.Component):
	...
```

What this should do is that if user installs a 3rd party django app, then user should be able to configure their components independently from the 3rd party app:

```django
{% load third_party_app %}

{% component "mycomp" %}             <--- Uses default tag formatter
	{% adivider / %}                 <--- Uses "shorthand" tag formatter
{% endcomponent %}
```

EmilStenstrom:
[...]

PR #615, opened 2024-08-26
feat: allow different template settings for ComponentRegistries

JuroOravec:
Ngl, I'm pretty excited about this! 🦜 

This MR allows template settings (`CONTEXT_BEHAVIOR` and `TAG_FORMATTER`), to be configured per registry / library.

So, assuming that `simple_a` comes from Library A, and `simple_b` comes from Library B, now following is possible:

```django
{% simple_a variable=123 %}
    SLOT 123
{% endsimple_a %}
{% component "simple_b" variable=123 %}
    SLOT ABC
{% endcomponent %}
```

Also:
- Added section on `ComponentRegistry` settings
- Added section on writing, publishing and sharing component libraries

Closes https://github.com/EmilStenstrom/django-components/issues/590

TODO:
- [x] Docs

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -33,7 +33,8 @@ from django.views import View
 
 from django_components.app_settings import ContextBehavior
 from django_components.component_media import ComponentMediaInput, MediaMeta
-from django_components.component_registry import ComponentRegistry, registry as registry_
+from django_components.component_registry import ComponentRegistry
+from django_components.component_registry import registry as registry_
 from django_components.context import (
     _FILLED_SLOTS_CONTENT_CONTEXT_KEY,
     _PARENT_COMP_CONTEXT_KEY,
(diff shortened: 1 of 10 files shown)


src/django_components/component.py
```diff
@@ -535,8 +541,10 @@ def _render_impl(
 
             with context.update(
                 {
+                    # Private context fields
                     _ROOT_CTX_CONTEXT_KEY: self.outer_context,
                     _FILLED_SLOTS_CONTENT_CONTEXT_KEY: updated_slots,
+                    _REGISTRY_CONTEXT_KEY: self.registry,
```
JuroOravec, line 547: Most of the changes were about wiring the registry instance to the component and slot, so we use the settings as defined on the registry, eg. `registry.settings.XXX`, instead of `app_settings.XXX`

src/django_components/component_registry.py
```diff
@@ -66,10 +77,16 @@ class ComponentRegistry:
     ```
     """
 
-    def __init__(self, library: Optional[Library] = None) -> None:
+    def __init__(
+        self,
+        library: Optional[Library] = None,
+        settings: Optional[Union[RegistrySettings, Callable[["ComponentRegistry"], RegistrySettings]]] = None,
```
JuroOravec, line 83: I allowed to define the settings also as a getter, so the latest value can be accessed.

JuroOravec:
@EmilStenstrom Another one for review 🚀 :)

EmilStenstrom (approved):
Nicely done, looks good to me!

Issue #612, opened 2024-08-25
Allow dynamic components


JuroOravec:
Similarly to https://github.com/EmilStenstrom/django-components/issues/587,  I was thinking of making it possible to pass variables as component names, to allow for dynamic components.

But it turns out that we would need to take a different approach.

This is because the TagFormatter feature, which works at template compile time. And the TagFormatter MUST work at compilation time, because it generates the "end tag" (e.g. `endcomponent` or `endbutton`), which is then used to inform Django how much of the remaining content to parse.

In theory we could have "start tag" generated at render-time, and "end tag" at compile-time, but that feels weird.

So instead I'm thinking of a standalone template tag. Similar to [django-cottons `<c-component :is="..." >`](https://github.com/wrabit/django-cotton#dynamic-components), or [Vue's `<component :is="..." />`](https://vuejs.org/guide/essentials/component-basics.html#dynamic-components) (tbh, django-cotton is inspired by Vue too). 

Since the `{% component %}` template tag is already taken by our regular components, it would have to be something else. `dynamic_component` is too long, but `xcomponent` could work.

How it works is that one would use this tag the same way as they would use a concrete component. So they would pass on all the args and kwargs they want, and optionally `only` flag or `/` for an inline block.

But they would pass also one more parameter, which specifies the component that should be invoked. At first I used `name`, but tbh I like `is` more, as Vue does it, simply because it's less likely that people will use a parameter called `is`.

So imagine a scenario where an icon can be rendered in multiple different ways, depending on the component input. And based on the input, `icon_component` is assigned a name of one of the components.

```django
{% xcomponent is=icon_component size="24" color="black" %}
{% endxcomponent %}
```

IMO `is` parameter should accept both string (registered component name) and Component class (pass directly by reference).

EmilStenstrom:
I'm not fond of leaving the simplicity of "django-component just adds one new template tag, component!" (yeye, slot/fill, but that doesn't count if we're doing marketing :)).

I think you could implement this inside a normal component if you wanted? So `{% component "dynamic" is=icon_component %}` would call the dynamic component (that the user would write themselves) and delegate to icon_component.

JuroOravec:
Hm, so the issue with the tag formatters. 

Your example works only for the default formatting. But what if someone used the "shorthand" one?

```django
{% button %}
{% endbutton %}
```

One way this could work is that, for the "shorthand" formatting, it would work like this:

```django
{% dynamic is=icon_component %}
[...]

PR #627, opened 2024-08-28
feat: Add dynamic component

JuroOravec:
Closes https://github.com/EmilStenstrom/django-components/issues/612

### DIFF
--- a/README.md
+++ b/README.md
@@ -1559,7 +1559,7 @@ This means that you can use `self.input` inside:
 - `get_template_name`
 - `get_template_string`
 
-`self.input` is defined only for the duration of `Component.render`, and returns `None` when called outside of this.
+`self.input` is defined only for the duration of `Component.render`, and raises `RuntimeError` when called outside of this.
 
 `self.input` has the same fields as the input to `Component.render`:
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -850,6 +855,54 @@ class MyComponent(Component):
...
+```
+
+These behave same way as regular components. You pass it the same args, kwargs, and slots as you would
+to the component that you want to render.
+
+The only exception is that also you supply 1-2 additional inputs:
+- `is` - Required - The component name or a component class to render
+- `registry` - Optional - The `ComponentRegistry` that will be searched if `is` is a component name. If omitted, ALL registries are searched.
+
+By default, the dynamic component is registered under the name `"dynamic"`. In case of a conflict, you can change the name used for the dynamic components by defining the [`COMPONENTS.dynamic_component_name` setting](#dynamic_component_name).
```
EmilStenstrom, line 888: Should we add a test for what happens on conflicts?

EmilStenstrom (approved):
Looks good to me, nice addition!

README.md
JuroOravec: Good idea! Added in https://github.com/EmilStenstrom/django-components/pull/627/commits/2bd105b03837efc4332704cf649b6b332d1a0e28

Issue #622, opened 2024-08-26
Use variables in JS / CSS and caching thereof


JuroOravec:
### Background
I feel I've addressed most of the items that were blocking me for the the UI component library (there's about 5-6 MRs I've got half-ready in the pipeline), and kinda the bigger pieces that are still remaining for v1 is the rendering of CSS / JS dependencies and then documentation. So I've started looking into the CSS / JS rendering.

IMO before I get back to some of the ideas we discussed before, regarding the dependency middleware and such (https://github.com/EmilStenstrom/django-components/issues/478), it first make sense to add desired CSS- / JS-related features to django_components , and only then think about how to send the CSS / JS over to the client.

So one thing I'd like to have is to use the variables from `get_context_data()` in CSS / JS files. Let's break it down.

### How static files are loaded

When it comes to static files, so those that are served by the server AS IS, those are managed by Django's `staticfiles` app. My experience was that the files are "collected" / generated using `collectstatic`, which physically puts them in a certain directory. And then, when the server starts, it serves the files from that directory.

Now, if we want to treat CSS / JS as Django templates, we cannot pre-process them with `collectstatic`. Instead, what would be generated would depend on the components inputs. Same as it is with the HTML template.

### How CSS / JS is inserted into the HTML

Both CSS and JS can be "injected" into an HTML file it two ways:
- Inlined, meaning that I embed the full definition within the HTML, e.g.
	```html
	<div>
		<!-- CSS -->
		<style>
			.my-class {
				background: red;
			}
		</style>

		<!-- JS -->
		<script>
			console.log("hello world!")
		</script>
	</div>
	```
- Linked , meaning that I only declare the URL where the CSS / JS can be found, and the browser fetches it
	```html
	<div>
		<!-- CSS -->
		<link media="all" rel="stylesheet" href="https://example.com/assets/light-efd2f2257c96.css">

		<!-- JS -->
		<script defer="defer" type="application/javascript" src="https://example.com/assets/wp-runtime-cc0837173e45.js"></script>
	</div>
	```

Static CSS / JS is linked. This is done by the `ComponentDependencyMiddleware`, which works like this:
1. When we render a component, we prepend the formatter HTML string with `<!-- _RENDERED {name} -->`
2. When the HTML response is being sent, the middleware looks at the response HTML, looks up all `<!-- _RENDERED {name} -->` comments within the HTML, and retrieves corresponding component classes from the registry.
	- TODO - How would this be handled if there was multiple registries?
[...]

PR #629, opened 2024-08-29
feat: validate component inputs if types are given

JuroOravec:
As I mention in https://github.com/EmilStenstrom/django-components/issues/622#issuecomment-2316849176, this MR adds validation of args, kwargs, slots, and data (returned from `get_context_data`) based on the types passed to the `Component`'s generics.

The reason I wanted to add this is following:

When I was adding the typing support to the components, I had two options:
1. Either use [`ParamSpec`](https://docs.python.org/3/library/typing.html#typing.ParamSpec) as one of the Component's generics, which would be bound to the `get_context_data`. This way, the `get_context_data` function would have been the source of truth for the type.
2. Or define args and kwargs as I eventually did, as a `Tuple` and `TypedDict`. This way, the `Component.render` would still be correctly typed, but the signature of `get_context_data` could diverge.

I went with option 2, because even I haven't worked with `ParamSpec` before, and I didn't want to ask users to pass around something which they might have never used before. It seemed complex. On the other hand, defining just tuples and typed dicts seems a lot more intuitive.

But that means that one could add typing for the Component, and the args and kwargs declared in the types could be totally different from the actual implementation of `get_context_data`.

So this MR adds some guard rails so that the component's implementation must be compatible with the declared args and kwargs types.

Altho, this is applicable only for Python 3.11 and later. In older versions, args are still validated, but TypedDicts like kwargs, slots, and data, are typed only, but not validated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
 - [Use components outside of templates](#use-components-outside-of-templates)
 - [Use components as views](#use-components-as-views)
 - [Pre-defined components](#pre-defined-components)
+- [Typing and validating components](#typing-and-validating-components)
+- [Pre-defined components](#pre-defined-components)
 - [Registering components](#registering-components)
 - [Autodiscovery](#autodiscovery)
 - [Using slots in templates](#using-slots-in-templates)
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -632,58 +634,6 @@ MyComponent.render_to_response(
 )
 ```
 
-### Adding type hints with Generics
```
JuroOravec, line 635: I moved this subsection to its own section on typing and validation.

EmilStenstrom (commented):
This is way above my head, so can't review. I trust it if the tests pass, so go ahead! :)

JuroOravec:
@EmilStenstrom Appreciate the honesty. I can be easily too deep in things, so I've added more comments to the code, hopefully making it clearer!

PR #632, opened 2024-08-31
feat: add `self` context var and make `is_filled` into attribute

JuroOravec:
Here is an MR to access the component instance from within the template, as mentioned in https://github.com/EmilStenstrom/django-components/discussions/540#discussioncomment-10496206

Changes:
- Access the component instance from within the template using `{{ self }}` ([inspired by django-viewcomponnets](https://django-viewcomponent.readthedocs.io/en/latest/context.html#self))

- To work with the previously-defined `{{ component_vars.is_filled }}`, I made `is_filled` into a property on Component instance, and made `component_vars` an alias for `self`. So `component_vars.is_filled` is the same as `self.is_filled`.

- Added a deprecation note that `component_vars` will be removed in v1. I think this makes sense as I don't see a need to add more pre-defined context variables in the near future. And if so, we can expose it via `self` like with `self.is_filled`.


README.md
```diff
@@ -45,7 +45,6 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
 - [Use components in templates](#use-components-in-templates)
 - [Use components outside of templates](#use-components-outside-of-templates)
 - [Use components as views](#use-components-as-views)
-- [Pre-defined components](#pre-defined-components)
```
JuroOravec, line 48: Same is defined on line 49.

tests/test_context.py
```diff
@@ -581,30 +598,16 @@ class ComponentWithComplexConditionalSlots(Component):
...
-                {% elif component_vars.is_filled.alt_subtitle %}
+                {% elif self.is_filled.alt_subtitle %}
                     <div class="subtitle">{% slot "alt_subtitle" %}Why would you want this?{% endslot %}</div>
                 {% else %}
                 <div class="warning">Nothing filled!</div>
                 {% endif %}
             </div>
         """
 
-    class ComponentWithNegatedConditionalSlot(Component):
```
JuroOravec, line 594: Inlined this component here https://github.com/EmilStenstrom/django-components/pull/632/files#diff-0739130667ad94af9f1ac9b3ab331c42b6e4c5b0fe7bc42a3dbb47f900b7cfcaR741

EmilStenstrom:
I'm not sure I like this API, passing around self and then accessing properties of this object. It's not something you would expect to do in regular Django, so feels very strange to me. 

In the discussion you mention this code:
```python
    def setup_context(self, *args, **kwargs):
        name = kwargs.get("name", None)
        return {
            "tabs": [],
            "name": name,
            "attrs": kwargs.get("attrs", None),
            "header_attrs": kwargs.get("header_attrs", None),
            "content_attrs": kwargs.get("content_attrs", None),
            "tabs_data": {"name": name},
        }
```

but wouldn't the repetition be easier solved with:
```python
    def setup_context(self, *args, **kwargs):
        name = kwargs.get("name", None)
        return {
            **kwargs,
            "tabs": [],
            "tabs_data"={"name": name},
        ]
```

JuroOravec:
> but wouldn't the repetition be easier solved with:
> 
> ```python
>     def setup_context(self, *args, **kwargs):
>         name = kwargs.get("name", None)
>         return {
>             **kwargs,
>             "tabs": [],
>             "tabs_data"={"name": name},
>         ]
> ```

It's an edge case, but in Python 3.8-3.10 this would allow any kwargs to be passed to the component and made accessible from the template. But it's true this could be solved more elegantly, e.g.

```py
    def setup_context(self, *args, **kwargs):
        name = kwargs.get("name", None)
        return {
            **pick(kwargs, "attrs", "header_attrs", "content_attrs"),
            "tabs": [],
            "tabs_data"={"name": name},
        ]
```

---

> I'm not sure I like this API, passing around self and then accessing properties of this object. It's not something you would expect to do in regular Django, so feels very strange to me.

Fair enough, I feel like this is probably skipping over one step in API evolution, and might be relevant later. For my needs for the component library, there's a lot of accessing of slots and props, so it would shave some boilerplate. The alternative is  this:

```py
from typing import Any, TypedDict

from django_components import Component, register

class TodoData(TypedDict):
    one: str
    self: "TodoComp"  # <-- Define self on Data dict

TodoComp = Component[Any, TodoKwargs, Any, TodoData]  # <-- Define typed Comp class

@register("todo")
class Todo(TodoComp):  # <-- Use the typed Comp class
    def get_context_data(self, *args: Any, **kwargs: Any) -> TodoData:
        return {
            "one": "abc",
            "self": self,  # <-- Expose `self`
        }
```

EmilStenstrom (approved):
The amount messing around with types... I'm almost starting to think it's not worth the code complexity... But you are the component library author here, so I trust you! :)

JuroOravec:
@EmilStenstrom Yeah, it was a pain tbh. But I'm lowkey hoping that having working typing support could one day lead to intellisense / linter for the templates 🤞 😄

PR #633, opened 2024-09-01
Fix docs action

### LINKED ISSUE #625: Release 0.93 did not trigger documentation build
The docs for the newest release 0.93 were not created. The job was skipped, see https://github.com/EmilStenstrom/django-components/actions/workflows/docs.yml

It could be a missing trigger on new tags. Something like:
```
on:
  push:
    tags:
      - '[0-9].[0-9]+'
    ....
```

I also wonder if mike expects versions with a leading "v", so a tag "v0.93" which is more common for version tags.

dalito:
Changes:

- Fix all triggers and logic (tested in my fork to work for commits, new tags, new releases)
- removed input option from manual dispatch. It us not useful to build old docs for old versions because it uses the (outdated) gh-action script present at that verion.

Other changes:
- Prevent pypi-publish action from running if forks
- Updated requirement for docs (dependencies were incompatible before)
- Activated showing of "latest" alias besides version tag (new feature in mkdocs only added recently), see screenshot

Add some point 0.92 should probably be deleted because it was build with an old config and older mike/mkdocs version.

Let me know if I should clean-up the git commit history.

image

----
Closes #625


.github/workflows/docs.yml
```diff
@@ -39,34 +42,51 @@ jobs:
...
           export SHORT_SHA=$(echo "${GITHUB_SHA}" | cut -c1-7)
-          hatch run docs:mike deploy dev --update-aliases --title "dev (${SHORT_SHA})" --push --alias-type=redirect
+          hatch run docs:mike deploy --push --update-aliases --title "dev (${SHORT_SHA})" dev
 
-      - name: Deploy docs (tag)
+      - name: Build & deploy docs for a new tag
         if: github.ref_type == 'tag' && github.event_name == 'push'
         run: |
-          hatch run docs:mike deploy ${{ github.ref_name }} latest --push --update-aliases --alias-type=redirect
+          # XXX next line removes docs wrongly deployed under latest identifier; remove once latest is used only as alias
```
JuroOravec, line 81: What is meant by the `"docs wrongly deployed under latest identifier; remove once latest is used only as alias"`? And how do we know when the `latest` identifier is removed?

dalito: mike deploys versions under an identifier and optionally also gives an alias to the identifier. Identifiers cannot be updated but have to be deleted and then recreated. Aliases on the other hand can be updated which means a new alias can be given to an existing identifier. For whatever reason, in current gh-pages "latest" was created as an identifier (there is a directory latest) which is not what we want; instead the alias latest should be attached to a version identifier.

.github/workflows/docs.yml
```diff
@@ -2,15 +2,16 @@
...
   push:
-    branches: [master]
-  workflow_dispatch:
-    inputs:
-      ref:
-        description: "The commit SHA, tag, or branch to publish. Uses the default branch if not specified."
-        default: ""
-        type: string
+    tags:
+      - '[0-9]+.[0-9]+'
```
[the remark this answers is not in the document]
dalito: Added!

dalito:
I hope that the step-wise dependency updates by dependabot that were auto-merged today have not created an inconsistent set of packages again. Maybe the auto-updates were also the reason for the inconsistent dependencies in the docs env.

EmilStenstrom (approved):
Looks good to me!

EmilStenstrom:
> I hope that the step-wise dependency updates by dependabot that were auto-merged today have not created an inconsistent set of packages again. Maybe the auto-updates were also the reason for the inconsistent dependencies in the docs env.

Yes, they were. How can we work around that?

dalito:
I just resolved the conflict in GitHub UI for now.

So far I have never used dependabot for Python package updates but just for github-actions.

Issue #634, opened 2024-09-02
RequestContext not populated with context processors on async request


larrybotha:
Component's rendered via an HTTP request that use `RequestContext` are not populated with values defined in `context_processors`

---

As per [Django's `RequestContext` docs](https://docs.djangoproject.com/en/5.1/ref/templates/api/#django.template.RequestContext), `RequestContext` differs from `Context` in that it makes the configured `context_processors` available when rendering templates

The context processors are added to context when `RequestContext.bind_template` is called, and things like `csrf_token` are then made available in templates

Since at least `django-components@0.95` (potentially since the introduction of the template monkey-patching) the context processors are not added to the context because [_monkeypatch_template._template_render](https://github.com/EmilStenstrom/django-components/blob/0cfc40231bed770cbe00256b2da0b9a9d2f37828/src/django_components/component.py#L838) doesn't call [`context.bind_template` for `RequestContext`](https://github.com/django/django/blob/64443f555f362ddf0c9094f8531c3da72a753827/django/template/context.py#L247) when the component is rendered via a network request. Because the component is not rendered as the child of a larger context, as when rendering a full page, there is no outer context available to the component, as when rendered during a 'normal' full-page request

## Demo

- https://replit.com/@larrybotha1/Django-Components-088#django_project/components/thing/index.html

## Steps to reproduce

- ensure `'django.contrib.auth.context_processors.auth'` is included in context processors in settings
- create a component that is rendered via a network request
- pass `context=RequestContext(request)` to `render_to_response`
- make a request for the component, rendering `{{ csrf_token }}`

## Expected

- `{{ csrf_token }}` is rendered when the component is rendered via a network request

## Actual

- `{{ csrf_token }}` is not available

## Possible solutions

I haven't yet worked out how this could be solved. The difference here is in what `Context.bind_template` and `RequestContext.bind_template` do. `Context` does little except yield the template, whereas `RequestContext` augments the context.

`django-components` does the correct thing in that it evaluates whether `template is None` before calling `bind_template`, so the problem may stem from [where `Component.template` is assigned](https://github.com/EmilStenstrom/django-components/blob/0cfc40231bed770cbe00256b2da0b9a9d2f37828/src/django_components/component.py#L524) - should `template` never be assigned directly, and only via `bind_template`?

## Workarounds

- instantiate `RequestContext` with the context processors manually

EDIT: improved clarity of description of problem

JuroOravec:
[...]

PR #643, opened 2024-09-03
fix: populate RequestContext with context processors

JuroOravec:
Fix for https://github.com/EmilStenstrom/django-components/issues/634


src/django_components/component.py
```diff
@@ -520,22 +522,7 @@ def _render_impl(
         context_data = self.get_context_data(*args, **kwargs)
         self._validate_outputs(context_data)
 
-        with context.update(context_data):
```
JuroOravec, line 523: All this patching is now hidden away in `_prepare_template`

src/django_components/component.py
```diff
@@ -843,3 +830,35 @@ def _template_render(self: Template, context: Context, *args: Any, **kwargs: Any
 
     # See https://stackoverflow.com/a/42154067/9788634
     template.render = types.MethodType(_template_render, template)
+
+
+@contextmanager
+def _maybe_bind_template(context: Context, template: Template) -> Generator[None, Any, None]:
```
JuroOravec, line 836: The body of `_maybe_bind_template` is actually what fixes https://github.com/EmilStenstrom/django-components/issues/634.

EmilStenstrom (approved):
Looks good to me! Impressed that you found this at all! :)

Issue #589, opened 2024-08-21
Template caching isn't working


JuroOravec:
Turns out that the `TEMPLATE_CACHE_SIZE` setting isn't being used ever since https://github.com/EmilStenstrom/django-components/commit/898d1483820778a05d7db9730b00d1702d241e6b (March 2023). We should put it back and test it.

JuroOravec:
Ok, continuing from https://github.com/EmilStenstrom/django-components/issues/613, there's 3 things on my mind:
1. What should be cached? I assume that only the inlined template (from `Component.template` or `Component.get_template_string`) needs to be cached, because templates loaded by the Django loader should be cached by Django ([see stackoverflow](https://stackoverflow.com/a/69677619/9788634))

2. To allow use cases like #613, I'd be great if the caching wasn't bound to a method, like it was [before](https://github.com/EmilStenstrom/django-components/commit/898d1483820778a05d7db9730b00d1702d241e6b#diff-569e56c0c056d2ede4e534ad3cae3717ff8e61ba7fdd6963a23df1adc8fd86d1L108), but could be used also independently, e.g.:
	```py
	from django_components import cached_template

	template = cached_template(
	    """
			<div>
				...
			</div>
		"""
	)
	```

	And we'd then call `cached_template` from within `Component.get_template()`.

3. While I'm at it, I also want to update the template-related Component API:

	Currently there's 5 attributes:
		- `template` - inlined string
		- `template_name` - template name string
		- `get_template_name` - function for template name string
		- `get_template_string` - function for inlined template string
		- `get_template` - Processes the above and returns an instance of `Template`

	To clean up the interface, I suggest to:
		- Merge `template` and `get_template_string` into `template` - And allow it to be string, a Template, or a function that returns one of.
		- Merge `template_name` and `get_template_name` into `template_name` - And allow it to be string, or a function that returns a string.
		- Make `get_template` private by renaming it to `_get_template` - After all, it isn't mentioned anywhere.
		- For compatibility `get_template_string` and `get_template_name` would still work until v1

PR #647, opened 2024-09-05
refactor: Fix template caching, expose `cached_template`, Component.template API changes

JuroOravec:
Changes:
- Reintroduce template caching using the `TEMPLATE_CACHE_SIZE` setting (fixes https://github.com/EmilStenstrom/django-components/issues/589)

- Also caches the process of extracting `{% fill %}` tags from `{% component %}` tags

- Add a way for users to manually add Templates to the cache using `cached_template`

Also relevant API cleanup / changes:
- The undocumented `Component.get_template` is made private as `_get_template`

- Introduced new `get_template`, which works like `get_template_string`, except it can return either a string or a Template instance. And I tagged `get_template_string` to be deprecated and will work till v1.

- Also updated the logic so that all 4 ways of defining the template are mutually exclusive, as previously it was ambiguous which one of e.g. `template_name` vs `get_template_name` should take priority.

Closes https://github.com/EmilStenstrom/django-components/issues/589

Note: Initially I tried to merge `template_name` and `get_template_name` into a single attribute, and same with `template` and `get_template`, to reduce the number of variables. And it mostly worked, but Mypy had issues with that. So I reverted back to all 4 ways.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,11 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
 
 ## Release notes
 
+**Version 0.97**
+- Fixed template caching. You can now also manually create cached templates with [`cached_template()`](#template_cache_size---tune-the-template-cache)
+- `get_template_name` and `get_template_string` were merged into `template_name` and `template`, and old formats will be removed in v1.
+- The undocumented `get_template` was made private.
+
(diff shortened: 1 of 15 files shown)


EmilStenstrom (approved):
Nice changes as always!

Issue #610, opened 2024-08-25
  Invalid block tag ...  'component'. Did you forget to register or load this tag? error


danjac:
1. Added "django_components" to `INSTALLED_APPS`
2. Set up `TEMPLATES` as follows:

```python
TEMPLATES = [
    {
        "BACKEND": "django.template.backends.django.DjangoTemplates",
        "DIRS": [BASE_DIR / "templates"],
        "OPTIONS": {
            "builtins": [
                "django_components.templatetags.component_tags",
               # my builtins...
            ],
            "debug": env.bool("TEMPLATE_DEBUG", default=False),
            "context_processors": [
                "django.template.context_processors.debug",
                "django.template.context_processors.request",
                "django.contrib.auth.context_processors.auth",
                "django.template.context_processors.i18n",
                "django.template.context_processors.media",
                "django.template.context_processors.static",
                "django.template.context_processors.tz",
                "django.contrib.messages.context_processors.messages",
            ],
            "loaders": [
                (
                    "django.template.loaders.cached.Loader",
                    [
                        "django.template.loaders.filesystem.Loader",
                        "django.template.loaders.app_directories.Loader",
                        "django_components.template_loader.Loader",
                    ],
                )
            ],
        },
    }
]
```

3. Created a simple component and added to template:

```html
{% component "card" %}
....
{% endcomponent %}
```

Got the error "Invalid block tag on line 2: 'component'. Did you forget to register or load this tag?", indicating the template tag was not being loaded correctly

4. Tried again with `{% load component_tags %}`, same error

Django 5.1,  `django-components==0.92`

danjac:
I'm trying if at all possible to avoid using `STATICFILES_DIRS` - I have no need to include CSS and JS as I'm using Tailwind and Alpine and IMHO it's a "code smell" to include templates and Python files among static assets. 

Is there a reason I can't just define them in the individual Python libraries e.g. under `components` ?

JuroOravec:
Please go over the documentation. See the [section on Autodiscovery](https://github.com/EmilStenstrom/django-components#autodiscovery). We use `STATICFILES_DIRS` to know which python files to search for.

It's possible not to use `STATICFILES_DIRS`, but then you have to import python files with the components yourself.

Like so in `apps.py`:

```py
from django.apps import AppConfig

class MyAppConfig(AppConfig):
    name = "my_app"

    def ready(self) -> None:
        from components.card.card import Card
        from components.list.list import List
[...]

PR #652, opened 2024-09-06
refactor: Remove safer_staticfiles, replace STATICFILES_DIRS with COMPONENTS.dirs, support `[app]/components`

JuroOravec:
As the title says, this one removes `safer_staticfiles`, and instead uses the staticfile's Finders to find additional static files. So people can have their usual static media configured by `STATICFILES_DIRS`, and components configured via `COMPONENTS.dirs`, independently.

`STATICFILES_DIRS` will still work with django_components until v1 though.

Closes https://github.com/EmilStenstrom/django-components/issues/610


README.md
```diff
@@ -71,6 +71,13 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
 
 ## Release notes
 
+🚨📢 **Version 0.100**
```
JuroOravec, line 74: Skipped to a round version number as this is a more fundamental change

README.md
```diff
@@ -191,14 +198,20 @@ _You are advised to read this section before using django-components in producti
```
JuroOravec, line 198: In v1, this section on static files could be removed, but until then it makes sense to keep it

README.md
```diff
@@ -230,11 +243,11 @@ For a step-by-step guide on deploying production server with static files,
 
 ## Installation
 
-1. Install the app into your environment:
+1. Install `django_components` into your environment:
```
JuroOravec, line 246: Changed the wording here and there, so the purpose of each step is clearer. Also, what was previously step 5 (`STATICFILES_DIRS`) is now split into steps 4 and 6 (`COMPONENTS.dirs` and `STATICFILES_FINDERS`)

sampleproject/sampleproject/settings.py
JuroOravec, line 1: This file shows nicely the changes required after this MR

src/django_components/finders.py
```diff
@@ -0,0 +1,144 @@
...
+from django.contrib.staticfiles.utils import get_files
+from django.core.checks import CheckMessage, Error, Warning
+from django.core.files.storage import FileSystemStorage
+from django.utils._os import safe_join
+
+from django_components.app_settings import app_settings
+from django_components.template_loader import get_dirs
+
+# To keep track on which directories the finder has searched the static files.
+searched_locations = []
```
JuroOravec, line 16: This global variable, and most of this file overall, was based on / taken from FileSystemFinder

src/django_components/finders.py
```diff
@@ -0,0 +1,144 @@
...
+    Differences:
+    - This finder uses ``COMPONENTS.dirs`` setting to locate files instead of ``STATICFILES_DIRS``.
+    - This finder ignores Python and HTML files (``.py``, ``.pyc``, ``.html``, ``.django``)
+    - If ``COMPONENTS.dirs`` is not set or empty, defaults to ``settings.BASE_DIR / "components"``
+    """
+
+    IGNORE_PATTERNS = StaticFilesConfig.ignore_patterns + ["*.html", "*.django", "*.py", "*.pyc"]
+
+    def __init__(self, app_names: Any = None, *args: Any, **kwargs: Any) -> None:
+        component_dirs = [str(p) for p in get_dirs()]
```
JuroOravec, line 38: It's still the Loader class where we define how we search the filesystem for component files. Here, inside the Finder, we use `get_dirs()`, which calls / uses the loader.

src/django_components/template_loader.py
```diff
@@ -3,59 +3,103 @@
...
+
+        component_dirs = component_dirs or [settings.BASE_DIR / "components"]
 
         logger.debug(
             "Template loader will search for valid template dirs from following options:\n"
             + "\n".join([f" - {str(d)}" for d in component_dirs])
         )
 
-        directories: Set[Path] = set()
+        # Add `[app]/components` to the directories
```
JuroOravec, line 60: This re-adds support for app-level components dir

tests/test_finders.py
```diff
@@ -60,44 +60,57 @@ def do_collect():
...
+    "INSTALLED_APPS": ("django_components", "django.contrib.staticfiles"),
 }
 
 
-# Check that .py and .html files are INCLUDED with the original staticfiles app
+# Check that the component files are NOT loaded when our finder is NOT added
 @override_settings(
     **common_settings,
-    INSTALLED_APPS=("django_components", "django.contrib.staticfiles"),
+    STATICFILES_FINDERS=[
```
JuroOravec, line 76: I reused the tests for `safer_staticfiles` to check that we indeed load only JS and CSS

tests/test_template_loader.py
```diff
@@ -24,7 +24,10 @@ def test_get_dirs__base_dir(self):
             sorted(dirs),
             sorted(
                 [
+                    # Top-level /components dir
                     Path(__file__).parent.resolve() / "components",
+                    # App-level /components dir
+                    Path(__file__).parent.resolve() / "test_app" / "components",
```
JuroOravec, line 30: This is where we test the app-level components dirs. For this to work, I added a Django app to `/tests/test_app` dir

README.md
```diff
@@ -71,6 +71,14 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
 
 ## Release notes
 
+🚨📢 **Version 0.100**
+- BREAKING CHANGE:
+    - `django_components.safer_staticfiles` app was removed. It is no longer needed.
+    - Installation change: Instead of defining component directories in `STATICFILES_DIRS`, set them to `COMPONENTS.dirs`.
+    - Installation change: You now must define `STATICFILES_FINDERS`
+    - [See here how to migrate your settings.py](https://github.com/EmilStenstrom/django-components/pull/652/files#diff-59eb658f2c5573a64306ac0adc96405344e029fd2df51573555c82e44b4d7062)
```
EmilStenstrom, line 79: I think this is too hard to follow. Don't think everyone is comfortable following git diffs around I'm afraid. Maybe we need a new page to link to?

README.md
```diff
@@ -243,13 +257,32 @@ For a step-by-step guide on deploying production server with static files,
...
+            ...,
+            os.path.join(BASE_DIR, "components"),
+        ],
+   }
+   ```
+
+   If `COMPONENTS.dirs` is omitted or empty, django-components will by default look for a top-level `/components` directory,
+   `{BASE_DIR}/components`.
+
+   Irrespective of `COMPONENTS.dirs`, django_components will also load components from app-level `/components` directories, e.g. `my-app/components/`.
```
EmilStenstrom, line 280: Hm. So I can set the directory for the global componets, but not for the app level ones? And searching for them can't be disabled?

README.md
```diff
@@ -243,13 +257,32 @@ For a step-by-step guide on deploying production server with static files,
...
+
+   NOTE: The input to `COMPONENTS.dirs` are the same as for `STATICFILES_DIRS`, and the paths must be full paths. [See Django docs](https://docs.djangoproject.com/en/5.0/ref/settings/#staticfiles-dirs).
+
+
+5. Next, to make Django load component HTML files as Django templates, modify `TEMPLATES` section of settings.py as follows:
```
EmilStenstrom, line 285: Since the defaults are so good, I think we can remove this from the basic setup. Everyone should be good with the defaults.

README.md
```diff
@@ -278,19 +311,18 @@ For a step-by-step guide on deploying production server with static files,
...
-   ```
+6. Lastly, be able to serve the component JS and CSS files as static files, modify `STATICFILES_FINDERS` section of settings.py as follows:
 
-   If `STATICFILES_DIRS` is omitted or empty, django-components will by default look for
-   `{BASE_DIR}/components`
+```py
+STATICFILES_FINDERS = [
+	# Default finders
+    "django.contrib.staticfiles.finders.FileSystemFinder",
+    "django.contrib.staticfiles.finders.AppDirectoriesFinder",
```
EmilStenstrom, line 320: Indentation.

src/django_components/finders.py
```diff
@@ -0,0 +1,144 @@
...
+# This is what makes it possible to define JS and CSS files in the directories as
+# defined by `COMPONENTS.dirs`, but still use the JS / CSS files with `static()` or
+# `collectstatic` command.
+class ComponentsFileSystemFinder(BaseFinder):
+    """
+    A static files finder based on ``FileSystemFinder``.
+
+    Differences:
+    - This finder uses ``COMPONENTS.dirs`` setting to locate files instead of ``STATICFILES_DIRS``.
+    - This finder ignores Python and HTML files (``.py``, ``.pyc``, ``.html``, ``.django``)
```
EmilStenstrom, line 31: I've never seen the .django ending, is that is use somewhere?

src/django_components/finders.py
```diff
@@ -0,0 +1,144 @@
...
+    - If ``COMPONENTS.dirs`` is not set or empty, defaults to ``settings.BASE_DIR / "components"``
+    """
+
+    IGNORE_PATTERNS = StaticFilesConfig.ignore_patterns + ["*.html", "*.django", "*.py", "*.pyc"]
```
EmilStenstrom, line 35: I'm a little worried about the security implications of this. If someone accidentally would put some other file type in these dirs, they could get exposed on the internet.

At the same time, explicitly including file types is hard, because the full list of potential static files is very large...

Any ideas of how to make this both safe and future-proof?

EmilStenstrom (commented):
Exciting! :D

Some docs questions and a bigger point about how to make the searching for files both safe for avoiding mistakes and future-proof.

src/django_components/finders.py
EmilStenstrom: Maybe a combo of both?

```python
ALLOWED_EXTENSIONS = getattr(settings, 'COMPONENTS_ALLOWED_EXTENSIONS', 
                             ['.css', '.js', '.png', '.jpg', '.jpeg', '.gif', '.svg', '.woff', '.woff2', '.ttf', '.eot'])
FORBIDDEN_EXTENSIONS = ['.py', '.pyc', '.html', '.django', '.env', '.db', '.sqlite3']
```

README.md
JuroOravec: We can't, we still need to insert `django_components.template_loader.Loader` so we can let Django handle the loading and caching of the HTML templates.

src/django_components/finders.py
JuroOravec: I didn't bother to change this, but agree with you, explicitly listing allowed extensions is better than the opposite. The reason why I used only ignored patterns was because then we can delegate the filtering to [`django.contrib.staticfiles.utils.get_files`](https://github.com/django/django/blob/main/django/contrib/staticfiles/utils.py#L16), which acepts a list of glob patterns.

But globbing doesn't support negations, so one cannot search for "everything except `.html`". So we can't use `get_files` to get only files with allowed patterns. And so we need to do the filtering ourselves.

And TBH, if we're doing the filtering ourselves, I would go with regex over globs, so users can be more expressive, e.g. maybe they would want to exclude CSS only from certain components, or omit static files that start with underscore.

But I just tried it, and writing regex for patterns like "everything except `.html`, `.py`, or `.pyc`" is not very easy to read. So I would still go with as you suggested, specifying both allowed and forbidden patterns. One approach I've seen that I liked was that first all files from allowed patterns are found, and then any forbidden files are removed. So, in other words, if a file was matched successfully against both `ALLOWED_EXTENSIONS` and `FORBIDDEN_EXTENSIONS`, then the file will be OMITTED.

```py
COMPONENTS = {
	"allowed_static_files": [
		r"\.{css|js|png|jpg|jpeg|svg}$",
	],
	"forbidden_static_files": [
		r"\.{html|py|pyc}$",
	],
}
```

The defaults would then be:
```py
COMPONENTS = {
	"allowed_static_files": [
		# Match all files
		".*",
	],
	"forbidden_static_files": [
		# Exclude .html, .py, .pyc
		r"\.{html|py|pyc}$",
	],
}
```

src/django_components/finders.py
JuroOravec: I've seen it used in this VSCode extension https://marketplace.visualstudio.com/items?itemName=junstyle.vscode-django-support.

They use `.django`, `.dj`, `.tpl`, but I forgot about the latter two.

src/django_components/app_settings.py
```diff
@@ -118,6 +126,25 @@ def RELOAD_ON_TEMPLATE_CHANGE(self) -> bool:
     def TEMPLATE_CACHE_SIZE(self) -> int:
         return self.settings.get("template_cache_size", 128)
 
+    @property
+    def STATIC_FILES_ALLOWED(self) -> List[str]:
+        default_static_files = [
+            # Allow suffixes: .js, .css
+            r"\.(?:js|css)$",
```
EmilStenstrom, line 133: Should we not also allow images by default?

EmilStenstrom (commented):
This is great work, as always! :)

1) I personally would have gone with just a list of file endings instead of regex. Developers are not as comfortable with regex as experienced developers think. They might mess up that ".html" means "[anything]html", they might mess up the raw string, the parenthesis, not understand how they would allow wildcard to the list? [When I advocated that Django move away from regex urlpatterns](https://groups.google.com/g/django-developers/c/u6sQax3sjO4/m/l3nvZCuuAAAJ) to path urlpatterns that was the exact same argument.

2) A generally question: What's a good default, shouldn't we allow images? Fonts?

src/django_components/app_settings.py
EmilStenstrom: Suggestion:

```python
default_static_files = [".css", ".gif", ".ico", ".jpeg", ".jpg", ".js", ".png", ".svg", ".webp"]
````

JuroOravec: Good question. We can add [common image types as defined in MDN](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/Image_types#common_image_file_types).

As for fonts, at first my reaction was that these should be probably at the root of the project, outside of components. But one could have an "app" component that would hold fonts and such, so I can see that now. We can go with the ones listed here https://stackoverflow.com/q/30572159/9788634

JuroOravec: For the regex patterns, what do you say if we supported both? That preferred way would be to pass in strings, which would be evaluated as suffixes like `.html`. But one could also pass an instance of `re.Pattern` (AKA the result of `re.compile`) for more control.

JuroOravec: Updated, see the changes here https://github.com/EmilStenstrom/django-components/pull/652/files/766fd047711463a32f2910cf578dbfcb27b5cd9d..16b418d8d78e56b78a68f0a6f184ade8dc6daec4

EmilStenstrom (approved):
Looks great! <3

PR #654, opened 2024-09-08
Minimize tox config and add black

EmilStenstrom:
I've gone through and made sure we're not including lots of unessesary cruft in the tox.ini, and also added black so that it doesn't just exist in the pre-commit config.

Experimented with adding bandit check but decided against it because it just added lots of false positives.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,63 +4,52 @@
 
 [tox]
 envlist =
-  py38-django{42}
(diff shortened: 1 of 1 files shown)


EmilStenstrom:
@coderabbitai review

JuroOravec (approved):

PR #655, opened 2024-09-08
Improve pitch in README

EmilStenstrom:
An attempt at modernizing the pitch when we're now aiming for practical parity with many frontend frameworks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,19 +4,24 @@
 
 [**Docs (Work in progress)**](https://EmilStenstrom.github.io/django-components/latest/)
 
-Create simple reusable template components in Django
+Django-components is a package that introduces component-based architecture to Django's server-side rendering. It aims to combine Django's templating system with the modularity seen in modern frontend frameworks.
 
 ## Features
 
-<!-- FIXME Links -->
+1. 🧩 **Reusability:** Allows creation of self-contained, reusable UI elements.
+2. 📦 **Encapsulation:** Each component can include its own HTML, CSS, and JavaScript.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,19 +4,24 @@
 
 [**Docs (Work in progress)**](https://EmilStenstrom.github.io/django-components/latest/)
 
-Create simple reusable template components in Django
+Django-components is a package that introduces component-based architecture to Django's server-side rendering. It aims to combine Django's templating system with the modularity seen in modern frontend frameworks.
```
JuroOravec, line 7: I think the second sentence can be omitted, and instead mention Vue / JSX, to give people a better idea of what we mean, e.g.:

"Django-components is a package that introduces component-based architecture seen in Vue / JSX to Django's server-side rendering. "

When I describe the project, I usually day it's like "Vue of the Django world 😃

EmilStenstrom: I kind of like that we talk about Django's templating. Would it work if we put some examples of frontend frameworks there?

JuroOravec: Makes sense with specifying the templating part. I suggested to remove the sentence because it sounded like repeating the same idea with different words. To me, "component-based architecture" and "modularity" mean the same in the context of frontend frameworks.

Maybe something like this?

"Django-components is a package that introduces to Django's tempting the component-based architecture as seen in frontend frameworks like Vue / JSX."

But ultimately any of these are good 😄

JuroOravec (approved):

Issue #657, opened 2024-09-08
Unexpected behaviour of `ComponentDependencyMiddleware` in combination with `as_view` usage


Indomian:
Hi! First I want to thank you for the great library.

I run into an issue which appears in combination of different features inside library. Issue appears if one uses `ComponentDependencyMiddleware` + `RENDER_DEPENDENCIES` setting + `as_view` usage of the component. 

Error provided: *The component "ClassName" is not registered* at django_components/component_registry.py, line 245

## How to reproduce

1. Configure `django_components` with dependencies middleware following default documentation
2. Create a simple component (doesn't matter which one)
3. Configure `RENDER_DEPENDENCIES: True`
4. Create any location which will be using given component
5. Try to open location - you will receive an exception *The component "MagicComponent" is not registered*

I created a simple repository with demo application illustrating combination of settings: https://github.com/Indomian/django_component_as_view_error

## Analysis

I belive problem is located somewhere in the combination of logic that is located here: https://github.com/EmilStenstrom/django-components/blob/b7e913a86087aee6f67d15805aca4a4252680e37/src/django_components/component.py#L673.
In general when component is rendered from `as_view` call it doesn't have any provided name, so value of `self.name` would be equal to the class name and this will lead to rendering of placeholder with name of the component which was not registered before. See https://github.com/EmilStenstrom/django-components/blob/b7e913a86087aee6f67d15805aca4a4252680e37/src/django_components/middleware.py#L70, here components registry will throw an exception that component with given class name is not registered (which is correct, due to the fact that is was registered under another name).

## How to approach to fix

I tried to check if it is possible to provide registred name to the `as_view` call, but right now it doesn't pass through any arguments. The only solution I came up with is:

```python
from django.urls import path
from django_components import ComponentView

from components.test_component.component import MagicComponent

component = MagicComponent(
    registered_name="test_component"
)


urlpatterns = [
    path(
        'magic/',
        ComponentView.as_view(
            component=component
        ),
        name="magic_component_name"
    )
]
```

But currently I'm not sure about downsides of this solution :). Also, I think I can create separate class with a component view which I will use later, but it would be definitely confusing.

JuroOravec:
Hey @Indomian, thanks for the report, you're right with the analysis. The middleware will undergo a revamp in a week or two, which will refactor the placeholders logic. But the component-name-passing behavior will still be there, so we can fix it before then.

[...]

PR #664, opened 2024-09-11
refactor: pass instantiated component to as_view

JuroOravec:
Currently the logic of middleware that inserts the JS and CSS dependencies works with REGISTERED components only (See https://github.com/EmilStenstrom/django-components/issues/657). This will be soon refactored, but since it'll be a large change, it may be 2 weeks or so. 

To allow to work around this limitation in the meantime, this MR refactors the `Component.as_view()` class method to work similar to `render` or `render_to_response`. That is, if the method is called on an INSTANCE, the component instance is passed. And if the method is called on the CLASS, we create an empty instance.

That way, one will be able to pass a dummy "registered_name" to the component:

```py
# Before
MyComponent.as_view()
# If instantiated, the registered name is ignored
MyComponent("my_comp").as_view()

# After
MyComponent.as_view()
# If instantiated, the registered name is used
MyComponent("my_comp").as_view()
```

Part of https://github.com/EmilStenstrom/django-components/issues/657


EmilStenstrom (approved):
Yay!

PR #668, opened 2024-09-12
Make sure BASE_DIR is a Path - fixesd #687

spapas:
This coverts BASE_DIR to a `pathlib.Path`. It works fine if BASE_DIR is already a Path.

### DIFF
--- a/src/django_components/app_settings.py
+++ b/src/django_components/app_settings.py
@@ -1,5 +1,6 @@
 import re
 from enum import Enum
+from pathlib import Path
 from typing import TYPE_CHECKING, Dict, List, Tuple, Union
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


EmilStenstrom:
@spapas looks good, would you mind adding a test as well? So we don't accidentally break this in the future :)

spapas:
@EmilStenstrom I've added two tests, one for string and one for path `BASE_DIR`

EmilStenstrom:
@JuroOravec How are we doing releases nowadays? :)

EmilStenstrom:
@spapas Thanks! Will be part of the next release.

JuroOravec:
@EmilStenstrom Same as before, tho I use Github UI to create and push the git tags:
1. Bump version in [pyproject.toml](https://github.com/EmilStenstrom/django-components/blob/master/pyproject.toml) and merge that to master
2. Go to [Releases](https://github.com/EmilStenstrom/django-components/releases) page and hit "Draft a new release"
3. When selecting a tag in "Choose a tag", I write the name of the new git tag, e.g. `0.101`.  It will show an option "Create new tag 0.101". Clicking that will create a new git tag when you create the release.
4. Then click "Generate release notes"
5. I tend to add `### feat`, `### fix`, etc... subheaders, and categorize and clean up the listed MRs
6. Click "Publish release", and you're done!

Issue #669, opened 2024-09-12
0.100: TypeError: the 'package' argument is required to perform a relative import 


danjac:
This error happens at startup:

```
  File "/PATH-TO-PROJECT/.venv/lib/python3.12/site-packages/django_components/autodiscover.py", line 36, in autodiscover
    return _import_modules(modules, map_module)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/PATH-TO-PROJECT/.venv/lib/python3.12/site-packages/django_components/autodiscover.py", line 65, in _import_modules
    importlib.import_module(module_name)
  File "/usr/lib64/python3.12/importlib/__init__.py", line 84, in import_module
    raise TypeError("the 'package' argument is required to perform a "
TypeError: the 'package' argument is required to perform a relative import for '.venv.lib.python3.12.site-packages.django_components.components.__init__'
```

Settings: 

```python
COMPONENTS = {
    "app_dirs": ["components"],
    "dirs": [BASE_DIR / "django_project" / "components"],
}
```

Previous (working) 0.97 settings:

```python
COMPONENTS = {
    "libraries": [
        "django_project.components.buttons",
        "django_project.components.cards",
        "django_project.components.page_header",
        "django_project.components.pagination",
        "django_project.components.search",
        "django_project.my_app.components.cards",
    ]
}
```

i.e. I have a number of components defined in modules under `django_project/components` e.g. `django_project/components/buttons.py` and app-specific components under  `django_project/my_app/components`  e.g. `django_project/my_app/components/cards.py`. 

Using Python 3.12.

EmilStenstrom:
@danjac What happens if you just remove all settings? The way you've set things up is the default, so you shouldn't need any settings, either app_dirs, dirs, or libraries.

We shouldn't break like that, so just trying to find a workaround.

danjac:
I get the same error with the `COMPONENTS`  setting commented out.

JuroOravec:
@danjac Could you create a demo project? I wasn't able to reproduce this.

JuroOravec:
My guess is that it has to do with this import path:

`.venv.lib.python3.12.site-packages.django_components.components.__init__`

It's the path to `django_components.components`, and it's there because we also look for 3rd party installed Django apps that are installed (`django_components`), and search for `app_dirs` in them (so `django_components.components`).

Then there's one safeguard to ignore those paths that are outside of the project root.

[...]

PR #672, opened 2024-09-14
fix: autoimport with nested apps

JuroOravec:
Closes https://github.com/EmilStenstrom/django-components/issues/669

### DIFF
--- a/src/django_components/apps.py
+++ b/src/django_components/apps.py
@@ -25,7 +25,7 @@ class ComponentsConfig(AppConfig):
         # See https://github.com/EmilStenstrom/django-components/discussions/567#discussioncomment-10273632
         # And https://stackoverflow.com/questions/42907285/66673186#66673186
         if app_settings.RELOAD_ON_TEMPLATE_CHANGE:
-            dirs = get_dirs()
+            dirs = get_dirs(include_apps=False)
             component_filepaths = search_dirs(dirs, "**/*")
             watch_files_for_autoreload(component_filepaths)
 
--- a/src/django_components/autodiscover.py
+++ b/src/django_components/autodiscover.py
@@ -4,8 +4,10 @@ import os
 from pathlib import Path
 from typing import Callable, List, Optional, Union
 
+from django.apps import apps
 from django.conf import settings
 
+from django_components.app_settings import app_settings
 from django_components.logger import logger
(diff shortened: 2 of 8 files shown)


EmilStenstrom (approved):
This is way beyond my comprehension, but I won't block this from merging. I trust the tests! :)

JuroOravec:
@EmilStenstrom Sorry didn't wrote description.

But basically, the definition of component directories comes from 2 places:
1. The `COMPONENTS.dirs` setting, (which defaults to `BASE_DIR / "components"`), 
2. For each installed app, we look for `[app]/components` (or, more generally, subfolders set by `COMPONENTS.app_dirs` within installed apps.

Both of these sources are practically filesystem paths (`./path/to/abc`). But to import the files as python module, we need to use `import_module`, which expects python import path (`path.to.abc`).

The `_filepath_to_python_module` function converts the filesystem path to python module path. How it works is that given that you know a filesystem path (`./path/to/abc`)  AND a python module path (`path.to.abc`) of some folder, then we can search for the contents of that folder, looking for Python files, and convert them to python module paths.

E.g. given that `./path/to/abc` -> `path.to.abc`, then `./path/to/abc/subdir/xyz.py` ->  `path.to.abc.subdir.xyz`

Similarly, `./path/to/abc/subdir/__init__.py` would be `path.to.abc.subdir`, because of how python works.

While the paths from `COMPONENTS.dirs` were fine, because we were resolving them relative to BASE_DIR, the paths from the installed apps were the issue. Because we cannot assume that they will be within BASE_DIR. E.g. third party packages will be defined whereever the Python executable has the packages defined. So for the installed apps, we have to use the apps name and path, to do the conversion above.

E.g. in case of the `django_components` package:

- app.name == `"django_components"`
- app.path == `/Users/myuser/Library/Python/python312/lib/site-packages/django_components/` (I'm making this path up)

So then, when we search for the `[app]/components`, we search in `app.path + "/components"`.

And if we for example find the `dynamic` component there, it's full path may be

`/Users/myuser/Library/Python/python312/lib/site-packages/django_components/components/dynamic.py`

And resolving this relatively to the app's root path, we can convert that to `django_components.components.dynamic`. Which is the correct import path.

EmilStenstrom:
Thanks for explaining!

PR #673, opened 2024-09-14
fix: Using empty default fill with no whitespace

JuroOravec:
Closes https://github.com/EmilStenstrom/django-components/discussions/671


EmilStenstrom (approved):
👍

Issue #635, opened 2024-09-02
First class support for HTML fragments


JuroOravec:
NOTE: I'm jumping way ahead with this one, so really this one becomes relevant only once https://github.com/EmilStenstrom/django-components/issues/622 and similar have been implemented. But since I looked into this today, I wanted to document it.

---

Building on top of https://github.com/EmilStenstrom/django-components/issues/622#issuecomment-2324233043,
I was thinking of what could be an intuitive use of HTML fragments with django_components.

For demonstration, imagine I have a component like so:

```py
class MyComp(Component):
    def get_context_data(self):
        return {
            "my_var": "hello",
        }

    def css_data(self):
        return {
            "bg_color": "blue",
        }

    template = """
        <div class="mycomp">
            <p>
                First - {{ my_var }}
            </p>
            <p class="myfrag">
                Second - {{ my_var }}
            </p>
            <p class="myfrag">
                Third - {{ my_var }}
            </p>
        </div>
    """

    js = """
        const isFragmentA = $els[0].classlist.includes('myfrag');
        const isFragmentB = $els[0].classlist.includes('myfrag-b');

        if (isFragmentA) {
            doSomething();
        } else if (isFragmentB) {
            doSomethingElse();
        } else {
            processAsFullComponent();
        }
    """

    css = """
        me {
            font-size: 12px;
        }

        me > p {
            background: red;
        }

        .myfrag {
            background: var(--bg_color);
        }
    """
```

### JS

1. **Allow to still run component JS, but allow to modify the logic depending on whether and which
   fragment it is.**

    As show in the example, where there is `isFragmentA` and `isFragmentB`, a single
    component can be rendered as multiple fragments.

    In all 3 cases `isFragmentA`, `isFragmentB`, and `else`, the `$els` would point to different elements.
    In full render, it is the top-level root elements. In `isFragmentA` and `isFragmentB`, it would
    be the fragments' root elements.

    We can save users some work by figuring for them which fragment is the JS script running.

    For this, I suggest to declare the possible fragments and their names on the `Component` class:

    ```py
    class MyComp(Component):
        class Fragments:
            header = "div > p"
            footer = "div:n-th(2) .row,div:n-th(3) .row"  # <-- Selector can point to multiple elems
    ```

    Thus, when rendering a component as a fragment, user would have to choose from pre-existing options:

    ```py
    MyComp.fragment("header")
    # or
    MyComp.fragment("footer")
    ```

[...]

Issue #277, opened 2023-04-25
Middleware lets script placeholders slip through undeleted in certain cases


lemontheme:
In the course of PR #271 it came to light that one user was observing script placeholder tags in their generated html, despite the middleware being enabled.

(For the uninitiated, these placeholders are added by the `component_js_dependencies` tag when render_dependencies=True. Later, the middleware finds-and-replaces them with script tags linking to the appropriate js scripts.)

So far, we haven't been able to reproduce this behavior, but the conversation in the PR includes an interesting hint:

> Just to be clear, I think this is an issue when you're loading a page that has components which don't have JS scripts defined, this script loads in anticipation but is not replaced. If there's a component on the page with a script, this will be replaced.

We most likely have a bug on our hands. The fact that we can't repro it probably indicates a blind spot in our tests. We're investigating the issue.

If you're noticing similar behavior in your projects, please let us know. Real-world code bases are much better for helping us understand subtle issues like this.

EmilStenstrom:
Here's a simple testcase that reproduces this:

```python
    def test_dependencies_with_no_components(self):
        template = Template(
            """
            {% load component_tags %}
            {% component_dependencies %}
            """
        )
        rendered = template.render(Context())

        # Assertions to ensure placeholders are properly replaced or removed
        self.assertNotIn(
            '<script name="JS_PLACEHOLDER"></script>', rendered, "JS placeholder should be removed"
        )
        self.assertNotIn(
            '<link name="CSS_PLACEHOLDER">', rendered, "CSS placeholder should be removed or replaced"
        )
```

JuroOravec:
I had a brief look into this and the JS/CSS dependencies rendering. The middleware approach seems to be kinda-sorta working (if I stream the response, or I use different content-type than `text/html`, it will not work). So I think it might be better to handle the dependencies insertion BEFORE we hit the Django response. So decoupling the dependency insertion from middleware.

For that, we need to be aware of the rendering execution - when it starts and when it stops. We also need to access the rendered content, so we can do string replacement like we do in the middleware.

Some ideas to achieve that:
1. Function with callback
	
	- Inside `component_dependencies`, we first mark on the Context that we're registering components.
	- Then we pass the context to render function, which would return a string.
	- Then again hidden inside `component_dependencies`, we replace the deps placeholders with actual JS/CSS deps

	```py
	track_dependencies(
		context,
		lambda ctx: Template("""
			{% component_dependencies %}
[...]

PR #688, opened 2024-10-02
feat: skeleton of dependency manager backend

JuroOravec:
This MR adds the basis for using the client-side dependency manager for loading JS and CSS files.

This is what, down the line, will allow for:
- Passing variables from Python to JS / CSS
- Defining and executing per-instance component's JS code with `$onLoad`
- HTML fragments
- Scoped CSS


.github/workflows/tests.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - 'master'
+      - 'dev'
```
JuroOravec, line 7: I'm introducing a `dev` branch, because while this MR is an encapsulated unit of work, it will be much better to merge everything into master once at the end.

README.md
```diff
@@ -76,6 +76,29 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
 
 ## Release notes
```
JuroOravec, line 77: I'm getting to the point for me where having the full release notes inlined in the README is making it hard to read. So as one of the MRs for the `dev`, I would like to move the release notes to their own page, and just link to them from here.

README.md
```diff
@@ -76,6 +76,29 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
...
 
+🚨📢 **Version 0.110**
+- BREAKING CHANGE:
+    - Installation changes:
+        - If your components include JS or CSS, you now must use the middleware and add django-components' URLs to your `urlpatterns`
+            (See "[Adding support for JS and CSS](#adding-support-for-js-and-css)")
+    - `preload` kwarg of `{% component_dependencies %}` tag was removed. This will be replaced with HTML fragment support.
```
JuroOravec, line 84: The removal of the `preload` kwarg from `{% component_dependencies %}` tags can be [seen here](https://github.com/EmilStenstrom/django-components/pull/688/files#diff-7345171ed38e1dd2e4b8199b65a2fd082323bdb331ff0ef0dc0dd3bd43f111aeL58).

It had to be removed, because in the new format, when we render component and insert a `<!-- _RENDERED -->` comment, we also need to know the ID of the component instance.

And we don't have any ID when we just specify the name of the component.

---

Moreover, I believe the `preload` kwarg was likely a workaround for allowing to render components as HTML fragments - E.g. that one would list the names of the components that *may* be loaded into the page AFTER the HTML document has loaded.

And this would be addressed separately in https://github.com/EmilStenstrom/django-components/issues/635, with which the `preload` kwarg would be no longer needed anyway.

---

Taking the HTML fragment support out of the consideration, I can imagine that there could still be a desire for "pre-loading" JS / CSS.

The use case would be one of optimization, for example if:
- A web dev used HTML fragments approach.
- One of the fragments would depend on large JS / CSS files that would be loading slowly.
- HTML fragments are sent on end-user's actions. And so if, during user actions, user would have to wait for the JS / CSS to load, this would be bad user experience.
- In such case the web dev may want to pre-load the fragments' JS / CSS, so at the point of user's action, everything is already loaded.

However, to support pre-loading PROPERLY, we'd have to take into the account that there can be multiple instances of component registries, and that there may be components NOT registered to any `ComponentRegistry` (which the old `preload` kwarg didn't account for).

So, instead, the preloading should be specified on the component, to avoid passing around the reference to it. But it may still be that for one page a component SHOULD be preloaded, and for a different page, it should NOT.

So IMO the component should specify for which "conditions" it should be preloaded and then link that up in the `{% component_dependencies %}`.

So it could look like this:
```py
class MyComp(Component):
    # Component specifies for which "preload_ids"
    # it should have its CSS / JS preloaded
    preload = ["page_1", "page_2", "page_3"]
```

```django
{# And the "preload_id" on `component_dependencies` #}
{# would load only those components that have a matching ID #}
{% component_dependencies id="page_1" %}
```

@EmilStenstrom Let me know what you think of this. IMO it's low on priority, but I can make it into a standalone issue to see if people would want it.

README.md
```diff
@@ -76,6 +76,29 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
...
+    - Component typing signature changed from `Component[Args, Kwargs, Data, Slots]` to `Component[Args, Kwargs, Slots, Data, JsData, CssData]`
+    - If you rendered a component A with `.render()` and then inserted that into another component B, now you must pass `kind=None` to component A:
```
JuroOravec, line 86: This is described more in-depth further down in the section [`Inserting pre-rendered HTML into another component`](https://github.com/EmilStenstrom/django-components/pull/688/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R3296).

As I was updating the README, I realized that there could be a way to be able to insert one rendered component into another without requiring the `kind=None`.

But I will look into that in a separate MR, since this one is already too big.

README.md
```diff
@@ -76,6 +76,29 @@ And this is what gets rendered (plus the CSS and Javascript you've specified):
...
+          kwargs={...},
+          kind=None,
+      )
+      html = CompB.render(
+          kwargs={
+              content=prerendered_a,
+          },
+      )
+      ```
+- `{% component_dependencies %}` tags are now OPTIONAL - If your components use JS and CSS, but you don't use `{% component_dependencies %}` tags, the JS and CSS will now be, by default, inserted at the end of `<body>` and at the end of `<head>` respectively.
```
JuroOravec, line 99: This is one change that was not discussed in the tech doc - I realized that we could make the use of `{% component_dependencies %}` tags optional. And, by default if no `{% component_dependencies %}` tags were inserted into the HTML document, we'd try to insert the CSS into the end of `<head>`, and JS at the end of `<body>`.

---

WIth this, hand-in-hand goes another question / consideration. IMO we should remove `{% component_dependencies %}` and instead keep only `{% component_js_dependencies %}` and `{% component_css_dependencies %}`.

Because having a single template tag that renders both JS and CSS is more convenient, so it's attractive to be used.

But from the perspective of web development best practices, placing the CSS and JS in the same places is NOT a good idea.

- Because either you put both JS and CSS into `<head>`, which means that execution of JS scripts will BLOCK the rendering of the rest of the page, potentially leading to a flash of empty page.
- And if both JS and CSS were placed at the end of `<body>`, then there may be a [flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content), because the HTML will be first rendered without any styling, and then re-rendered with the styling once the parser reaches the CSS styles at the end of the `<body>`.

So, instead, if we keep only `{% component_js_dependencies %}` and `{% component_css_dependencies %}`, users are forced to split the two up. And with a bit of luck, they google the best practices and insert the CSS / JS as we do so by default.

README.md
```diff
@@ -337,6 +360,34 @@ STATICFILES_FINDERS = [
 ]
 ```
 
+### Adding support for JS and CSS
+
+If you want to use JS or CSS with components, you will need to:
+
+1. Add [`ComponentDependencyMiddleware`](#setting-up-componentdependencymiddleware) to `MIDDLEWARE` setting.
```
JuroOravec, line 367: The installation now tells people to ALWAYS use the middleware. And it's only if users read [the docs further](https://github.com/EmilStenstrom/django-components/pull/688/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R3225), where they will learn that there are cases where they don't actually need to use the middleware.

[Also there is no more a setting that decides whether to use the middleware or not. ](https://github.com/EmilStenstrom/django-components/pull/688/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L3120)

README.md
```diff
@@ -3986,5 +4206,8 @@ twine upload --repository pypi dist/* -u __token__ -p <PyPI_TOKEN>
 
 ### Development guides
 
-- [Slot rendering flot](https://github.com/EmilStenstrom/django-components/blob/master/docs/slot_rendering.md)
-- [Slots and blocks](https://github.com/EmilStenstrom/django-components/blob/master/docs/slots_and_blocks.md)
+Deep dive into how django_components' features are implemented.
+
+- [Slot rendering](https://github.com/EmilStenstrom/django-components/blob/master/docs/devguides/slot_rendering.md)
+- [Slots and blocks](https://github.com/EmilStenstrom/django-components/blob/master/docs/devguides/slots_and_blocks.md)
+- [JS and CSS dependency management](https://github.com/EmilStenstrom/django-components/blob/master/docs/devguides/dependency_mgmt.md)
```
JuroOravec, line 4213: I added a doc on [how the JS / CSS dependency management works](https://github.com/EmilStenstrom/django-components/pull/688/files#diff-59a841a79c146e0f8f478fb78480b7930d6bf09da7fc59605ac4836c9527a6ef).

Also I moved these "dev guides" from `docs` to `docs/devguides`, so they are distinguishable from user-facing docs.

pyproject.toml
```diff
@@ -27,6 +27,8 @@ classifiers = [
 ]
 dependencies = [
     'Django>=4.2',
+    'nanoid>=2.0',
+    'selectolax>=0.3',
```
JuroOravec, line 31: Why `nanoid` and `selectolax`?

- Nanoid:

  For many features like CSS scoping, CSS / JS vars, or per-instance JS with `$onLoad`, we need to insert the component's ID into the HTML.

  Previously the component ID was a simply incremental (0, 1, 2, ...). But making this compatible with HTML fragments makes the incremental IDs not suitable.

  Because I could render a full HTML document with components with IDs 0, 1, 2, 3, 4. Then, for whatever reason, the Django server crashes and restarts. And only then user makes an action that triggers a rendering of an HTML fragment.

  In the case if the IDs were incremental, the fragment's HTML could contain component IDs e.g. 0 and 1. But components with IDs 0 and 1 are already in the HTML from the initial render.

  So, instead, we need to use global / universal IDs, so the IDs are not dependent on the server's state like incrementing IDs. And for that I used nanoid.

  As described in a comment here https://github.com/EmilStenstrom/django-components/pull/688/files#diff-b07310aac6ba2705304fd226eda2e8ca6ac5d0324b2d84d6ba83b47f06744169R14, just 6 digit-long ID already has a chance of collision 1 in 3M

- Seletolax:

  For the features above, we need to insert various marks into the generated HTML. E.g. to mark each component's HTML with the respective IDs.

  For the scoped CSS, we decided to implement it like Vue did, by marking all HTML nodes with the component's ID. For something like that, regex is not sufficient, and we need to properly parse the HTML.

  There are multiple libraries for that, but it seems that selectolax is by far the faster ([See comparison](https://medium.com/@yahyamrafe202/4f268ddea8df))

sampleproject/sampleproject/settings.py
```diff
@@ -44,6 +44,7 @@
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
+    "django_components.middleware.ComponentDependencyMiddleware",
```
JuroOravec, line 47: The sample project has also been updated

src/django_components/apps.py
```diff
@@ -25,7 +26,7 @@ def ready(self) -> None:
         # See https://github.com/EmilStenstrom/django-components/discussions/567#discussioncomment-10273632
         # And https://stackoverflow.com/questions/42907285/66673186#66673186
         if app_settings.RELOAD_ON_TEMPLATE_CHANGE:
-            dirs = get_dirs(include_apps=False)
+            dirs = get_component_dirs(include_apps=False)
```
JuroOravec, line 29: I also renamed `get_dirs` to `get_component_dirs`, so it's purpose is clearer.

Initially I wanted to keep the renaming as a separate MR, but it was already too complicated to untangle the two.

src/django_components/component.py
```diff
@@ -76,14 +75,15 @@
 
 # isort: on
 
-RENDERED_COMMENT_TEMPLATE = "<!-- _RENDERED {name} -->"
```
JuroOravec, line 79: Moved `RENDERED_COMMENT_TEMPLATE` inside `dependencies.py`, so all related code is in that file.

src/django_components/component.py
```diff
@@ -538,6 +541,11 @@ def render(
           is rendered. The keys on the context can be accessed from within the template.
             - NOTE: In "isolated" mode, context is NOT accessible, and data MUST be passed via
               component's args and kwargs.
+        - `kind` - Configure how to handle JS and CSS dependencies.
```
JuroOravec, line 544: Another difference from the tech doc - in the tech doc I suggested to add kwarg named `render_dependencies` to `Component.render()` and `Component.render_to_response()`.

During the preparation of the MR, I realized that a better name would be if the kwarg reflected the intent.

So using `kind="document"` means "process the rendered string as HTML document", and it is the default behaviour.

```py
Component.render(kind="document")
```

This way down the line it will be possible to add `kind="fragment"` to render the JS / CSS dependencies for the HTML fragment use case.

```py
Component.render(kind="fragment")
```

In this MR one can also pass `kind=None`, which is [currently a workaround as described here](https://github.com/EmilStenstrom/django-components/pull/688/files#r1784292398), and hopefully I will be able to remove it in one of the following MRs.

src/django_components/component.py
```diff
@@ -618,7 +629,11 @@ def _render_impl(
         self._validate_inputs()
 
         context_data = self.get_context_data(*args, **kwargs)
-        self._validate_outputs(context_data)
+        self._validate_outputs(data=context_data)
+
+        # Process JS and CSS files
+        cache_inlined_js(self.__class__, self.js or "")
+        cache_inlined_css(self.__class__, self.css or "")
```
JuroOravec, line 636: This caches the Component's inlined JS and CSS, so it will be retrievable from within [the endpoint that serves components' inlined JS / CSS](https://github.com/EmilStenstrom/django-components/pull/688/files#diff-371d604ce82d2a01458065f1b137fd981b12af1b0b5c5fb2c1f7f5b45c20003fR699).

However, this will make more sense in the following MRs. Because in this MR we don't post-process the inlined JS / CSS in any way.

src/django_components/component.py
```diff
@@ -672,14 +687,19 @@ def _render_impl(
...
 
-        if is_dependency_middleware_active():
-            output = RENDERED_COMMENT_TEMPLATE.format(name=self.name) + rendered_component
-        else:
-            output = rendered_component
+                # Allow to optionally override/modify the rendered content
+                new_output = self.on_render_after(context, template, html_content)
+                html_content = new_output if new_output is not None else html_content
+
+                output = postprocess_component_html(
```
JuroOravec, line 697: This is where we hand off the rendered HTML to the logic that takes care of inserting component IDs and what not. Inside `postprocess_component_html` will be also where we will implement the scoped CSS feature, for example.

src/django_components/component.py
```diff
@@ -805,7 +821,7 @@ def _validate_outputs(self, data: Any) -> None:
         maybe_inputs = self._get_types()
         if maybe_inputs is None:
             return
-        args_type, kwargs_type, data_type, slots_type = maybe_inputs
+        args_type, kwargs_type, slots_type, data_type, js_data_type, css_data_type = maybe_inputs
```
JuroOravec, line 824: While the generics for the `Component` class were already updated in this MR, the actual code that will validate the data from `get_js_data` and `get_css_data` will be added in the MR where I introduce the JS / CSS variables.

src/django_components/dependencies.py
```diff
@@ -0,0 +1,764 @@
+"""All code related to management of component dependencies (JS and CSS scripts)"""
```
JuroOravec, line 1: As the comment says, this file contains all the code related to managing the JS / CSS dependencies.

I tried to organize it into sections what flow in chronological order, as there's a lot to it:
1. Caching of the component's inlined JS / CSS
2. Inserting of the `<!-- _RENDERED -->` comment into the HTML, so we know which components were used to render an HTML, so we know which JS and CSS we need to load.
3. Extracting of the `<!-- _RENDERED -->` comments from HTML using `render_dependencies` - This is what was previously done in the middleware, but ofc it was expanded upon here.
4. Defining a Django view that we can request to get component's inlined JS / CSS. That way we can treat the inlined JS / CSS similarly to the JS / CSS defined in `Media.js/css`
5. Defining the middleware that wraps `render_dependencies` - This is the same middleware as was previously defined in `middleware.py`

src/django_components/dependencies.py
```diff
@@ -0,0 +1,764 @@
...
+        return self._data.get(key, None)
+
+    def has(self, key: str) -> bool:
+        return key in self._data
+
+    def set(self, key: str, value: str) -> None:
+        self._data[key] = value
+
+
+comp_media_cache = InMemoryComponentMediaCache()
```
JuroOravec, line 81: Initially I used Django's cache, and only then I refactored it to our own class. That's why I also defined it as an abstract class, so if there is ever a need to support different implementations for the cache (e.g. that users could supply their own), then we know that this is where to do it.

src/django_components/dependencies.py
```diff
@@ -0,0 +1,764 @@
...
+
+#########################################################
+# 2. Modify the HTML to use the same IDs defined in previous
+#    step for the inlined CSS and JS scripts, so the scripts
+#    can be applied to the correct HTML elements. And embed
+#    component + JS/CSS relationships as HTML comments.
+#########################################################
+
+
+class Dependencies(NamedTuple):
```
JuroOravec, line 187: Initially I had a list of `Dependencies` objects, hence this class. However, then I figured out we don't need topass around that list, but forgot to remove the `Dependencies` class.

This will be removed in one of the subsequent MRs.

src/django_components/dependencies.py
```diff
@@ -0,0 +1,764 @@
...
+    else:
+        content_ = cast(bytes, content)
+
+    content_, js_dependencies, css_dependencies = _process_dep_declarations(content_, kind)
+
+    # Replace the placeholders with the actual content
+    did_find_js_placeholder = False
+    did_find_css_placeholder = False
+
+    def on_replace_match(match: "re.Match[bytes]") -> bytes:
```
JuroOravec, line 341: Similar replace logic was previously in the middleware. However, there we replaced only the first occurence of the CSS or JS placeholder.

IMO this is probably the cause this bug https://github.com/EmilStenstrom/django-components/issues/277. So instead, now we will replace ALL placeholders we encounter, so the behaviour is more predictable.

src/django_components/dependencies.py
```diff
@@ -0,0 +1,764 @@
...
+        script = comp_media_cache.get(cache_key)
+
+        if script_type == "js":
+            script = mark_safe(f"<script>{_escape_js(script)}</script>")
+        elif script_type == "css":
+            script = mark_safe(f"<style>{script}</style>")
+    return script
+
+
+def _gen_exec_script(
```
JuroOravec, line 616: And here we generate the JS code that calls the client-side dependency manager.

src/django_components/html.py
```diff
@@ -0,0 +1,98 @@
+import re
```
JuroOravec, line 1: Helpers for working with the `selectolax` HTML objects. In the future I want to break down the current `utils.py` file and instead make it into a directory, so this file could be moved to `utils/html.py`.

src/django_components/templatetags/component_tags.py
```diff
@@ -81,55 +65,34 @@ def _get_components_from_preload_str(preload_str: str) -> List["Component"]:
     return components
 
 
-@register.simple_tag(name="component_dependencies")
-def component_dependencies(preload: str = "") -> SafeString:
+def _component_dependencies(kind: Literal["js", "css", "both"]) -> SafeString:
```
JuroOravec, line 68: Refactored the logic for the `{% component_dependencies %}`  tags so that the logic is defined only once.

src/django_components/urls.py
```diff
@@ -0,0 +1,5 @@
+from django.urls import include, path
+
+urlpatterns = [
+    path("components/", include("django_components.dependencies")),
```
JuroOravec, line 4: I didn't mention it explicitly in other comments, but with this change, if users want to use JS or CSS with the components (whether as inlined JS / CSS or inside the `Media` class, they will MUST add our URL router to theirs.

tests/e2e/testserver/testserver/components/__init__.py
```diff
@@ -0,0 +1,80 @@
+from django_components import Component, register, types
```
JuroOravec, line 1: Components here are used in the E2E tests

tests/e2e/testserver/testserver/views.py
```diff
@@ -0,0 +1,51 @@
+from django.http import HttpResponse
```
JuroOravec, line 1: Views here are used in the E2E tests

tests/test_component_media.py
```diff
@@ -9,115 +9,47 @@
...
-            template = "<div class='html-css-only'>Content</div>"
-            css = ".html-css-only { color: blue; }"
-
-        comp = HTMLCSSComponent("html_css_component")
-        self.assertHTMLEqual(
-            comp.render(Context({})),
-            "<div class='html-css-only'>Content</div>",
-        )
-        self.assertHTMLEqual(
-            comp.render_css_dependencies(),
```
JuroOravec, line 42: I removed / replaced the tests that replied on `Component. render_dependencies` and its variants, since those methods were removed.

tests/test_dependency_rendering.py
```diff
@@ -114,13 +174,15 @@ def test_preload_dependencies_render_when_no_components_used(self):
...
         rendered = create_and_process_template_response(template)
+
         self.assertInHTML('<script src="script.js">', rendered, count=1)
         self.assertInHTML(
             '<link href="style.css" media="all" rel="stylesheet"/>',
             rendered,
             count=1,
         )
 
+    @skip("Old implementation of preload no longer compatible - needs rework")
```
JuroOravec, line 185: As I mentioned in [this comment](https://github.com/EmilStenstrom/django-components/pull/688/files#r1784289318), I removed the `preload` kwarg from `{% component_dependencies %}` tags.

However, instead of removing the corresponding tests, I only skipped them. So that, if we decide to implement the preloading feature as outlined in the other comment, then we would come back to these tests.

tests/test_expression.py
```diff
@@ -150,7 +150,7 @@ def get_context_data(
 
         self.assertEqual(
             rendered.strip(),
-            "<div>lorem</div>\n                <div>True</div>\n                <div>[{'a': 1}, {'a': 2}]</div>",
+            "<!-- _RENDERED SimpleComponent_5b8d97,a1bc3e -->\n                <div>lorem</div>\n                <div>True</div>\n                <div>[{'a': 1}, {'a': 2}]</div>",  # noqa: E501
```
JuroOravec, line 153: NOTE: Since the components now leave a comment when rendered, we had to add those here and there.

tests/testutils.py
```diff
@@ -20,14 +20,36 @@
...
+
         super().tearDown()
         registry.clear()
 
         from django_components.template import _create_template
 
         _create_template.cache_remove()  # type: ignore[attr-defined]
 
+    # Mock the `gen_id` function so it returns deterministic IDs
+    def _start_gen_id_patch(self):
```
JuroOravec, line 38: Snce I'm using the UUIDs from nanoid, I needed to mock the `gen_id` function (which itself calls `nanoid`), so that the component IDs in the tests are deterministic.

tests/test_dependency_rendering_e2e.py
```diff
@@ -0,0 +1,140 @@
+"""
+Here we check that all parts of managing JS and CSS dependencies work together
+in an actual browser.
+"""
```
JuroOravec, line 4: And in this file we check if it all works together correctly.

JuroOravec:
@EmilStenstrom This is now ready for review. I know it's a lot, so no rush. The subsequent MRs will be much smaller, as they will be only adding to the backbone that's defined here.

As mentioned in one of the comments above, I'm merging this to `dev`, so we can make MRs for the remaining features separately.

EmilStenstrom (approved):
Love the smell of a 3000 line MR in the morning! Having looked through this, it's thoughtful and feels safe with all the tests. Taking on JS and CSS more seriously feels like a step change here, big things coming to django-components! Great work!

Only high-level thought is if we really need nanoid? There is something nice with having few dependencies, and we're very unlikely to need any updates to nanoid in the future.

README.md
JuroOravec: No harm in checking, so I made an issue for it here https://github.com/EmilStenstrom/django-components/issues/701

src/django_components/component.py
JuroOravec: Turns out that `as` is a reserved keyword :D 

https://realpython.com/lessons/reserved-keywords/

docs/devguides/dependency_mgmt.md
```diff
@@ -0,0 +1,223 @@
...
+
+   `render_dependencies()`, whether called directly, via middleware or other way, does the following:
+
+   1. Find all `<!-- _RENDERED -->` comments, and for each comment:
+
+   2. Look up the corresponding component class.
+
+   3. Get the component's inlined JS / CSS from `Component.js/css`, and linked JS / CSS from `Component.Media.js/css`.
+
+   4. Generate JS script that loads the JS / CSS dependencies.
```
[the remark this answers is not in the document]
JuroOravec: Short answer, no, because the CSS is also inserted as `<style>` and `<link>` tags.

But this raises a good question on whether components would still render correctly if the browser had disabled execution of javascript. [Playwright supports this option](https://stackoverflow.com/questions/65958243/disable-javascript-in-playwright), so we should be able to test that, so I'll do that in one of the follow-up MRs.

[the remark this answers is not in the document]
JuroOravec: To elaborate a bit more, currently there's the `kind="document"` mode, which renders the CSS as said above (inserted as `<style>` and `<link>` tags). So components with `kind="document"` SHOULD be renderable even in no-JS browsers.

However, this is not true for `kind="fragment"`. There, we do need JS to load the JS / CSS dependencies.

docs/devguides/dependency_mgmt.md
```diff
@@ -0,0 +1,223 @@
...
+   then this way we make sure that the `defer` attribute will be present on the `<script>` tag when
+   it is inserted into the DOM at the time of loading the JS script.
+
+5. To be able to fetch component's inlined JS and CSS, django-components adds a URL path under:
+
+   `/components/cache/<str:comp_cls_hash>.<str:script_type>/`
+
+   E.g. `/components/cache/my_table_10bc2c.js/`
+
+   This endpoint takes the component's unique hash, e.g. `my_table_10bc2c`, and looks up the component's inlined JS or CSS.
```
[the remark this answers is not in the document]
JuroOravec: The endpoint path looks like a file, but it's not ([see definition](https://github.com/EmilStenstrom/django-components/pull/688/files#diff-371d604ce82d2a01458065f1b137fd981b12af1b0b5c5fb2c1f7f5b45c20003fR720)). We actually take the component's JS / CSS string from the cache.

pyproject.toml
JuroOravec: I managed to remove nanoid! It was actually really small library, so I copied to code in https://github.com/EmilStenstrom/django-components/pull/688#discussion_r1785985901.

As for why I didn't just use `random.random()` or similar - I don't know enough about cryptography, so I prefer to choose a solution that claims to have what I'm looking for. And nanoid claims to have low chance of collision and optimized for small IDs, which is the case here.

I explored a bit here https://chatgpt.com/share/670512bb-e4b8-8004-9df5-0c9dc60762fb

src/django_components/html.py
```diff
@@ -0,0 +1,98 @@
...
+html_tag_escapes_reverse = {val: key for key, val in html_tag_escapes.items()}
+
+
+class TransformHtmlCallback(Protocol):
+    def __call__(self, head: Optional[LexborNode], body: Optional[LexborNode]) -> None: ...  # noqa: #704
+
+
+def transform_html_document(html: str, transform: TransformHtmlCallback) -> str:
+    # Escape <!doctype>, <html>, <head>, and <body> tags, because Selectolax treats them
+    # specially, which makes it impossible to edit them.
```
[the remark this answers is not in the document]
JuroOravec: > Since we are already detecting the head and body end tags, couldn't we insert our HTML directly with a string replace, and not need selectolax at all?

Unfortunately inserting with string replace its not that simple. Consider this example:

```html
<!-- <body> is bla bla bla -->
<html>
  <body>
      <script>
        console.log("Searching for <body>");
        document.querySelector("body");
      </script>
  </body>
</html>
```

With a naive regex, we wouldn't be able to sensibly select only the correct `<body>` without replacing also also all entries of `<body>` in comments or scripts.

In my case I could use regex, because I'm not inserting anything extra, I'm just temporarily renaming a string and then renaming it back. And even then I'm not 100% if there won't be some edge case.

> So in the first usage of this new library we have to work around very basic limitations... Doesn't bode well for the future I think. Are you still confident this is the right choice?

Yeah, the Python wrapper is not that great, but the underlying C libraries are solid. [See this Reddit thread](https://www.reddit.com/r/Python/comments/vu9fgz/comment/ifg3pj1/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button) (which links to [this performance comparison](https://gist.github.com/MercuryRising/4061368?permalink_comment_id=3004245#gistcomment-3004245)).

Here are my own tests:

Selectolax is 4-4.5x as fast as lxml when it comes to pure parsing:
```
==== Total trials: 5000 =====
LXML total time: 12.8 s
LEXBOR total time: 2.8 s
```

And about 3x as fast as lxml when it comes to mutating HTML:

```
==== Total trials: 5000 =====
LEXBOR - Total time: 6.2 s
LXML - Total time: 19.2 s
```

But the HTML mutation with my workarounds is actually slower than lxml:

```
==== Total trials: 5000 =====
LEXBOR - Total time: 20.0 s
LXML - Total time: 16.9 s
```

So if we got the issues resolved upstream, we'd get a speed-up of 3-4x. At the same time, implementing the features like scoped CSS and whatnot will overall require the HTML to be parsed 2-3x. So in the end we should end up with comparable performance than if we parsed the HTML once with lxml.

The upstream issue consists of 2 parts, and I've reported it in https://github.com/rushter/selectolax/issues/128:

1. API for inserting one node into another
   - Here, [Lexbor DOES implement that](https://github.com/lexbor/lexbor/blob/72236d31da1a4c3d83c621446dcf4f70345fda05/source/lexbor/dom/interfaces/node.c#L385). It's just missing from the Selectolax wrapper, but adding it should be as simple as [copying this function and updating function calls](https://github.com/rushter/selectolax/blob/4370cddf04a6ed3e8cad97f2d588bb5f6c854511/selectolax/lexbor/node.pxi#L556).

2. Making sure methods like `Node.replace_with()` or `Node.insert_child()` work with `<head>` and `<body>` tags.
   - This is an unknown for me, and I will need to see what the maintainers of Selectolax have to say.
      - Tho I can imagine that it could turn out that `Node.replace_with()` might not work with head / body, while methods like `Node.insert_child()` would work. Because `Node.insert_after/before()` work fine.

README.md
```diff
@@ -3113,13 +3222,124 @@ MIDDLEWARE = [
...
+```
+
+In this case, as mentioned above, each time we call `Component.render()`, we also call `render_dependencies()`.
+
+However, there is a problem here - When we call `render_dependencies()` inside `CardActions.render`,
+we extract the info on components' JS and CSS from the HTML. But the template of `CardActions`
+contains no `{% component_depedencies %}` tags, and nor `<head>` nor `<body>` HTML tags.
+So the component's JS and CSS will NOT be inserted, and will be lost.
+
+To work around this, you must set `kind=None` when rendering pieces of HTML with `Component.render()`
```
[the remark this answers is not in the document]
JuroOravec: So I tried different approaches today, but TBH all of them are unnecessarily complex and felt error prone, or would require that we re-parse the HTML every time we insert one component into another.

So that means that we DO need to pass some kind of flag to `Component.render()` to tell it that the component will be nested into another one.

So what about `nested=True`?

JuroOravec:
Lot's of comments, so to summarize, these are outstanding points for this PR:
- Rename `kind` to something else? (`as` is not available; Personally I'm fine with `kind`)
- [Use `nested=True` instead of `kind=None`?](https://github.com/EmilStenstrom/django-components/pull/688#discussion_r1792436961)

And these are changes I will make in subsequent MRs:
- [Move release notes to separate file](https://github.com/EmilStenstrom/django-components/pull/688#discussion_r1784264634)
- ~[Allow to insert one rendered component into another without needing `kind=None`](https://github.com/EmilStenstrom/django-components/pull/688#discussion_r1784292398)~
- [Remove `{% component_dependencies %}` and keep only the JS / CSS variants](https://github.com/EmilStenstrom/django-components/pull/688#discussion_r1784302542)
- [Remove the `Dependencies` class](https://github.com/EmilStenstrom/django-components/pull/688#discussion_r1784580982)
- [Refactor the utils files](https://github.com/EmilStenstrom/django-components/pull/688#discussion_r1784629551)
- [Test component behaviour in no-JS browser](https://github.com/EmilStenstrom/django-components/pull/688#discussion_r1791411783)
- [Replace workarounds around Selectolax once implemented upstream](https://github.com/EmilStenstrom/django-components/pull/688#discussion_r1792114756)

src/django_components/html.py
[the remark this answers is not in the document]
JuroOravec: Note: [I've added `Node.insert_child()` upstream](https://github.com/rushter/selectolax/pull/130) so once its released, we should be able to remove `transform_html_document`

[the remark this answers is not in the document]
JuroOravec: Sorry for spam, but good news, Selectolax fix was released, and in my local code (not in this PR) I managed to replace the workaround code with the code below. 

```py
tree = LexborHTMLParser(html_content)

# Parse the head and body snippets into elements
_, head_snippet_elements = parse_multiroot_html(head_snippet)
_, body_snippet_elements = parse_multiroot_html(body_snippet)

# Append each element in the head snippet to the <head> tag
for element in head_snippet_elements:
    tree.head.insert_child(element)

# Append each element in the body snippet to the <body> tag
for element in body_snippet_elements:
    tree.body.insert_child(element)

# Return the modified HTML content as a string
return tree.html
```

And this is still 3x faster than lxml:

```
==== Total trials: 5000 =====
LEXBOR - Total time: 8.5
LXML - Total time: 24.3
```

README.md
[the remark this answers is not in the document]
EmilStenstrom: What about render_dependencies=False?

docs/devguides/dependency_mgmt.md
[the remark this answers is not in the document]
EmilStenstrom: Understood!

docs/devguides/dependency_mgmt.md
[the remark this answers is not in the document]
EmilStenstrom: "To be able to fetch component's inlined JS and CSS, django-components adds a URL"... How is the URL connected to INLINE JS/CSS?

src/django_components/html.py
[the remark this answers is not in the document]
EmilStenstrom: Awesome!

EmilStenstrom:
@JuroOravec I'm fine with it being merged as is (that's why I approved). As for the naming, "kind" sounds a bit too much like "you are so kind", so I need to mentally parse it to mean "type" instead. So type is an option. But really, you decide.

docs/devguides/dependency_mgmt.md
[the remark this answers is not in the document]
JuroOravec: The endpoint returns the component's inlined JS / CSS.

To understand why:
1. The JS / CSS from `Component.Media.js/css` is formatted by Django as HTML tags. E.g.
    - `<script src="/path/to/my/script.js"></script>`
    - `<link href="/path/to/my/style.css" />`

2. But the inlined JS / CSS is just plain text, e.g.
   ```js
   console.log("hi!")
   ```
   ```css
   .my-table { color: red; }
   ```

3. But we want to allow the client-side dependency manager to 1. Fetch the inlined JS / CSS, and 2. To decide if it needs to be fetched at all.

   To achieve that, the endpoint above returns the INLINED JS / CSS.

4. That way, instead of
   ```js
   console.log("hi!")
   ```
   ```css
   .my-table { color: red; }
   ```

	we can treat the component's inlined JS / CSS also as HTML tags:

    - `<script src="/components/cache/my_table_10bc2c.js"></script>`
    - `<link href="/components/cache/my_table_10bc2c.css" />`

PR #702, opened 2024-10-09
refactor: selectolax update and tests cleanup

JuroOravec:
Some of a cleanup as mentioned here https://github.com/EmilStenstrom/django-components/pull/688#issuecomment-2400732761

Changes:
- Add test for noJS browser (https://github.com/EmilStenstrom/django-components/commit/3ac74d22a77994fa5c75ccf193cdcc7387355d39)
- Remove skipped tests that tested preloading (https://github.com/EmilStenstrom/django-components/commit/2d2ec48b9c2c6b154c277c756691be805ea23f7b)
- Remove leftover code for preloading (https://github.com/EmilStenstrom/django-components/commit/696f027b5298b3e320e20250bb4f57041ab198c8)
- Update selectolax to use Node.insert_child (https://github.com/EmilStenstrom/django-components/commit/20612e2d0e5a2e01bd330a1b148216a377f5295f)

### DIFF
--- a/src/django_components/html.py
+++ b/src/django_components/html.py
@@ -1,5 +1,4 @@
-import re
-from typing import Dict, Optional, Protocol, Union, cast
+from typing import List, Tuple, cast
 
 from selectolax.lexbor import LexborHTMLParser, LexborNode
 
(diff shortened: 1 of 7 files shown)


tests/test_dependency_rendering_e2e.py
```diff
@@ -138,3 +138,80 @@ async def test_multiple_component_dependencies(self):
         self.assertEqual("rgb(255, 0, 0)", data["myStyle2Color"])  # AKA 'color: red'
 
         await page.close()
+
+    @with_playwright
+    async def test_renders_css_nojs_env(self):
```
JuroOravec, line 143: This is the test that checks that the CSS works even without JS

src/django_components/html.py
```diff
@@ -1,101 +1,100 @@
...
+    # top-level elements, we wrap them in a dummy singular root.
+    parser = LexborHTMLParser(f"<root>{html}</root>")
+
+    # Get all contents of the root
+    root_elem = parser.css_first("root")
+    elems = [*root_elem.iter()] if root_elem else []
+    return cast(LexborNode, root_elem), elems
+
+
+def is_html_parser_fragment(html: str, tree: LexborHTMLParser) -> bool:
```
JuroOravec, line 50: One thing that could still be improved in selectolax is that if we use the `HTMLParser`, then it expects a well-formatted HTML with `<html>`, <head>`, and `<body>` tags. And it adds them if these are missing.

So then I have to do the workaround below to figure out which one we're dealing with.

I've raised this https://github.com/rushter/selectolax/issues/74#issuecomment-2404470344

tests/test_html.py
```diff
@@ -32,59 +35,236 @@ def test_parse_node(self):
             """,
         )
 
-    def test_insert_before_end(self):
-        node = parse_node(
+    def test_parse_multiroot_html(self):
```
JuroOravec, line 38: Removed old and added new tests

src/django_components/html.py
```diff
@@ -11,91 +10,12 @@ def parse_node(html: str) -> LexborNode:
...
-    final_html = HTML_ROOT_TAGS_REVERSE_REGEX.sub(on_reverse_replace_match, transformed_html)
-
-    # NOTE: Because of how we work around selectolax to enable modifying the <body> and <head>
-    # tags, selectolax also adds `</doctype_>` at the end of the content when we serialize the HTML.
-    # So we have to remove that.
-    return final_html[: -len("</doctype_>")]
+    # Get all contents of the root
+    root_elem = parser.css_first("root")
+    elems = [*root_elem.iter()] if root_elem else []
+    return cast(LexborNode, root_elem), elems
```
EmilStenstrom: Really nice update, well done! :)

EmilStenstrom: Why are we returning the `<root>`-wrapped LexborHTMLParser here? Seems it's not used anywhere, so maybe remove? More elegant to keep the root elem an implementation detail.

tests/test_html.py
```diff
@@ -32,59 +32,57 @@ def test_parse_node(self):
...
+        root, nodes = parse_multiroot_html(html)
 
         self.assertHTMLEqual(
-            node.html,
+            root.html,
+            f"""
+            <root>
+                {html}
+            </root>
+            """,
```
EmilStenstrom, line 62: Internal detail, not worth it to test.

EmilStenstrom (approved):
Yay!

src/django_components/html.py
JuroOravec: In this MR there's code only to parse the multi-root HTML, but later on there will be also a need to serialize it back to string. Back when I was young and foolish (few days back), I used the `root` for serializing it. But I can also use just the elements.

Updated in https://github.com/EmilStenstrom/django-components/pull/702/commits/4bff9e73026d7f826dbbde60ef2a3ba85bca36ac!

PR #707, opened 2024-10-12
refactor: split up utils file and move utils to util dir

JuroOravec:
And one more for today, again a cleanup one, as mentioned here https://github.com/EmilStenstrom/django-components/pull/688#issuecomment-2400732761

In this MR, I clean up the utilities / helper functiions, so that:
- The files for main concepts, like `component` or `autodiscover` remain directly under the `src/django_components`
- The utilities are in `src/django_components/util`
- Utilities are further split based on what they focus on, e.g. `logger` or `html`

No logic changes.


src/django_components/apps.py
```diff
@@ -14,7 +14,7 @@ def ready(self) -> None:
         from django_components.component_registry import registry
         from django_components.components.dynamic import DynamicComponent
         from django_components.template_loader import get_component_dirs
-        from django_components.utils import search_dirs, watch_files_for_autoreload
+        from django_components.util.utils import search_dirs, watch_files_for_autoreload
```
EmilStenstrom, line 17: Maybe call utils something else when you are there messing about? util.utils doesn't look good :)

EmilStenstrom (approved):
Looks good to me.

Maybe split utils into many smaller utils?

src/django_components/apps.py
JuroOravec: Agree, but wasn't sure what other name to give it. Maybe can rename it to `misc.py`, so it'd be `util.misc`?

As for the idea of splitting utils into (i assume) single-function files, I'm not personally a favourite, because I find it harder to find common patterns and navigate at such granular level.

PR #708, opened 2024-10-13
docs: Move docs inside src/ to allow imports in python scripts

JuroOravec:
WIth those smaller changes out of the way, next in line for the [Big Features™️](https://docs.google.com/document/d/1A9GfJrYHpmbcXUkQfk0wJFmjJyJs5vlfFESWHPb1w-k) would be to add support for CSS / JS variables.

But in the meantime I started polishing the documentation. So I thought I'd first merge those changes to clean up my workspace (it's not ALL documentation reworked, just about 3 pages).

As first step of that, I moved the `docs` directory inside `src` dir, so we can have python scripts inside the `docs` directory, and these files can import each other.

My train of thought is that:
1. For the API reference, I want to reduce what is documented, so only the public API is, so we have the freedom to change stuff internally.
2. Secondly, currently API reference justs lists all python modules and their contents. But for some API references, like settings, there's two issues:
	1. How we define the settings in the app is not captured by mkdocs.
	2. Moreover, I want to tweak how the doc entries are generated. E.g. user shouldn't have to navigate to `django_components` > `app_settings` > to find the settings. The settings should be available e.g. on the `References > Settings` page.
3. Then there also some other tweaks, like linking to the source code, or [pre-processing the docstrings](https://github.com/mkdocstrings/python/discussions/190)
4. And so overall, there's couple of python scripts I wrote to make that work. And they import one another. And for that to work they need to be within `src` directory.

---

PS: After these changes, the docs still look the same:

Screenshot 2024-10-13 at 09 50 43


mkdocs.yml
```diff
@@ -175,13 +174,7 @@ plugins:
             signature_crossrefs: true
             summary: true
             unwrap_annotated: true
-            # show_root_heading: true
```
JuroOravec, line 178: Turns out all these commented-out bits were already defined here, and with the same values. So can be safely removed.

EmilStenstrom (approved):
Safe enought to try!

Q: Does this mean we now also ship the docs as part of a release?

JuroOravec:
> Q: Does this mean we now also ship the docs as part of a release?

Good question, didn't thought of that. But I've checked, and no :)

Screenshot 2024-10-15 at 23 27 19

PR #715, opened 2024-10-16
refactor: Docs prep 1

JuroOravec:
Various changes to make it possible to document programmatically:
- the top-level exports,
- management commands,
- and the variables available inside the templates under `{{ component_vars }}`

I also renamed the `autodiscover.py` file to `autodiscovery.py` ("Y" at the end), so the name doesn't clash with the `autodiscover()` function. Otherwise, since `autodiscover` is in `__all__`, then the MODULE was also part of the export.


src/django_components/__init__.py
```diff
@@ -1,52 +1,67 @@
...
-from django_components.util.types import (
-    EmptyTuple as EmptyTuple,
-    EmptyDict as EmptyDict,
-)
+from django_components.util.types import EmptyTuple, EmptyDict
 
-# isort: on
 
-if django.VERSION < (3, 2):
+if _django.VERSION < (3, 2):
```
JuroOravec, line 36: Also, I wonder if we can remove this logic, since Django 3.2 isn't supported since Apr 2024?

src/django_components/component.py
```diff
@@ -110,6 +111,38 @@ class ViewFn(Protocol):
     def __call__(self, request: HttpRequest, *args: Any, **kwargs: Any) -> Any: ...  # noqa: E704
 
 
+class ComponentVars(NamedTuple):
```
JuroOravec, line 114: In the script that will generate the API reference, I will be importing this class and use its docstrings to document the variables available in the django templates.

src/django_components/components/__init__.py
```diff
@@ -1,3 +1,4 @@
-# flake8: noqa F401
+# NOTE: Components exported here are documented in
+from django_components.components.dynamic import DynamicComponent
```
JuroOravec, line 2: And I will use the exports of this file to generate documentation on pre-defined components.

src/django_components/management/commands/startcomponent.py
```diff
@@ -7,49 +7,117 @@
 
 
 class Command(BaseCommand):
```
JuroOravec, line 9: Same as mentioned in other comments. I want to auto-generate reference docs also for the management commands. So I've put the docs from README in here.

Plus for commands, I generate the inputs section by letting django / argparse to format the arguments, and then I use that to generate the documentation.

src/django_components/middleware.py
```diff
@@ -1,2 +1,4 @@
 # These middlewares are part of public API
-from django_components.dependencies import ComponentDependencyMiddleware as ComponentDependencyMiddleware  # noqa: F401
+from django_components.dependencies import ComponentDependencyMiddleware
+
+__all__ = ["ComponentDependencyMiddleware"]
```
JuroOravec, line 4: And same for the middlewares. All middlewares defined in this file will be picked up by the docs-generating script.

JuroOravec:
Here's demo of how the respective sections will look like in the docs:

Screenshot 2024-10-16 at 08 53 06
Screenshot 2024-10-16 at 08 52 50
Screenshot 2024-10-16 at 08 52 08
Screenshot 2024-10-16 at 08 52 25
Screenshot 2024-10-16 at 08 51 48

src/django_components/__init__.py
EmilStenstrom: Definitely! Anything not here can be removed: https://github.com/EmilStenstrom/django-components?tab=readme-ov-file#compatibility

EmilStenstrom (approved):
Looks really good! 👍

src/django_components/__init__.py
JuroOravec: Alright, updated!

PR #716, opened 2024-10-19
refactor: Document template tags

JuroOravec:
- Documented individual tags like `{% component %}`, `{% slot %}`, and other.
- Refactored the tags so that the parsing of tags inputs (what args, kwargs or flags it accepts) is now declarative. So each tag defines a `TagSpec` instance, which is the source of truth and is also used for generating documentation.

### DIFF
--- a/src/django_components/templatetags/component_tags.py
+++ b/src/django_components/templatetags/component_tags.py
@@ -973,9 +973,7 @@ def _parse_tag(
         else:
             is_key_allowed = (
                 tag_spec.keywordonly_args == True or key in tag_spec.keywordonly_args  # noqa: E712
-            ) or bool(
-                tag_spec.pos_or_keyword_args and key in tag_spec.pos_or_keyword_args
-            )
+            ) or bool(tag_spec.pos_or_keyword_args and key in tag_spec.pos_or_keyword_args)
         if not is_key_allowed:
             is_optional = key in tag_spec.optional_kwargs if tag_spec.optional_kwargs else False
             if not is_optional:
--- a/src/django_components/templatetags/component_tags.py
+++ b/src/django_components/templatetags/component_tags.py
@@ -135,11 +135,11 @@ def _component_dependencies(type: Literal["js", "css"]) -> SafeString:
     return TextNode(mark_safe(placeholder))
 
 
+@register.tag("component_css_dependencies")
 @with_tag_spec(TagSpec(
     tag="component_css_dependencies",
     end_tag=None,  # inline-only
 ))
-@register.tag("component_css_dependencies")
 def component_css_dependencies(parser: Parser, token: Token, tag_spec: TagSpec) -> TextNode:
     """
     Marks location where CSS link tags should be rendered after the whole HTML has been generated.
@@ -158,11 +158,11 @@ def component_css_dependencies(parser: Parser, token: Token, tag_spec: TagSpec) 
     return _component_dependencies("css")
 
 
+@register.tag(name="component_js_dependencies")
 @with_tag_spec(TagSpec(
     tag="component_js_dependencies",
     end_tag=None,  # inline-only
 ))
-@register.tag(name="component_js_dependencies")
 def component_js_dependencies(parser: Parser, token: Token, tag_spec: TagSpec) -> TextNode:
     """
     Marks location where JS link tags should be rendered after the whole HTML has been generated.
@@ -181,6 +181,7 @@ def component_js_dependencies(parser: Parser, token: Token, tag_spec: TagSpec) -
     return _component_dependencies("js")
 
 
+@register.tag("slot")
 @with_tag_spec(TagSpec(
     tag="slot",
     end_tag="endslot",
@@ -190,7 +191,6 @@ def component_js_dependencies(parser: Parser, token: Token, tag_spec: TagSpec) -
     repeatable_kwargs=False,
     flags=[SLOT_DEFAULT_KEYWORD, SLOT_REQUIRED_KEYWORD],
 ))
-@register.tag("slot")
 def slot(parser: Parser, token: Token, tag_spec: TagSpec) -> SlotNode:
     """
     Slot tag marks a place inside a component where content can be inserted
@@ -331,6 +331,7 @@ def slot(parser: Parser, token: Token, tag_spec: TagSpec) -> SlotNode:
     return slot_node
 
 
+@register.tag("fill")
 @with_tag_spec(TagSpec(
     tag="fill",
     end_tag="endfill",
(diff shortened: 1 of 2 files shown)


src/django_components/templatetags/component_tags.py
```diff
@@ -57,34 +136,191 @@ def _component_dependencies(type: Literal["js", "css"]) -> SafeString:
             f"Unknown dependency type in {{% component_dependencies %}}. Must be one of 'css' or 'js', got {type}"
         )
 
-    return mark_safe(placeholder)
+    return TextNode(mark_safe(placeholder))
 
 
-@register.simple_tag(name="component_css_dependencies")
```
JuroOravec, line 63: I've changed `component_dependencies` tags to use `@register.tag()` instead of `@register.simple_tag()`, so that the documentation could be auto-generated.

src/django_components/templatetags/component_tags.py
```diff
@@ -46,6 +60,71 @@
 register = django.template.Library()
 
 
+class TagSpec(NamedTuple):
```
JuroOravec, line 63: TagSpec is what was previously the input to the [`_parse_tag()`](https://github.com/EmilStenstrom/django-components/pull/716/files#diff-7345171ed38e1dd2e4b8199b65a2fd082323bdb331ff0ef0dc0dd3bd43f111aeR797) function. But it's now declared as standalone objects, so we can grab these objects when generating documentation.

I also found it confusing trying to remember what was the difference between `params` and `extra_params`. So I renamed the inputs and aligned them with how [Python represents function parameters](https://docs.python.org/3/library/inspect.html#inspect.Parameter.kind) the inputs, e.g. `positional_only_args`, `pos_or_keyword_args`, `keywordonly_args`.

JuroOravec:
Here's a recording of how it looks:

https://github.com/user-attachments/assets/0cc0a4bf-a624-41ed-8da1-cb3e46a25732

src/django_components/templatetags/component_tags.py
```diff
@@ -298,29 +797,22 @@ class ParsedTag(NamedTuple):
 
 
 def _parse_tag(
```
JuroOravec, line 799: I might move `_parse_tag` to a sepate file in `util/` eventually, but not in this MR for simplicity's sake.

src/django_components/templatetags/component_tags.py
```diff
@@ -57,34 +136,192 @@ def _component_dependencies(type: Literal["js", "css"]) -> SafeString:
...
+    )
+)
+def component_css_dependencies(parser: Parser, token: Token, tag_spec: TagSpec) -> TextNode:
+    """
+    Marks location where CSS link tags should be rendered after the whole HTML has been generated.
+
+    Generally, this should be inserted into the `<head>` tag of the HTML.
 
+    If the generated HTML does NOT contain any `{% component_css_dependencies %}` tags, CSS links
+    are by default inserted into the `<head>` tag of the HTML. (See
```
EmilStenstrom, line 156: Reading this again, are we sure this is the right behavior? Feels very magical that this happens automatically and I'm not sure it outweighs the user not understanding how the framework works.

src/django_components/templatetags/component_tags.py
```diff
@@ -57,34 +136,192 @@ def _component_dependencies(type: Literal["js", "css"]) -> SafeString:
...
+    )
+)
+def component_js_dependencies(parser: Parser, token: Token, tag_spec: TagSpec) -> TextNode:
+    """
+    Marks location where JS link tags should be rendered after the whole HTML has been generated.
+
+    Generally, this should be inserted at the end of the `<body>` tag of the HTML.
+
+    If the generated HTML does NOT contain any `{% component_js_dependencies %}` tags, JS scripts
+    are by default inserted at the end of the `<body>` tag of the HTML. (See
```
EmilStenstrom, line 181: See above.

src/django_components/templatetags/component_tags.py
```diff
@@ -57,34 +136,192 @@ def _component_dependencies(type: Literal["js", "css"]) -> SafeString:
...
+        repeatable_kwargs=False,
+        flags=[SLOT_DEFAULT_KEYWORD, SLOT_REQUIRED_KEYWORD],
     )
+)
+def slot(parser: Parser, token: Token, tag_spec: TagSpec) -> SlotNode:
+    """
+    Slot tag marks a place inside a component where content can be inserted
+    from outside.
+
+    Similar to [Vue's slots](https://vuejs.org/guide/components/slots.html) or React's `children`.
```
EmilStenstrom, line 209: Or the Web Component with the same name: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot - Maybe link to our own docs first, and have these references there instead?

EmilStenstrom (approved):
Yay! Some small suggestions.

src/django_components/templatetags/component_tags.py
JuroOravec: Whaaaat, I didn't know Web components had slots! Such a shame that the content of web components is not queriable with `document.querySelector`, otherwise I'd be already using it :/

Changed the order and added link to web components.

src/django_components/templatetags/component_tags.py
JuroOravec: IMO this is a matter of preference, I personally don't mind if a framework does things for me automatically, as long as I can still modify stuff when needed.

Also it feels like there's already quite an upfront cost, requiring 8 steps to fully install the package with JS / CSS support. Which would be 9 with configuring `{% component_dependencies %}`.

If I imagine I'm starting a new project and installing django_components right from the start, then I have no HTML files yet. And so if I'm told to insert `{% component_css_dependencies %}`, that'd be even more steps for me - create the `templates` dir, look up how the base HTML structure looks like (I always google that), load the `component_tags` library, and only finally insert the tags.

So actually I think the default behaviour removes quite a bit friction there.

On the other hand, I think this should be mentioned early on so it's not magical, but just sane defaults.

The installation steps can be split into two parts to make it less overwhelming:
1. Basic installation (5 steps)
2. Adding support for JS and CSS (3 steps)

So I think that as 4th step of the 2nd section, we can mention that "optionally you can use `{% component_dependencies %}` and by default they are inserted here and here".

This is how that looks (the installation page still needs polishing)

Screenshot 2024-10-20 at 09 08 20

Screenshot 2024-10-20 at 09 08 09

Screenshot 2024-10-20 at 09 07 54

JuroOravec: > Step 4 in the first block of setup is optional too right?

Yup! Then there's also step 5 in the first block, but I just realized it's not in the screenshots 😅

PR #719, opened 2024-10-21
refactor: pass slot fills in template via slots param

JuroOravec:
## Side note

As I was updating the docs, I've found 4 bugs, plus did some tweaks. I want to first fix the bugs / push the tweaks, before I finish the docs, so that I won't have to re-visit the docs afterwards.

So when it comes to PRs, my course of actions is to:

1. Fix the bugs
2. Submit the tweaks
3. And only then I will be again making PRs documenting the API with docstrings (like the [last one](https://github.com/EmilStenstrom/django-components/pull/716))
4. After that, it will be possible to make PRs in parallel for both documentation (breaking up and migrating the README)
   and the rest of the[ features like JS/CSS vars](https://docs.google.com/document/d/1A9GfJrYHpmbcXUkQfk0wJFmjJyJs5vlfFESWHPb1w-k/edit?usp=sharing), etc.

Bugs:
- `self.input.slots` is empty when component is given slots inside the template via `{% fill %}` tags
- Using multiple `{% slot "slot_name" default %}` tags raises error even if those slots use the same name.
- Type validation doesn't handle nested dictionaries
- `reload_on_template_change` doesn't actually work. And the logic is watching only templates.

Tweaks:
- Rename `forbidden_static_files` to `static_files_forbidden` so it aligns with `static_files_allowed`
- Add `js_name` and `css_name` attributes, which are the JS / CSS equivalents of `template_name`.
  - Same as with `template_name`, `js/css_name` and `js/css` are mutually exclusive
  - The difference between `js/css_name` and `Media.js/css` is that `js/css_name` will be treated
    specially (e.g. apply scoped CSS, etc), while `Media.js/css` is not post-processed in any way.
- Lower-case the settings for `ComponentRegistry`, so it's the same casing as the main settings.
  E.g. `CONTEXT_BEHAVIOR` -> `context_behavior`. Because the upper-case variant is only how we refer
  to the settings internally.
- Rename `reload_on_template_change` to `reload_on_file_change`, since it woul now watch all files, not just template files.

...And now onto the actual PR:

---

## Description

`self.input.slots` is empty when component is given slots inside the template using `{% fill %}` tags.

In the UI component library, there's often logic that goes like "render A if slot is given,
or render B if slot is NOT given".

I found that I cannot use `Component.is_filled` for this, because `is_filled` is actually
populated LATER than I need. (This is also one of the topics to discuss, but I'm leaving
this for later). So I have to use `self.input.slots`.

But I found that when a component is defined inside a template, e.g.

```django
{% component "table" %}
    {% fill "headers" %}
    bla bla
    {% endfill %}
{% endcomponent %}
```

Then internally, we store these as `self.fill_content`, and not as `self.input.slots`.

On the other hand, `self.input.slots` is populated when we pass in slots in Python:

```python
Table.render(
    slots={
        "headers": "bla bla",
    },
)
```

So to align the two, I refactored the slots logic, so that the `fill_content` is normalized
as slot functions, and stored under `self.input.slots`.

`fill_content` has not been documented, so should be safe to remove.

### DIFF
--- a/src/django_components/slots.py
+++ b/src/django_components/slots.py
@@ -697,7 +697,7 @@ def _nodelist_to_slot_render_func(
 
         # If slot fill is using `{% fill "myslot" default="abc" %}`, then set the "abc" to
         # the context, so users can refer to the default slot from within the fill content.
-        if data_var:
+        if default_var:
             ctx[default_var] = slot_ref
 
         return nodelist.render(ctx)
--- a/src/django_components/slots.py
+++ b/src/django_components/slots.py
@@ -67,29 +67,6 @@ SlotContent = Union[SlotResult, SlotFunc[TSlotData]]
 # Internal type aliases
 SlotId = str
 SlotName = str
-SlotDefaultName = str
-SlotDataName = str
-
-
-@dataclass(frozen=True)
-class FillContent(Generic[TSlotData]):
-    """
-    This represents content set with the `{% fill %}` tag, e.g.:
-
-    ```django
-    {% component "my_comp" %}
-        {% fill "first_slot" %} <--- This
-            hi
-            {{ my_var }}
-            hello
-        {% endfill %}
-    {% endcomponent %}
-    ```
-    """
-
-    content_func: SlotFunc[TSlotData]
-    slot_default_var: Optional[SlotDefaultName]
-    slot_data_var: Optional[SlotDataName]
 
 
 class Slot(NamedTuple):
@@ -124,8 +101,6 @@ class SlotFill(Generic[TSlotData]):
     escaped_name: str
     is_filled: bool
     content_func: SlotFunc[TSlotData]
-    slot_default_var: Optional[SlotDefaultName]
-    slot_data_var: Optional[SlotDataName]
 
 
 class SlotRef:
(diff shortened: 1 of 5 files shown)


src/django_components/slots.py
```diff
@@ -67,29 +67,6 @@ def __call__(self, ctx: Context, slot_data: TSlotData, slot_ref: "SlotRef") -> S
 # Internal type aliases
 SlotId = str
 SlotName = str
-SlotDefaultName = str
-SlotDataName = str
-
-
-@dataclass(frozen=True)
-class FillContent(Generic[TSlotData]):
```
JuroOravec, line 75: No longer needed, the `slot_default_var` and `slot_data_var` are now part of the slot function.

src/django_components/slots.py
```diff
@@ -195,27 +170,7 @@ def render(self, context: Context) -> SafeString:
             if key.startswith(_INJECT_CONTEXT_KEY_PREFIX):
                 extra_context[key] = value
 
-        # If slot fill is using `{% fill "myslot" default="abc" %}`, then set the "abc" to
```
JuroOravec, line 198: Moved inside `_nodelist_to_slot_render_func`

EmilStenstrom (approved):
Looks good to me!

PR #726, opened 2024-10-28
fix(docs) Replace `shorthand_component_formatter` with `component_sho…

larsent:
…rthand_formatter` in README.md

The proper name appears to be [`component_shorthand_formatter`](https://github.com/EmilStenstrom/django-components/blob/7911daa89ed61132b2a819e74769d748efb3ab03/src/django_components/tag_formatter.py#L222)

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ it allows to return either a string or a Template instance.
     {% endcomponent %}
     ```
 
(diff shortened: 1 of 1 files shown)


EmilStenstrom:
@larsent Thanks for submitting a PR, good catch!

PR #727, opened 2024-10-30
fix: Do not raise error if multiple slots with same name are flagged as default

JuroOravec:
This MR makes it OK if a template contains multiple slots with the same name and flagged as default, e.g.:

```django
{% load component_tags %}
<div>
    <main>{% slot "main" default %}1{% endslot %}</main>
    <div>{% slot "main" default %}2{% endslot %}</div>
</div>
```

### DIFF
--- a/src/django_components/slots.py
+++ b/src/django_components/slots.py
@@ -568,20 +568,25 @@ def _resolve_default_slot(
     else:
         default_fill = None
 
-    default_slot_encountered: bool = False
+    default_slot_name: Optional[str] = None
 
     # Check for errors
     for slot in slots.values():
         if not slot.is_default:
             continue
 
-        if default_slot_encountered:
+        # NOTE: Allow one slot to be marked as 'default', or multiple slots but with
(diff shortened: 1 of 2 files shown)


JuroOravec:
Ah, I forgot about the broken CI pipeline, I will have to fix that before this MR can go in.

JuroOravec:
Btw, @EmilStenstrom, a lil update. It's crazy, but with the tweaks / fixes I made in the last 1-2 weeks (yet to be merged), I got to a point where I can convert Vue to AlpineJS + django-components 😄

EmilStenstrom (approved):
The code looks good. But could you complete this with the WHY of this change?

JuroOravec:
@EmilStenstrom As for why, I came across it when I was playing around and copy-pasting. It feels like a bug, because there's no reason not to allow that. Alteranative would be to require the `default` flag to be on all instances of the same slot, but that feels cumbersome. So allowing one or more slots (of the same name) with the `default` flags feels like the most user friendly approach

PR #728, opened 2024-10-30
refactor: pin version of playwright in tests

JuroOravec:
Turns out that the error was following:
1. When playwright installs a browser, it caches it at
  `/home/runner/.cache/ms-playwright/chromium-1140/chrome-linux/chrome`
2. The cache path contains the browser version.
3. Different versions of playwright point to different versions of cache. 
4. When we installed the browser, we did it using `requirements-ci.txt`, which pinned playwright version to `1.47.0`.
5. But when we ran tox, tox has it's own set of dependencies specified, adn there the playwright version was not pinned and used newer version of playwright.
6. And so, inside tox, the latter playwright version was used, which expected browser to be installed inside `chromium-1140`. But we installed the browser using the older playwright version, which installed version `chromium-v1134`.

So by pinning the playwright version in `tox.ini`, the two should be in sync

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -30,7 +30,10 @@ deps =
   django50: Django>=5.0,<5.1
   pytest
   pytest-xdist
-  playwright
+  # NOTE: Keep playwright is sync with the version in requirements-ci.txt
+  # Othrwise we get error:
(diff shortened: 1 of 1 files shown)


EmilStenstrom:
Great find, as always!

PR #730, opened 2024-10-31
refactor: tag formatter - allow fwd slash in end tag

JuroOravec:
For the UI component library I was playing around with the idea of defining component similarly to how Vue / React does it - with PascalCase and forward slash to indicate the end tag:

```django
{% MyComp %}
{% /MyComp %}
```

But the forward slash was not recognized. So this MR adds `/` to the set of character allowed for start / end tags.

Also updated docs for tag formatter API.


EmilStenstrom (approved):
Looks good to me.

PR #731, opened 2024-10-31
refactor: Use lowercase names for registry settings

JuroOravec:
In this MR I suggest to rename the inputs passed to ComponentRegistry, from all-uppercase to lowercase. The reason is so that the API is aligned with the global settings, which are also all-lowercase. I noticed this difference when I was writing up the documentation.

The reason why I originally defined the settings as uppercase was because I got confused by our internal `app_settings`, where the fields are defined as upper case.

NOTE: This MR has similar structure as the last one - One commit is the change itself, the other is documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,11 @@
     )
     ```
 
+- The [`ComponentRegistry`](../api#django_components.ComponentRegistry) settings [`RegistrySettings`](../api#django_components.RegistrySettings) were lowercased
+  to align with the global settings:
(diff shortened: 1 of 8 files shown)


EmilStenstrom (approved):
Looks good to me! :)

EmilStenstrom:
Should we accept the old formats as well as to not break people's setups unnecessarily??

JuroOravec:
@EmilStenstrom That's a fair point! I've updated it to kept both upper- and lower-case settings, and documented the uppercase as deprecated.

PR #733, opened 2024-11-03
feat: define settings as a data class for type hints, intellisense, and docs

JuroOravec:
I wanted the settings documentation to live close to the code, using the docstrings to generate the API reference docs. However, the way we parsed the django_component's settings, I wasn't able to access the settings fields. There is the `AppSettings` class, but its fields do not match the fields that the user fills in (e.g. user supplies `app_dirs`, while we access it as `app_settings.APP_DIRS`.

So what I did was to define a class (`ComponentsSettings`) that's purely for defining the shape of the user-facing settings. And this becomes the source of truth of what settings are there. This class can then have its fields documented with docstrings, and this can be then used to generate the settings docs. Plus, it also offers users intellisense and type checking if users decide to use `ComponentSettings` to define the settings instead of a plain dict.

So in users' Django settings, instead of

```py
COMPONENTS = {
    autodiscover: False,
}
```

Then can now do

```py
from django_components import ComponentsSettings

COMPONENTS = ComponentsSettings(
    autodiscover=False,
)
```

Users can still supply only the plain dict, so this is not a breaking change.

### DIFF
--- a/src/django_components/app_settings.py
+++ b/src/django_components/app_settings.py
@@ -147,7 +147,7 @@ class Dynamic(Generic[T]):
 #       for `COMPONENTS.dirs`, we do it lazily.
 # NOTE 2: We show the defaults in the documentation, together with the comments
 #        (except for the `Dynamic` instances and comments like `type: ignore`).
-#        So `fmt: off` turns off Black formatting and `--snippet:defaults--` allows
+#        So `fmt: off` turns off Black formatting and `snippet:defaults` allows
 #        us to extract the snippet from the file.
 #
 # fmt: off
--- a/src/django_components/app_settings.py
+++ b/src/django_components/app_settings.py
@@ -1,12 +1,32 @@
 import re
+from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
-from typing import TYPE_CHECKING, Dict, List, Literal, Tuple, Union
+from typing import (
+    TYPE_CHECKING,
+    Callable,
+    Generic,
+    List,
+    Literal,
+    NamedTuple,
+    Optional,
+    Sequence,
+    Tuple,
+    TypeVar,
+    Union,
+    cast,
+)
 
 from django.conf import settings
 
+from django_components.util.misc import default
+
 if TYPE_CHECKING:
     from django_components.tag_formatter import TagFormatterABC
+
+
+T = TypeVar("T")
+
 
 ContextBehaviorType = Literal["django", "isolated"]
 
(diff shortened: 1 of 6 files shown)


src/django_components/app_settings.py
```diff
@@ -93,95 +114,134 @@ class ContextBehavior(str, Enum):
...
 
     @property
     def RELOAD_ON_TEMPLATE_CHANGE(self) -> bool:
-        return self.settings.get("reload_on_template_change", False)
+        return default(self._settings.reload_on_template_change, cast(bool, defaults.reload_on_template_change))
 
     @property
     def TEMPLATE_CACHE_SIZE(self) -> int:
-        return self.settings.get("template_cache_size", 128)
+        return default(self._settings.template_cache_size, cast(int, defaults.template_cache_size))
```
JuroOravec, line 229: (Following is outside of the scope for this MR, just sharing my thoughts)

Since the default values now live outside of `AppSettings` (`InternalSettings`), you can see that there's a lot of repetition here, and most of the properties do the same. The only difference is CONTEXT_BEHAVIOR, where we also validate the input.

This could be simplified by:
1. Moving the `CONTEXT_BEHAVIOR` validation to `ComponentsSettings`.

   At that point, all of these properties would do the same - either return value from given settings or return the default value.

2. Instead of defining all properties on `AppSettings`, we could do away with defining just a single [`__getattr__`](https://docs.python.org/3/reference/datamodel.html#object.__getattr__), that would do the same - return the key from given settings or from the defaults.

---

And when it comes to having validation defined on `ComponentsSettings`, I'd use [pydantic](https://docs.pydantic.dev/latest/) for that.

I also want to introduce pydantic as dependency down the line, because our [custom validation that I wrote](https://github.com/EmilStenstrom/django-components/blob/1fd95708d1239a46d5de19af5aa319942648f54d/src/django_components/util/validation.py) for validating Component's args and kwargs is not good enough. It doesn't handle nested TypedDict / Tuples. And I use a pattern where I define JavaScript component inputs by putting them all inside a `js` dictionary, e.g.:

```py
class TableJsProps(TypedDict):
    items: str
    items_per_page: NotRequired[str]

class TableProps(TypedDict):
    js: TableJsProps
    attrs: dict

@register("table")
class Table(Component[EmptyTuple, TableProps, ...]):
    ...
```

Which I can then use as:

```django
{% component "table"
  js:items="someJsVar"
  js:items_per_page="20"
  attrs:class="flex"
%}
```

Pydantic can handle nested structures, and it can take TypedDict and Tuples as inputs. So we wouldn't have to change the API for how the component types are defined. Only the internal implementation. Plus we could remove the whole [validation.py](https://github.com/EmilStenstrom/django-components/blob/1fd95708d1239a46d5de19af5aa319942648f54d/src/django_components/util/validation.py) file.

src/django_components/app_settings.py
```diff
@@ -93,95 +114,134 @@ class ContextBehavior(str, Enum):
...
+#
+# NOTE: Because we need to access Django settings to generate default dirs
+#       for `COMPONENTS.dirs`, we do it lazily.
+# NOTE 2: We show the defaults in the documentation, together with the comments
+#        (except for the `Dynamic` instances and comments like `type: ignore`).
+#        So `fmt: off` turns off Black formatting and `--snippet:defaults--` allows
+#        us to extract the snippet from the file.
+#
+# fmt: off
+# --snippet:defaults--
```
JuroOravec, line 154: In the docs, I embedded this snippet of default settings, so that it's kept in sync with the actual default values. How I did it was that at docs generation, this file is read, and the section between `--snippet:defaults--` and `--endsnippet:defaults--` is taken (and slightly modified).

Here's how it will look in the docs

Screenshot 2024-11-03 at 10 20 46

src/django_components/app_settings.py
```diff
@@ -93,95 +114,134 @@ class ContextBehavior(str, Enum):
...
+        ".py", ".pyc",
+    ],
+    tag_formatter="django_components.component_formatter",
+    template_cache_size=128,
+)
+# --endsnippet:defaults--
+# fmt: on
+
+
+class InternalSettings:
```
JuroOravec, line 189: I also renamed `AppSettings` to `InternalSettings`, so there's no confusion beteen this and `ComponentsSettings`

JuroOravec:
One more note: I did not add `ComponentsSettings` to the README in this MR. But in the updated docs I updated the code blocks to preferentially use `ComponentsSettings`, and `ComponentsSettings` will be also mentioned on the settings page

Screenshot 2024-11-03 at 10 26 52

src/django_components/app_settings.py
EmilStenstrom: Simplifying code around ComponentsSettings sounds like a good suggestion!

About adding pydantic as a dependency I'm more cautious. Remember that I would like to go for inclusion in Django one day. And Django is not typed. So adding pydantic to Django as a dependency is not going to happen. 

Overall I'm a little worried of all the complexity that types are adding across the library. I would not be comfortable touching any of the internals that deal with type validation that exists now.

src/django_components/app_settings.py
EmilStenstrom: Really nice! :)

EmilStenstrom (approved):

src/django_components/app_settings.py
JuroOravec: Good to know! Since I'm now thinking also about the Vue port as a standalone package that'd be build on top of django_components (let's call it django_vue), potentially the type validation could eventually live there 👀

As I'm having a look at the code, if there was the plugin support, and it would allow to hook into the flow both BEFORE and AFTER we call `get_context_data()`, then the type validation could be moved out of django_components and into django_vue.

I think there was one or two things that I wanted to experiment with that would require the use of pydantic elsewhere, but I don't remember right now what it was.

---

Although that makes me wonder, whether it makes sense, without type validation, to keep the Component generics. (TL;DR: yes, for intellisense / mypy support)

What I have in mind is this construct:

```py
Component[ArgsType, KwargsType, SlotsType, DataType, JsDataType, CssDataType]
```

My intention for the component generics was three-fold:
1. Runtime input validation
2. Static input validation (mypy) / intellisense, when using `Component.render(...)`
3. As source of truth for language server, so there could be intellisense for the django templates when using `{% component %}`.

However, with django_vue in the equation, then:
- Use case 1. would be moved to django_vue
- If I get to writing a language server (use case 3. ), then instead of writing one for Django templates, I would write one for Vue files, extending it so it supports a `<python>` (or similar) section to define the Python component.

But that means that for the use case 2., it still makes sense for the Component generics to live in django_components.

PR #734, opened 2024-11-04
refactor: fix reload-on-change logic, expose autodiscover's dirs-getting logic, rename settings

JuroOravec:
This MR is a mix of a couple of changes, it grew a bit out of scope, sorry about that.

Initially I wanted to only rename the settings:
- `forbidden_static_files` -> `static_files_forbidden`, so it's aligned with `static_files_allowed`
- `reload_on_template_change` -> `reload_on_file_change`, because using the word "template" suggests that this would apply only to HTML files.

Similarly to the last MR, the old names of the settings are kept, but tagged as deprecated, and would be removed in v1.

Moreover, the reload-on-change feature (`reload_on_template_change`) wasn't working, so I wanted to fix that. But for that I needed to obtain the same list of directories as we use in `autodiscover()`.

Moreover, for django_vue, I also need to access the same list of directories, so I can search them for `.vue` files.

So what this resulted into is that I extracted the logic that was inside `template_loader.py` and `autodiscovery.py`, and moved it to `util/loader.py`. And I've added 2 functions to the public API:
- `get_component_dirs()` - This returns the same directories as used by `autodiscover()`
- `get_component_files(suffix)` - This searches the directories from `get_component_dirs()` for files with a given suffix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,9 @@
   )
   ```
 
+- Use `get_component_dirs()` and `get_component_files()` to get the same list of dirs / files that would be imported by `autodiscover()`, but without actually
+importing them.
+
 #### Fix
 
 - Slots defined with `{% fill %}` tags are now properly accessible via `self.input.slots` in `get_context_data()`
@@ -85,6 +88,17 @@
   - `RegistrySettings.TAG_FORMATTER` -> `RegistrySettings.tag_formatter`
 
   The old uppercase settings `CONTEXT_BEHAVIOR` and `TAG_FORMATTER` are deprecated and will be removed in v1.
+
+- The setting `reload_on_template_change` was renamed to
+  [`reload_on_file_change`](../settings#django_components.app_settings.ComponentsSettings#reload_on_file_change).
+  And now it triggers server reload when any file in `BASE_DIR` changes. The old name `reload_on_template_change`
+  is deprecated and will be removed in v1.
+
+- The setting `forbidden_static_files` was renamed to
+  [`static_files_forbidden`](../settings#django_components.app_settings.ComponentsSettings#static_files_forbidden)
+  to align with [`static_files_allowed`](../settings#django_components.app_settings.ComponentsSettings#static_files_allowed)
+  The old name `forbidden_static_files` is deprecated and will be removed in v1.
+
 
 ## 🚨📢 v0.100
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,7 +91,7 @@ importing them.
 
 - The setting `reload_on_template_change` was renamed to
   [`reload_on_file_change`](../settings#django_components.app_settings.ComponentsSettings#reload_on_file_change).
-  And now it triggers server reload when any file in `BASE_DIR` changes. The old name `reload_on_template_change`
+  And now it properly triggers server reload when any file in the component dirs change. The old name `reload_on_template_change`
   is deprecated and will be removed in v1.
 
 - The setting `forbidden_static_files` was renamed to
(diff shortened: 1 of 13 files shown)


src/django_components/autodiscovery.py
```diff
@@ -30,56 +23,19 @@ def autodiscover(
...
+    To get the same list of modules that `autodiscover()` would return, but without importing them, use
+    [`get_component_files()`](../api#django_components.get_component_files):
+
+    ```python
+    from django_components import get_component_files
+
+    modules = get_component_files(".py")
+    ```
     """
-    dirs = get_component_dirs(include_apps=False)
```
JuroOravec, line 34: This logic was refactored into `get_component_files`

src/django_components/template_loader.py
```diff
@@ -32,72 +25,4 @@ def get_dirs(self, include_apps: bool = True) -> List[Path]:
 
         `BASE_DIR` setting is required.
         """
-        # Allow to configure from settings which dirs should be checked for components
```
JuroOravec, line 35: This logic was refatored into `get_component_dirs`

tests/test_autodiscover.py
```diff
@@ -116,50 +115,3 @@ def test_import_libraries_map_modules(self):
         self.assertIn("multi_file_component", all_components)
 
         settings.COMPONENTS["libraries"] = []
-
-
-class TestFilepathToPythonModule(_TestCase):
```
JuroOravec, line 121: These tests were moved to `test_loader.py`

src/django_components/apps.py
```diff
@@ -49,3 +47,19 @@ def ready(self) -> None:
...
         registry.register(app_settings.DYNAMIC_COMPONENT_NAME, DynamicComponent)
+
+
+# See https://github.com/EmilStenstrom/django-components/issues/586#issue-2472678136
+def _watch_component_files_for_autoreload() -> None:
+    from django_components.util.loader import get_component_dirs
+
+    component_dirs = set(get_component_dirs())
+
+    def template_changed(sender: Any, file_path: Path, **kwargs: Any) -> None:
```
JuroOravec, line 58: This function was moved from `util/misc.py`. Also, the logic itself changed. Before, I tried to use the `autoreload_started` signal, and `sender.extra_files.add()` to watch the files.

Now, I connect to the `file_changed` signal, and use `trigger_reload()`.

I think I've seen this in either Django or Tetra codebase.

CHANGELOG.md
```diff
@@ -86,6 +89,17 @@
 
   The old uppercase settings `CONTEXT_BEHAVIOR` and `TAG_FORMATTER` are deprecated and will be removed in v1.
 
+- The setting `reload_on_template_change` was renamed to
+  [`reload_on_file_change`](../settings#django_components.app_settings.ComponentsSettings#reload_on_file_change).
+  And now it triggers server reload when any file in `BASE_DIR` changes. The old name `reload_on_template_change`
```
EmilStenstrom: You mean "when any **component** file changes" here right?

EmilStenstrom (approved):
One small doc fix (I think), otherwise #shipit!

CHANGELOG.md
JuroOravec: Looks like you were looking at an older commit, the latest wording is "...when **any file in the component dirs** change".

At first, before I decided to extract `get_component_dirs()` and `get_component_files()` out of `autodiscover()` and `TemplateLoader`,  I tried implementing it by watching all files in `BASE_DIR`.

I then later refactored it, so only the files within `COMPONENTS.dirs` and `COMPONENTS.app_dirs` are watched, but forgot to update the changelog.

EmilStenstrom: Yes, my fault.

PR #742, opened 2024-11-06
Improve first example the users sees

EmilStenstrom:
This PR improves the first example the user sees to be easier to understand. I removed the references to JS/CSS to keep things simple.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,9 @@ And this is what gets rendered:
 <div class="calendar-component">Today's date is <span>2024-11-06</span></div>
 ```
 
-Read on to learn about the details!
+Read on to learn about all the exciting details and configuration possibilities!
+
+(If you instead prefer to jump right into the code, [check out the example project](https://github.com/EmilStenstrom/django-components/tree/master/sampleproject))
 
 ## Table of Contents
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,21 +23,39 @@ Potential benefits:
...
 
-And this is what gets rendered (plus the CSS and Javascript you've specified):
+And this is what gets rendered:
 
 ```html
-<div class="calendar-component">Today's date is <span>2015-06-19</span></div>
+<div class="calendar-component">Today's date is <span>2024-11-06</span></div>
 ```
 
-[See the example project](https://github.com/EmilStenstrom/django-components/tree/master/sampleproject) or read on to learn about the details!
```
JuroOravec, line 40: I'd keep the link to the example project, because this is the only prominent place where it's mentioned.

EmilStenstrom: Yeah, but where? I don't think it should be part of the quickstart, it's way to hard to differentiate the stuff in there when you hardly know what the projects is about. I think it's more of an advanced resource for config issues...

JuroOravec: What do you think of this?

```markdown
Read on to learn about the details!

If you instead prefer to jump right into the code, [check out the example project](...)
```

I agree that for someone who comes across django-components for the first time, they wouldn't find the example useful. But I also think that as soon as they start the installation steps, or then when they're learning how to define components, or later to remind themselves of what the correct syntax of something is... In all those cases they might want to check how things are implemented in an actual (demo) project. And since it's something that they might repeatedly come back to, that's why I feel like it should be somewhere high up in an obvious place.

PR #758, opened 2024-11-18
feat: passthrough slots and more

JuroOravec:
~NOTE: I haven't updated the README / docstrings yet. There's a lot going on, so I want to run that with you @EmilStenstrom first. Plus writing out the comments in this MR also helped me to figure out what info needs to be communicated in the README.~

README and CHANGELOG have been updated too, and this is now ready for review.

---

## Summary

All-in-all, the changes in this MR make the slot and fill resolution to be done dynamically
at render time, as opposed to statically walking down the `Template` / NodeList definition
at compile time.

So we let Django's templating to handle everything, and only register a `FillNode` or a `SlotNode`
when we reach their `render()` functions.

This considerably simplifies the logic around fills and slots, while making it also more permissive.
E.g. we can now place `{% fill %}` or `{% slot %}` tags inside for loops, and users can use their
own custom tags too.

The only tradeoff is that we no longer raise an error on unused fills, because we no longer have
the knowledge of ALL slots in the template. We know only about those slots that WERE rendered.

## Problem

All the changes in this MR stem from the need for 2 features that are present in Vue:
1. Accessing slot data on default slot
2. "Passthrough slots"

And a bug when passing forloop variables to `{% slot %}`.

### 1. Accessing slot data on default slot.

```vue
<MyTable>
    <template #default="{ item }">
        {{ item }}
    </template>
</MyTable>
```

### 2. Support for "passthrough slots" - dynamically passing all (or subset) slots to a child component.

```vue
<ChildComponent>
    <template v-for="slotName in $slots">
        <template #[slotName]="slotProps">
            <slot :name="slotName" v-bind="slotProps" />
        </template>
    </template>
</ChildComponent>
```

The passthrough slots feature is actually composed of multiple smaller features:

1. Allowing for-loops inside `{% component %} .. {% endcomponent %}` to build up the `{% fill %}` tags dynamically

2. Resolving the `{% fill %}` tags dynamically by capturing the fill INSTANCES instead of NODES.
    - NOTE: Distinction between _node_ and _instance_
      
      When we have something like
      ```django
      {% for name in list %}
        {% fill name=name / %}
      {% endfor %}
      ```

      Then the template contains 1 `ForNode` and 1 `FillNode`. HOWEVER, this will call `FillNode.render()`
      as many times as is the size of `list`.

      Each time a `FillNode.render()` is called, I call that a separate _INSTANCE_.
    
    - Previously to find the fills, we looked inside the `{% component %} .. {% endcomponent %}`,
      and statically searched for all `FillNodes`. However, since now there can be a FillNode inside for loops,
      the static search doesn't account for that.

      I achieved that by evaluating the body inside `{% component %} .. {% endcomponent %}`
      as `NodeList.render(outer_context)`, and adding logic that "registers" a fill when the rendering
      reaches `FillNode.render()`.

3. Resolving the `{% slot %}` tags dynamically
    - In case of `SlotNodes` it was a similar case as to resolving fills. Previously we statically searched
      the `Component.template` to find any `{% slot %}` tags (AKA `SlotNodes`).

      While the slot name was already allowed to be a variable, we still used NODES instead of INSTANCES.
      So putting `{% slot %}` inside a for loop didn't work:

      ```django
      {% for slot_name in slots %}
        {% slot name=slot_name / %}
      {% endfor %}
      ```

    - I resolved this similarly to how fills were handled - Instead of collecting the `{% slot %}` tags
      ahead of the time, we now just render the template. And only when / if we come across a `{% slot %}`
      tag, then only at that point we:
      1. Resolve the slot name (thus taking variables from `{% for %}` or `{% with %}` into the account)
      2. Check that we're inside a `Component.template`
      3. Check whether a fill for this slot was provided, and render the fill if so.

### 3. Bug when passing forloop vars to `{% slot %}`

This refactor also fixes following error I encountered (below). The problem here was that
when I tried to access the slot's variable `item` from within the `{% fill %}` tag, it was
undefined (empty string).

It was undefined because the variable `item` was generated dynamically in the loop, and the slot
was not able to resolve it, because it was parsed statically.

This is the template of the `list` component:

```django
{# list.html #}
<ul role="list" {% html_attrs attrs class="flex flex-col gap-4" %}>
    {% for item in items %}
        <li {% html_attrs item.attrs item_attrs class="group flex justify-between gap-x-6 border border-gray-300 pl-4 pr-6 bg-white" %}>
            {% if item_handle_class %}
                {% slot "handle" item=item %}
                    {% component "icon"
                        icon='outline_arrows_pointing_out'
                        size=18
                        color="text-gray-500 hover:text-gray-400"
                        svg_attrs:class="inline"
                        text_attrs:class="p-0"
                        attrs:class=item_handle_class|add:" invisible group-hover:visible self-center cursor-pointer rotate-45"
                    / %}
                {% endslot %}
            {% endif %}
        </li>
    {% endfor %}
</ul>
```

The template of the `list` component above can be distilled to having this inside a component template:

```django
{% for item in items %}
    {% slot "handle "item=item / %}
{% endfor %}
```

When I then used the component and tried to access the slot data, the `data.item` variable was undefine, which was the bug:

```django
{# index.html #}
{% component "list" %}
    {% fill "handle" data="data" %}
        {{ data.item }}
    {% endfill %}
{% endcomponent %}
```

## Changes

~1. Allow to define a `{% fill %}` tag without a name to be able to access the slot's data / default content
   on default slots:~

1. Use `{% fill "default" %}` tag to be able to access the slot's data / default content
   on default slots:

   ```django
   {% component "my_table" %}
     {% fill "default" data="slot_data" default="slot_default" %}
       There is {{ slot_data.headers|len }} columns.
       <br />
       {{ slot_default }}
     {% endfill %}
   {% endcomponent %}
   ```

2. Instead of a private namespace, the default fill is now stored under `slots.default`.

   When a slot is marked as `default`, it will render the fill named `"default"`,
   if there is such fill. If not, we use the fill with slot's original name (`header`).

    ```django
    {% slot "header" default / %}
    ```

    Below, all 3 cases will render the content in the same place:

    ```django
   {% component "my_table" %}
       1
   {% endcomponent %}
   ```

    ```django
   {% component "my_table" %}
     {% fill name="default" %}
       1
     {% endfill %}
   {% endcomponent %}
   ```

    ```django
   {% component "my_table" %}
     {% fill name="header" %}
       1
     {% endfill %}
   {% endcomponent %}
   ```

   HOWEVER, double-filling a slot raises an error:

   ```django
   {% component "my_table" %}
     {% fill name="default" %}
       1
     {% endfill %}
     {% fill name="header" %}
       2
     {% endfill %}
   {% endcomponent %}
   ```

3. Allow the slot fills from `Component.input.slots` to be passed through the Django templates
   - Achieved by wrapping it in an object and setting `do_not_call_in_templates = True` on that object.
     See https://docs.djangoproject.com/en/5.1/ref/templates/api/#variables-and-lookups

6. We no longer raise error on extra fills.

7. The body of `{% component %} ... {% endcomponent %}` is now less restrictive. Before, if we wanted to use `{% fill %}`,
   inside of a body, it had to be placed directly underneath:

   ```django
   {% component "my_table" %}
     {% fill "a" %}
       Hello from slot A
     {% endfill %}
     {% fill "b" %}
       Hello from slot B
     {% endfill %}
   {% endcomponent %}
   ```

   Now, the body can contain also any other tags, and the `{% fill %}` tags can be nested,
   as long as those tags don't generate any text.

   So all of this is fine, as long as they rendered tags do not leave any text in their place:
   ```django
   {% component "my_table" %}
     {% expr "[1, 2, 3]" as my_list %}
     {% with alpha="a" beta="b" %}
        {% for item in my_list %}
          {# Custom tags are fine as long as they do not output any content #}
          {% custom_tag alpha item as result %}

          {% fill name=result %}
            Hello from slot {{ result }}
          {% endfill %}

          {% include result|add:item %}

          {# Even other components! As long as they do not output any content #}
          {% component "other" / %}
        {% endfor %}
     {% endwith %}
   {% endcomponent %}
   ```

   But this would NOT be fine, because it contains both `{% fill %}` AND outputs some text:

   ```django
   {% component "my_table" %}
     Some text

     {% lorem w 3 %}

     {% component "icon" / %}

     {% fill name=result %}
       Hello from slot {{ result }}
     {% endfill %}
   {% endcomponent %}
   ```

   But if we removed the `{% fill %}`, so the body does not contain ANY `{% fill %}` tags,
   it's all alright again, and this will be interpreted as a default fill.

   ```django
   {% component "my_table" %}
     Some text

     {% lorem w 3 %}

     {% component "icon" / %}
   {% endcomponent %}
   ```


src/django_components/component.py
```diff
@@ -96,15 +95,14 @@ class RenderInput(Generic[ArgsType, KwargsType, SlotsType]):
     args: ArgsType
     kwargs: KwargsType
     slots: SlotsType
-    escape_slots_content: bool
```
JuroOravec, line 99: `escape_slots_content` can be passed to `Component.render()`, so that the results from user-supplied slots will be sanitised if not SafeText.

The way it is processed, though, is that if `escape_slots_content` is True, we pre-process the slot functions such that their output will be serialized.

So keeping the record of what value of `escape_slots_content` was passed to the component IMO doesn't provide any value. I can't imagine how I would use it. That's why I removed it.

src/django_components/slots.py
```diff
@@ -62,45 +60,52 @@ class SlotFunc(Protocol, Generic[TSlotData]):
...
 
-# Internal type aliases
-SlotId = str
-SlotName = str
 
+@dataclass
+class Slot(Generic[TSlotData]):
+    """This class holds the slot content function along with related metadata."""
 
-class Slot(NamedTuple):
```
JuroOravec, line 72: NOTE: The old and new `Slot` classes share only the name, but they are used for different purposes.

src/django_components/component.py
```diff
@@ -630,17 +628,21 @@ def _render_impl(
...
     ) -> str:
+        # NOTE: We must run validation before we normalize the slots, because the normalization
+        #       wraps them in functions.
+        self._validate_inputs(args or (), kwargs or {}, slots or {})
+
         # Allow to provide no args/kwargs/slots/context
         args = cast(ArgsType, args or ())
         kwargs = cast(KwargsType, kwargs or {})
-        slots = cast(SlotsType, slots or {})
+        slots_untyped = self._normalize_slot_fills(slots or {}, escape_slots_content)
```
JuroOravec, line 638: The slot fills are normalize in two steps:
1. They are normalized to functions. So if user supplies a plain string via calling 
`Component.render(slots={...})`, then we convert it to dummy function that returns the said function.
2. Slot fill functions are then wrapped in a `Slot` instance.

In this MR I added the step 2.

The `Slot` class behaves as a function on the surface (calling the associated slot fill function), so interface doesn't change. But it allows us to associate extra metadata with those slot fill functions.

Moreover, `Slot` class is used to make it possible to pass the slot fill functions as variables within the Django template.
   - The `Slot` class defines a `do_not_call_in_templates = True` property, which tells Django to NOT automatically call the function

See the definition here
https://github.com/EmilStenstrom/django-components/pull/758/files#diff-c7f4341cbd58fb82a995fdba37703267d2dbbdea83ac86c7cb2a67c258fd835cR67

src/django_components/component.py
```diff
@@ -670,39 +669,37 @@ def _render_impl(
         cache_inlined_css(self.__class__, self.css or "")
 
         with _prepare_template(self, context, context_data) as template:
-            fills = self._escape_slot_fills(slots, escape_slots_content)
+            # For users, we expose boolean variables that they may check
+            # to see if given slot was filled, e.g.:
+            # `{% if variable > 8 and component_vars.is_filled.header %}`
+            is_filled = SlotIsFilled(slots_untyped)
```
JuroOravec, line 675: Since all slots and fills are now resolved at runtime (we no longer statically analyse the Django Template for Nodes), then it's no longer possible to define all available slots. Instead, we only know about the fills that the user provided.

So if we rendered a component like so:

```django
{% component "my_table" %}
  {% fill "header" %}
    Hello!
  {% endfill %}
{% endcomponent %}
```

Then, BEFORE,  `is_filled` dictionary would have been something like this:

```py
{
    "header": True,
    "main": False,
    "footer": False,
}
```

But NOW, the `is_filled` dictionary does NOT have the falsy values:

```py
{
    "header": True,
}
```

So, to keep the behavior consistent, I added a `SlotIsFilled` class, which behaves like a regular dictionary, but returns False on missing key.

So the difference is minor, but what was previously an invalid key (and coerced by Django into empty string) will now return False:

Before:
```django
{{ component_vars.is_filled.header }} -> True
{{ component_vars.is_filled.footer }} -> False
{{ component_vars.is_filled.nonexist }} -> "" (invalid key)
```

After:
```django
{{ component_vars.is_filled.header }} -> True
{{ component_vars.is_filled.footer }} -> False
{{ component_vars.is_filled.nonexist }} -> False
```

src/django_components/component.py
```diff
@@ -670,39 +669,37 @@ def _render_impl(
...
+            self._render_stack[-1].is_filled = is_filled
 
-            _, resolved_fills = resolve_slots(
-                context,
-                template,
+            component_slot_ctx = ComponentSlotContext(
```
JuroOravec, line 678: How the slot rendering works now is that:

1. When we reach `Component.render()`, we set
    ```py
    context[_COMPONENT_SLOT_CTX_CONTEXT_KEY] = component_slot_ctx
    ```
   which holds data that slots need from the component (e.g. component name or available slot fills).
2. Then, as we're rendering the Component's content, we may come across `{% slot %}` tags (AKA `SlotNodes`).
3. Inside `SlotNode.render()`, we check if the `_COMPONENT_SLOT_CTX_CONTEXT_KEY` is accessible.
4. If it is, and there is a fill with the same name as the slot name, we render that fill in place of the slot.
5. If no fill matches, we render the slot's default content.

src/django_components/component.py
```diff
@@ -836,19 +833,18 @@ class MyComp(Component):
         self._types = args_type, kwargs_type, slots_type, data_type, js_data_type, css_data_type
         return self._types
 
-    def _validate_inputs(self) -> None:
-
+    def _validate_inputs(self, args: Tuple, kwargs: Any, slots: Any) -> None:
```
JuroOravec, line 836: Passed in the values of args, kwargs, & slots, directly into the function, so it can be called before we populate `Component.input`.

src/django_components/component.py
```diff
@@ -897,18 +892,7 @@ def render(self, context: Context) -> str:
...
-                DEFAULT_SLOT_KEY: _nodelist_to_slot_render_func(
-                    "<default>",
-                    self.fill_nodes[0].nodelist,
-                    data_var=None,
-                    default_var=None,
-                ),
-            }
-        else:
-            slots = resolve_fill_nodes(context, self.fill_nodes, self.name)
+        slot_fills = resolve_fills(context, self.nodelist, self.name)
```
JuroOravec, line 895: The logic that extracts `{% fill %}` tags from in between the `{% component %} {% endcomponent %}` tags was also reworked, and the logic that assigns the default fill is now inside `resolve_fills()`

src/docs/devguides/slot_rendering.md
```diff
@@ -75,134 +75,149 @@ This doc serves as a primer on how component slots and fills are resolved.
    | -------
    ```
 
-5. I want to render the slots with `{% fill %}` tag that were defined OUTSIDE of this template. How do I do that?
+6. The names of the slots and fills may be defined using variables
```
JuroOravec, line 78: Updated the guide to describe the new implementation

src/django_components/templatetags/component_tags.py
```diff
@@ -427,6 +429,20 @@ def fill(parser: Parser, token: Token, tag_spec: TagSpec) -> FillNode:
...
     ```
+
+    ### Accessing slot data and default content on the default slot
+
+    To access slot data and the default slot content on the default slot,
+    use `{% fill %}` without the `name` kwarg:
+
+    ```django
+    {% component "button" %}
+      {% fill data="slot_data" default="default_slot" %}
```
JuroOravec, line 440: For django-vue, beside the "passthrough slots" feature, I also needed a way to use the `data` and `default` kwargs on a default fill.

E.g. in Vue you can do `#slot_name="{ item }"` to access the slot data of the slot `slot_name`

  ```vue
  <MyTable>
      <template #default="{ item }">
          {{ item }}
      </template>
  </MyTable>
  ```

Because in Vue, when you pass in a default fill, it's as if the fill was assigned to a fill named "default".

So in Vue, these two are identical:

  ```vue
  <MyTable>
          {{ item }}
  </MyTable>

  <MyTable>
      <template #default>
          {{ item }}
      </template>
  </MyTable>
  ```

---

Now, in my head, the point of the default fill is that you don't know / care what is the slot's name. So my initial approach for accessing slot data of a default fill in django-components was to simply omit the `name=...` kwarg:

```django
{% component "button" %}
  {% fill data="slot_data" default="default_slot" %}
    You clicked me {{ slot_data.count }} times!
    {{ default_slot }}
  {% endfill %}
{% endcomponent %}
```

However, later, as I had to redo the slot implementation, I realized that omitting the name would not work nicely with the "passthrough slots" feature. Because the default slot was stored under an internal key. So we'd either allow something like this to target the default slot:
```django
{% fill name="_DJANGO_COMPONENTS_DEFAULT_SLOT" %}
```
Or for the passthrough slots, there'd have to be an extra if / else:
```django
{% component "slotted" %}
    {% for slot_name in slots %}
      {% if slot_name == "_DJANGO_COMPONENTS_DEFAULT_SLOT" %}
        {% fill %}
            {% slot name=slot_name default / %}
        {% endfill %}
      {% else %}
        {% fill name=slot_name %}
            {% slot name=slot_name / %}
        {% endfill %}
      {% endif %}
    {% endfor %}
{% endcomponent %}
```

So in the end I decided to go with Vue's implementation, which is that:
1. Default fill is assigned to a fill named "default"
2. If a slot is marked as default, it will use the "default" fill.

---

In the end, I kept both approaches (omitting fill's name and using "default" fill, which means that fills can now be written in 3 ways:

1. Fill with regular name
  ```django
  {# This will target the `header` slot, irrespective of whether it's default #}
  {% fill name="header" / %}
  ```
2. Fill with name "default"
```django
  {# This will target the slot with the `default` flag #}
  {% fill name="default" / %}
  ```
3. Fill without a name
```django
  {# This will target the slot with the `default` flag #}
  {% fill / %}
  ```
---

Note, though, that our implementation differs from Vue's. Because in Vue, the default slot is defined by omitting the name:

```vue
<slot />
```

In which case Vue gives this slot a name `default`.

However, our case, we use the `default` flag, but still require some name:

```django
{% slot "my_slot" default / %}
```

Which means that a single slot may be targetted in two ways:

1. By the slot name 
2. As the "default" slot.

E.g. in the case below, both fills target the same slot.

```django
{% component 'test' %}
    {% fill %} ABC {% endfill %}
    {% fill "my_slot" %} XYZ {% endfill %}
{% endcomponent %}
```

In such cases, we raise an error. (See tests like this one https://github.com/EmilStenstrom/django-components/pull/758/files#diff-ea7f91c03e3b214aa30b65d58c3ea193c51868981b298410097f11e6f457fa01R366)

tests/test_templatetags_slot_fill.py
```diff
@@ -481,98 +698,297 @@ def test_component_without_default_slot_refuses_implicit_fill(self):
...
 
-        with self.assertRaises(TemplateSyntaxError):
-            c.render(Context({}))
+            def get_context_data(self, name: Optional[str] = None) -> Dict[str, Any]:
+                return {
+                    "name": name,
+                }
 
-    @parametrize_context_behavior(["django", "isolated"])
-    def test_slot_name_fill_typo_gives_helpful_error_message(self):
```
JuroOravec, line 506: The tradeoff for the changes in this MR is that we can no longer know all slots defined in a component's template ahead of the time. So if a fill with an invalid name is given, we won't raise an error, but simply ignore it.

However, to salvage at least something, when a slot is marked as required, and the corresponding fill is NOT present, then there we still show a helpful message. See https://github.com/EmilStenstrom/django-components/pull/758/files#diff-c7f4341cbd58fb82a995fdba37703267d2dbbdea83ac86c7cb2a67c258fd835cR300

tests/test_templatetags_slot_fill.py
```diff
@@ -481,98 +698,297 @@ def test_component_without_default_slot_refuses_implicit_fill(self):
...
+                </header>
+                <main>Default main</main>
+                <footer>Default footer</footer>
+            </custom-template>
+            """,
+        )
 
-    # NOTE: This is relevant only for the "isolated" mode
-    @parametrize_context_behavior(["isolated"])
-    def test_slots_of_top_level_comps_can_access_full_outer_ctx(self):
```
JuroOravec, line 534: This test was moved to here https://github.com/EmilStenstrom/django-components/pull/758/files#diff-ea7f91c03e3b214aa30b65d58c3ea193c51868981b298410097f11e6f457fa01R263

tests/test_templatetags_slot_fill.py
```diff
@@ -481,98 +698,297 @@ def test_component_without_default_slot_refuses_implicit_fill(self):
...
+            def get_context_data(self, name: Optional[str] = None) -> Dict[str, Any]:
+                return {
+                    "slots": self.input.slots,
+                }
+
+            template: types.django_html = """
+                {% load component_tags %}
+                <div>
+                    {% component "slotted" %}
+                        {% for slot_name in slots %}
```
JuroOravec, line 908: One nice side-effect of this new render-time slot and fill resolution is that we can now allow ANY tags to be inside the `{% component %} {% endcomponent %}` tags, **as long as they don't output any text.**

So users can use for-loops to generate the fills:

```django
{% component %}
  {% for item in items %}
    {% fill name=item %}
      {{ item }}
    {% endfill %}
  {% endfor %}
{% endcomponent %}
```

But also e.g. `{% with %}`:

```django
{% component %}
  {% with abc="xyz" %}
    {% fill name=abc %}
      {{ abc }}
    {% endfill %}
  {% endwith %}
{% endcomponent %}
```

And even their custom tags, or using the `as var` syntax:

```django
{% component %}
  {% expr "['header', 'main', 'footer']" as slots %}
  {% for slot in slots %}
    {% fill name=slot %}
      {{ slot }}
    {% endfill %}
  {% endfor %}
{% endcomponent %}
```

src/django_components/slots.py
```diff
@@ -122,6 +127,63 @@ def __str__(self) -> str:
...
+
+@dataclass
+class ComponentSlotContext:
+    component_name: str
+    template_name: str
+    is_dynamic_component: bool
+    default_slot: Optional[str]
+    fills: Dict[SlotName, Slot]
+
+    def post_render_validation(self) -> None:
```
JuroOravec, line 151: This method runs after the component's template is done rendering. It's only at this point that we have the knowledge of all slots that were rendered, and we can potentially do more validation based on that.

src/django_components/slots.py
```diff
@@ -152,14 +214,120 @@ def __repr__(self) -> str:
...
+                "Encountered a SlotNode outside of a ComponentNode context. "
+                "Make sure that all {% slot %} tags are nested within {% component %} tags.\n"
+                f"SlotNode: {self.__repr__()}"
+            )
+
+        component_ctx: ComponentSlotContext = context[_COMPONENT_SLOT_CTX_CONTEXT_KEY]
+        slot_name, kwargs = self.resolve_kwargs(context, component_ctx.component_name)
+
+        # Check for errors
+        if self.is_default and not component_ctx.is_dynamic_component:
```
JuroOravec, line 229: The `default` and `required` flags of the `{% slot %}` tag are now handled at render time

src/django_components/slots.py
```diff
@@ -227,75 +395,181 @@ def __init__(
...
-        is_implicit: bool = False,
     ):
         super().__init__(nodelist=nodelist, args=None, kwargs=kwargs, node_id=node_id)
 
-        self.is_implicit = is_implicit
         self.trace_id = trace_id
-        self.component_id: Optional[str] = None
 
     def render(self, context: Context) -> str:
+        if self._is_extracting_fill(context):
```
JuroOravec, line 404: Similarly to slots, also fills are resolve at render-time.

1. We render the content between `{% component %} .. {% endcomponent %}` as `NodeList.render()`.
2. We also set a special variable `FILL_GEN_CONTEXT_KEY` to the context
3. When we come across a `{% fill %}` tag, we enter `FillNode.render()`. Here, we check if we're in the "fill collection mode" by checking for the special variable.
4. If so, we extract the data like the fill name, and push it to a list of fills that's available under `FILL_GEN_CONTEXT_KEY`.
5. Once we're done rendering, the list that was set under `FILL_GEN_CONTEXT_KEY` is now populated with the fills that we've discovered during rendering.

src/django_components/slots.py
```diff
@@ -227,75 +395,181 @@ def __init__(
...
+        # tag, so one can access the `data` and `default` vars.
+        # E.g.
+        # ```django
+        # {% component "mytable" %}
+        #   {% fill default="default" data="data" %}
+        #     {{ data.some_var }}
+        #   {% endfill %}
+        # {% endcomponent %}
+        # ```
+        if name is None:
```
JuroOravec, line 433: NOTE: If we decided to roll back the feature of allowing to specify the default fill by omitting the `name` kwarg, then we'd also remove this if branch

JuroOravec:
@EmilStenstrom Let me know your thoughts. Sorry about the size, what initially seemed like a simple change grew out of proportions

CHANGELOG.md
```diff
@@ -2,24 +2,15 @@
 
 ## 🚨📢 v0.110
 
-#### BREAKING CHANGES
+### General
```
JuroOravec, line 5: I've grouped the changes by themes so it's easier to skim

CHANGELOG.md
```diff
@@ -99,6 +79,254 @@ importing them.
...
+
+  ```django
+  {% for slot_name in slots %}
+    {% slot name=slot_name %}
+  {% endfor %}
+  ```
+
+#### Refactor
+
+- When you define multiple slots with the same name inside a template,
```
JuroOravec, line 214: As I was going through the README, I realized there's one more change of behavior - when we have multiple `{% slot %}` tags with the same name, the `default` and `required` flag are now applied only to the individual tags

src/django_components/templatetags/component_tags.py
JuroOravec: UPDATE: So I reverted the "`{% fill %}` without a name is a default slot" feature in https://github.com/EmilStenstrom/django-components/pull/758/commits/072c4f45659fecdab228f99c43dd540dba0fa22c. It was just simpler to document the changes while always using `name="default"`.

JuroOravec:
UPDATE: So I reverted the "`{% fill %}` without a name is a default slot" feature in https://github.com/EmilStenstrom/django-components/pull/758/commits/072c4f45659fecdab228f99c43dd540dba0fa22c. It was just simpler to document the changes while always using `name="default"`.

JuroOravec:
Closes  https://github.com/EmilStenstrom/django-components/issues/763

src/django_components/slots.py
```diff
@@ -152,14 +216,120 @@ def __repr__(self) -> str:
...
+                    default_var=None,
+                ),
+            )
+
+        # Check: If a slot is marked as 'required', it must be filled.
+        #
+        # To help with easy-to-overlook typos, we fuzzy match unresolvable fills to
+        # those slots for which no matching fill was encountered. In the event of
+        # a close match, we include the name of the matched unfilled slot as a
+        # hint in the error message.
```
EmilStenstrom, line 290: Gold star! ⭐

JuroOravec:
@EmilStenstrom Appreciate that!

Issue #752, opened 2024-11-11
Bug: Incorrect parsing of apostrophes


MartinBoge:
Hi Emil

First off - thank you very much for this sweet sweet library. If this issue is redundant or you find it irrelevant just go ahead and close it.

I currently have a card header component that i use like so:

```
{% component "card_header" ...other code... tooltip_text=_("Responsible for the organisation's data protection compliance") / %}
```

If the tag is not closing via the `/ %}` syntax like so:

```
{% component "card_header" ...other code... tooltip_text=_("Responsible for the organisation's data protection compliance") %} {% endcomponent %}
```

Then i actually get the following error:
IndexError: pop from empty list

I have narrowed it down to the fact that i have an apostrophe ( ' ) in my text and causes and issue when the template is being parsed. That means i do NOT get the error if i:

1. Remove the apostrophe from my str variable.
2. Let the tag be self-closing.

Is this a known issue?

EmilStenstrom:
Hi Martin, it sounds like a bug that should be fixed. 

@JuroOravec Mind helping me understand what the right syntax should be here? `tooltip_text="{% trans "Responsible..." %}"`?

JuroOravec:
@MartinBoge Do you have access to the full stack trace? Could you post it here?

Martin's example SHOULD work, so this is indeed a bug.

Btw, @MartinBoge, so if you omit the single apostrophe, then you are able to pass the text like so?

```django
{% component "card_header" tooltip_text=_("Responsible for the organisations data protection compliance") / %}
```

I haven't worked with Django's translation / internationalisation features, so personally I never tried passing the `key=_("some text")` syntax through components. I wonder if that has something to do with it?

@MartinBoge could you check which of the following syntaxes work correctly? The one you mentioned above is case 7. below. I expect that cases 1-4 SHOULD work, but no idea how 5-8 behave.

1. Without `_()`, without single apostrophe, self-closing:
	```django
	{% component "card_header" tooltip_text="organisations" / %}
	```

2. Without `_()`, without single apostrophe, end-tag:
	```django
	{% component "card_header" tooltip_text="organisations" %} {% endcomponent %}
	```

3. Without `_()`, with single apostrophe, self-closing:
	```django
	{% component "card_header" tooltip_text="organisation's" / %}
	```

4. Without `_()`, with single apostrophe, end-tag:
	```django
	{% component "card_header" tooltip_text="organisation's" %} {% endcomponent %}
	```

[...]

PR #765, opened 2024-11-23
fix: apostrophes in tags

JuroOravec:
## Background

To support the use of Django template tags inside component inputs, e.g.

```django
{% component
    display_name="{{ first_name }} {{ last_name }}"
    description="{% lorem w 3 %}"
/ %}
```

We have to fix Django's tokenization, because Django simply cuts up the template based on the first found `{% ... %}` pairs. So, in the example above, even though the `{% lorem w 3 %}` is wrapped by quotes, Django wrongly interprets the `{% component %}` tag to be:

```django
{% component
    display_name="{{ first_name }} {{ last_name }}"
    description="{% lorem w 3 %}"
```

## Problem

When Django cuts up a tag incorrectly like above, we can identify it because there's a missing closing quote (`"`).

But the fix I implemented previously is too simplistic, because it just checks if there's an even or odd number of quotes in the tag.

But in https://github.com/EmilStenstrom/django-components/issues/752, we got an example where there's a apostrophe in one of the strings passed to the component, which throws off the quote-counting. With quote-counting, we cannot distinguish whether an extra quote is part of a string or not.

## Solution

So instead here I implemented a proper solution - a low-level parser that processes the tag character-by-character. Using this parser we can distinguish and ignore extra apostrophes that are part of strings. And we also get to know if there is or is not an unclosed quote at the end.

- NOTE: While parsing the content character-by-character felt as an overkill at first. But I don't know how else it could be approached. And, yesterday, I found that I also need this parser for the plugin feature - to allow plugins to pre-process component inputs. So IMO the parser is justified. 

- NOTE 2: This may still leave some obscure edge cases. But to prevent them, we'd have to do the tokenization of the whole template ourselves. So this is good enough for now.

Fixes https://github.com/EmilStenstrom/django-components/issues/752

### DIFF
--- a/tests/test_templatetags.py
+++ b/tests/test_templatetags.py
@@ -516,26 +516,83 @@ class MultilineTagsTests(BaseTestCase):
 
 
 class NestedTagsTests(BaseTestCase):
+    class SimpleComponent(Component):
+        template: types.django_html = """
+            Variable: <strong>{{ var }}</strong>
+        """
+
+        def get_context_data(self, var):
+            return {
+                "var": var,
+            }
+
     # See https://github.com/EmilStenstrom/django-components/discussions/671
     @parametrize_context_behavior(["django", "isolated"])
     def test_nested_tags(self):
-        @register("test_component")
-        class SimpleComponent(Component):
-            template: types.django_html = """
-                Variable: <strong>{{ var }}</strong>
-            """
-
-            def get_context_data(self, var):
-                return {
-                    "var": var,
-                }
+        registry.register("test", self.SimpleComponent)
 
         template: types.django_html = """
             {% load component_tags %}
-            {% component "test_component" var="{% lorem 1 w %}" %}{% endcomponent %}
+            {% component "test" var="{% lorem 1 w %}" %}{% endcomponent %}
         """
         rendered = Template(template).render(Context())
         expected = """
             Variable: <strong>lorem</strong>
+        """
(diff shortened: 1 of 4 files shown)


tests/test_templatetags.py
```diff
@@ -516,26 +516,83 @@ def get_context_data(self, variable, variable2="default"):
...
         """
         self.assertHTMLEqual(rendered, expected)
+
+    @parametrize_context_behavior(["django", "isolated"])
+    def test_nested_quote_single(self):
+        registry.register("test", self.SimpleComponent)
+
+        template: types.django_html = """
+            {% load component_tags %}
+            {% component "test" var=_("organisation's") %} {% endcomponent %}
```
JuroOravec, line 550: The tests in this file cover the cases from #752

EmilStenstrom (approved):
Nice fix, feels like a lot more stable too.

JuroOravec:
Btw, just a note, that as I mentioned in the PR description that

> NOTE 2: This may still leave some obscure edge cases. But to prevent them, we'd have to do the tokenization of the whole template ourselves. So this is good enough for now.

Then when it comes to merging django-components into Django codebase, IMO for that we would have to implement the "tokenization of the whole template". It would then replace the regex splitting in [`Lexer.tokenize()`](https://github.com/django/django/blob/857b1048d53ebf5fc5581c110e85c212b81ca83a/django/template/base.py#L359) and `DebugLexer.tokenize()`

PR #766, opened 2024-11-23
Update README.md

### LINKED ISSUE #764: component_shorthand_formatter and dynamic components
Are dynamic components supported with the component_shorthand_formatter?
In the docs the only example is with the classical `component` tag.

I've also looked in issues and discussions but didn't find anything.

As a workaround, I'm rendering the component inside the django function like `output = MyComponent.render(kwargs={...})` and then i pass `output` as context and just render it as `{{ output }}`.

Thanks.

vb8448:
Updated documentation for dynamic components.
Close #764 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -1085,6 +1085,20 @@ Instead, you can use **dynamic components**. Dynamic components are used in plac
 {% endcomponent %}
 ```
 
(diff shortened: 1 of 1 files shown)


JuroOravec:
@vb8448 Thanks for the help! :)

PR #778, opened 2024-11-26
fix: do NOT render component tags when parsing {% component %} body for fills

JuroOravec:
Small bug I came across when there is a component nested in another component, whereas the inner one uses `inject()` to use data provided by the outer component.

E.g.

```django
{% component "data_provider" %}
  {%  component "data_user" / %}
{% endcomponent %}
```

This errored, because we tried to render `data_user` component as part of searching for `{% fill %}` tags. But at that point the `data_provider` has not yet provided the data.

So the solution is NOT to render the `{% component %}` tags at that stage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1311,7 +1311,7 @@ tags:
 {% endcomponent %}
 ```
 
-You can also use `{% for %}`, `{% with %}`, or other tags (even `{% include %}`)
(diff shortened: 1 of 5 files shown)


JuroOravec:
@EmilStenstrom Tagging for visibility. I will go ahead and merge and release this if the tests pass, as it's blocking me from updating my work project to v0.110 .

PR #780, opened 2024-11-26
fix: allow components to accept default fill even if no default slot was encountered during rendering

JuroOravec:
Another bug I encountered while trying to update to v0.110.

When I had a following template:

```django
{% if component_vars.is_filled.activator %}
  <div>
    {% slot "activator" default / %}
  </div>
{% endif %}
```

And I used the component WITHOUT the implicit / default fill:

```django
{% component "Menu" /  %}
```

Then I was getting error

```
Component 'Menu' passed default fill content (i.e. without explicit 'name' kwarg), even though none of its slots is marked as 'default'.
```

While this error made sense when slots were defined statically, it doesn't make sense anymore. Because as you can see above in the template, the default slot still IS defined in the template. But it's just that it was not rendered, because the `{% if %}` branch didn't run.

So I removed the error.

### DIFF
--- a/src/django_components/slots.py
+++ b/src/django_components/slots.py
@@ -148,41 +148,6 @@ class ComponentSlotContext:
     default_slot: Optional[str]
     fills: Dict[SlotName, Slot]
 
-    def post_render_validation(self) -> None:
-        if self.is_dynamic_component:
-            return
-
-        default_fill = self.fills.get(DEFAULT_SLOT_KEY, None)
-
-        # Check: Only component templates that include a 'default' slot
-        # can be invoked with implicit filling.
-        if default_fill and not self.default_slot:
-            raise TemplateSyntaxError(
-                f"Component '{self.component_name}' passed default fill content "
-                f"(i.e. without explicit 'name' kwarg), "
-                f"even though none of its slots is marked as 'default'."
-            )
-
-        # NOTE:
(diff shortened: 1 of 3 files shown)


src/django_components/slots.py
```diff
@@ -148,41 +148,6 @@ class ComponentSlotContext:
     default_slot: Optional[str]
     fills: Dict[SlotName, Slot]
 
-    def post_render_validation(self) -> None:
```
JuroOravec, line 151: I removed the whole method, because it has no use anymore.

src/django_components/slots.py
```diff
@@ -214,6 +179,27 @@ def active_flags(self) -> List[str]:
     def __repr__(self) -> str:
         return f"<Slot Node: {self.node_id}. Contents: {repr(self.nodelist)}. Options: {self.active_flags}>"
 
+    # NOTE:
```
JuroOravec, line 182: Moved the note to a new place

JuroOravec:
@EmilStenstrom Another bug. I will again go ahead and merge and release this, as this is blocking me on my work project.

Also, I realized I accidentally broke the documentation, because as I released v0.111, it included the PR from yesterday, which put the documentation into broken state.

So after merging this, I will revert the yesterday's documentation PR, and then release v0.112. So documentation should be work again. And then I will re-apply the yesterday's documentation PR.

PR #783, opened 2024-11-26
fix: ensure consistent order of scripts in Component.Media.js

JuroOravec:
One more bug fix as I was trying to update my project to v0.110. Turns out that Python's Set does NOT preserve order, which was causing JS and CSS defined via `Component.Media.js/css` to change positions.


JuroOravec:
@EmilStenstrom Sorry for spam today. This is a tiny fix, so I will again go ahead and merge and release this.

EmilStenstrom:
@JuroOravec Don't worry about the "spam", you're fixing stuff and making the library better! :)

PR #786, opened 2024-11-27
fix: do not render slot tag when extracting fill tags

JuroOravec:
One more bug found. This bug is practically the same as the one in https://github.com/EmilStenstrom/django-components/pull/778/files, where I had to add following snippet to `ComponentNode`:

```py
if _is_extracting_fill(context):
    return ""
```

Except this time this snippet is needed for `SlotNode`.

Beside that, I also updated the error message when an error occurs while rendering a component. Now it shows the component path like so:

```
KeyError: "An error occured while rendering components ProjectPage > ProjectLayoutTabbed >
Layout > RenderContextProvider > Base > TabItem:
Component 'TabItem' tried to inject a variable '_tab' before it was provided."
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,31 @@
 # Release notes
 
+## v0.114
+
+#### Fix
+
+- Prevent rendering Slot tags during fill discovery stage to fix a case when a component inside a slot
+  fill tried to access provided data too early.
+
+## v0.113
+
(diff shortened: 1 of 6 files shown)


JuroOravec:
@EmilStenstrom Another bug, again tagging for visibility. My work project is working now, so this should be the last of 'em. 🚀

PR #775, opened 2024-11-25
docs: Split README

JuroOravec:
There's still some more things to document and small tweaks and whatnot.

But next in line I want to add support for plugins, so I can unblock my work towards the UI component library and django-vue ([as mentioned here](https://github.com/EmilStenstrom/django-components/discussions/547#discussioncomment-11371335)), so I can work on that in parallel with other django-components features.
- NOTE: I want to get a proof of concept for django-vue before finishing off features likes scoped CSS, JS / CSS vars, etc, in order to avoid having to needlessly refactor it.

But before adding support for plugins, I want to push the documentation a bit, because it's hard to navigate the big README file, and frankly I'm not totally sure where I would insert a documention section on plugins.

So I will do a bit of documentation cleanup now. That will consist of 4 steps:
1. Split README into smaller articles and move those articles into the documentation site.
2. Generate documentation and API reference from docstrings
3. Other changes to make it all work and deploy the docs.
4. Once the docs site was successfully updated and deployed, clean up the README, removing the sections that were migrated in step 1).

At the end of those 4 steps, the documentation site should be much more appealing and easier to follow.

So in this MR I will do the step 1), splitting the README.

I'm merely splitting up the content. Some time later, I want to go over these pages:
1. Fix links + add links to API reference
2. Review the content.

---

Here's overview of the sectios in the current README, and how they will be grouped / nested afterwards:

With the documentation articles categorisation like `overview`, `concepts/fundamentals, and `concepts/advanced`,  I took inspiration from e.g. Vue, where they have ["Essentials" and "In-depth" categories](https://vuejs.org/guide/essentials/application.html), and Pydantic regarding how they have "concepts" and "API documentation" separate.

README chapter | New section
---|---
Release notes | `/changelog`
Security notes | `/overview/security_notes`
Installation | `/overview/installation`
Compatibility | `/overview/compatibility`
Community examples | `/overview/community`
Running django-components project locally | `/overview/development`
Development guides | `/overview/development`
Create your first component | `/concepts/fundamentals/your_first_component`
Using single-file components | `/concepts/fundamentals/single_file_components`
Use components in templates | `/concepts/fundamentals/components_in_templates`
Use components outside of templates | `/concepts/fundamentals/components_in_python`
Use components as views | `/concepts/fundamentals/components_as_views`
Autodiscovery | `/concepts/fundamentals/autodiscovery`
Using slots in templates | `/concepts/fundamentals/slots`
Accessing data passed to the component | `/concepts/fundamentals/access_component_input`
[...]

PR #788, opened 2024-11-27
docs: add script to generate API reference from the source code

JuroOravec:
Scripts to generate API reference from the source code. This is the step 2 of 4, as outlined in https://github.com/EmilStenstrom/django-components/pull/775

Demo screenshots:

Screenshot 2024-11-27 at 20 35 31

Screenshot 2024-11-27 at 20 35 41

Screenshot 2024-11-27 at 20 36 09

Screenshot 2024-11-27 at 20 36 27

Screenshot 2024-11-27 at 20 36 35

Screenshot 2024-11-27 at 20 36 56

Screenshot 2024-11-27 at 20 38 13

Screenshot 2024-11-27 at 20 38 27

Screenshot 2024-11-27 at 20 38 42

Screenshot 2024-11-27 at 20 38 48


src/docs/scripts/extensions.py
```diff
@@ -0,0 +1,61 @@
...
+SOURCE_CODE_GIT_BRANCH = "master"
+
+
+mkdocstrings_config = get_mkdocstrings_plugin_handler_options() or {}
+is_skip_docstring: bool = mkdocstrings_config.get("show_if_no_docstring", "false") != "true"
+
+
+# Replacement for `show_bases: true`. Our implementation takes base names from the runtime
+# objects, effectively resolving aliases / reassignments.
+class RuntimeBasesExtension(griffe.Extension):
```
JuroOravec, line 18: These are extensions for `mkdocstrings`. I don't have a very deep grasp of how they work, I got help from the maintainer of mkdocstrings about how to write the plugins.

One important thing to keep in mind is that `mkdocstrings` / `griffe` goes through the source code and parses the AST to discover what's defined, and use the docstrings for documentation.

That's also the reason why I added this `RuntimeBasesExtension`. Since `mkdocstrings` by default uses only the AST, then if there is a class that subclasses another:

```py
class Dog(Animal):
    ...
```

Then `mkdocstrings` is unable to resolve the reference class `Animal`.

So if the original file actually contained this:
```py
Animal = Rock

class Dog(Animal):
    ...
```

Then the documentation would still show `Animal` as the base for `Dog`, although `Animal` is just an alias for `Rock`.

And instead, this custom `RuntimeBasesExtension` imports the class as a Python object to get the actual name of the base classes, and show those.

---

Note: I might copy the code for the documentation for other projects like the component library, that's why I took the time to get this right. Because there, I use aliases so that the end-user could either just the component type or the component class:

```py
class ADividerProps(TypedDict):
    abc: int

class ADividerSlots(TypedDict):
    default: NotRequired[SlotContent]

ADividerType = Component[EmptyTuple, ADividerProps, ADividerSlots, Any, Any, Any]

class ADivider(ADividerType):
    ...
```

And there, using `show_bases: true` would show `ADividerType` as the base class. But this custom extension would successfully resolve that to `Component`.

src/docs/templates/reference_commands.md
```diff
@@ -0,0 +1,4 @@
+# Commands
```
JuroOravec, line 1: For each reference page, there's a corresponding template in `src/docs/templates`, which is prepended to the generated content.

src/docs/scripts/reference.py
```diff
@@ -1,47 +1,715 @@
-# Adapted from https://github.com/mkdocstrings/python/blob/main/scripts/gen_ref_nav.py
```
JuroOravec, line 1: The old file was practically removed, it has no connection to the new implementation, only in file name, so it was also ok to remove this license comment.

EmilStenstrom (approved):
Reads very thoughtful as usual. Safe to merge IMHO.

Issue #793, opened 2024-11-29
Extra <html><head><body> inserted after updating to v0.110


PouriaMzn:
First of all, thank you for the awesome work on this library!
i just recently update from v0.94 to v0.114 and encountered an issue when rendering partial templates that do not contain any components. These templates are now being altered by the ```ComponentDependencyMiddleware```.

I have a main template that includes multiple components. Within it, I dynamically include a partial template to add rows to a table. Below is an example:

Main Template:
```
<tbody id="items" class="divide-y divide-slate-300">
    {% for form in formset %}
	    {% include "partials/items_list.html" with form=form row_number=forloop.counter %}
    {% endfor %}
</tbody>

<button 
	type="button"
	class="btn btn-sm btn-info"
	hx-get="{{ custom_url }}"
	hx-trigger="click"
	hx-target="#items"
	hx-swap="beforeend"
	hx-vals="{% jsonify form_count=formset|length %}"
	id="add-item-button">
		Add Row
</button>
```

Partial Template (partials/items_list.html):

```
{% load widget_tweaks %}
{% url 'products:search_product_service' as product_search_url %}

<tr class=" hover:bg-gray-200 py-0 {% cycle 'bg-white' 'bg-gray-50' %} divide-x "
    aria-rowindex="{{ row_number }}">
    <!-- row num -->
    <td class="whitespace-nowrap w-fit text-center px-4 w-px"
        aria-colindex="1">
        {{ row_number }}
        {% render_field form.id placeholder="" %}
        {% render_field form.DELETE type="hidden" %}
    </td>
	.....
</tr>
```

so with this setup i use the button to add new rows to the main template from the partial template that itself does not have any components. but now when i updated to v0.114, when i retrieve the response from the backend is :

```
<html><head></head><body>
    <!-- row num -->
    
        2
        <input type="hidden" name="items-1-id" placeholder="" id="id_items-1-id">
        <input type="hidden" name="items-1-DELETE" id="id_items-1-DELETE">
		
		.......
     
<script src="/static/django_components/django_components.min.js"></script><script>eval(Components.unescapeJs(`(() =&gt; {
        const loadedJsScripts = [];
        const loadedCssScripts = [];
        const toLoadJsScripts = [];
        const toLoadCssScripts = [];

        loadedJsScripts.forEach((s) =&gt; Components.manager.markScriptLoaded(&quot;js&quot;, s));
[...]

PR #797, opened 2024-11-30
fix: do not insert dependencies script when template does not contain any components

JuroOravec:
This fixes the problem described in https://github.com/EmilStenstrom/django-components/issues/793, which was caused by faulty parsing of the HTML.

This faulty parsing was causing Selectolax to insert extra `<html>`, `<body>` and `<head>` tags.

This was causing issues downstream. Because when we are given an HTML document, then if it contains no `{% component_js_dependenies %}`, then we try to insert them into the `<body>` (and insert css into head).

And so because Selectolax was adding those tags, then our logic was detecting `<body>` and `<head>` tags in templates where it should NOT be.

I fixed this problem maybe a month or two upstream, in Selectolax. There I wrote an approach that more smartly parses the HTML, and knows when it's given an HTML fragment, or a full HTML document.

So here in django_components it was only a matter of bumping selectolax, and using API from there instead of our custom functions from `util/html.py`.

Closes https://github.com/EmilStenstrom/django-components/issues/793


src/django_components/util/misc.py
```diff
@@ -77,3 +77,10 @@ def get_last_index(lst: List, key: Callable[[Any], bool]) -> Optional[int]:
         if key(item):
             return len(lst) - 1 - index
     return None
+
+
+def _escape_js(js: str, eval: bool = True) -> str:
```
JuroOravec, line 82: This was previously in the `dependencies.py` file. Moved it here

src/django_components/dependencies.py
```diff
@@ -771,30 +789,28 @@ class ComponentDependencyMiddleware:
...
-        response = self.process_response(response)
+        response = await self._get_response(request)
+        response = self._process_response(response)
         return response
 
-    def process_response(self, response: HttpResponse) -> HttpResponse:
-        if not isinstance(response, StreamingHttpResponse) and response.get("Content-Type", "").startswith(
-            "text/html"
-        ):
+    def _process_response(self, response: HttpResponse) -> HttpResponse:
```
JuroOravec, line 812: Not related to the bug. Prefixed this method with underscore, so it does not appear in the generated documentation.

src/django_components/util/html.py
```diff
@@ -1,100 +0,0 @@
-from typing import List, Union
```
JuroOravec, line 1: These functions can be removed, as their equivalent was implemented in selectolax. So I removed also the test file.

JuroOravec:
Superseded by https://github.com/EmilStenstrom/django-components/pull/823

PR #798, opened 2024-12-01
docs: Fix build + various

JuroOravec:
Third step of https://github.com/EmilStenstrom/django-components/pull/775. After this is merged to `dev`, I would merge it to `master` to update the docs. If the docs build and deploy correctly, I would make another MR to delete the content from README, and instead redirect people to the docs site.

Changes:
- Added custom styling
- Added CI job to test that docs build without error
- Fixed imports in docs scripts.

And other small changes.


.github/workflows/tests.yml
```diff
@@ -31,3 +31,28 @@ jobs:
         playwright install chromium --with-deps
     - name: Run tests
       run: tox
+
+  # Verify that docs build
+  test_docs:
```
JuroOravec, line 36: Added a test to check that the docs do build, to avoid accidentlly breaking them again.

mkdocs.yml
```diff
@@ -136,6 +141,7 @@ plugins:
         - code_of_conduct.md
         - license.md
   - markdown-exec
+  - mkdocs-nav-weight
```
JuroOravec, line 144: What `mkdocs-nav-weight` does is that it sorts pages by the value of `weight` that can be specified in the front matter, e.g.

```txt
weight: 5
---
```

src/docs/SUMMARY.md
```diff
@@ -7,6 +7,4 @@
   - [Setup](guides/setup/)
   - [Cookbook](guides/cookbook/)
 - [API Documentation](reference/)
-- [Changelog](changelog.md)
-
-
+- [Release notes](release_notes.md)
```
JuroOravec, line 10: Renamed "Changelog" to "Release notes" in the docs.

src/docs/scripts/reference.py
```diff
@@ -50,7 +52,21 @@
...
+#         - We can't import from packages found in the `src` directory (e.g. `docs.scripts.extensions`)
+#
+#       But we need to import from another module.
+#       Hence we add the `src` directory to `sys.path` ourselves, so we can import from the
+#       `docs.scripts.extensions` module.
+current_dir = os.path.dirname(os.path.abspath(__file__))
+src_dir = os.path.abspath(os.path.join(current_dir, "..", ".."))
+sys.path.insert(0, src_dir)
+
+from docs.scripts.extensions import _format_source_code_html  # noqa: E402
```
JuroOravec, line 69: As the docs say, I had to modify the `sys.path`, so that the script can import from other files in `src/docs/scripts`.

I came across this only now, as it turns out I had pip-installed django-components in my django-components project, which made the python import paths under `docs` valid.

But when I started with a clean install, this script could no longer import from `docs`. Hence the workaround.

JuroOravec:
@EmilStenstrom This is now ready for review, after this we'll be able to see the updated docs live :)

src/docs/scripts/reference.py
dalito: Was it this missing sys.path modification that made you move the docs under src? (which is unusual)-  If so you could make docs a root folder again like in most other projects to not break expectations.

JuroOravec: Not sure I understand the wording, but my original intention to move the docs under `src` was to avoid having to do hacks with `sys.path`. But turns out I had to do it anyway. So yeah, the `docs` dir could be moved back to where it was before.

@dalito Would you do write up the MR for moving the `docs` dir back up (after this MR is merged)? There's still two other bugs after introducing v0.110 that I want to focus on first.

dalito: Ok. I'll look into moving the docs to root after this is merged.

.github/workflows/tests.yml
```diff
@@ -31,3 +31,28 @@ jobs:
...
+        python-version: ${{ matrix.python-version }}
+        cache: "pip"
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -r requirements-docs.txt
+        # Install your package locally
+        python -m pip install -e .
+    - name: Run tests
```
EmilStenstrom, line 57: Lets change this too.

EmilStenstrom (approved):
One small suggestion, but go ahead!

PR #802, opened 2024-12-02
refactor: add generated API docs to git

JuroOravec:
One more issue with the new documentation, the API reference is not showing up. I think the reason is because the directory they are in is .gitignore, and so I think that's why it doesn't get pushed to the `gh-pages` branch.

So this MR adds the generated files to source control.

I also added a little

`<!-- Autogenerated by reference.py -->`

at the top of those files to make it clear where they come from.


JuroOravec:
@EmilStenstrom Again, tagging for visibility

PR #803, opened 2024-12-02
docs: slim down project README

JuroOravec:
The docs are working over at https://emilstenstrom.github.io/django-components/dev/

So the last step of https://github.com/EmilStenstrom/django-components/pull/775 is to remove the duplicate content from the project README.

After this, I'd make a release so that this new documentation becomes the default / latest one, so it will be available also under

https://emilstenstrom.github.io/django-components


EmilStenstrom (approved):
Ready to go!

One strange thing in the new docs is that:
https://emilstenstrom.github.io/django-components/dev/
and https://emilstenstrom.github.io/django-components/dev/overview/welcome/
is a bit different. Know how to fix?

JuroOravec:
@EmilStenstrom Noticed that too. What's going on there is that they are two different pages, it "just so happens" that we show the same content on both. And one of them is in a directory, so shows all other pages in the same directory on the side. Whereas the other does not.

But there HAS to be the `README.md` file that is used to populate the top-level page [emilstenstrom.github.io/django-components/dev](https://emilstenstrom.github.io/django-components/dev/)

So one option  could be to remove the nested `welcome` page - [emilstenstrom.github.io/django-components/dev/overview/welcome](https://emilstenstrom.github.io/django-components/dev/overview/welcome/)

But I imagine that down the line this project could have flashier landing page, like [Tetra](https://tetraframework.com/) or [django-cotton](https://django-cotton.com/) (or every other tech product ever). And in that case, what's currently the front page would actually make more sense to be in "/overview/welcome/".

---

I came across https://github.com/mkdocs/mkdocs-redirects, and I think this should make it possible to redirect from the homepage to the "/overview/welcome" page. So then people would be redirected to the version of the page that has the left panel.

I'll do that in a separate PR.

Issue #808, opened 2024-12-02
Add support for Python 3.13


JuroOravec:
Python 3.13 was released 2 months ago https://devguide.python.org/versions/, so we should add it to the supported list.

Django's docs now include support for Python 3.13 in 5.1 and 5.2 here https://docs.djangoproject.com/en/dev/faq/install/.

JuroOravec:
Python 3.8 is no longer supported, but Django 4.2 still uses it. And Django 4.2 will be deprecated in April 2026 ([see docs](https://www.djangoproject.com/download/#ft2)). By then Python 3.14 will be introduced, and Python 3.9 will be deprecated.

So the next in line will be:
- Add support for Python 3.14 on Nov 2025
- Drop support for Python 3.8 and 3.9, and Django 4.2 in April 2026

PR #809, opened 2024-12-02
chore: update supported versions - add Python 3.13, Django 5.1

JuroOravec:
Closes https://github.com/EmilStenstrom/django-components/issues/808


scripts/supported_versions.py
```diff
@@ -13,7 +13,7 @@ def cut_by_content(content: str, cut_from: str, cut_to: str):
 
 
 def keys_from_content(content: str):
-    return re.findall(r"<td>(.*?)</td>", content)
+    return re.findall(r"<td><p>(.*?)</p></td>", content)
```
JuroOravec, line 16: Had to update the script a little

Issue #806, opened 2024-12-02
Run CI tests also on Windows image


JuroOravec:
https://github.com/EmilStenstrom/django-components/issues/796 is a Windows-specific issue, which I have harder time checking, because my main machine is Unix.

I have only yesterday learnt that you can run Windows environemnt in Github actions ([see docs](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#using-a-github-hosted-runner)).

I tried to add that in https://github.com/EmilStenstrom/django-components/pull/799 ([See diff](https://github.com/EmilStenstrom/django-components/pull/799/commits/f0e50d7d54e10c5ffa73916bacd39ae6f0911a8c)), but it looks like it won't be as simple as just modifying the yaml file. Hence I'm moving this into a separate issue.

PR #799, opened 2024-12-01
fix: set paths for component media as posix paths even on windows

JuroOravec:
Closes https://github.com/EmilStenstrom/django-components/issues/796

PR #815, opened 2024-12-02
refactor: Run CI tests in both unix and windows environments

JuroOravec:
To ensure that bugs like that in https://github.com/EmilStenstrom/django-components/pull/799 don't regress, this MR configures the CI to run the tests also in Windows.

Closes https://github.com/EmilStenstrom/django-components/issues/806


JuroOravec:
Btw @EmilStenstrom it seems that for this one, the protection rules will have to be updated. Because the "pending checks" are waiting for jobs like `build (3.10)`, but the actual jobs have names like `Run tests / build (3.10, ...)`. However, don't know why it's blocking here, and not on other MRs.

See this post - https://github.com/orgs/community/discussions/26698#discussioncomment-3252954

Screenshot 2024-12-02 at 19 56 44

EmilStenstrom:
@JuroOravec Disabled requiring status checks to pass. We should still never merge with failing checks, but it's such a mess to always keep them up to date, so let's check that manually.

EmilStenstrom:
@JuroOravec Added the requirements back up!

PR #816, opened 2024-12-02
docs: Move docs-folder to root

dalito:
As a follow-up of #798 this pull request moves the docs from `src/docs/` to `docs/` to match expectations.

I used Windows and observed two problems in `scripts/reference.py`:
- The file-encoding was assumed to be "utf-8" by default but this is not the case on Windows.
- The URLs were generated from relative paths which contain backslashes on Windows. Paths should be converted to POSIX notation if they are used to construct URLs.

ToDo:
- [x] Check / update gh-actions --> I think there is nothing to do.


dalito:
Somehow GitHub has a problem processing this. I'll update / force push to see if this fixes it

requirements-docs.txt
```diff
@@ -141,7 +146,7 @@ mkdocs-material-extensions==1.3.1
 mkdocs-minify-plugin==0.8.0
     # via hatch.envs.docs
 mkdocs-nav-weight==0.2.0
-    # manually added
+    # via hatch.envs.docs
```
JuroOravec, line 149: Thanks for updating this file, I was too lazy to remember how to use hatch when I  made the changes :D

docs/scripts/extensions.py
```diff
@@ -2,9 +2,9 @@
 from typing import List, Optional, Type
 
 import griffe
+from mkdocs_util import get_mkdocstrings_plugin_handler_options, import_object, load_config
```
JuroOravec, line 5: Minor thing, but I'd move this import two lines down back under the `django_components.util.misc`. Because right now it looks as if `mkdocs_util` was an external package.

A convention I follow is:
```py
# Python's standard standard library packages like `os`, `typing`, etc
import os
from typing import Any

# Third-party packages like `django`, or `griffe`
import griffe

# Our own module, like `django_components` or in this case `mkdocs_util`
from django_components.util.misc import get_import_path
```

docs/reference/template_tags.md
```diff
@@ -21,7 +21,7 @@ Import as
 
 
 
-<a href="https://github.com/EmilStenstrom/django-components/tree/master/.venv/lib/python3.11/site-packages/django_components/templatetags/component_tags.py#L117" target="_blank">See source code</a>
+<a href="https://github.com/EmilStenstrom/django-components/tree/master/src/django_components/templatetags/component_tags.py#L117" target="_blank">See source code</a>
```
JuroOravec, line 24: Did you have to fix this manually, or did this get updated when you ran `mkdocs build` or `mkdocs serve`?

JuroOravec:
@dalito Looks great, appreciate that! Just two minor questions, otherwise good to go. As for the stuck processing, did you check if the branch on the latest `master`?

docs/reference/template_tags.md
dalito: This got changed/fixed when running `hatch run docs:build`

docs/scripts/extensions.py
dalito: I initially forgot to install pre-commit to sort imports before committing. This bad sorting then triggered pre-commit auto fix here causing probably the hanging merge.

JuroOravec (approved):

JuroOravec:
I just tested it locally to double check and it run :)

@EmilStenstrom Tagging for visibility. This is just moving the docs back to their original place, so I'll go ahead and merge it.

PR #817, opened 2024-12-03
Adding `request` arg to render

lhole:
The [django components docs](https://emilstenstrom.github.io/django-components/latest/overview/installation/) mention `context_processors`, however I don't seem to be able to get them to work when only using `render_to_response` to render my page.

The [django `render` function](https://docs.djangoproject.com/en/5.1/topics/http/shortcuts/#render) takes `request` as an argument, however django components' `render_to_response` and `render` [don't](https://emilstenstrom.github.io/django-components/latest/concepts/fundamentals/components_in_python/?h=render#inputs-of-render-and-render_to_response). 
I believe this makes the `context_processors` not run because they default to using `django.template.context.Context` instead of `django.template.context.RequestContext`.  Only [RequestContext calls the context processors](https://github.com/django/django/blob/871e1ee5ff0b75aee5dd1bd3e88e349ca0ddc60d/django/template/context.py#L257)

This PR adds passing `request` through the `render` and `render_to_response` methods. I've made it optional to minimise backwards incompatibility. 

Before investing more time on this (for things like tests), I want to get your opinion on whether this it the right route to go down.


src/django_components/component.py
```diff
@@ -588,7 +593,8 @@ def render(
               or to the end of the `<body>` tag. CSS dependencies are inserted into
               `{% component_css_dependencies %}`, or the end of the `<head>` tag.
         - `render_dependencies` - Set this to `False` if you want to insert the resulting HTML into another component.
-
+        - `request` - The request object. This is only required when needing to use RequestContext,
```
JuroOravec, line 596: We should add that `request` is ignored when `context` is given and it is already an instance of `Context` (so we don't have to initialize `Context`.

JuroOravec: And we should also update the docs in src/docs/concepts/fundamentals/components_in_python.md. Tho note that that file will be moved in https://github.com/EmilStenstrom/django-components/pull/816

JuroOravec:
@lhole Thanks for the suggestion, IMO we can add it to make the API more convenient.

Personally I try not to use the context processors, as I prefer to have all context variables explicitly defined on the components. That way, when there is a bug, I know that I need to consider only the variables that go into the component. But it also means that sometimes I have to "drill down" some info through several components, like user info.

Theoretically one could define`RenderContext` outside of a component, and pass it as `context`:

```py
context = RenderContext({ ... }, request)
MyComponent.render(
    context=context,
)
```

But I can imagine that using context processors is fairly standard thing in other projects.
And as you said, django's `render()` requires only the `request` kwarg to be given, and it probably creates the RenderContext behind the scenes (?). So I think the `request` kwarg makes sense.

lhole:
> Personally I try not to use the context processors, as I prefer to have all context variables explicitly defined on the components. 

That makes sense. I had assumed that because they were explicitly mentioned in the [installation instructions](https://emilstenstrom.github.io/django-components/latest/overview/installation/) that they were considered _recommended_. It might be worth nuancing the docs there. 

In my case, all I was wanting was the `debug` variable to be set, which is easy to accomplish in a bunch of other ways.

EmilStenstrom:
I’m +1 on supporting this!

JuroOravec:
Great! Oki, in that case @lhole, would you add test(s) for this? I think we can put the tests inside `tests/test_component.py` inside the `ComponentRenderTest` class. You will know better, what comes to my mind is that we should test 3 scenarios:
1. `request` given but no `context` given -> Should create empty `RenderContext`
2. `request` given and `context` given as dictionary -> Should create `RenderContext` with given dict
3. `request` given and `context` as Context -> Should ignore `request`, `RenderContext` NOT created.

lhole:
> @lhole for reference, have you seen the support for provide/inject? https://github.com/EmilStenstrom/django-components?tab=readme-ov-file#prop-drilling-and-dependency-injection-provide--inject

Thanks. I spotted that after doing a bit more digging. For now, I'm just reading `debug` from settings in the component that needs it.

lhole:
> Great! Oki, in that case @lhole, would you add test(s) for this? I think we can put the tests inside `tests/test_component.py` inside the `ComponentRenderTest` class. You will know better, what comes to my mind is that we should test 3 scenarios:
> 
> 1. `request` given but no `context` given -> Should create empty `RenderContext`
> 2. `request` given and `context` given as dictionary -> Should create `RenderContext` with given dict
> 3. `request` given and `context` as Context -> Should ignore `request`, `RenderContext` NOT created.

I've added the suggested tests. They're a bit duplicative as I didn't see how common behaviour was extracted in the test suite. I can refactor if you have suggestions.

I've updated the docs as suggested

JuroOravec (approved):

JuroOravec:
@lhole Looks great and the tests are passing too! I'll go ahead and merge / release this

JuroOravec:
This is now live in https://github.com/EmilStenstrom/django-components/releases/tag/0.118 🎉

Issue #789, opened 2024-11-27
Problem with inlined (pre-rendered) component and dependencies (v0.110+)


PavelPancocha:
Hi!

First thanks for all nice progress in last days!

After a recent update to v0.110+ (v0.113) I found error, which I am not happy with and I am not able to solve it. Maybe there is some magic trick I misseen?

Historically we had some html-fragments rendered directly in python, then moved them to "html-fragments" and rendered still in methods and now we switched that to components. Eg. if you want to show some components in django-admin changelist you do not have much more options.

So we have component, we call it in method with render -> display that in django admin changelist.

Before this change the fragment looked liked this:

![image](https://github.com/user-attachments/assets/3ebe596e-1df3-4775-bbac-5fc2b3482b71)

But now it is like this:

![image](https://github.com/user-attachments/assets/820782fb-937f-4c1f-9c07-9bc9df1960dd)

Is that on purpose? Is this really the best way how to do that? We have two badges, per row, 50 rows = this code is there 100x + the tendencies are injected on the top.

Can I somehow render just the component **without** dependencies?

Thank you!

PavelPancocha:
Hi, thanks for swift reaction.

First things first: `render_dependencies=False` works, now it does the same thing as before. 

I will try to share some code with you, lets start from the bottom:

```python
def get_badge(label: StrOrPromise, color: BadgeColor = BadgeColor.INFO) -> SafeString:
    """
    Return an html badge.
    Color options: primary, success, info, warning, danger
    """
    from components.badge import Badge

    return mark_safe(  # nosec
        Badge.render(
            kwargs={"style": color.value},
            slots={"content": label},
            render_dependencies=False
        )
    )
```

We than use it (with some method in between that just fill the label and color depending on conditions) in admin as method, to be used to display that in admin (in changelist or change page):

```python
    @display(description=_("State"))
    def state(self, obj: CustomUser) -> SafeString:
        """
        Return user’s state as badge.
        """
        return get_user_state_badge(obj)

```

The code is partly historic and being transformed to use components step by step. But the usage pattern seems complete normal for django admin. Feel free to suggest better solution, if you see some! :)

JuroOravec:
@PavelPancocha Glad to hear that you're rendering the badge with `Badge.render()`. If it was embedded inside a template as `{% component %}`, that'd be more difficult to migrate.

Btw where / how do you insert the `badge/style.css` into the HML?

JuroOravec:
[...]

PR #821, opened 2024-12-05
fix: Fix broken JS execution order

### LINKED ISSUE #791: v0.110+ no longer loads JS scripts one after the other
The JS scripts in `Component.Media.js` are now loaded in parallel. I initially made the change thinking that it could help save some time.

But in my project it's making things more difficult, as each file now needs to poll for if the variables / functions from other files are loaded or not.

And I'm also having a hard time to imagine a case where the parallel loading would be preferable over serial.

I want to fix this over the weekend, making the JS scripts in `Component.Media.js` load serially. And also, the JS in `Component.js` should run only AFTER all the JS dependencies have been loaded.

### LINKED ISSUE #818: alpine:init issue
I've got a super weird issue and i was wondering if anyone else had encountered it?

I have a number of Django Components that create and register their Alpine.js x-data in the Django Components js file. 

Something like this:

```
 document.addEventListener('alpine:init', () => {
        Alpine.data('somexdata', () => ({
            somevalue: false
        }))
    })
```

This works fine when I'm referencing Alpine.js externally via its CDN link. However, if I’m referencing Alpine locally then the Alpine.data method is never called. It’s like the alpine:init method has already fired before the code that subscribes to the alpine:init event is added to the page

Normally you would need to add this event hook after the Alpine.js script is loaded but before any Alpine components are initialised. 

In the previous version of Django Components I was using (0.73 i think?) I could call component_js_dependencies and have the Django Component scripts added into the head of the page. Using component_js_dependencies to place the scripts in the head now results in a JavaScript error. If i put component_js_dependencies at the top of the body it’s still fails. I'm thinking this is because rather than writing script tags there is now a JS loader that loads the scripts and the scripts are now added to the page after the apline:init event has fired. 

Is there a way to have component_js_dependencies just out put script tags rather than use the JS script loader?

JuroOravec:
For context, see https://github.com/EmilStenstrom/django-components/issues/818#issuecomment-2514496357.

In summary (assuming we're rendering a "document", and NOT a "fragment"):
- Before, when a component had `Media.js` defined, those JS scripts were ALWAYS loaded dynamically via our JS dependency manager
- Now, the JS scripts in `Media.js` are actually inserted as `<script>` tags directly into the HTML server-side.

So before we'd replace `{% component_js_dependencies %}` with something like this:

```html
<script src="django_components/django_components.min.js"></script>
<!-- Component.js scripts-->
<script>
  console.log("hello")
</script>
<script>
  ...
</script>
<!-- Media.js scripts-->
<script>
   loadJs("/path/to/script1.js");
   loadJs("/path/to/script2.js");
   ...
</script>
```

Now, we replace `{% component_js_dependencies %}` with:

```html
<script src="django_components/django_components.min.js"></script>
<!-- Internal script to mark scripts as loaded-->
<script>
   markLoaded("/path/to/script1.js");
   markLoaded("/path/to/script2.js");
   ...
</script>
<!-- Media.js scripts-->
<script src="/path/to/script1.js"></script>
<script src="/path/to/script2.js"></script>
...
<!-- Component.js scripts-->
<script>
  console.log("hello")
</script>
<script>
  ...
</script>
```

What that means is that all JS scripts, whether from `Component.js` or `Media.js` will be executed at the first pass.

This also solves the issue around using AlpineJS. E.g. see valid usage [here](https://github.com/EmilStenstrom/django-components/pull/821/files#diff-1bdbe68104fea6a9d8fbe2377d9ccdaf51604312e468f258b5ce4c43d05c552fR136) or [here](https://github.com/EmilStenstrom/django-components/pull/821/files#diff-1bdbe68104fea6a9d8fbe2377d9ccdaf51604312e468f258b5ce4c43d05c552fR183).

Closes #818, closes #791


docs/devguides/dependency_mgmt.md
```diff
@@ -189,15 +189,15 @@ This is how we achieve that:
 
    ```js
    // Load JS or CSS script if not loaded already
-   Components.loadScript("js", '<script src="/abc/xyz/script.js">');
-   Components.loadScript("css", '<link href="/abc/xyz/style.css">');
+   Components.loadJs('<script src="/abc/xyz/script.js">');
```
JuroOravec, line 192: The JS function `loadScript` was pratically two functions and a switch, so I've split it into two explicit functions, `loadJs` and `loadCss`

src/django_components/dependencies.py
```diff
@@ -492,16 +493,25 @@ def get_component_media(comp_cls_hash: str) -> Media:
...
+            *(to_load_css_urls if type == "document" else []),
+        ]
+    )
+    loaded_js_urls = sorted(
+        [
+            *loaded_component_js_urls,
+            # NOTE: When rendering a document, the initial JS is inserted directly into the HTML
+            # so the scripts are executed at proper order. In the dependency manager, we only mark those
+            # scripts as loaded.
+            *(to_load_js_urls if type == "document" else []),
```
JuroOravec, line 508: What this bit of changes does is that before, we've sent the JS scripts from `Media.js` to be loaded via our client-side `loadJs()` function (previously called `loadScript()`.

But since now we insert them as regular `<script>` tags into the HTML, we just let the dependency manager know that these scripts have been loaded by calling `markScriptLoaded()` with the URL of the resource.

src/django_components/dependencies.py
```diff
@@ -551,8 +574,9 @@ def _postprocess_media_tags(
 
         if not _is_nonempty_str(maybe_url):
             raise RuntimeError(
-                f"One of entries for `Component.Media.{script_type}` media is missing "
-                f"value for attribute '{attr}'. Got:\n{tag}"
+                f"One of entries for `Component.Media.{script_type}` media is missing a "
```
JuroOravec, line 577: Updated this error message. This is raised if user defined an inlined scripts inside `Media.js`, e.g.

```py
class MyTable(Component):
    class Media:
        js = mark_safe("""
          <script>
            console.log('helo!')
          </script>
        """)
```

The thing is that we allow to pass `SafeString` to `Media.js`. And in such case we assume that we're given a string of the `<script>` tag.

But the `<script>` tag MUST contain the `src` attribute. So we ask user to move the content into a separate file and to only link to that file with this `<script>` tag.

So the result would be something like this:

```js
// hello.js
console.log('helo!')
```

```py
class MyTable(Component):
    class Media:
        js = mark_safe("""
          <script src="static/hello.js"></script>
        """)
```

src/django_components/dependencies.py
```diff
@@ -672,39 +709,38 @@ def _gen_exec_script(
...
-            toLoadJsScripts.map((s) => Components.manager.loadScript("js", s))
-        ).catch(console.error);
-        Promise.all(
-            toLoadCssScripts.map((s) => Components.manager.loadScript("css", s))
-        ).catch(console.error);
-    }})();
+    # NOTE: Wrap the body in self-executing function to avoid polluting the global scope.
+    exec_script: types.js = f"""
+        (() => {{
+            Components.manager._loadComponentScripts({{
```
JuroOravec, line 730: I moved the entirety of what was in this script before, and defined it in the JS-side dependency manager here https://github.com/EmilStenstrom/django-components/pull/821/files#diff-5b8acf408bfdc9d80635c5efe6b1b636da344349820e975c948e0214cf736b98R230

src/django_components/dependencies.py
```diff
@@ -672,39 +709,38 @@ def _gen_exec_script(
...
+                loadedCssUrls: {json.dumps(loaded_css_urls)},
+                loadedJsUrls: {json.dumps(loaded_js_urls)},
+                toLoadCssTags: {js_arr(escaped_to_load_css_tags)},
+                toLoadJsTags: {js_arr(escaped_to_load_js_tags)},
+            }});
+            document.currentScript.remove();
```
JuroOravec, line 736: This line makes it so that after the script is executed the parent `<script>` is removed from the HTML.

This is relevant mostly for the HTML fragment use case, where people may be surprised by finding extra `<script>` tags, like seen here https://github.com/EmilStenstrom/django-components/issues/789

tests/test_dependency_rendering_e2e.py
```diff
@@ -215,3 +215,132 @@ async def test_renders_css_nojs_env(self):
         self.assertEqual("rgb(255, 0, 0)", data["myStyle2Color"])  # AKA 'color: red'
 
         await page.close()
+
+    @with_playwright
+    async def test_js_executed_in_order__js(self):
```
JuroOravec, line 220: Tests in this file are new. Other test files are just updates.

JuroOravec:
@EmilStenstrom Altho this is just a bug fix, it's relatively big, so would you give this one a look?

EmilStenstrom (approved):
Looks good to me! With script tags the browser will also be able to load the scripts faster via prefetch, so I think this is a very good improvement overall.

PR #823, opened 2024-12-06
refactor: replace selectolax with beautifulsoup

JuroOravec:
This supersedes the PR https://github.com/EmilStenstrom/django-components/pull/797.

In short, this replaces [Selectolax](https://github.com/rushter/selectolax) library with [BeautifulSoup](https://www.crummy.com/software/BeautifulSoup/).

My cup hath runneth over with Selectolax / Lexbor for 3 reasons:

1. It modifies the HTML, e.g. inserting the extra `<html>`, `<head>`, `<body> tags.

2. It tries to fix / modify the HTML, e.g. adding a closing tag when it comes across an unclosed one.

3. I cannot access body of `<template>` tags (see https://github.com/rushter/selectolax/issues/146).

Reasons 1 and 2 are annoying, but 3) is flat-out a deal-breaker.

---

Initially I tried to replaces Selectolax with [lxml](https://lxml.de/), because lxml is also compiled to C and has similar performance as Selectolax. But it turned out that lxml too makes changes to the given HTML fragment.

I don't want to use a library that will make extra modifications to the users' HTML. That will just lead to confusion as one won't be able to figure out where the changes are coming from.

So I ended up with BeautifulSoup, which does just that.

Closes https://github.com/EmilStenstrom/django-components/issues/793

### DIFF
--- a/src/django_components/dependencies.py
+++ b/src/django_components/dependencies.py
@@ -33,10 +33,9 @@ from django.templatetags.static import static
 from django.urls import path, reverse
 from django.utils.decorators import sync_and_async_middleware
 from django.utils.safestring import SafeString, mark_safe
-from selectolax.lexbor import LexborHTMLParser
 
 import django_components.types as types
-from django_components.util.html import parse_document_or_nodes, parse_multiroot_html, parse_node
+from django_components.util.html import SoupNode
 from django_components.util.misc import escape_js_string_literal, get_import_path
 
 if TYPE_CHECKING:
@@ -362,26 +361,14 @@ def render_dependencies(content: TContent, type: RenderType = "document") -> TCo
     # then try to insert the JS scripts at the end of <body> and CSS sheets at the end
     # of <head>
     if type == "document" and (not did_find_js_placeholder or not did_find_css_placeholder):
-        tree = parse_document_or_nodes(content_.decode())
+        maybe_transformed = _insert_js_css_to_default_locations(
+            content_.decode(),
+            css_content=None if did_find_css_placeholder else css_dependencies.decode(),
+            js_content=None if did_find_js_placeholder else js_dependencies.decode(),
+        )
 
-        if isinstance(tree, LexborHTMLParser):
-            did_modify_html = False
-
-            if not did_find_css_placeholder and tree.head:
-                css_elems = parse_multiroot_html(css_dependencies.decode())
-                for css_elem in css_elems:
-                    tree.head.insert_child(css_elem)  # type: ignore # TODO: Update to selectolax 0.3.25
-                did_modify_html = True
-
-            if not did_find_js_placeholder and tree.body:
-                js_elems = parse_multiroot_html(js_dependencies.decode())
-                for js_elem in js_elems:
-                    tree.body.insert_child(js_elem)  # type: ignore # TODO: Update to selectolax 0.3.25
-                did_modify_html = True
-
-            transformed = cast(str, tree.html)
-            if did_modify_html:
-                content_ = transformed.encode()
+        if maybe_transformed is not None:
+            content_ = maybe_transformed.encode()
 
     # Return the same type as we were given
     output = content_.decode() if isinstance(content, str) else content_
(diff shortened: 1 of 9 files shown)


src/django_components/dependencies.py
```diff
@@ -362,26 +361,14 @@ def on_replace_match(match: "re.Match[bytes]") -> bytes:
     # then try to insert the JS scripts at the end of <body> and CSS sheets at the end
     # of <head>
     if type == "document" and (not did_find_js_placeholder or not did_find_css_placeholder):
-        tree = parse_document_or_nodes(content_.decode())
-
-        if isinstance(tree, LexborHTMLParser):
```
JuroOravec, line 367: Moved this section into own function `_insert_js_css_to_default_locations`

src/django_components/dependencies.py
```diff
@@ -802,27 +826,27 @@ class ComponentDependencyMiddleware:
     """
 
     def __init__(self, get_response: "Callable[[HttpRequest], HttpResponse]") -> None:
-        self.get_response = get_response
+        self._get_response = get_response
```
JuroOravec, line 829: Not related to the bug. Prefixed this method with underscore, so it does not appear in the generated documentation.

src/django_components/util/html.py
```diff
@@ -1,100 +1,108 @@
-from typing import List, Union
+from abc import ABC, abstractmethod
+from typing import Any, List, Optional, Sequence
 
-from selectolax.lexbor import LexborHTMLParser, LexborNode
+from bs4 import BeautifulSoup, CData, Comment, Doctype, Tag
 
 
-def parse_node(html: str) -> LexborNode:
+class HTMLNode(ABC):
```
JuroOravec, line 7: Since I changed from Selectolax to lxml to BS4, I defined the interface for the HTML parser.

<details>
<summary>
E.g. click here to see implementation for lxml
</summary>

```py
class LXMLNode(HTMLNode):
    def __init__(self, node: lxml.etree.ElementBase):
        self.node = node

    @classmethod
    def from_fragment(self, html: str) -> List["LXMLNode"]:
        elems = lxml.html.fragments_fromstring(html)
        return [LXMLNode(elem) for elem in elems]

    def to_html(self) -> str:
        return lxml.html.tostring(self.node, encoding="unicode")

    def find_tag(self, tag: str) -> Optional["LXMLNode"]:
        if self.node.tag == tag:
            return self
        else:
            matches = self.node.xpath(tag)
            if matches:
                return LXMLNode(matches[0])

        return None

    def append_children(self, children: Sequence["LXMLNode"]) -> None:
        for child in children:
            self.node.append(child.node)

    def get_attr(self, attr: str, default: Any = None) -> Any:
        return self.node.get(attr, default)

    def set_attr(self, attr: str, value: Any) -> None:
        if value is True:
            self.node.set(attr, None)
        elif value is False:
            self.node.attrib.pop(attr, None)
        else:
            self.node.set(attr, value)

    def is_element(self) -> bool:
        return isinstance(self.node, lxml.html.HtmlElement)
```

</details>

I also added a test for this interface, in case we want to change the implementation in the future.

That's because one thing that I keep in mind is that if in the future we get to measure performance for bottlenecks, and it turns out that HTML parsing is a bottleneck, then I have written a lightweight HTML parser for django-vue. So in case of a bottleneck, that parser could be migrated to Cython, and then it should have similar performance as lxml or Selectolax.

src/django_components/util/misc.py
```diff
@@ -77,3 +77,10 @@ def get_last_index(lst: List, key: Callable[[Any], bool]) -> Optional[int]:
         if key(item):
             return len(lst) - 1 - index
     return None
+
+
+def _escape_js(js: str, eval: bool = True) -> str:
```
JuroOravec, line 82: This was previously in the `dependencies.py` file. Moved it here

JuroOravec:
@EmilStenstrom This is a last bug fix related to v0.110, now ready for review :)

EmilStenstrom (approved):
Looks good to me! Understand the rationale and appreciate switching to a "safe bet" like bs4.

I still feel it's a bit annoying that we have to take on a dependency like selectolax/lxml/beautifulsoup. There's all sorts of dependency problems that could happen because of this, since the user might already be using bs4 and now have to upgrade it to match our version. Django would have to take it on to merge this, which compounds the problem.

Yes, I remember that we need this for doing JS/CSS stuff later on, so that improves the tradeoffs. And yes, this is not a new problem, could as well have happened together with selectolax as well. Just restating the problem so we know it's a conscious choice we're making.

BTW: Shipping things with C dependencies (lxml, Cython) is not fun, not sure I think doing that is a good idea. But we could explore that if we ever get there.

JuroOravec:
@EmilStenstrom Totally agree. I've made a separate issue for that https://github.com/EmilStenstrom/django-components/issues/831.

CleitonDeLima:
Hello,
I've had performance issues parsing XML using bs4 in my personal project (since it is a simple parser I replaced everything with pure `lxml`.). I wondered if this could happen here too?

This post comments a little on the problem.
https://thehftguy.com/2020/07/28/making-beautifulsoup-parsing-10-times-faster/

EmilStenstrom:
I (and @JuroOravec) think our best bet would probably be to just write a very simple parser that is specialized at parsing HTML outputs by Django: https://github.com/EmilStenstrom/django-components/issues/831

PR #825, opened 2024-12-06
refactor: apply Template patch at AppsConfig.ready()

JuroOravec:
As mentioned here https://github.com/EmilStenstrom/django-components/discussions/819#discussioncomment-11458984, this PR applies the monkeypatching of `Template.render()` at the `ready()` hook. This should unblock integration with `django-debug-toolbar-template-profiler`, as described in https://github.com/EmilStenstrom/django-components/discussions/819


EmilStenstrom (approved):
Looks good to me! :)

PR #827, opened 2024-12-07
refactor: use the tag parser to streamline the tag handlers

JuroOravec:
This MR refactors the parsing of tag inputs. No public API changes.

When we have component inputs within a template:

```django
{% component "table"
  data
  headers=headers
  ...table_attrs
  attrs:class="text-red pa-4"
/ %}
```

Up until now we relied on Django's logic to decide which of whese are args and which are kwargs.

But to allow plugins to pre-process tag inputs (e.g. add a kwarg, or allow extra arguments), I needed to be able to obtain the raw args, kwargs and flags BEFORE we do validation. And Django's logic coupled parsing with validation. So I needed to separate this.

With the tag parser in the codebase (the logic that parses tag's input character-by-character), I used that instead of the Django's piece of logic. 

In other words, the flow before was:
```py
args, kwargs = parse_and_validate(token)
```

And now it will be:
```py
raw_args, raw_kwargs, raw_flags = parse(token)

# This is where plugins will, well, plug-in
# raw_args, raw_kwargs, raw_flags = plugins.on_tag(raw_args, raw_kwargs, raw_flags)

args, kwargs, flags = post_process_and_validate(raw_args, raw_kwargs, raw_flags)
```


JuroOravec:
With this change our tag parsing becomes effectively a superset of Django's. So when it comes to merging django-components into Django, then there it would make sense to replace Django's  [`token_kwargs()`](https://github.com/django/django/blob/c075d4c2c8cef3c9b28180c749d421c63544a9dd/django/template/base.py#L1088) function with our [`parse_tag()`](https://github.com/JuroOravec/django-components/blob/250c60b6aca9c6b2bdefea78b19f0a8b0ad5ea74/src/django_components/templatetags/component_tags.py#L823) and [`TagSpec`](https://github.com/JuroOravec/django-components/blob/250c60b6aca9c6b2bdefea78b19f0a8b0ad5ea74/src/django_components/templatetags/component_tags.py#L59)

src/django_components/expression.py
```diff
@@ -214,24 +210,6 @@ def is_spread_operator(value: Any) -> bool:
...
-# it as a kwarg, so it remains in the original position.
-#
-# So from `...dict`, we make `...1=dict`
-#
-# That way it's trivial to merge the kwargs after the spread
-# operator is replaced with actual values.
-INTERNAL_SPREAD_OPERATOR_RE = re.compile(r"^\.\.\.\d+=")
-
-
-def is_internal_spread_operator(value: Any) -> bool:
```
JuroOravec, line 228: Spread operators are now detected at tokenization in `util/tag_parser.py`

src/django_components/template_parser.py
JuroOravec, line 1: This file contained modifications to Django's functions / classes. No longer needed, as now  the tag parsing is independent.

src/django_components/templatetags/component_tags.py
```diff
@@ -809,164 +815,279 @@ class ParsedTag(NamedTuple):
...
     token: Token,
     tag_spec: TagSpec,
+    tag_id: str,
 ) -> ParsedTag:
-    # Use a unique ID to be able to tie the fill nodes with components and slots
-    # NOTE: MUST be called BEFORE `parse_body()` to ensure predictable numbering
-    tag_id = gen_id()
+    tag_name, raw_args, raw_kwargs, raw_flags, is_inline = _parse_tag_preprocess(parser, token, tag_spec)
+
+    parsed_tag = _parse_tag_process(
```
JuroOravec, line 831: Here, after `_parse_tag_preprocess()` and before `_parse_tag_process()`, is where one of the plugin's hooks would go, in order to be able to e.g. add extra args / kwargs.

EmilStenstrom (approved):
Woop woop, good change!

PR #828, opened 2024-12-07
refactor: Add own LRU cache impl for template caching

JuroOravec:
There was one thing that irked me about the templating caching for some time.

Previously, I used the [`@functools.lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache) decorator to implement caching of templates.

But the problem was that it cached the results based on the inputs to the function. What this meant was that internally we had to initialise `Template` class first, and then manually assign additional arguments.

Django's `Template` class can accept e.g. the `origin` object that tracks where the Template came from.

So previously the flow was:

```py
@lru_cache
def cached_create_template(...):
    ...

template = cached_create_template(TemplateCls, template_str)
template.origin = origin
template.name = name
```

The problem with that was that if someone subclassed a `Template` class in order to pre-process the `name` or `origin`, it wouldn't work, e.g.:

```py
class MyTemplate(Template):
    def __init__(self, template_str: str, origin: Origin, name: str):
        origin = self.transform_origin(origin)
        super.__init__(template_str, origin, name)
```

The issue is because `lru_cache` caches based on its input to its function. You cannot pass e.g. 6 arguments, and say that only first 3 should be used for caching. So if we wanted to pass `name` or `origin` args at `Template` initialisation, then those values would have to be cached too. But that would be wrong.

So instead I added an implementation of a LRU cache, so that the cache is now as a standalone object. So we can derive the cache key from whichever arguments we want.

So now the flow is:

```py
cache_key = (template_cls_id, template_str, engine)
if cache.has(cache_key):
    template = cache.get(cache_key)
else:
    template = TemplateCls(template_str, name, origin, ...)
    cache_key.set(cache_key, template)
```

So even something like the `MyTemplate` shown above would work correctly.

### DIFF
--- a/tests/testutils.py
+++ b/tests/testutils.py
@@ -30,9 +30,10 @@ class BaseTestCase(SimpleTestCase):
         super().tearDown()
         registry.clear()
 
-        from django_components.template import _create_template
+        from django_components.template import template_cache
 
-        _create_template.cache_remove()  # type: ignore[attr-defined]
+        if template_cache:
+            template_cache.clear()
 
     # Mock the `generate` function used inside `gen_id` so it returns deterministic IDs
     def _start_gen_id_patch(self):
--- a/src/django_components/template.py
+++ b/src/django_components/template.py
@@ -1,24 +1,16 @@
-from functools import lru_cache
 from typing import Any, Optional, Type, TypeVar
 
 from django.template import Origin, Template
-from django.template.base import UNKNOWN_SOURCE
 
 from django_components.app_settings import app_settings
-from django_components.util.cache import lazy_cache
+from django_components.util.cache import LRUCache
+from django_components.util.misc import get_import_path
 
 TTemplate = TypeVar("TTemplate", bound=Template)
 
 
-# Lazily initialize the cache. The cached function takes only the parts that can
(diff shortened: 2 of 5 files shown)


tests/testutils.py
```diff
@@ -30,9 +30,10 @@ def tearDown(self):
         super().tearDown()
         registry.clear()
 
-        from django_components.template import _create_template
+        from django_components.template import template_cache
 
-        _create_template.cache_remove()  # type: ignore[attr-defined]
+        if template_cache:
```
EmilStenstrom, line 35: Is this if needed here?

JuroOravec: Yeah, I just checked, and there are two tests in `tests/test_component.py` that break otherwise. I've added a comment.

EmilStenstrom: No, I meant just the if. We’re importing a module, and then saying ”if this module is truthy”? What does that add?

JuroOravec: Ah, right 😄 The cache is initiated lazily, because it needs info from the settings. [At the beginning the variable may be None](https://github.com/JuroOravec/django-components/blob/a1bd29f7a8fc63fb8c183d1e3f52477b670ff485/src/django_components/template.py#L13)

Issue #847, opened 2024-12-13
Version 0.119 seems to have broken the translation mechanism for component inputs


penenkel:
Im using the following pattern to provide translated inputs to my components: `{% component "my-component" label=_("The Label") %}`

This works with version 0.117 and 0.118 but not 0.119.

I'm translating in the parent template (as opposed to in the component itself) so that the normal django makemessages tooling can discover the translatable strings.

JuroOravec:
Hi @penenkel could you add more details? What error did you get? Could you show (minimal) example of the `my-component` component, as well as the parent template?

penenkel:
A stripped down version of my component:
```
@register("my-component")
class MyComponent(Component):
    def get_context_data(self, label: str):
        return {"label": label}

    _template: types.django_html = """
        <div class="component">{{ label }}</div>
    """
    template = _template
```
The usage is as I described before:
```
<div class="wrapper">
    {% component "my-component" label=_("label_Location") %}
    {% endcomponent %}
</div>
``` 
where `label_Location` is the `msgid` in the django.po file.

With version 0.118 this results in the expected rendered http snippet of
```
<div class="wrapper">
  <div class="component">Location</div>
</div>
```
where `Location` is the corresponding `msgstr` from the django.po file.

With version 0.119 the key remains untranslated:
```
<div class="wrapper">
  <div class="component">label_Location</div>
</div>
```

JuroOravec:
This was fixed in #849 and the fix was now released in [v0.120](https://github.com/EmilStenstrom/django-components/releases/tag/0.120) 🚀 

Btw, @penenkel, would you write a test for django-components to check the integration of django-components with Django's translation machinery? Currently there's a risk that the integration may break in the future, as there are no tests for it. Even [the test](https://github.com/EmilStenstrom/django-components/pull/849/files#diff-e52281ab9296959b1e923035c39297b272421932e36736e58a13730db9bc4e19R157) I wrote for this fix is not quite covering this behaviour. I tried to write a test that would cover Django actually translating the string in the template, but I wasn't able to figure out how to use translations in tests (see [chatgpt](https://chatgpt.com/share/675e7c28-b9ec-8004-94bb-2391acfa4b29)), as I'm not using translation files in my project. For inspiration see the tests in [BlockCompatTests](https://github.com/JuroOravec/django-components/blob/a2101ea48c99d5f415083b23e2118d34137320de/tests/test_templatetags.py#L97).

PR #849, opened 2024-12-15
fix: format tag param with translation correctly

JuroOravec:
Fix a bug in tag parser where if a value was a translation, e.g.

`{% component _("my_text") %}`

then the `_()` were being accidentally stripped.

Closes https://github.com/EmilStenstrom/django-components/issues/847

### DIFF
--- a/src/django_components/util/tag_parser.py
+++ b/src/django_components/util/tag_parser.py
@@ -145,6 +145,7 @@ def parse_tag_attrs(text: str) -> Tuple[str, List[TagAttr]]:
         take_while(TAG_WHITESPACE)
 
         start_index = len(normalized)
+        is_translation = False
 
         # If token starts with a quote, we assume it's a value without key part.
         # e.g. `component 'my_comp'`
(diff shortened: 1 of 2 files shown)


src/django_components/util/tag_parser.py
```diff
@@ -216,7 +222,7 @@ def take_while(tokens: Sequence[Union[str, Tuple[str, ...]]]) -> str:
                 start_index=start_index,
                 quoted=quoted,
                 spread=is_spread,
-                translation=False,
+                translation=is_translation,
```
JuroOravec, line 225: This is the bug. The detected state about whether a variable in a template tag is a translation or not was not being forwarded.

tests/test_tag_parser.py
```diff
@@ -9,96 +9,178 @@
...
+            TagAttr(key=None, value="component", start_index=0, quoted=None, spread=False, translation=False),
+            TagAttr(key=None, value="my_comp", start_index=10, quoted="'", spread=False, translation=False),
+            TagAttr(key="key", value="val", start_index=20, quoted=None, spread=False, translation=False),
+            TagAttr(key="key2", value="val2 two", start_index=28, quoted="'", spread=False, translation=False),
+        ]
+
+        self.assertEqual(attrs, expected_attrs)
         self.assertEqual(
-            attrs,
+            [a.formatted() for a in attrs],
```
JuroOravec, line 22: Updated the tests to also check that we get the original formatting once again when we call `TagAttr.formatted()`

JuroOravec:
@EmilStenstrom Tagging for visibility. Tiny bug, so I'll merge it right away.

Issue #855, opened 2024-12-17
Bug when updating from 0.118 to 0.119.


spapas:
Hey friends, I've got the following code in one of my components:

```
{% component "collapsible-card" "Στοιχεία συντάκτη / εκκινητή διαδικασίας "|add:app.get_initiator_details header_color="bg-info" header_text_color="text-white" border_color="border-info" %}
```

This worked fine on 0.118. When upgrading to 0.119 I get:

```
django.template.exceptions.TemplateSyntaxError: Could not find variable at start of |add:app.get_initiator_details.
```

This can be fixed using a {% with %} however I'd rather avoid changing all my components.

My understanding is that this bug was introduced because of https://github.com/EmilStenstrom/django-components/pull/827. Can you please rollback these breaking changes? 

TIA

EmilStenstrom:
@spapas thanks for reporting! I’d say a better plan to fix this is to add a test for this case (it’s likely missing, and then fix the bug. The fix you are referring to solves other problems that we still want. Could you write up a small test case that we can then fix?

spapas:
Hello @EmilStenstrom the PR of the test case needed is here https://github.com/EmilStenstrom/django-components/pull/856

JuroOravec:
Hi, @spapas for reporting this. It was my bad, I forgot consider Django's template filters in the last refactor.  I wrote fix in https://github.com/EmilStenstrom/django-components/pull/857

JuroOravec:
This has been released in [v0.121](https://github.com/EmilStenstrom/django-components/releases/tag/0.121) :)

PR #857, opened 2024-12-17
refactor: Fix the use of filters with component inputs

JuroOravec:
Closes https://github.com/EmilStenstrom/django-components/issues/855, closes https://github.com/EmilStenstrom/django-components/pull/856

### DIFF
--- a/src/django_components/util/tag_parser.py
+++ b/src/django_components/util/tag_parser.py
@@ -1,37 +1,71 @@
 from dataclasses import dataclass
 from typing import List, Optional, Sequence, Tuple, Union
 
-TAG_WHITESPACE = (" ", "\t", "\n", "\r", "\f")
+TAG_WHITESPACES = (" ", "\t", "\n", "\r", "\f")
+TAG_QUOTES = ("'", '"', '_("', "_('")
+TAG_FILTER_JOINERS = ("|", ":")
+TAG_SPREAD = "..."
 
 
 @dataclass
-class TagAttr:
-    key: Optional[str]
-    value: str
-    start_index: int
+class TagAttrPart:
     """
-    Start index of the attribute (include both key and value),
-    relative to the start of the owner Tag.
+    Django tag attributes may consist of multiple parts, being separated by filter pipes (`|`)
+    or filter arguments (`:`). This class represents a single part of the attribute value.
+
+    E.g. in the following tag:
+    ```django
+    {% component "my_comp" key="my val's" key2=val2|filter1:"one" %}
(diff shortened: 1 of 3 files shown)


src/django_components/util/tag_parser.py
```diff
@@ -1,37 +1,71 @@
...
+TAG_WHITESPACES = (" ", "\t", "\n", "\r", "\f")
+TAG_QUOTES = ("'", '"', '_("', "_('")
+TAG_FILTER_JOINERS = ("|", ":")
+TAG_SPREAD = "..."
 
 
 @dataclass
-class TagAttr:
-    key: Optional[str]
+class TagAttrPart:
```
JuroOravec, line 11: Previously I considered the cases of args (`"abc"`), kwargs (`key=value`), spread (`...my_attrs`) and translation (`_("my_text")`). But I didn't consider that values can have filters with `|`, and filter can have args with `:`.

Before this PR, the implementation expected that the value can be only EITHER string literal like `"abc"`, or a variable / value like `my_var`.

But with filters, both of these can be present simultaneously, e.g. `key="MY_TEXT"|lower` or `key=my_var|yesno:"1,2"`.

So the parser now recognizes the "parts".
- In `key="MY_TEXT"|lower`, the parts are text `"MY_TEXT"` and filter `lower`
- In `key=my_var|yesno:"1,2"`, the parts are variable `my_var`, filter `yesno`, and filter arg text `"1,2"`

src/django_components/util/tag_parser.py
```diff
@@ -138,90 +172,109 @@ def take_while(tokens: Sequence[Union[str, Tuple[str, ...]]]) -> str:
 
         return result
 
-    # Parse
+    def parse_attr_parts() -> List[TagAttrPart]:
```
JuroOravec, line 175: This function was move from [line 189](https://github.com/EmilStenstrom/django-components/pull/857/files#diff-fcef21070412acf8475f5c6880736500491a9b4fbc5d77a37d46dd63814c158fL189) and  and updated.

tests/test_tag_parser.py
```diff
@@ -174,3 +360,75 @@ def test_tag_parser_translation(self):
                 'key=_("two")',
             ],
         )
+
+    def test_filter(self):
```
JuroOravec, line 364: Added this test to ensure filter support works

JuroOravec:
@EmilStenstrom Tagging for visibility, I'll merge this one as it's a bug fix

PR #859, opened 2024-12-19
bug: TemplateDoesNotExist when using extends on main template and two components with same parent template

ldurey:
Hello everyone,

Again i have to thank you for your great work!
The project improve daily and i'm glad to see all the improvements that have been done since June (The last time i came here to report an issue).

I come today with another specific use case which cause an Exception while rendering templates.
I have made a TestCase to try to illustrate my words and to show you the edges of the issue.

## Description
There is a circular issue or maybe a bad cleaning of origin after rendering a component when you try to use two components which `extends` the same template in a template which is himself extended.

- It seems like it doesn't matter if it is the same component or another using the same base template.
- It does not trigger if the two components are in a not extended template, so it seems like it is in a nested extend pattern.

### Non components templates
- `base.html`
```
<!DOCTYPE html>
<html lang="en">
  <body>
    <main role="main">
      <div class='container main-container'>
        {% block body %}
        {% endblock %}
      </div>
    </main>
  </body>
</html>
```
- `main.html`
```
{% extends 'base.html' %}
{% block body %}
	{% component 'test_1' / %}
	{% component 'test_2' / %}
{% endblock %}
```

### Components templates
- `base_component.html`
```
<div>BASE</div>
{% block test %}
{% endblock %}
```
- `test_1.html`
```
{% extends 'base_component.html' %}
{% block test %}
	override 1
{% endblock
```
- `test_2.html`
```
{% extends 'base_component.html' %}
{% block test %}
	override 2
{% endblock
```

## Exception raised
A `TemplateDoesNotExist` is raised when the template loader is trying to find the common template `base_component.html`.
The debug mode shows that this template have been found but it is `Skipped to avoid recursion` even if there is no recursion here.

I'm trying now to find from where it could come from.
Tell me if you need anything else to understand the issue.
Many thanks in advance

### DIFF
--- a/tests/templates/included.html
+++ b/tests/templates/included.html
@@ -1,0 +1,4 @@
+{% extends "simple_template.html" %}
+{% block before_custom %}
+    <div>BLOCK OVERRIDEN</div>
+{% endblock %}
--- a/tests/test_templatetags_extends.py
+++ b/tests/test_templatetags_extends.py
@@ -400,3 +400,39 @@ class ExtendsCompatTests(BaseTestCase):
             </html>
         """
         self.assertHTMLEqual(rendered, expected)
+
+    @parametrize_context_behavior(["isolated", "django"])
+    def test_double_extends_on_main_template_and_nested_component_and_include(self):
+        registry.register("slotted_component", SlottedComponent)
+        registry.register("blocked_and_slotted_component", BlockedAndSlottedComponent)
+
+        @register("extended_component")
+        class _ExtendedComponent(Component):
+            template_name = "included.html"
+
+        template: types.django_html = """
+            {% extends 'block.html' %}
+            {% load component_tags %}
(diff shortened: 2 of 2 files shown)


JuroOravec:
@ldurey Thanks for the high-quality PR for the tests! I might be able to look into this at the end of the week.

ldurey:
My please to help as i can, i just added a commit to show some weird interaction with `include` tag.
My guess from that point is that the problem may come from caching as it does not show up on first render but on the  second one.

ldurey:
Hi, i have made the same conclusion than you about the persistence of `render_context` but i did not find a way to solve it properly.
Thank you for your reactivity.
Have a good end of year celebrations.

JuroOravec:
Happy holidays to you too! :) 🎄

PR #861, opened 2024-12-19
refactor: Backbone for passing JS and CSS variables

JuroOravec:
Stepping stone towards having JS and CSS variables, and other features. This PR adds the internal piping of the JS and CSS variables.

Only change that's public is the removal of the undocumented `Component.component_id`, and instead adding `Component.id`.

After this PR, next step could be already adding support for JS or CSS variables itself. But before that, I first want to allow to define `Component.js/css` in separate files, so that way people could make use of the upcoming features right away.

There's a lot going on, so please see the comments below.


src/django_components/component.py
```diff
@@ -318,7 +325,6 @@ def __init__(
 
         self.registered_name: Optional[str] = registered_name
         self.outer_context: Context = outer_context or Context()
-        self.component_id = component_id or gen_id()
```
JuroOravec, line 321: Previously, there was a not-yet-documented `Component.component_id`, whereas the ID was unique to every instance of class `Component`.

Now, the ID is:
1. Available under `Component.id`
2. Available only during render
3. It is unique for every time we call render

Thus, when we have a forloop:

```django
{% for ... %}
  {% component "table" / %}
{% endfor %}
```

Then each time we iterate over the loop, the component being rendered will have a different ID.

This is important, because each iteration could yield different JS or CSS variables. And so we may need to distinguish between the different renders, and we use the ID for that.

E.g. the resulting HTML may look like this:
```html
<table data-djc-id-1b3caf>
  ...
</table>

<table data-djc-id-7dacf2>
  ...
</table>

<table data-djc-id-2bc3d4>
  ...
</table>
```

src/django_components/provide.py
```diff
@@ -28,10 +27,7 @@ def __init__(
     ):
         super().__init__(nodelist=nodelist, args=None, kwargs=kwargs, node_id=node_id)
 
-        self.nodelist = nodelist
-        self.node_id = node_id or gen_id()
         self.trace_id = trace_id
-        self.kwargs = kwargs or RuntimeKwargs({})
```
JuroOravec, line 34: Not related to the PR, just noticed that this is extra, as the same assignment happens in `super().__init__`

tests/test_attributes.py
```diff
@@ -101,7 +101,7 @@ def get_context_data(self, *args, attrs):
         self.assertHTMLEqual(
             rendered,
             """
-            <div @click.stop="dispatch('click_event')" x-data="{hello: 'world'}" class="padding-top-8 added_class another-class" data-id=123>
+            <div @click.stop="dispatch('click_event')" x-data="{hello: 'world'}" class="padding-top-8 added_class another-class" data-djc-id-a1bc3f data-id=123>
```
JuroOravec, line 104: There's A LOT of changes to the tests, because now, when you render a component, it's root elements will be tagged with the component ID.

So if a component has template

```djangohtml
<div>{{ message }}<div>
<span> Sent from <b>iPhone</b> </span>
```

Then the end result will be

```djangohtml
<div data-djc-id-1bc3df> hi! <div>
<span data-djc-id-1bc3df> Sent from <b>iPhone</b> </span>
```

tests/test_dependencies.py
```diff
@@ -54,7 +61,7 @@ def test_standalone_render_dependencies(self):
         rendered_raw = template.render(Context({}))
 
         # Placeholders
-        self.assertEqual(rendered_raw.count('<link name="CSS_PLACEHOLDER">'), 1)
+        self.assertEqual(rendered_raw.count('<link name="CSS_PLACEHOLDER"/>'), 1)
```
JuroOravec, line 64: Tiny change, but added `/` at the end of the placeholder for the dependencies CSS, so it's properly formatted.

Because otherwise, as I used beautiful soup to parse and serialize the HTML, BS4 added the trailing `/` there anyway upon re-serialization.

tests/test_dependencies.py
```diff
@@ -471,37 +478,59 @@ def test_middleware_response_without_content_type(self):
...
+        registry.register("test-component", component=SimpleComponent)
+        registry.register("test/component", component=SimpleComponent)
+        registry.register("test_component", component=SimpleComponent)
 
-        component_names = [
-            "test-component",
-            "test/component",
-            "test_component",
-        ]
-        for component_name in component_names:
```
JuroOravec, line 484: I refactored this test to remove the for loop. The reason being that in each loop iteration, the resulting HTML would have different component IDs, so it was hard to make sense of which IDs were expected in which iteration.

tests/test_dependency_rendering.py
```diff
@@ -491,3 +491,131 @@ def test_multiple_components_all_placeholders_removed(self):
         template = Template(template_str)
         rendered = create_and_process_template_response(template)
         self.assertNotIn("_RENDERED", rendered)
+
+    def test_adds_component_id_html_attr_single(self):
```
JuroOravec, line 495: Here are tests that are being added as part of this PR

tests/test_dependency_rendering.py
```diff
@@ -491,3 +491,131 @@ def test_multiple_components_all_placeholders_removed(self):
...
+            {% load component_tags %}
+            {% component 'outer' variable='foo' / %}
+        """
+        template = Template(template_str)
+        rendered = create_and_process_template_response(template)
+
+        self.assertHTMLEqual(
+            rendered,
+            """
+            Variable: <strong data-djc-id-a1bc3f data-djc-id-a1bc41>foo</strong>
```
JuroOravec, line 563: Here's an interesting thing.

Vue also uses component IDs format like `data-djc-id-a1bc3f`, so as boolean attributes with no value

At some point, before I realized why, I tried to set the component IDs as key=value attributs, e.g. `data-djc-id="a1bc3f"`.

But the problem with that is that a single HTML element may be a root element for multiple components.

E.g. consider this:

```py
@register("inner")
class Inner(Component):
    template = """
    <div>123</div>
    """

class Outer(Component):
    template = """
    {% component "inner" %}
    """
```

In this case, the `<div>123</div>` will be the root element for both the `Outer` and `Inner` components.

But if the component ID was set as key=value pair, they would overwrite each other. Because the browser parses:

```html
<div data-djc-id="a1bc3f" data-djc-id="d3d3d3">123</div>
```

as:

```html
<div data-djc-id="d3d3d3">123</div>
```

Now, on the other hand, if we define the IDs as boolean attributes, we can have as many of them without fearing that they would overwrite one-another:

```html
<div data-djc-id-a1bc3f data-djc-id-d3d3d3>123</div>
```

tests/test_dependency_rendering_e2e.py
```diff
@@ -106,17 +111,34 @@ async def test_multiple_component_dependencies(self):
...
+            #         variable
+            #     </strong>
+            #     XYZ:
+            #     <strong class="other" data-djc-id-IYirHK>
+            #         variable_inner
+            #     </strong>
+            # </div>
+            # <div class="my-style">123</div>
+            # <div class="my-style2">xyz</div>
+            re.compile(
```
JuroOravec, line 128: Had to update some tests to check via regex because in the E2E tests the component IDs are random.

tests/test_expression.py
```diff
@@ -150,7 +150,7 @@ def get_context_data(
 
         self.assertEqual(
             rendered.strip(),
-            "<!-- _RENDERED SimpleComponent_5b8d97,a1bc3e -->\n                <div>lorem</div>\n                <div>True</div>\n                <div>[{'a': 1}, {'a': 2}]</div>",  # noqa: E501
+            "<!-- _RENDERED SimpleComponent_5b8d97,a1bc3f,, -->\n<div data-djc-id-a1bc3f>lorem</div>\n<div data-djc-id-a1bc3f>True</div>\n<div data-djc-id-a1bc3f>[{'a': 1}, {'a': 2}]</div>",  # noqa: E501
```
JuroOravec, line 153: The HTML comment that we leave behind after a component was rendered, e.g. `<!-- _RENDERED SimpleComponent_5b8d97`, now contains also info about which JS and CSS variables should be loaded with the specific rendering of the component.

tests/test_templatetags_component.py
```diff
@@ -488,75 +509,149 @@ def test_raises_on_invalid_args(self):
...
+            <custom-template data-djc-id-a1bc44>
+                <header>
+                    Default header
+                </header>
                 <main>Default main</main>
                 <footer>Default footer</footer>
             </custom-template>
-        """
-
-    def wrap_with_slot_tags(self, s):
```
JuroOravec, line 523: I refactored these tests to remove the methods like `wrap_with_slot_tags` here, and instead inlined the code into the tests. So the tests are now dumber and more verbose, but easier to read.

src/django_components_js/README.md
```diff
@@ -26,7 +26,7 @@ Components.registerComponentData(
 Components.callComponent(
   "table",  // Component name
   12345,    // Component ID - An HTML element with corresponding
-            //                attribute (`data-comp-id-12345`) MUST
+            //                attribute (`data-djc-id-12345`) MUST
```
JuroOravec, line 29: I changed the attribute that will hold component's IDs from `data-comp-id-` to `data-djc-id-`.

src/django_components/dependencies.py
```diff
@@ -112,23 +111,29 @@ def _hash_comp_cls(comp_cls: Type["Component"]) -> str:
 def _gen_cache_key(
     comp_cls_hash: str,
     script_type: ScriptType,
+    input_hash: Optional[str],
```
JuroOravec, line 114: Many of the functions in this file now accept additional `input_hash` argument, which refers to a hash of the serialized JS or CSS variables.

So e.g. if I defined CSS variables like so:
```py
def get_css_data(self):
    return {"border_width": "3px", "button_color": "blue" }
```

Then this would be serialized into JSON as:
```json
{
  "border_width": "3px",
  "button_color": "blue"
}
```

And since JSON is a string, the input hash would be derived from the JSON as:
```py
input_hash = md5(json_data.encode()).hexdigest()[0:6]
```

So the first 6 characters of the MD5 hash.

This is then passed around as the ID that points to the file that holds the CSS / JS variables.

E.g. If I had a component `Table`, and I generated the CSS variables above, and I got hash `a1b2c3`, then I would know that:

1. An endpoint like `/components/cache/table.css` would fetch the CSS from `Table.css`
2. An endpoint like `/components/cache/table.a1b2c3.css` would fetch the CSS variables for component `Table` as described above.

src/django_components/dependencies.py
```diff
@@ -147,32 +152,114 @@ def _cache_script(
...
 
-    # Prepare the script that's common to all instances of the same component
-    # E.g. `my_table.js`
-    if not _is_script_in_cache(comp_cls, "js"):
+    # Generate and cache a JS script that contains the JS variables.
+    if not _is_script_in_cache(comp_cls, "js", input_hash):
         _cache_script(
             comp_cls=comp_cls,
-            script=content,
+            script="",  # TODO
```
JuroOravec, line 198: One thing that's missing from `cache_component_js_vars` and `cache_component_css_vars` is that we need to pre-process the JS / CSS vars into the correct shape so they could be imported.

E.g. for CSS, when we get a dictionary of values, we actually want to format it as proper CSS, so as
```css
[data-djc-css-a1b2c3] {
  --border_width: 3px;
  --button_color: blue;
}
```

I didn't include that in this PR for simplicity, hence why the `script="",  # TODO`.

This will be included in the PRs that actually add the JS / CSS vars features.

src/django_components/dependencies.py
```diff
@@ -595,7 +755,7 @@ def _postprocess_media_tags(
 
 
 def _prepare_tags_and_urls(
-    data: Iterable[str],
+    data: List[Tuple[str, ScriptType, Optional[str]]],
```
JuroOravec, line 758: Changed how the data is passed to this function, so we don't needlessly generate `<script>` or `<link>` tags for JS / CSS variables, when the component doesn't define any.  But the logic is still the same

src/django_components/dependencies.py
```diff
@@ -266,21 +389,30 @@ def postprocess_component_html(
...
+# E.g. `data-djc-id-a1b2c3`
+MAYBE_COMP_ID = r"(?: data-djc-id-\w{6})?"
+# E.g. `data-djc-css-99914b`
+MAYBE_COMP_CSS_ID = r"(?: data-djc-css-\w{6})?"
+
 PLACEHOLDER_REGEX = re.compile(
     r"{css_placeholder}|{js_placeholder}".format(
-        css_placeholder=CSS_DEPENDENCY_PLACEHOLDER,
-        js_placeholder=JS_DEPENDENCY_PLACEHOLDER,
+        css_placeholder=f'<link{MAYBE_COMP_CSS_ID}{MAYBE_COMP_ID} name="{CSS_PLACEHOLDER_NAME}"/>',
```
JuroOravec, line 414: One peculiar thing that now happens is that we add the `data-djc-id-` attributes to the root elements. But if we define the `{% component_js/css_dependencies %}` template tags at the root of component's template, e.g.

```py
class Table(Component):
    template = """
    {% component_js_dependencies %}
    <div> Hello </div>
    """
```

Then since `{% component_js_dependencies %}` simply renders a dummy `<script>` tag, then this script tag will be assigned with `data-djc-id-` too.

And so in the regex for finding and removing these placeholder tags, we have to account for that the tags may have that attribute.

src/django_components/dependencies.py
```diff
@@ -183,31 +270,58 @@ def cache_inlined_css(comp_cls: Type["Component"], content: str) -> None:
...
+        #
+        # NOTE: We use `data-djc-css-a1b2c3` and `data-djc-id-a1b2c3` instead of
+        # `data-djc-css="a1b2c3"` and `data-djc-id="a1b2c3"`, to allow
+        # multiple values to be associated with the same element, which may happen if
+        # One component renders another.
+        elem.set_attr(f"data-djc-id-{component_id}", True)
+
+        # Attribute by which we bind the CSS variables to the component's CSS,
+        # e.g. `data-djc-css-a1b2c3`
+        if css_input_hash:
```
JuroOravec, line 296: We now add `data-djc-id-xxxxxx` to each component, where `xxxxxx` stands for the component's render ID.

If the component defines CSS variables (and defines `Component.css`), we also insert `data-djc-css-yyyyyy` attribute to tell the browser which HTML elements should have access to the CSS variables identified by the `yyyyyy` hash.

tests/test_expression.py
EmilStenstrom: What do the three values mean? Is it just a list of component ids?

tests/test_expression.py
```diff
@@ -220,7 +220,13 @@ def get_context_data(
 
         self.assertEqual(
             rendered.strip(),
-            "<!-- _RENDERED SimpleComponent_743413,a1bc3e -->\n                <div>lorem ipsum dolor</div>\n                <div>True</div>\n                <div>[{'a': 1}, {'a': 2}]</div>\n                <div>{'a': 3}</div>",  # noqa E501
+            (
+                "<!-- _RENDERED SimpleComponent_743413,a1bc3f,, -->\n"
```
EmilStenstrom, line 224: Why is this comment ending with two commas?

EmilStenstrom (commented):
Couple of questions in the comments, but generally looks good!

tests/test_expression.py
JuroOravec: Fields are:
- `comp_cls_hash` - Cache key of the component class that was rendered
- `id` - Component render ID - Will be used for passing JS data to individual component instances.
- `js` - Cache key for the JS data from `get_js_data()` - part of JS variables feature
- `css` - Cache key for the CSS data from `get_css_data()` - part of CSS variables feature

I added a comment here to make it clear https://github.com/EmilStenstrom/django-components/pull/861/files#diff-371d604ce82d2a01458065f1b137fd981b12af1b0b5c5fb2c1f7f5b45c20003fR403

src/django_components/component.py
JuroOravec: Yeah, for CSS I do it that way, that if the component doesn't define any CSS variables, then there's no need for `data-djc-css-...` attribute.

The `data-djc-id-...` attribute is needed though. Because I'm using it also in the CSS scoping feature to avoid applying the scoping attribute to nested components.

E.g. if I have a component like this

```py
class MyTable(Component):
    template = """
        <h1>
          <strong> Hi! </strong>
        </h1>
        <div class="x">
          {% component "button" / %}
        </div>
    """
    css = """
        .x {
          color: red;
        }
    """
    css_scoped = True
```

Then the insides of the template will be transformed to:

```html
<h1 data-djc-scope-1b2c3d>
  <strong data-djc-scope-1b2c3d> Hi! </strong>
</h1>
<div class="x" data-djc-scope-1b2c3d>
 ...
</div>
```

BUT we don't want to apply `data-djc-scope-1b2c3d` to the insides of `{% component "button" / %}`.

Since I know that all components will have the `data-djc-id-xxx`, then for the CSS scoping, I walk down the HTML tree, and keep applying `data-djc-scope-1b2c3d` until I come across `data-djc-id`, in which case I stop walking down that branch. So the final HTML may then look like this:

```html
<h1 data-djc-id-12345 data-djc-scope-1b2c3d>
  <strong data-djc-scope-1b2c3d> Hi! </strong>
</h1>
<div class="x" data-djc-id-12345 data-djc-scope-1b2c3d>
  <!-- START OF {% component "button" / %} -->
  <button data-djc-id-efefef>
    <strong>Click me!</strong>
  </button>
  <!-- END OF {% component "button" / %} -->
</div>

File: tests/test_component_as_view.py [hunk 18]
```diff
@@ -56,9 +56,9 @@ def render_template_view(request):
         client = CustomClient(urlpatterns=[path("test_template/", render_template_view)])
         response = client.get("/test_template/")
         self.assertEqual(response.status_code, 200)
-        self.assertIn(
-            b'<input type="text" name="variable" value="TEMPLATE">',
-            response.content,
+        self.assertInHTML(
+            '<input type="text" name="variable" value="TEMPLATE">',
+            response.content.decode(),
```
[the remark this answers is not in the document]
JuroOravec: I change the assertion from `self.assertIn()` to `self.assertInHTML()` because if I remember correctly `assertInHTML` allows the HTML attributes to be in any order, and pre-processing the HTML with BS4 sometimes changed the order of the attributes.

And `assertInHTML()` accepts only strings, thus the `.decode()`.

tests/test_dependencies.py
```diff
@@ -471,37 +478,59 @@ def test_middleware_response_without_content_type(self):
...
-            self.assertEqual(rendered.count("Variable: <strong>value</strong>"), 1)
+        rendered1 = create_and_process_template_response(
+            template,
+            context=Context({"component_name": "test-component"}),
+        )
+
+        assert_dependencies(rendered1)
+        self.assertEqual(
+            rendered1.count('Variable: <strong data-djc-id-a1bc41 data-djc-id-a1bc42="">value</strong>'),
+            1,
```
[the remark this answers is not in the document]
JuroOravec: There's a couple of patterns to unpack in

    rendered1.count('Variable: <strong data-djc-id-a1bc41 data-djc-id-a1bc42="">value</strong>')

1. Where possible I used `self.assertInHTML` or `self.assertHTMLEqual`, but I think the latter doesn't work when I want to search for a combination of text + HTML, e.g. like how above I searched for `Variable: <strong>...`.

   So in those cases I opted for `self.assertEqual(str.count('Variable: <strong>...), 1)`.

   But when using `str.count()` or `self.assertEqual(str)`, I have to put in there the exact same string.

   On the other hand, `self.assertHTMLEqual()` includes some nicities like:
   - HTML attributes can be in any order
   - Whitespace between elements is ignored
   - Boolean HTML attributes like `data-djc-id-a1bc42=""` and `data-djc-id-a1bc42` are considered as identical

   So, because I used `str.count()`, I had to write extact string, so I had to write `data-djc-id-a1bc42=""` instead of simply `data-djc-id-a1bc42`.

2. As for why there are even both `data-djc-id-a1bc42=""` and `data-djc-id-a1bc42` formats present:

   When we use BS4 to process the HTML to insert the ID attributes, then what I think happens is that, the first time the boolean attribute like `data-djc-id-a1bc42` is inserted, it's inserted without the value part (without `=""`). But when HTML already contains boolean HTML attributes without values, then BS4 will re-format them into `data-djc-id-a1bc42=""`. Semantically they are the same.

3. And as for why both `data-djc-id-a1bc41` and `data-djc-id-a1bc42=""` (two different IDs) are on the same HTML element is because that `<strong>` is the root HTML element for two components, where one is nested in another, e.g. something like this:

   ```py
   @register("outer")
   class Outer(Component):
       template = """
         {% component "inner" / %}
       """
   
   @register("inner")
   class Inner(Component):
       template = """
          <strong>
            Hi!
          </strong>
       """
   ```

tests/test_expression.py
JuroOravec: I treat the text after the `_RENDERED` as a row of CSS ([as described here](https://github.com/EmilStenstrom/django-components/pull/861#discussion_r1898891705)). So the two commas means that there's no data for those fields, which means that the rendered component doesn't define any JS nor any CSS variables.

tests/test_dependencies.py
JuroOravec: Oh nice! I just tried it, and using `html5` will also make BS4 stop adding empty value (`=""`) to boolean attributes (e.g. `data-djc-id-a1bc40=""`).

But it also breaks around 30 tests where the `data-djc-id-a1bc40=""` will need to be updated to `data-djc-id-a1bc40`. So I would make this change in a separate PR for clarity.

tests/test_expression.py
JuroOravec: Btw I just realized that, since the "component ID" is now different for each time we call `render()`, then the way we pass the information above could be simplified, such that:

1. We'd keep only the render ID inside the HTML:
    ```html
    <!-- _RENDERED a1bc3e -->
    ```
2. When a component is rendered, we'd cache the other related data (`comp_cls_hash`, `js`, `css`) in an actual cache, instead of inserting it into the HTML:
    ```py
    cache.set(
        "a1bc3e",
        {"comp_cls_hash": "SimpleComponent_5b8d97", "js": None, "css": None},
    )
    ```
3. Then, when the rendered HTML gets processed by the middleware or `render_dependencies()`, it would extract only the render ID from `<!-- _RENDERED a1bc3e -->`, and get (pop) the rest of the data from the cache.

This was not possible before, when multiple calls to `render()` used the same `component_id`, because it wasn't possible to tell how long the entries in the cache would have to remain alive for.

I won't make the changes in this PR, tho, but sometime later.

JuroOravec:
I resolved conflicts, after which I had to update tests in [tests/test_templatetags_extends.py](https://github.com/EmilStenstrom/django-components/pull/861/commits/0e4e049deb5fd818a7eadb83eedbb01b34d85c71).

@EmilStenstrom I think this now good to go, let me know if you have any more questions

tests/test_dependencies.py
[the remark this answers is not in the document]
EmilStenstrom: Thanks, I should have been more explicit in my question. I was wondering about the id="" part. But this seems to be a quirk of the bs4 parser, check.

tests/test_expression.py
EmilStenstrom: Using the built-in Django cache requires configuration on the user part, so I'm not sure the trade-off is worth it. Let's talk in a separate PR about that :)

tests/test_dependencies.py
EmilStenstrom: Happy to hear that using html5 as a formatter solves those small annoyances. Will also make things easier if we switch to another (possibly home-brewd) parser in the future.

EmilStenstrom:
@JuroOravec Happy to have this merged! :)

tests/test_expression.py
JuroOravec: Ah, by cache here I meant just a simple dictionary :D

PR #862, opened 2024-12-23
fix: TemplateDoesNotExist when using {% extends %} on main template and two components with same parent template

JuroOravec:
See here for the explanation of the bug and the fix https://github.com/EmilStenstrom/django-components/pull/859#issuecomment-2559487398

Closes https://github.com/EmilStenstrom/django-components/pull/859


tests/test_templatetags.py
```diff
@@ -94,396 +94,6 @@ def test_nested_component_templates_all_shown_as_used(self):
         self.assertIn("simple_template.html", templates_used)
 
 
-class BlockCompatTests(BaseTestCase):
```
JuroOravec, line 97: Since https://github.com/EmilStenstrom/django-components/pull/859 introduced some tests for compatibility with `{% extends %}` in the file `test_templatetags_extends.py`, I moved also other existing tests there too.

JuroOravec:
@EmilStenstrom Tagging for visibility

PR #870, opened 2024-12-28
feat: allow to set main JS and CSS from files + lazy-load component m…

### LINKED ISSUE #807: Resolve component HTML / JS / CSS paths lazily
As mentioned in https://github.com/EmilStenstrom/django-components/discussions/790#discussioncomment-11421036, I'm having an issue creating a package with a standalone component, because the component's `template_name` is being resolved at class creation. But for some reason this happens before Django settings are resolved, and so it raises an error.

I could create the component by inlining the HTML, but I'd rather fix this.

The solution is to resolve the paths like `Component.template_name` only once they are actually required, which is when:
1. We're rendering the component for the first time
2. We need to access `Component.media`

Lazily resolving the filepaths might also give some performance boost for larger project, although the first render might be a bit slower. But this is only in theory. I don't have the performance metrics, so the difference may as well be tiny.

JuroOravec:
This PR is mainly abouut 2 things:

### 1. Allowing to define the "main" or "inlined" CSS / JS (`Component.js/css`) in separate files, similarly how we link to JS / CSS files in `Media.js/css`

   So while before the HTML, JS and CSS could be inlined like this:

   ```py
   class Calendar(Component):
       template = """
         <div class="calendar">
           Today's date is <span>{{ date }}</span>
         </div>
       """

       css = """
         .calendar {
           width: 200px;
           background: pink;
         }
       """

       js = """
         document.querySelector(".calendar").onclick = function () {
           alert("Clicked calendar!");
         };
       """
   ```

   Now we will be able to move them all to separate files:
   ```py
   class Calendar(Component):
       template_name = "calendar.html"
       js_file = "calendar.js"
       css_file = "calendar.css"
   ```

   Previously we used `Media.js/css` for that:
   ```py
   class Calendar(Component):
       template_name = "calendar.html"
       
       class Media:
           js = "calendar.js"
           css = "calendar.css"
   ```

   But with the oncoming features, there will be differences between these two:
   - JS and CSS files linked to in `Media` will NOT be modified in any way
   - JS and CSS in `Component.js/css` or `Component.js/css_file` WILL be used with (and potentially modified for):
     - JS / CSS variables
     - CSS scoping
     - Wrapping JS in self-invoking function

   So the role of `Media.js/css` will become more about providing "additional" JS / CSS, while the "main" JS / CSS will be in `Component.js/css` or `Component.js/css_file`.

   So large part of this PR is also about updating the docs to preferrentially use `Component.js/css` or `Component.js/css_file`.

### 2. Lazily resolve component HTML / JS / CSS paths (#807)

At first I didn't mean to address this in this PR, but while working on the first part, I bumped into this issue again.

As a reminder, all the CSS/JS/HTML associated with a component - e.g. the `js`, `js_file`, `template_name` or `Media` class,
are all class attributes. And some of these attributes need to be resolved, e.g. to find the files
that `js_file`, `css_file` and `template_name` point to.

Some of the resolutions we need to do is:
- Component's HTML/JS/CSS files can be defined as relative to the component class file. So for each file,
  we check the relative path points to an actual file, and if so, we use that path.
- If the component defines `js_file` or `css_file`, we load the content of the file and set it to `js` or `css`.
  - Note 1: These file paths still may be relative to the component, so the paths are resolved as above,
            before we load the content.

At the same time, we need to do so lazily, otherwise we hit a problem with circular imports when trying to
use Django settings. This is because the settings are not available at the time when the component class is defined
(Assuming that components are defined at the top-level of modules).

We achieve this by:
1. At class creation, we define a private `ComponentMedia` object that holds all the media-related attributes.
3. At the same time, we replace the actual media-related attributes (like `js`) with descriptors that intercept the access to them.
4. When the user tries to access the media-related attributes, we resolve the media files if they haven't been resolved yet.
5. Any further access to the media-related attributes will return the resolved values.

Closes #807

This will also unblock creation of third-party packages, like what I tried for [an icons package](https://github.com/EmilStenstrom/django-components/discussions/790#discussioncomment-11421036).


README.md
```diff
@@ -6,6 +6,78 @@
...
+  alert("Clicked calendar!");
+};
+```
+
+```py title="calendar.py"
+from django_components import Component
+
+class Calendar(Component):
+    template_name = "calendar.html"
+    js_file = "calendar.js"
```
JuroOravec, line 50: At first I wanted to go with `js_name` and `css_name` as attribute names, so it would be in line with `template` and `template_name`. But TBH, `js_name` doesn't sound intuitive, while `js_file` is more self-explanatory. I'm also considering `js_src`, or open to other ideas.

README.md
```diff
@@ -6,6 +6,78 @@
 
 Django-components is a package that introduces component-based architecture to Django's server-side rendering. It aims to combine Django's templating system with the modularity seen in modern frontend frameworks.
 
+A component in django-components can be as simple as a Django template and Python code to declare the component:
```
JuroOravec, line 9: Recently when I shared an update on Reddit about new feature in django-components, there was one guy who thought, based on the link I shared, that django-components supports only the "inlined" HTML/CSS/JS. What probably led him to it is that for most of the documentation I wrote, I used the inlined format.

So I think it would be good to show both formats right at the start. So I added this section both here to the root README, and to the documentation welcome page (`welcome.md`).

docs/concepts/getting_started/adding_js_and_css.md
```diff
@@ -177,25 +177,103 @@ So in your HTML, you may see something like this:
...
+    - CSS scoping [a la Vue](https://vuejs.org/api/sfc-css-features.html#scoped-css)
+    - CSS variables from Python are available
+    - JS variables from Python are available
+    - JS can pass a callback to special JS method `$onLoad()`, which will be called every time
+      a component is rendered on the page.
+
+    This is not true for JS and CSS defined in `Media.js/css`, where the linked JS / CSS are rendered as is.
+-->
+
+### 5. Link additional JS and CSS to a component
```
JuroOravec, line 228: Added this section to the tutorial which now describes the files linked to with `Media.js/css` as "additional JS/CSS"

README.md
EmilStenstrom: This is tricky. template_name is used for Django's generic views and should be familiar to Django users. I guess the idea is that you just give the file *name*, and then the template finders will expand that to a path, and fetch the actual file. name -> path -> file. In that context, I think js_(file)name makes sense.

But just looking at "js_name", names usually don't map to files in the file system. It could mean a name of a variable, a function, an alias for something. Not knowing that we are referring to a filename likely makes this confusing for beginners. 

And js_filename is too long. So I'm leaning towards js_file, just like you do.

What's left is the inconsistency of template_name and js_file. I think the easiest way forward for that is to add a template_file alias, that does the same thing as template_name. Then things would make sense imho.

README.md
```diff
@@ -6,6 +6,78 @@
...
+```py title="calendar.py"
+from django_components import Component
+
+class Calendar(Component):
+    template_name = "calendar.html"
+    js_file = "calendar.js"
+    css_file = "calendar.css"
+```
+
+Alternatively, you can "inline" HTML, JS, and CSS right into the component class:
```
EmilStenstrom, line 54: Not to over-index on something one guy said on reddit, I think this format should be the default. It makes it very easy to get a full overview of what a component is doing, without opening four parallel files and trying to pattern-match across them.

docs/concepts/getting_started/adding_js_and_css.md
```diff
@@ -177,25 +177,103 @@ So in your HTML, you may see something like this:
...
+    (e.g. `[your apps]/components` dir and `[project root]/components`)
+    3. Relative to any of the directories defined by `STATICFILES_DIRS`.
+
+
+<!-- TODO: UPDATE AFTER AT LEAST ONE IMPLEMENTED
+!!! info
+
+    Special role of `css` and `js`:
+
+    The "primary" JS and CSS you that specify via `js/css` and `js/css_file` have special role in multitudes of django-components'
```
EmilStenstrom, line 217: Nit: js_file/css_file

EmilStenstrom: multitudes -> many

docs/concepts/getting_started/adding_js_and_css.md
```diff
@@ -223,18 +301,57 @@ class Calendar(Component):
...
+        document.querySelector(".calendar").onclick = () => {
+          alert("Clicked calendar!");
+        };
+      })();
+    </script>
+  </body>
+</html>
+```
+
+---
```
EmilStenstrom, line 355: Very nice explanation!

sampleproject/components/calendar/calendar.py
```diff
@@ -8,9 +8,9 @@ class Calendar(Component):
     #
     # `template_name` can be relative to dir where `calendar.py` is, or relative to COMPONENTS.dirs
     template_name = "calendar/calendar.html"
-    # Or
-    # def get_template_name(context):
-    #     return f"template-{context['name']}.html"
+
+    css_file = "calendar/calendar.css"
+    js = "calendar/calendar.js"
```
EmilStenstrom, line 13: js_file, why doesn't this break?

docs/concepts/fundamentals/defining_js_css_html_files.md
```diff
@@ -40,17 +38,24 @@ from django_components import Component, register
...
+    css_file = "calendar/style.css"
+    js_file = "calendar/script.js"
 ```
 
 NOTE: In case of conflict, the preference goes to resolving the files relative to the component's directory.
 
 ## Defining multiple paths
 
-Each component can have only a single template. However, you can define as many JS or CSS files as you want using a list.
+Each component can have only a single template, and single main JS and CSS. However, you can define additional JS or CSS
```
EmilStenstrom, line 49: Why are we not allowing multiple files here?

EmilStenstrom (approved):
Really nice! I think we should add a template_file alias for consistency, and shift the docs towards the inline version, with external files always afterwards as an option. None of them are blocking merge.

docs/concepts/fundamentals/defining_js_css_html_files.md
JuroOravec: IMO complexity of supporting multiple "main" JS or CSS is not worth it.

Might not seem like that right now, but down the line there are 2 features where it could be problematic:

1. For the JS variables feature, with only 1 file allowed for the main JS / CSS, there is currently a single entrypoint to the JS variables (`$onLoad`), e.g.:

   ```py
   class MyComponent(Component):
       js = """
       $onLoad(({ data }) => {
         console.log(data);
       });
       """
   ```

   If we allowed multiple files, then the `$onLoad` could be triggered from any of the files, and it feels more confusing to me.

2. Support for "pre-compilers"

   I also want to have one-to-one mapping between the inlined code, e.g. `Component.template`, and the file, e.g. `Component.template_name`, because down the line I'd want to have support for "pre-compilers".

   E.g. that one could write the "main" JS code block as TypeScript, the CSS block as SASS, or the HTML block as Markdown. And all they would need for that would be to install a third-party package, and plug it into django-components (similarly how django-components is plugged into Django).

   I got a proof of concept working with TypeScript. There it looked something like this:
   ```py
   class MyComponent(Component):
       js_lang = "ts"
       js = """
       function add(a: onLoad, b: number) {
           return a + b;
       }
       """
   ```

   And as part of the flow, if the TypeScript code was initially inlined, then I had to export it into the static directory.

   And while overall it could eventually work even with multiple "main" JS / CSS files, it sure would be a lot more complex.

README.md
JuroOravec: Good idea, I'll make an alias for `template_file`.

docs/concepts/getting_started/adding_js_and_css.md
JuroOravec: Fixed, thanks!

sampleproject/components/calendar/calendar.py
JuroOravec: I did actually break, thanks for spotting that :D

JuroOravec:
@EmilStenstrom Thanks for having a look at this one! I'll add the `template_file` alias in next MR.

TBH it took me until I was writing up the documentation to fully grasp the changes introduced with `js/css_file`  and the lazy loading. So there will be a few more MRs in this area:

1. Describe the lazy-loading in the documentation, because it's not mentioned anywhere yet.

2. Add the `template_file` alias to align it with `js_file` and `css_file`

3. Assign the file content to `Component.template` when the file is defined with `template_name`, to align it with `js/js_file` and `css/css_file`. And test this behavior.

4. Add support for Django's [`Media.extend`](https://docs.djangoproject.com/en/5.1/topics/forms/media/#extend) for completeness, with the difference that in django-components, one would specify boolean or list of Components. And thus, when one Component subclasses another, the child component will contain all auxiliary JS / CSS from `Media.js/css` from all its parents.

5. Test and document subclassing - Because with `Media.extend`, and having multiple mutually-exclusive attributes defining the HTML / CSS / JS (`js` vs `js_file`), subclassing will work a bit different - e.g. if parent class contains EITHER `template` OR `template_name`, then we assume that that parent DOES define HTML. And we keep searching up the parent tree only if the parent doesn't have NEITHER.

docs/concepts/fundamentals/defining_js_css_html_files.md
EmilStenstrom: I wonder if we could have python functions as the API for both of those features...

```python
class MyComponent(Component):
    js = Component.js_onload("""
      console.log(data);
    """)

class MyComponent(Component):
    js = Component.ts_to_js("""
    function add(a: onLoad, b: number) {
        return a + b;
    }
    """)
```

If that was the case, I'm thinking we could allow those methods to take files as well:

```python
    js_file = [Component.js_onload("file.js"), "file2.js"]
```

EmilStenstrom:
>     2. Add the `template_file` alias to align it with `js_file` and `css_file`

Maybe we should think about it as changing the default to template_file, and aliasing in template_name for backwards compatibility?

docs/concepts/fundamentals/defining_js_css_html_files.md
JuroOravec: Sorry, just noticed your reply. Interesting takes though!

- Ultimately I think it should be possible to have lists for `js_file` and others.

- As for `js_onload()`, I think that kinda blurs the line between `Component.js_file` and `Component.Media.js`. Instead, I think that if we defined

   ```js
   js_file = ["file.js", "file2.js"]
   ```

    we could assume that both `file.js` and `file2.js` would be able to make use of `$onLoad()` within their body.

   If we instead had `Component.js_onload()`, it's not clear to me when should `file2.js` run - Because since it's defined in `js_file`, I would guess that it would run AFTER `file.js`, HOWEVER, since it doesn't need to wait for the component data, there's no reason why it couldn't run at the same time as when we execute JS from `Component.Media.js`.

- As for using TS files, I think what might be possible eventually is to simply specify the TS file, e.g. like:

   ```js
   js_file = ["file.ts", "file2.js"]
   ```

   Because how I'm currently thinking of the implementation is that there might need to be something similar to the [JSON](https://github.com/EmilStenstrom/django-components/blob/5fd45ab424908f7c3565139709c8eae86bdd9533/tests/static_root/staticfiles.json) generated with Django's [ManifestStaticFilesStorage](https://docs.djangoproject.com/en/5.1/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage). The JSON would track that, if I had `file.ts`, which JS files would that `file.ts` be split into.

   E.g. in the example below, both `file.ts` and `file2.ts` had shared code, which Esbuild extracted into separate file `chunk_0eeb72042b59.js`. This would be represented by a JSON like so:

   ```json
	{
	    "file.ts": {
            "entrypoint": "path/to/compiled/file.js",
            "chunks": ["path/to/compiled/chunk_0eeb72042b59.js"],
        },
	    "file2.ts": {
            "entrypoint": "path/to/compiled/file2.js",
            "chunks": ["path/to/compiled/chunk_0eeb72042b59.js"],
        },
	}
   ```

   And if there was such JSON, then at runtime, we'd come across `MyComponent.js_file`, and see the `file.ts` there. And we'd:
   1. Look up the corresponding entry in the JSON
   2. Found that entrypoint for `file.ts` is `path/to/compiled/file.js`
   3. Found that `file.ts` had some chunks, whch we'd insert into `Component.Media.js`

   So if we had a component like:
   ```py
   class MyTable(Component):
       js_file = ["file.ts", "file2".ts]
   ```

   Then, at runtime at the same time as when we lazy-resolve the media, we'd transform the component into:
   ```py
   class MyTable(Component):
       js_file = ["path/to/compiled/file.js", "path/to/compiled/file2.js"]
       class Media:
           js = ["path/to/compiled/chunk_0eeb72042b59.js"]
   ```

   Tho, we'll how things will pan out once productionized.

JuroOravec: Hm, btw one more argument against `js_file/css_file` being a list is that `template_file` cannot be a list. There should be only a single template for a component.

EmilStenstrom: Agreed, let’s skip multiple files for now.

About $onload, I assumed that was an entry point that would be called when all other files where loaded, like ’<body onload=func>’. If we had two files, we could still have dependencies that used JS or css variables to load, but only one of them would be the entry point for starting the execution. I might be missing something basic, but that was my expectation.

About using functions: the idea was simply to make it transparent what’s happening. It would of course be possible through some plugin + config setup, but that would be less transparent imho. not that important, I just liked that API.

Do you have some other idea in mind for how to apply a plugin to just some files in some components?

JuroOravec: > About $onload, I assumed that was an entry point that would be called when all other files where loaded

Yup, I'm thinking the same.

> If we had two files, we could still have dependencies that used JS or css variables to load, but only one of them would be the entry point for starting the execution. I might be missing something basic, but that was my expectation.

Accessing JS and "starting the execution" both occur within `$onLoad`. In other words, you cannot access the JS variables outside or without `$onLoad`. So if we had two files in `js_file`, it'd make sense to allow both of them to use `$onLoad`, so that both files can access the JS variables.

CSS case is a bit different, because as long as the CSS rules select the component's HTML, the CSS rule will have access to those CSS variables.


<details>
<summary>
(Actually, over the holidays I dumped my working branch <a href="https://github.com/JuroOravec/django-components/tree/dev-compilers" target="_blank">here</a>, so I can show you how it works 😄 Click to expand)

</summary>
</br>

I initially implemented it to have a single `$onLoad` per component class.

How it's now:

1. So the `$onLoad` would simply wrap the `registerComponent()` function defined in `manager.ts` ([source](https://github.com/JuroOravec/django-components/blob/7462b1b58344fd30ed036e57e6537190b0a0305b/src/django_components/dependencies.py#L183)).

   User would use the `$onLoad`, passing it a callback. And within this callback, they'll be able to access the data defined in `Component.get_js_data()`, e.g.:
   ```js
   $onLoad(({ data }) => {
     console.log(data);
   })
   ```

2. `registerComponent()` simply stores the callback for later ([source](https://github.com/JuroOravec/django-components/blob/7462b1b58344fd30ed036e57e6537190b0a0305b/src/django_components_js/src/manager.ts#L227))

3. Lastly, when we actually get around to calling the components `$onLoad` callbacks, we access the stored callback and call it ([source](https://github.com/JuroOravec/django-components/blob/7462b1b58344fd30ed036e57e6537190b0a0305b/src/django_components_js/src/manager.ts#L242))

Now, the above could be easily transformed to support multiple callbacks for `$onLoad`. So it would then work similarly to registering listeners in JS with [`addEventListener()`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener), where you can attach multiple listeners to the same event ([See StackOverflow](https://stackoverflow.com/a/5411130/9788634)):

```js
const el = document.querySelector(".calendar");
el.addEventListener("click", () => console.log("Clicked!")) 
el.addEventListener("click", () => console.log("Yahoo!")) 
```

To implement that, we'd set it such that `registerComponent` doesn't store a single callback for a given component, but an array of callbacks. Thus, when we get around to calling `$onLoad` callbacks, we'd call them all one after another.

</details>


> Do you have some other idea in mind for how to apply a plugin to just some files in some components?

Are you asking with regard to `$onLoad`, or the TypeScript files?

PR #877, opened 2024-12-30
docs: add documentation around component media

JuroOravec:
Adds missing documentation for https://github.com/EmilStenstrom/django-components/pull/870

### DIFF
--- a/docs/css/style.css
+++ b/docs/css/style.css
@@ -11,6 +11,19 @@ h6 {
   padding-top: 40px;
 }
 
+.md-typeset h3 {
+  /* Original styling */
+  font-size: 1.25em;
+  font-weight: 400;
+  letter-spacing: -.01em;
+  line-height: 1.5;
+  margin: 1.6em 0 0.8em;
+
+  /* Custom */
+  border-top: 0.5px solid var(--md-typeset-color);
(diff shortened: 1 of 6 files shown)


docs/css/style.css
```diff
@@ -11,6 +11,19 @@ h6 {
   padding-top: 40px;
 }
 
+.md-typeset h3 {
```
JuroOravec, line 14: Added this so that there's a horizontal line even between `<h3>` headers

src/django_components/__init__.py
```diff
@@ -44,6 +45,8 @@
     "Component",
     "ComponentFileEntry",
     "ComponentFormatter",
+    "ComponentMediaInput",
+    "ComponentMediaInputPath",
```
JuroOravec, line 49: Exported these, so that the typings for `Media.js` and `Media.css` show up in the API reference.

JuroOravec:
Some examples:

Screenshot 2024-12-30 at 10 57 31


https://github.com/user-attachments/assets/00654e00-37c7-4aed-bf06-611819f4cbaa

EmilStenstrom (approved):
Looks great, thanks for the video (including background music!)

PR #878, opened 2024-12-31
refactor: rename template_name to template_file

JuroOravec:
Follow up from https://github.com/EmilStenstrom/django-components/pull/870#issuecomment-2565828739

Changes:
- Renamed `Component.template_name` to `Component.template_file`
- Some linting changes here and there, triggered by search-and-replace
- Added new `template_name` which forward all access and setting to `template_file` via descriptors (see https://github.com/EmilStenstrom/django-components/pull/878/commits/7cb60c89bd0b5a42913c29279ac3ee8a5af41c9c)

### DIFF
--- a/docs/concepts/fundamentals/single_file_components.md
+++ b/docs/concepts/fundamentals/single_file_components.md
@@ -3,7 +3,7 @@ title: Single-file components
 weight: 1
 ---
 
-Components can be defined in a single file, which is useful for small components. To do this, you can use the `template`, `js`, and `css` class attributes instead of the `template_name` and `Media`. For example, here's the calendar component from above, defined in a single file:
+Components can be defined in a single file, which is useful for small components. To do this, you can use the `template`, `js`, and `css` class attributes instead of the `template_file` and `Media`. For example, here's the calendar component from above, defined in a single file:
 
 ```python title="[project root]/components/calendar.py"
 # In a file called [project root]/components/calendar.py
--- a/docs/concepts/fundamentals/single_file_components.md
+++ b/docs/concepts/fundamentals/single_file_components.md
@@ -3,10 +3,13 @@ title: Single-file components
 weight: 1
 ---
 
-Components can be defined in a single file, which is useful for small components. To do this, you can use the `template`, `js`, and `css` class attributes instead of the `template_file` and `Media`. For example, here's the calendar component from above, defined in a single file:
+Components can be defined in a single file, which is useful for small components. To do this, you can use the `template`, `js`, and `css` class attributes instead of the `template_file`, `js_file`,  and `css_file`.
+
+For example, here's the calendar component from
+the [Getting started](../getting_started/your_first_component.md) tutorial,
+defined in a single file:
 
 ```python title="[project root]/components/calendar.py"
-# In a file called [project root]/components/calendar.py
 from django_components import Component, register, types
 
 @register("calendar")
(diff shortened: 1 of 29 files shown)


src/django_components/component.py
```diff
@@ -158,14 +158,32 @@ def get_context_data(self, *args, **kwargs):
     """
 
 
+# Descriptor to pass getting/setting of `template_name` onto `template_file`
+class ComponentTemplateNameDescriptor:
```
JuroOravec, line 162: Because the `template_name` is a class attribute, not an instance property, I had to use descriptors to intercept the access, same as I did in `component_media.py`

tests/test_component.py
```diff
@@ -206,7 +206,29 @@ class Media:
...
+        rendered = SimpleComponent.render(kwargs={"variable": "test"})
+        self.assertHTMLEqual(
+            rendered,
+            """
+            Variable: <strong data-djc-id-a1bc3e>test</strong>
+            """,
+        )
+
+    @parametrize_context_behavior(["django", "isolated"])
+    def test_template_file_static__compat(self):
```
JuroOravec, line 231: The diff is wrong, the added that tests the backwards compatibility test is this one below

docs/concepts/fundamentals/single_file_components.md
```diff
@@ -3,7 +3,7 @@ title: Single-file components
 weight: 1
 ---
 
-Components can be defined in a single file, which is useful for small components. To do this, you can use the `template`, `js`, and `css` class attributes instead of the `template_name` and `Media`. For example, here's the calendar component from above, defined in a single file:
+Components can be defined in a single file, which is useful for small components. To do this, you can use the `template`, `js`, and `css` class attributes instead of the `template_file` and `Media`. For example, here's the calendar component from above, defined in a single file:
```
EmilStenstrom, line 6: We should reference js_file and css_file here too I think?

src/django_components/component.py
```diff
@@ -158,14 +158,32 @@ def get_context_data(self, *args, **kwargs):
...
+    def __get__(self, instance: Optional["Component"], cls: Type["Component"]) -> Any:
+        obj = instance if instance is not None else cls
+        return obj.template_file  # type: ignore[attr-defined]
```
EmilStenstrom, line 165: Do we need error handling here?

src/django_components/component_media.py
```diff
@@ -240,7 +240,14 @@ def __setattr__(cls, name: str, value: Any) -> None:
...
 
-        super().__setattr__(name, value)
+        # NOTE: When a metaclass specifies a `__setattr__` method, this overrides the normal behavior of
+        #       setting an attribute on the class with Descriptors. So we need to call the normal behavior explicitly.
+        # NOTE 2: `__dict__` is used to access the class attributes directly, without triggering the descriptors.
+        desc = cls.__dict__.get(name, None)
+        if hasattr(desc, "__set__"):
+            desc.__set__(cls, value)
+        else:
+            super().__setattr__(name, value)
```
EmilStenstrom, line 250: Tricky! :)

EmilStenstrom (approved):
Looking good!

src/django_components/component.py
JuroOravec: I don't think so, because the `Component.template_file` is set to `None` ([source](https://github.com/JuroOravec/django-components/blob/818aed3d205dba132b5d848c6bc61ad83289bf66/src/django_components/component.py#L226)), and this descriptor should be used only with the Component class, so accessing `template_file` should be always valid.

src/django_components/component_media.py
JuroOravec: Yeah, so this (django-components) is the first I ever had to use the descriptors in Python, but overall it feels unfortunate.

Like, why can't I use `@property` on a metaclass to configure a class attributes? Why can't metaclass be to class the same as class is to an instance? And why does `__set__` works only for instances while `__setattr__` works for everything? It feels like these things were piled onto each other without an overarching design in mind.

Anyway, rant's over :D

docs/concepts/fundamentals/single_file_components.md
JuroOravec: Great catch! Fixed now :)

PR #880, opened 2025-01-02
refactor: Assign content of file from Component.template_file to Component.template

JuroOravec:
This aligns the `template_file` behaviour with `js_file` and `css_file`, where the content of the file is assigned to `Component.js` / `Component.css` upon loading of the assets / media.

This addresses the point 3. from https://github.com/EmilStenstrom/django-components/pull/870#issuecomment-2565712822.

The main reason for loading the contents of the template file as plain string is to eventually be able to support plugins like django-cotton, where a user could define the components as valid HTML, and the plugin would convert it to valid Django template.

This will become possible as plugins will be able to pre-process the contents BEFORE we hand it over to Django.

Moreover, because of the changed flow, I had to explicitly add a call to the [`template_rendered`](https://docs.djangoproject.com/en/5.1/ref/signals/#template-rendered) signal to fix the tests. So while I was at it also:

- Added API reference page to describe the Signals that django-components defines or touches
- Moved the tests for the `template_rendered` signal into `test_signals.py` file, so it's clearer what it does. Previously those tests were called "Instrumentation", and at least to me it wasn't clear what was their purpose. Now I know.

Screenshot 2025-01-02 at 12 08 36


tests/test_component_media.py
```diff
@@ -20,29 +20,7 @@
 # "Main media" refer to the HTML, JS, and CSS set on the Component class itself
 # (as opposed via the `Media` class). These have special handling in the Component.
 class MainMediaTest(BaseTestCase):
-    def test_html_inlined(self):
```
JuroOravec, line 23: Removed these two tests, as they are now part of the updated tests

tests/test_templatetags_component.py
```diff
@@ -581,8 +581,16 @@ def test_both_components_render_correctly_with_slots(self):
             """,
         )
 
-    @parametrize_context_behavior(["django", "isolated"])
-    def test_both_components_render_correctly_when_only_first_has_slots(self):
+    @parametrize_context_behavior(
+        # TODO: Why is this the only place where this needs to be parametrized?
```
JuroOravec, line 585: As the TODO says, I'm not sure why this happens

tests/test_templatetags.py
```diff
@@ -21,79 +19,6 @@ class SlottedComponent(Component):
 #######################
 
 
-class TemplateInstrumentationTest(BaseTestCase):
```
JuroOravec, line 24: Moved these tests to `test_signals.py`

src/django_components/component_media.py
```diff
@@ -377,14 +379,24 @@ def get_context_data(self):
...
-    # If the component defined `js_file` or `css_file`, instead of `js`/`css` resolve them now.
-    # Effectively, even if the Component class defined `js_file`, at "runtime" the `js` attribute
+    # If the component defined `template_file`, `js_file` or `css_file`, instead of `template`/`js`/`css`,
+    # we resolve them now.
+    # Effectively, even if the Component class defined `js_file` (or others), at "runtime" the `js` attribute
     # will be set to the content of the file.
-    comp_media.js = _get_static_asset(
-        comp_cls, comp_media, inlined_attr="js", file_attr="js_file", comp_dirs=comp_dirs
+    # So users can access `Component.js` even if they defined `Component.js_file`.
+    comp_media.template = _get_asset(
```
JuroOravec, line 387: This here sets the contents of the file set by `Component.template_file` into `Component.template`.

tests/test_templatetags_component.py
EmilStenstrom: With some help from claude-3.5-sonnet and o1-mini:

### **Issue Summary**
- **Different Modes**: Your tests run in two modes: **django** and **isolated**.
- **ID Generation**: Both modes generate unique IDs for components using a global counter.
- **Inconsistent Rendering**: When only some components use slots, the **isolated** mode triggers additional rendering steps, leading to different ID sequences compared to the **django** mode.

### **Why the Analysis is Correct**

1. **ID Generation Timing**:
   - **Current Approach**: IDs are generated during the rendering process (`gen_id()` is called each time a component or slot is rendered).
   - **Problem**: In **isolated** mode, components with slots create additional isolated contexts, which results in extra calls to `gen_id()`. This means more IDs are generated, altering the sequence and causing inconsistencies.

2. **Context Behavior**:
   - **Django Mode**: Components share the same context, leading to a predictable and sequential ID generation.
   - **Isolated Mode**: Components with slots create separate contexts, leading to extra IDs being generated during the rendering process. This disrupts the expected ID sequence.

3. **Test Specifics**:
   - **Tests with All Components Using Slots**: Both modes handle slot processing similarly, maintaining consistent ID generation.
   - **Tests with Some Components Using Slots**: Only the components with slots in **isolated** mode trigger additional context handling, leading to different ID outcomes compared to **django** mode. This is why parametrization is needed in these cases to account for the differing ID values.

### **Why Pre-Generating IDs Doesn't Solve the Issue**

While pre-generating and storing IDs at the time of component registration might seem like a solution, it doesn't address the root cause:

- **Root Cause**: The inconsistency arises from how contexts are managed and how many times components are rendered (and thus, how many IDs are generated) in different modes.
- **Pre-Generating IDs**: This approach would stabilize the ID assignment but wouldn't prevent the extra IDs from being generated due to additional rendering steps in **isolated** mode. The core problem of context-based rendering differences remains unresolved.

### **Recommended Approach**

To achieve consistent ID generation across both **django** and **isolated** modes, consider the following strategies:

1. **Centralize ID Generation**:
   - Instead of generating IDs during rendering, manage them centrally to ensure consistency regardless of the rendering path.
   
2. **Adjust Context Handling**:
   - Modify how contexts are isolated in **isolated** mode to prevent additional unintended rendering steps from affecting ID generation.
   
3. **Synchronize Rendering Steps**:
   - Ensure that both modes follow a similar rendering pipeline, minimizing differences in how many times components are processed and IDs are generated.

4. **Implement ID Mapping**:
   - Maintain a mapping between component instances and their assigned IDs to prevent discrepancies caused by different rendering paths.

By addressing the context handling and centralizing ID management, you can eliminate the inconsistencies in ID generation between the two modes.

### **Conclusion**

The initial analysis correctly pinpointed that the discrepancy in ID generation stems from how **isolated** mode handles contexts and rendering differently when dealing with slots. To resolve this, focus on harmonizing the rendering process and managing ID generation in a centralized manner rather than merely pre-generating IDs at registration.

If you need further assistance with specific code implementations or exploring these strategies in more detail, feel free to ask!

EmilStenstrom (approved):
This looks good to me!

PR #888, opened 2025-01-07
refactor: exclude underscored files from autodiscovery

JuroOravec:
Tiny change to exclude those file and directories from autodiscovery that start with `_` (e.g. `_ignore_me.py` or `_privatedir/my_file.py`). `__init__.py` is NOT excluded.

I came across this because sometimes I follow the pattern where I define test files close to the actual file in `__tests__` subdirectory, and I was that the test files were being picked up.


EmilStenstrom:
> Tiny change to exclude those file and directories from autodiscovery that start with `_` (e.g. `_ignore_me.py` or `_privatedir/my_file.py`). `__init__.py` is NOT excluded.
> 
> I came across this because sometimes I follow the pattern where I define test files close to the actual file in `__tests__` subdirectory, and I was that the test files were being picked up.

I’m not sure I understand the reasoning here. Isnöt this unintuitive, that different files are imported differently depending on their file name? Is this a pattern somewhere else that I’ve missed?

JuroOravec:
@EmilStenstrom Looks like the pattern I follow I got it from working with [Jest](https://stackoverflow.com/questions/55551350/jest-test-formatting-tests-vs-test-js), where the default location is by placing the tests close to their definings inside `__tests__` subdirectory

Screenshot 2025-01-07 at 20 55 19


---

Though I think it's good to have some escape hatch to prevent files from being auto-imported:
- For internal files like in my case
- Or a pattern I've seen is that for background tasks (e.g. think [this](https://django-background-tasks.readthedocs.io/en/latest/) or Celery) is that the worker may have their own files, which sometimes shouldn't be imported elsewhere (could be that a worker needs different dependencies). So that could be potentially solved by renaming a `tasks` directory to `_tasks`  in a Django app - that way the background worker files could live in the app that relates to them, without the overhead of unintentionally importing those files.

EmilStenstrom:
@JuroOravec Thanks for explaining. Still feels a bit strange and "unpythonic", but I see the point, so let's go with it :)

PR #890, opened 2025-01-07
feat: add Media.extend

JuroOravec:
This PR:

- Makes the `Component.media` instance to inherit the JS and CSS from parent components
- Users can opt-out or configure the inheritance behaviour with the `extend` attribute on `Component.Media`, which behaves similarly to Django's [Media.extend](https://docs.djangoproject.com/en/5.1/topics/forms/media/#extend).
- Clears up behaviour around subclassing components by documenting it

This addresses points 4. and 5. from [this discussion](https://github.com/EmilStenstrom/django-components/pull/870#issuecomment-2565712822).

**Examples:**

  Normal media inheritance:
```python
class ParentComponent(Component):
    class Media:
        js = ["parent.js"]

class MyComponent(ParentComponent):
    class Media:
        js = ["script.js"]

print(MyComponent.media._js)  # ["script.js", "parent.js"]
```

  Disable media inheritance:
```python
class ParentComponent(Component):
    class Media:
        js = ["parent.js"]

class MyComponent(ParentComponent):
    class Media:
        extend = False  # Don't inherit parent media
        js = ["script.js"]
print(MyComponent.media._js)  # ["script.js"]
```

Specific set of parent Media classes

In this case, the media files are inherited ONLY
from the specified components, and NOT from the original parent components:

```python
class ParentComponent(Component):
    class Media:
        js = ["parent.js"]

class MyComponent(ParentComponent):
    class Media:
        # Only inherit from these, ignoring the files from the parent
        extend = [OtherComponent1, OtherComponent2]
        js = ["script.js"]

print(MyComponent.media._js)  # ["script.js", "other1.js", "other2.js"]
```

And some documentation :)

https://github.com/user-attachments/assets/21e8149d-6d46-446e-a8fe-32d88966e1bc

Screenshot 2025-01-07 at 20 19 11


src/django_components/component_media.py
```diff
@@ -399,11 +532,6 @@ def get_context_data(self):
         comp_cls, comp_media, inlined_attr="css", file_attr="css_file", comp_dirs=comp_dirs, type="static"
     )
 
-    media_cls = comp_media.media_class or MediaCls
-    media_js = getattr(comp_media.Media, "js", [])
-    media_css = getattr(comp_media.Media, "css", {})
-    comp_media.media = media_cls(js=media_js, css=media_css) if comp_media.Media is not None else None
```
JuroOravec, line 405: Previously, the `media` instance contained only the JS / CSS files from the current Component class, so it was enough to just instantiate it. Now, the [`_get_comp_cls_media`](https://github.com/EmilStenstrom/django-components/pull/890/files#diff-bdb1917504a9c50a51e7fdf49498d5db3ce0233be9eb81b13938d4b3dcd5efc3R385) travels up the Component's parent classes to find all the JS / CSS. And so the child component class end up inheriting them all.

EmilStenstrom (approved):
This looks awesome. Ready to merge. Really like the tests, simple and effective.

I think I remember you describing that we've implemented Media "except for the extends" somewhere in the docs, but I don't see that removed here?

JuroOravec:
@EmilStenstrom Yeah that's, thanks for reminding me that!

PR #891, opened 2025-01-07
refactor: use HTML5 parser for BeautifulSoup

JuroOravec:
This updates BS4 to use the HTML5 parser, as mentioned in [this discussion](https://github.com/EmilStenstrom/django-components/pull/870#issuecomment-2565712822).

Turns out that there was far fewer tests to fix after the change.


EmilStenstrom:
@JuroOravec Nice! :)

Issue #894, opened 2025-01-08
In-template component input that has filter(s) with whitespace don't work


JuroOravec:
I found that the tag parser inproperly parses the component tag inputs that have filters to them.

This is fine:

```django
{% component "table" key=val key2="val2" %}
```

Even filters without spaces are fine

```django
{% component "table" key=val|default:"123" key2="val2"|yesno:"a,b" %}
```

However, the parser doesn't account for whitespaces in filters, whether between `|` and `:`, or within strings after the `:`

```django
{% component "table"
	key=val | default
	key=val|default:" 123 "
    key2="val2" | yesno : " a,b "
%}
```

Also, currently the parser expects no whitespace for translations:

```django
{% component "table" key=_("welcome") %}
```

But Django allows that (according to ChatGPT), but currently, this would cause error:

```django
{% component "table" key=_(   "welcome"   ) %}
```

I got a fix for this drafted up.

unikol-salm70900:
Hello, we tried to solve the issue.

This is what we did:

Modified the _parse_tag_input function to properly handle filters with whitespace and translation functions. This change allows for more flexible syntax in component tags, including whitespace around filters and within translation function calls.

You can review changes in this commit: https://github.com/unikol-salm70900/EmilStenstrom-django-components-894/commit/48acb2d92ea80846572d3d418905dca91477c913.


> [!CAUTION]
> Disclaimer: The concept of solution was created by AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---

Latta AI seeks to solve problems in open source projects as part of its mission to support developers around the world. Learn more about our mission at https://latta.ai/ourmission . If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

Issue #872, opened 2024-12-30
Allow literal arrays and dictionaries as component inputs and better setting of conditional CSS classes


JuroOravec:
## Background

The use of `{% html_attrs %}` still feels clunky, especially when it comes to adding classes that should be applied only conditionally. 

Taking inspiration from Vue and Alpine, ideally I could use an inlined dictonary to be able to "toggle" whether a class should / should not be rendered. 

In the example below, if `is_hidden` is truthy, then the class `hidden` IS applied, otherwise it is NOT.

```django
<div {% html_attrs attrs class={"hidden": is_hidden } %}>
  Hi
</div>
```

Last time I was looking into similar things, I ended up creating [django-expr](https://pypi.org/project/django-expr/). Unfortunately, this use case is not easily solved by django-expr, because it gets messy if I want to conditionally apply more than one class:

```django
{% expr "'hidden' if is_hidden else None" as hidden_cls %}
{% expr "'disabled' if is_disabled else None" as disabled_cls %}
<div {% html_attrs attrs class=hidden_cls class=disabled_cls %}>
  Hi
</div>
```

So I again end up defining the CSS classes in Python:

```py
class MyTable(Component):
    def get_context_data(self, is_hidden, is_disabled, **attrs):
        if is_hidden:
            attrs["class"] = attrs.get("class", "") + "hidden"
        if is_disabled:
            attrs["class"] = attrs.get("class", "") + "disabled"
        return {
            "attrs": attrs,
        }
```
```django
<div {% html_attrs attrs %}>
  Hi
</div>
```

---

### Inlined arrays and dicts

Since our parsing of template tags is now a superset of Django's, I don't see a problem with adding support for inlined arrays and dicts.

Unlike django-expr, however, these arrays / dicts would NOT be parsed as fully-valid Python AST. Instead, the values and keys would still be the same as what's allowed within `{{ }}` tags (e.g. `{{ my_var|upper }}`:

```django
{% component
   {"first_name": "John", "last_name": "Wick"}
   abc=["mystr", 123, False, my_var, other_var|yesno:":),:("]
   xyz={
     "key1": "mystr",
     key_var: 123,
   }
   nested=[
     {"ids": [1, 2, 3]},
     {"ids": [4]},
     {"ids": [5, 6]},
   ]
%}
{% endcomponent %}
```

So the parsing of arrays / dicts would be handled by our parser. How it would work is that:
[...]

PR #898, opened 2025-01-09
feat: list and dict literals in tags + fix tag parser

JuroOravec:
Following up on https://github.com/EmilStenstrom/django-components/issues/872, because, frankly, it was fun to solve 😄 

This PR refactors the tag parser, which turns Django's template tag content of e.g.

`{% component key=value key2="abc def" / %}`

into an AST which can be worked with downstream.

1. This PR fixes the bug in https://github.com/EmilStenstrom/django-components/issues/894 - after this PR, it will be possible to have whitespace around filters and translation syntax:

   `{% component key=_(   "hello"  ) key2="abc def"   |   lower / %}`

2. I added support for defining list and dict literals as inputs, and allowing to spread variables into those literal lists and dicts.

	I then took this idea to its logical conclusion, allowing nested lists, dicts, etc. So the syntax works as follows:

    Supported syntax:
    - Variables: `val`, `key`
    - Kwargs (attributes): `key=val`, `key2='val2 two'`
    - Quoted strings: `"my string"`, `'my string'`
    - Translation: `_("my string")`
    - Filters: `val|filter`, `val|filter:arg`
    - List literals: `[value1, value2]`, `key=[value1, [1, 2, 3]]`
    - Dict literals: `{"key1": value1, "key2": value2}`, `key={"key1": value1, "key2": {"nested": "value"}}`
    - Trailing commas: `[1, 2, 3,]`, `{"key": "value", "key2": "value2",}`
    - Spread operators: `...`, `*`, `**`
    - Spread inside lists and dicts: `key=[1, *val, 3]`, `key={"key": val, **kwargs, "key2": 3}`
    - Spread with list and dict literals: `{**{"key": val2}, "key": val1}`, `[ ...[val1], val2 ]`
    - Spread list and dict literals as attributes: `{% component ...[val1] %}`, `{% component ...{"key" val1 } %}`

    Invalid syntax:
    - Spread inside a filter: `val|...filter`
    - Spread inside a dictionary key: `attr={...attrs: "value"}`
    - Spread inside a dictionary value: `attr={"key": ...val}`
    - Misplaced spread: `attr=[...val]`, `attr={...val}`, `attr=[**val]`, `attr={*val}`
    - Spreading lists and dicts: `...[1, 2, 3]`, `...{"key": "value"}`

	This tag content is parsed into an AST that looks like this:
    ```py
	[
    TagAttr(
        key=None,
        value=TagValueStruct(
            type="simple",
            spread=None,
            meta={},
            entries=[
                TagValue(
                    parts=[
                        TagValuePart(
                            value="component", quoted=None, spread=None, translation=False, filter=None
                        ),
                    ]
                ),
            ],
        ),
        start_index=13,
    )
    ...
    ]
    ```

NOTE: I didn't make changes to documentation yet. For one, this is already a huge PR. For second, while I updated the parser, as of this PR, users cannot yet use the literal dicts and lists. Or rather, they can write it, but it will be treated as a string. So I plan to make a second PR to update that part and update documentation with it.

Closes https://github.com/EmilStenstrom/django-components/issues/894

### DIFF
--- a/src/django_components/templatetags/component_tags.py
+++ b/src/django_components/templatetags/component_tags.py
@@ -12,7 +12,7 @@
 #   During documentation generation, we access the `fn._tag_spec`.
 
 import functools
-from typing import Any, Callable, Dict, List, Literal, NamedTuple, Optional, Set, Tuple, Union
+from typing import Any, Callable, Dict, List, Literal, NamedTuple, Optional, Set, Tuple, Union, cast
 
 import django.template
 from django.template.base import FilterExpression, NodeList, Parser, TextNode, Token, TokenType
@@ -34,7 +34,6 @@ from django_components.expression import (
     SpreadOperator,
     is_aggregate_key,
     is_dynamic_expression,
-    is_spread_operator,
 )
 from django_components.provide import PROVIDE_NAME_KWARG, ProvideNode
 from django_components.slots import (
@@ -49,7 +48,7 @@ from django_components.slots import (
 from django_components.tag_formatter import get_tag_formatter
 from django_components.util.logger import trace_msg
 from django_components.util.misc import gen_id
-from django_components.util.tag_parser import TagAttr, parse_tag_attrs
+from django_components.util.tag_parser import TagAttr, TagValue, parse_tag
 
 # NOTE: Variable name `register` is required by Django to recognize this as a template tag library
 # See https://docs.djangoproject.com/en/dev/howto/custom-template-tags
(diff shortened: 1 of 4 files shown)


EmilStenstrom (commented):
This is a nice feature, but with ~3000 lines, it's not worth reviewing imho. I'll just trust the tests in this case :)

JuroOravec:
Fair point. Just a note, that 2200 lines of that are purely tests, because it's about 30 tests, and the AST  takes up a lot of space.

PR #899, opened 2025-01-09
refactor: move the utils from component_tags to util.template_tag

JuroOravec:
Bit of cleanup before I do next part of https://github.com/EmilStenstrom/django-components/issues/872. 

I moved the helper functions from the file where the template tags are defined to its own file in the `util` folder. No logic change.


JuroOravec:
cc @EmilStenstrom

Issue #881, opened 2025-01-04
Document how to define defaults Component inputs


JuroOravec:
Before, when there was only `get_context_data()`, it was sufficient to define the defaults directly on the method:

```python
class Icon(Component):
    def get_context_data(
        self,
        name: IconName,
        variant: Literal["outline", "solid"] = "outline",
        size: int = 24,
        color: str = "currentColor",
        stroke_width: float = 1.5,
        viewbox: str = "0 0 24 24",
        attrs: Optional[Dict] = None,
    ) -> Dict:
        ...
```

However, with the upcoming `get_js_data()` and `get_css_data()`, since they all receive the same input, all these methods have to declare the function signature and defaults themselves, duplicately:

```python
class Icon(Component):
    def get_context_data(
        self,
        name: IconName,
        variant: Literal["outline", "solid"] = "outline",
        ...
    ) -> Dict:
        ...

    def get_js_data(
        self,
        name: IconName,
        variant: Literal["outline", "solid"] = "outline",
        ...
    ) -> Dict:
        ...

    def get_css_data(
        self,
        name: IconName,
        variant: Literal["outline", "solid"] = "outline",
        ...
    ) -> Dict:
        ...
```

We can avoid having to re-declare the function signature by typing the Component, and accessing the args and kwargs as `self.input.args` and `self.input.kwargs`. However, individual functions still have to define the defaults, duplicately:

```py
class IconKwargs(TypedDict):
    name: IconName
    variant: NotRequired[Literal["outline", "solid"]]
    size: NotRequired[int]
    color: NotRequired[str]
    stroke_width: NotRequired[float]
    viewbox: NotRequired[str]
    attrs: NotRequired[Optional[Dict]]

IconType = Component[EmptyTuple, IconKwargs, EmptyDict, Any, Any, Any]

class Icon(IconType):
    def get_context_data(self, *args, **kwargs): -> Dict:
        name = self.input.kwargs["name"]
	    variant = self.input.kwargs.get("variant", "outline")
	    size = self.input.kwargs.get("size", 24)
	    color = self.input.kwargs.get("color", "currentColor")
	    stroke_width = self.input.kwargs.get("stroke_width", 1.5)
	    viewbox = self.input.kwargs.get("viewbox", "0 0 24 24")
	    viewbox = self.input.kwargs.get("attrs", None)
        ...

    def get_js_data(self, *args, **kwargs): -> Dict:
        name = self.input.kwargs["name"]
[...]

PR #902, opened 2025-01-13
feat: Literal dicts and lists part 2

JuroOravec:
Changes:
- Inputs to template tags now behave closer to python functions:
  - Overriding kwargs:
       ```django
       {% component **attrs1 **attrs2 %}
       ```
    - Before, kwargs worked that latter kwargs overrode earlier. So, below, if `attrs1` and `attr2` contained the same key, the key in `attrs1` would be ignored and overriden by the value of the key in `attrs2`
    - Now, this works the same way as calling python function like `component(**attrs1, **attrs2)`, which will raise TypeError if both contain the same key.
    - To work around this, one can merge the two dictionaries before spreading them onto the template tag:
       ```django
       {% component
          ...{
              **attrs1,
              **attrs2,
          }
       %}
       ```
- One can now spread both lists and dicts
   - Previously, only dicts could be spread with `...`
- One may now put positional args after kwargs.
- It is no longer possible to start key names with `:`, because it's indistinguishable from template filters
   - E.g. `{% component key1=val1|lower :placeholder="No text" %}`
   - Instead, users will now have to scope such keys under dictionaries, as `attrs::placeholder="No text"`


docs/scripts/reference.py
```diff
@@ -593,30 +593,19 @@ def _format_tag_signature(tag_spec: TagSpec) -> str:
...
-    if tag_spec.positional_args_allow_extra:
-        params.append("[arg, ...]")
-
-    if tag_spec.keywordonly_args is True:
-        params.append("**kwargs")
-    elif tag_spec.keywordonly_args:
-        params.extend(
-            [f"{name}=None" if name in optional_kwargs else name for name in (tag_spec.keywordonly_args or [])]
-        )
+    # The signature returns a string like:
```
JuroOravec, line 596: This updates the logic how we generate the API reference for template tags (e.g. `{% component  %}`, etc).

src/django_components/attributes.py
```diff
@@ -18,32 +18,19 @@
 class HtmlAttrsNode(BaseNode):
     def __init__(
         self,
-        kwargs: RuntimeKwargs,
-        kwarg_pairs: RuntimeKwargPairs,
```
JuroOravec, line 22: Previously, one could pass in the same key multiple times to key defined on `{% html_attrs %}`, e.g.:

```django
{% html_attrs
    class="pt-4"
    class=class_from_var
    class="extra-class"
%}
```

However, the inputs to template tags now follow how regular python functions accept input. And so, it's no longer valid to pass in multiple kwargs with the same key.

To keep  the `{% html_attrs %}` working the same, I added this behaviour upstream in https://github.com/EmilStenstrom/django-components/pull/902/files#diff-56d60a8b1d11d8ff86b912ddde6d8ee539f71a1fae917bc39a1fd049ea22705cR309.

With the support for literal lists and dicts, I suggest to deprecate the behaviour of supporting multiple same-name keys, and then remove it in v1.

And instead, once in v1, one would do:

```django
{% html_attrs
    class=["pt-4", class_from_var, "extra-class"]
%}
```

src/django_components/expression.py
```diff
@@ -203,14 +138,8 @@ def is_dynamic_expression(value: Any) -> bool:
...
-def is_spread_operator(value: Any) -> bool:
-    if not isinstance(value, str) or not value:
-        return False
-
-    return value.startswith("...")
-
-
-def process_aggregate_kwargs(kwargs: Mapping[str, Any]) -> Dict[str, Any]:
+# TODO - Move this out into a plugin?
+def process_aggregate_kwargs(params: List["TagParam"]) -> List["TagParam"]:
```
JuroOravec, line 142: Interestingly, with the new flow, the `attr:key=val` syntax could be later redefined as an (internal) plugin.

src/django_components/expression.py
```diff
@@ -103,72 +104,6 @@ def render(self, context: Context) -> str:
         return str(result)
 
 
-class RuntimeKwargs:
```
JuroOravec, line 106: Much of the logic around resolving args and kwargs passed to template tags was removed and replaced by the [code here](https://github.com/EmilStenstrom/django-components/pull/902/files#diff-fcef21070412acf8475f5c6880736500491a9b4fbc5d77a37d46dd63814c158fR275).

Personally I think the flow is now a bit simpler. Because where before we passed around half-resolved values like `SpreadOperator` or `DynamicFilterExpression`, all of that is now encapsulated by [`TagAttr`](https://github.com/EmilStenstrom/django-components/pull/902/files#diff-fcef21070412acf8475f5c6880736500491a9b4fbc5d77a37d46dd63814c158fR54). And at render time, we resolve it all into `TagParam`, which just holds the key and value of arg / kwarg.

src/django_components/templatetags/component_tags.py
```diff
@@ -103,20 +100,19 @@ def component_js_dependencies(parser: Parser, token: Token, tag_spec: TagSpec) -
...
 
+def slot_signature(name: str, **kwargs: Any) -> None: ...  # noqa: E704
+
+
 @register.tag("slot")
 @with_tag_spec(
     TagSpec(
         tag="slot",
         end_tag="endslot",
-        positional_only_args=[],
```
JuroOravec, line 116: The whole point of the `TagSpec` is to:
1. validate the parameters passed to template tags
2. as source of truth for the documentation

As you can see, the old values were already based on how Python describes function input - there's positional-only args, keyword-only args, and params which can be either as positional or as kwargs.

Instead of listing all of those, I switched to simply use the already-available `inspect.Signature`, which captures allof this info too.

The only difference is that for `{% html_attrs %}` we allowed to pass multiple kwargs with the same key, which is not allowed with a normal python function. That's why I want to deprecate that eventually, so that calling template tags from within the template would eventually behave similarly to calling python functions.

As for why I introduced the `TagSpec.signature`, see [this code](https://github.com/EmilStenstrom/django-components/pull/902/files#diff-56d60a8b1d11d8ff86b912ddde6d8ee539f71a1fae917bc39a1fd049ea22705cR49). Basically that function constructs another function ("validator"), and passing in the parameters. And the `validator` has its signature set to `TagSpec.signature`. This makes it so that when then we pass in parameters to the `validator` function, it raises an error the same way as when you pass wrong input to a python function.

src/django_components/util/tag_parser.py
```diff
@@ -226,8 +267,99 @@ def render_value(value: Union[TagValue, TagValueStruct]) -> str:
...
+                elif isinstance(entry, TagValue) and entry.is_spread:
+                    resolved_list.extend(resolved)
+                # Case: Plain value: [ val ]
+                else:
+                    resolved_list.append(resolved)
+            return resolved_list
+
+        elif self.type == "dict":
+            resolved_dict: Dict = {}
+            dict_pair: List = []
```
JuroOravec, line 330: Fun fact:

How I handled the literal dictonaries is that the the keys and values are stored as a list. But not as a list of key-value tuples, but a flat list.

The reason for that is because dictionaries may contain spread syntax, like so:

```
{"key1": val1, **attrs, "key2": val2}
```

So one way to look at the dictionary is that as if both `,` and `:` were delimeters. So the example above would be parsed into a list like so (omitting the classes that wrap the values for brevity):

```
["key1", val1, **attrs, "key2", val2]
```

However, this means that we then need to validate that:
- the values between start / end / spreads should always be in pairs (e.g. "key1" and val1). E.g.:
   ```
   ["key1", val1, **attrs, "key2", val2]
      (key)  (val) (spread) (key)   (val)
   ```
- The spead must be on the position of dictionary key.

Thus, with this structure, it's only here in `resolve()` (AKA at render time), when we replace the spread with the actual values, and where we finally construct a python dictionary out of this flat representation.

src/django_components/util/tag_parser.py
```diff
@@ -625,10 +757,15 @@ def extract_spread_token(curr_struct: TagValueStruct, filter_token: Optional[str
...
                 # E.g. imagine:    `height="20" | yesno : "1,2,3" | lower`
                 # and we're here:               ^
+                # (or here)                             ^
+                # (or here)                                       ^
                 # and we want to parse `yesno` next
                 if not is_first_part:
                     filter_token = taken_n(1)  # | or :
                     take_while(TAG_WHITESPACE)  # Allow whitespace after filter
+
+                    if filter_token == ":" and values_parts[-1].filter != "|":
```
JuroOravec, line 767: Django's filters have structure `var|filter:arg`, where the filters can be chained. This addition catches the case when the expression uses `:` without previously using `|`.

So e.g. `var|filter:arg` is fine, while `var:arg` will raise an error.

src/django_components/util/template_tag.py
```diff
@@ -87,9 +35,106 @@ class TagSpec(NamedTuple):
     - and treated as `only=False` and `required=False` if omitted
     """
 
+    def copy(self) -> "TagSpec":
```
JuroOravec, line 38: We allow each component to potentially use a different template tag. E.g. `{% table ... / %}` instead of `{%  component "table" ... / %}`.

For that, we have to set the `tag` and `endtag` fields on the component's `TagSpec` instance.

But that means we also have to make copies of the original TagSpec. Hence this `copy()` function.

src/django_components/util/template_tag.py
```diff
@@ -87,9 +35,106 @@ class TagSpec(NamedTuple):
...
+        #       we pass in only the list index.
+        validator_call_script = "args, kwargs = validator(*call_args, "
+        for kw_index, _ in enumerate(call_kwargs):
+            validator_call_script += f"**call_kwargs[{kw_index}], "
+        validator_call_script += ")"
+
+        try:
+            # Create function namespace
+            namespace: Dict[str, Any] = {"validator": validator, "call_args": call_args, "call_kwargs": call_kwargs}
+            exec(validator_call_script, namespace)
```
JuroOravec, line 122: This is the second part of the validation magic. I wanted the args and kwargs being passed to a template tag to behave same way sa it would behave if they were passed to a function.

So e.g. if we have a template tag `{% slot name="abc" data=[1, 2, 3] %}`, then this should behave as if passing the data as if `{% slot %}` was a fucntion: 

`slot(name="abc", data=[1, 2, 3])`

But when it comes to dealing with spreads, e.g.:

`slot(**attrs1, **attrs2)`

Then if the two dictionaries had at least one shared key, then python would raise a TypeError. Because Python would evaluate this as

```py
slot(
    # kwargs from attrs1
    key1=1,
    key2=2,
    # kwargs from attrs2
    key1=3,
    key4=4,
)
```

which contains two `key1` kwargs, and thus raises.

---

Initially I wanted to pass the params to the `validator` as simply spreading them:

`validator(*args, **kwargs)`

But for that, the `kwargs` would already contain ALL combined kwargs. And while multiple same keys raises in a function, it does NOT in a dictionary. In other words

`{ **attrs1, **attrs2 }`

doesn't raise, and it keeps only the `key1` from `attrs2`.

So this is not the behaviour I wanted.

---

Instead, the trick I did to achieve python function validation behaviour is that I wrapped each kwarg in it's own dictionary.

So instead of passing in

```py
validator(key=value)
```

I instead did something like:

```py
kwarg = {key: value}
validator(**kwarg)
```

But the challenge here was that I don't know of any way how to apply a dynamic number of spreads to a python function. So instead I turned to `exec()`

I dynamically generate a script like this, that contains as many `**call_kwargs[0]` as the kwargs I need to spread:

```py
args, kwargs = validator(
    *call_args,
    **call_kwargs[0],
    **call_kwargs[1],
    **call_kwargs[2],
)
```

And then I execute the script, passing in the definition of `call_args` and `call_kwargs`.

Thus, if any of `call_kwargs[0]`, `call_kwargs[1]`, etc... contain kwargs with the same key, Python's function validation will capture that.

src/django_components/util/template_tag.py
```diff
@@ -87,9 +35,106 @@ class TagSpec(NamedTuple):
...
+
+        Raises:
+            TypeError: If the parameters don't match the tag spec's rules.
+        """
+
+        # Create a function with this signature that captures the input and sorts
+        # it into args and kwargs
+        def validator(*args: Any, **kwargs: Any) -> Tuple[List[Any], Dict[str, Any]]:
+            # Let Python do the signature validation
+            bound = self.signature.bind(*args, **kwargs)
```
JuroOravec, line 70: Third part of the validation magic is that inside the `validator`, we use the given `TagSpec.signature` to figure out which parameters are args and kwargs.

Distinguishing between args and kwargs is important, because some of the Component features require this distinction.
- E.g. for the type hints when passing inputs to `Component.render()`, args and kwargs have to be defined separately due to Python typing limitations.
- Or the [feature about component defaults](https://github.com/EmilStenstrom/django-components/issues/881), that will likely (or at least at first) apply only to kwargs.

src/django_components/util/template_tag.py
```diff
@@ -155,287 +246,99 @@ def _parse_tag_preprocess(
...
+        params=TagParams(params=raw_params, tag_spec=tag_spec),
+        flags=flags,
+        # NOTE: We defer parsing of the body, so we have the chance to call the tracing
+        # loggers before the parsing. This is because, if the body contains any other
+        # tags, it will trigger their tag handlers. So the code called AFTER
+        # `parse_body()` is already after all the nested tags were processed.
+        parse_body=lambda: _parse_tag_body(parser, tag_spec.end_tag, is_inline) if tag_spec.end_tag else NodeList(),
+    )
 
-def _parse_tag_input(tag_name: str, attrs: List[TagAttr]) -> Tuple[List[str], List[TagKwarg], Set[str]]:
```
JuroOravec, line 171: Lots of changes and deletions in this file, as previously we tried to extract and separate args and kwargs BEFORE resolving the values, while now we do it only at render time. Only flags (e.g. `{% slot "name" default %}`) are still extracted at the time when the template is being parsed / compiled.

tests/test_expression.py
```diff
@@ -46,38 +46,6 @@ def make_context(d: Dict):
 #######################
 
 
-class ResolveTests(BaseTestCase):
```
JuroOravec, line 49: Removed, the tested functions were removeed.

tests/test_expression.py
```diff
@@ -729,7 +697,7 @@ def test_html_attrs(self):
         )
 
     @parametrize_context_behavior(["django", "isolated"])
-    def test_later_spreads_overwrite_earlier(self):
+    def test_later_spreads_do_not_overwrite_earlier(self):
```
JuroOravec, line 700: One of potentially breaking changes - Previously, when there was

`{% component ...attrs1 ...attr2 %}`

then if there was the same key in both `attrs1` and `attrs2`, the first one would be overwritten.

But as described in some of the other comments here, now this raises an error

tests/test_expression.py
```diff
@@ -851,25 +866,63 @@ class SimpleComponent(Component):
...
-            template.render(
-                Context(
-                    {
-                        "var_a": "abc",
-                        "var_b": "def",
-                    }
-                )
-            )
+
+class AggregateKwargsTest(BaseTestCase):
```
JuroOravec, line 873: This group of tests was moved here from tests/test_template_parser.py

tests/test_tag_parser.py
```diff
@@ -2414,3 +2496,361 @@ def test_spread_list_literal_as_attribute(self):
                 "...[val1]",
             ],
         )
+
+    def test_dynamic_expressions(self):
```
JuroOravec, line 2500: There's a lot of trivial code changes in this file as the  `TagValueStruct` class now accepts a `Parser`, to be able to evaluate template tags and filters.

But the genuinely new tests start here.

JuroOravec:
@EmilStenstrom This one is now ready for review. Sorry, this one is fairly complex as ti touches a lot of files. Unfortunately it felt like this was the minimum to make this work 

I'll update the docs in a separate PR

src/django_components/attributes.py
EmilStenstrom: Similar to our discussion of attrs:class=X I like this syntax. It isn't that DRY, but it reads better than python syntax imho. I like the idea (later in the thread) of it being an automatically enabled plugin that converts one syntax to the other, instead of being removed.

tests/test_expression.py
EmilStenstrom: I think this is reasonable.

tests/test_expression.py
```diff
@@ -762,32 +743,47 @@ def get_context_data(
...
-            <div data-djc-id-a1bc3f>OVERWRITTEN_X</div>
+            <div data-djc-id-a1bc40>{'@click': '() =&gt; {}', 'style': 'OVERWRITTEN'}</div>
+            <div data-djc-id-a1bc40>[1, 2, 3]</div>
+            <div data-djc-id-a1bc40>1</div>
+            <div data-djc-id-a1bc40>OVERWRITTEN_X</div>
             """,
         )
 
     @parametrize_context_behavior(["django", "isolated"])
-    def test_raises_if_positional_arg_after_spread(self):
```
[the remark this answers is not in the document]
EmilStenstrom: I don't think this is allowed in python?

```
>>> def x(x, y=None): print(x, y)
...
>>> x(5, 6)
5 6
>>> x(y=4, 6)
  File "<stdin>", line 1
    x(y=4, 6)
            ^
SyntaxError: positional argument follows keyword argument
```

tests/test_expression.py
```diff
@@ -851,25 +866,63 @@ class SimpleComponent(Component):
...
+                captured = args, kwargs
+                return {}
+
+        template_str: types.django_html = """
+            {% load component_tags %}
+            {% component 'test'
+                attrs:@click.stop="dispatch('click_event')"
+                attrs:x-data="{hello: 'world'}"
+                attrs:class=class_var
+                attrs::placeholder="No text"
```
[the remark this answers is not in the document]
EmilStenstrom: Agreed.

EmilStenstrom (approved):
Hi, happy with this feature!

I've read through the comments, but the code is way too much to get through in reasonable time. I've read through the comments and written a few things. Go ahead and merge.

tests/test_expression.py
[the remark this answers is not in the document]
JuroOravec: Ah, good catch! So it's a behaviour of the changes I made, not of python functions. Looking again, it's because I re-sort the inputs [here](https://github.com/EmilStenstrom/django-components/pull/902/files#diff-56d60a8b1d11d8ff86b912ddde6d8ee539f71a1fae917bc39a1fd049ea22705cR102) to put all args into a single list, that then goes before the kwargs.

So if you were to call `{% component "test" y=4 6 %}`

it's as if the function was called like `x(  *[6], **{"y":4}  )`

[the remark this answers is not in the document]
EmilStenstrom: I think that's quite strange behavior frankly. That y=4 could have been a mistake (should have written x=4), and now the function is being called with another random parameter. Stuff could be deleted because of an error like that, if you send True to the wrong function. I think it would be safer to crash with a syntax error...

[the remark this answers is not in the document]
JuroOravec: Yup, in the next PR one will have to again put args before kwargs!

PR #908, opened 2025-01-14
refactor: Patch Template.compile_nodelist with custom template parser

JuroOravec:
This PR adds a template parser ([see file](https://github.com/EmilStenstrom/django-components/pull/908/files#diff-e4b741cf09ac5f994de44baaa54cefa221ee318b4c0368c5527d8564145b40ca)) that is capable of correctly parsing Templates that contain `{% ... %}` inside of component input. This parser is then used inside a patched `Template.compile_nodelist()`.

As taken from the `template_parser.py` file:

The strategy is to use Django's lexer, and use our own parser
only when necessary, for the shortest time possible.

How it works is that:
  1. We use Django's lexer to get the tokens.
  2. We check them one-by-one, and if we find a broken token, we switch to our parser to fix it.
  3. Once the broken token is fixed, we find it's end position, and switch back to the Django lexer
     for the remaining text (step 1).

Before I switched to this strategy, my initial parser was about 50x slower than Django's lexer.
I assume it was because I was walking character by character, instead of using a regex.

The overall speed might then vary with the number of broken tokens in the template.

Performance of the new strategy on a real-world example:
- A template with about 110 lines and 6 components
- Components spanning ~35 lines in total, so roughly 1/3 of the template
- The custom parser is about 8x slower than Django's Debug lexer.
- For a mid-sized project of 200 templates, it would take 7-8 seconds to load all the templates
    (from 1 second with Django's lexer).
    - However, thanks to django-component's lazy-loading, this should not be a problem.


benchmarks/test_lexer_performance.py
```diff
@@ -0,0 +1,195 @@
+# NOTE: This file is more of a playground than a proper test
```
JuroOravec, line 1: I've also included a script I used for testing performance

src/django_components/component.py
```diff
@@ -1272,54 +1273,6 @@ def render(self, context: Context) -> str:
         return output
 
 
-def monkeypatch_template(template_cls: Type[Template]) -> None:
```
JuroOravec, line 1275: Since the applying of the custom template parser required another method of the `Template` class to be overriden, I moved them both to `django_monkeypatch.py`

src/django_components/expression.py
```diff
@@ -48,8 +50,7 @@ def __init__(self, parser: Parser, expr_str: str) -> None:
         # Copy the Parser, and pass through the tags and filters available
         # in the current context. Thus, if user calls `{% load %}` inside
         # the expression, it won't spill outside.
-        lexer = Lexer(self.expr)
-        tokens = lexer.tokenize()
+        tokens = parse_template(self.expr)
```
JuroOravec, line 53: This ensures that, when we use a template tag inside a component input like so:

```django
{% component
    description="{% lorem 3 w %}"
%}
```

then the inner `{% lorem 3 w %}` will be parsed by the updated parser too.

src/django_components/util/django_monkeypatch.py
```diff
@@ -0,0 +1,110 @@
...
+    monkeypatch_template_compile_nodelist(template_cls)
+    monkeypatch_template_render(template_cls)
+    template_cls._djc_patched = True
+
+
+# Patch `Template.compile_nodelist` to use our custom parser. Our parser makes it possible
+# to use template tags as inputs to the component tag:
+#
+# {% component "my-component" description="{% lorem 3 w %}" / %}
+def monkeypatch_template_compile_nodelist(template_cls: Type[Template]) -> None:
```
JuroOravec, line 20: This is where we override `Template.compile_nodelist`

src/django_components/util/template_parser.py
```diff
@@ -0,0 +1,227 @@
...
+#   - For a mid-sized project of 200 templates, it would take 7-8 seconds to load all the templates
+#     (from 1 second with Django's lexer).
+#     - However, thanks to django-component's lazy-loading, this should not be a problem.
+#
+#   How it works is that:
+#   1. We use Django's lexer to get the tokens.
+#   2. We check them one-by-one, and if we find a broken token, we switch to our parser to fix it.
+#   3. Once the broken token is fixed, we find it's end position, and switch back to the Django lexer
+#      for the remaining text (step 1).
+def parse_template(text: str) -> List[Token]:
```
JuroOravec, line 64: Please see the info above.

src/django_components/util/template_parser.py
```diff
@@ -0,0 +1,227 @@
...
+        index += 1
+        return char
+
+    def peek_char(offset: int = 0) -> str:
+        peek_index = index + offset
+        if peek_index >= length:
+            return ""
+        return text[peek_index]
+
+    # This is an optimized version that uses regex to find the next stop character
```
JuroOravec, line 130: I used ChatGPT to optimize the `_detailed_tag_parser` function, hence comments like this one.

src/django_components/util/template_parser.py
EmilStenstrom: Thanks for compiling this info and benchmarking it. It sounds a bit worrying to be sure, that we might add a startup cost of 8 seconds to a project, it spreading out the loading costs across the project when templates are loading. Could this mean that django-components doesn't work for larger projects that have even more templates?

EmilStenstrom (approved):
Code looks good. I'm a little worried about the perf, but I trust you if you say this will still work.

src/django_components/util/template_parser.py
JuroOravec: Hard to tell. I jsut did some extra timing on my work project, running in Docker, and there the speed penalty was about ~10%:
- before: 0.45-0.55s 
- after: 0.48-0.56s

The page I tested it on contained in total 17 templates, and I tested the execution of the whole `Template.compile_nodelist()`, not just the isolated parser.

Definitely better than the expected 8x speed difference. Though there's definitely a room for improvement. The tested numbers above were achieved by reloading the server, so no templates were in the cache. But when I reloaded a page to load it a second time, for some reason this triggered about 400 calls to `Template.compile_nodelist()`, although it took only 0.10s to load these. But no idea where these are coming from.

Plus this raises a question about performance of not this parser, but the other one (I call it the tag parser, that's the one that enables literal lists inside `{% ... %}`), as that one is not using regex, but goes char by char in a plain loop. And the tag parser covers a similar amount of code as our "slow" template parser (basically any time we come across `{% ... %}` of one of our own template tags). I didn't stop to think about performance at that one. And the numbers above, the 0.48-0.56s, since that times the whole execution of `compile_nodelist`, that includes also the execution of the tag parser...

Buuuut, I don't want to go too deep into this just yet. There's still quite a few features to build out. It'll be easier to understand what's going on once things are settled.

EmilStenstrom: Yeah, I think it makes sense to take a 10% performance hit now, and come back to performance later.

PR #910, opened 2025-01-18
feat: @template_tag and refactor how template tags are defined

JuroOravec:


docs/concepts/advanced/template_tags.md
```diff
@@ -0,0 +1,197 @@
+---
```
JuroOravec, line 1: Added a documentation page on creating custom template tags that make use of django-component's features.

Here's a preview:


https://github.com/user-attachments/assets/1e1a1dd6-8f89-452a-89a9-bcd76552a897

docs/concepts/advanced/template_tags.md
```diff
@@ -0,0 +1,197 @@
...
+{# Missing required parameter - will raise error #}
+{% greet "John" %}  {# Error: missing 'msg' #}
+```
+
+When you pass input to a template tag, it behaves the same way as if you passed the input to a function:
+
+- If required parameters are missing, an error is raised
+- If unexpected parameters are passed, an error is raised
+
+To accept keys that are not valid Python identifiers (e.g. `data-id`), or would conflict with Python keywords (e.g. `is`), you can use the `**kwargs` syntax:
```
JuroOravec, line 108: This section describes what we outlined in https://github.com/EmilStenstrom/django-components/discussions/900#discussioncomment-11859970, that kwargs like `data-id` or `class` should still be accepted. But when doing so, these kwargs must be collected onto `**kwargs`.

docs/scripts/reference.py
```diff
@@ -585,29 +586,29 @@ def _list_urls(urlpatterns: Sequence[Union[URLPattern, URLResolver]], prefix="")
     return urls
 
 
-def _format_tag_signature(tag_spec: TagSpec) -> str:
+def _format_tag_signature(node_cls: BaseNode) -> str:
```
JuroOravec, line 589: Before, the API reference for template tags was created from the `TagSpec` objects that were atttached to the template tag definitions. Now the same information is held directly by the `Node` classes

src/django_components/attributes.py
```diff
@@ -2,39 +2,84 @@
...
 from django.utils.safestring import SafeString, mark_safe
 
 from django_components.node import BaseNode
-from django_components.util.template_tag import TagParams
-
-HTML_ATTRS_DEFAULTS_KEY = "defaults"
-HTML_ATTRS_ATTRS_KEY = "attrs"
 
 
 class HtmlAttrsNode(BaseNode):
```
JuroOravec, line 14: Now, all these 4 are co-located:
1. Tag definition (e.g. whether it's `{% slot %}` or `{% provide %}`),
2. The Node class that the tag is converted to
3. The `render` logic when said Node is encountered during render
4. The source of truth for the API documentation of this template tag

So I refactored all our template tags, from `templatetags/component_tags.py`, so that the actual definitions are stored in the respective files. So now, `templatetags/component_tags.py` contains no logic, and it only declares which template tags we define.

Previously the documentation for individual tags was in `templatetags/component_tags.py`, so that's why you can see the docstrings here. The docstrings didn't change.

src/django_components/attributes.py
```diff
@@ -2,39 +2,84 @@
...
-        params: TagParams,
-        node_id: Optional[str] = None,
-    ):
-        super().__init__(nodelist=None, params=params, node_id=node_id)
-
-    def render(self, context: Context) -> str:
-        append_attrs: List[Tuple[str, Any]] = []
-
-        # Resolve all data
-        args, kwargs = self.params.resolve(context)
```
JuroOravec, line 30: Also the `render()` methods are simpler now, because they no longer have to contain logic for figuring out what inputs the template tag received. This is now done behind the scenes by `BaseNode`. And the individual template tag implementations only need to declare the signature on the `render()` method.

src/django_components/component.py
```diff
@@ -1209,44 +1209,155 @@ def _validate_outputs(self, data: Any) -> None:
...
+        params: List[TagAttr],
+        flags: Optional[Dict[str, bool]] = None,
+        nodelist: Optional[NodeList] = None,
         node_id: Optional[str] = None,
     ) -> None:
-        super().__init__(nodelist=nodelist or NodeList(), params=params, node_id=node_id)
+        super().__init__(params=params, flags=flags, nodelist=nodelist, node_id=node_id)
 
         self.name = name
-        self.isolated_context = isolated_context
```
JuroOravec, line 1226: Flags are not stored on the Node class directly, but instead can be accessed as `self.flags["flag_name"]`

src/django_components/component.py
```diff
@@ -1209,44 +1209,155 @@ def _validate_outputs(self, data: Any) -> None:
...
         self.registry = registry
 
-    def __repr__(self) -> str:
```
JuroOravec, line 1229: Common methods like `__repr__` were moved up to `BaseNode`

src/django_components/component.py
```diff
@@ -1209,44 +1209,155 @@ def _validate_outputs(self, data: Any) -> None:
...
-        return "<ComponentNode: {}. Contents: {!r}>".format(
-            self.name,
-            getattr(self, "nodelist", None),  # 'nodelist' attribute only assigned later.
+    @classmethod
+    def parse(  # type: ignore[override]
```
JuroOravec, line 1330: The `ComponentNode` is used when we use `{% component %}` template tag.

Remember that we use ComponentRegistry:
1. To allow components to be rendered either as `{% component %}` or `{% comp_name %}`
2. To allow to share components as standalone packages.

Because of that, when we use the `{% comp_name %}` ("shorthand") syntax, then to play nice with the new definition of template tags, we dynamically create a subclass of `BaseNode` / `ComponentNode`, that defines the `comp_name` / `endcomp_name` as it's start / end tag.

---

Looking at the code below, I realize that this creates a subclass of `ComponentNode` on every `{% component %}` (or custom named tag). But in reality, we need to define only one subclass for `component`, etc.

So this could be improved by storing the created subclasses, and reusing them when the `start_tag` matches (and raise if the `end_tag` or `registry` don't match).

But this is already a big MR, so I'll leave that for later.

src/django_components/component.py
```diff
@@ -1209,44 +1209,155 @@ def _validate_outputs(self, data: Any) -> None:
...
+            subcls,
+            parser,
+            token,
+            registry=registry,
+            name=name,
         )
+        return node
 
-    def render(self, context: Context) -> str:
-        trace_msg("RENDR", "COMP", self.name, self.node_id)
```
JuroOravec, line 1236: Also the tracing logs were moved into `BaseNode`.

src/django_components/util/logger.py
```diff
@@ -62,27 +62,18 @@ def trace(logger: logging.Logger, message: str, *args: Any, **kwargs: Any) -> No
 
 
 def trace_msg(
-    action: Literal["PARSE", "RENDR", "GET", "SET"],
```
JuroOravec, line 65: Removed unused bits from this function

tests/django_test_setup.py
```diff
@@ -22,6 +22,11 @@ def setup_test_config(
                     "tests/templates/",
                     "tests/components/",  # Required for template relative imports in tests
                 ],
+                "OPTIONS": {
+                    "builtins": [
+                        "django_components.templatetags.component_tags",
```
JuroOravec, line 27: This makes it so that we don't have to call `{% load component_tags %}` in tests. See https://github.com/EmilStenstrom/django-components/discussions/900#discussioncomment-11834328

src/django_components/component_registry.py
```diff
@@ -462,12 +463,39 @@ def _register_to_library(
...
-        return ComponentRegistryEntry(cls=component, tag=tag)
+        from django_components.component import ComponentNode
+
+        registry = self
+
+        # Define a tag function that pre-processes the tokens, extracting
+        # the component name and passing the rest to the actual tag function.
+        def tag_fn(parser: Parser, token: Token) -> ComponentNode:
+            # Let the TagFormatter pre-process the tokens
+            bits = token.split_contents()
```
JuroOravec, line 474: This couple of next lines previously lived in `templatetags/component_tags.py`  (https://github.com/EmilStenstrom/django-components/pull/910/files#diff-7345171ed38e1dd2e4b8199b65a2fd082323bdb331ff0ef0dc0dd3bd43f111aeL498).

It calls the TagFormatter to figure out which start and end template tags should be used with a particular component (e.g. `{% component "table" %} ... {% endcomponent %}` vs `{% table %} ... {% endtable %}`.

I moved it up here, so that, instead, the `ComponentNode` only accepts the already-resolved `start_tag` and `end_tag`.

src/django_components/dependencies.py
```diff
@@ -1036,3 +1038,66 @@ def _process_response(self, response: HttpResponse) -> HttpResponse:
...
 
         return response
+
+
+#########################################################
+# 6. Template tags
+#########################################################
+
+
+def _component_dependencies(type: Literal["js", "css"]) -> SafeString:
```
JuroOravec, line 1048: The `{% component_js_dependencies %}` and  `{% component_css_dependencies %}` were moved to this file for cosistency.

src/django_components/templatetags/component_tags.py
```diff
@@ -1,707 +1,43 @@
...
-    )
+# All tags are defined with our BaseNode class. Reasons for that are:
+# - This ensures they all have the same set of features, like supporting flags,
+#   or literal lists and dicts as parameters.
+# - The individual Node classes double as a source of truth for the tag's documentation.
+#
+# NOTE: The documentation generation script in `docs/scripts/reference.py` actually
+#   searches this file for all `Node` classes and uses them to generate the documentation.
+#   The docstring on the Node classes is used as the tag's documentation.
+ComponentNode.register(register)
```
JuroOravec, line 22: This file now only register the template tags onto the registry.

src/django_components/slots.py
```diff
@@ -150,34 +147,132 @@ class ComponentSlotContext:
...
-
-        self.is_required = is_required
-        self.is_default = is_default
-        self.trace_id = trace_id
+    """
+    Slot tag marks a place inside a component where content can be inserted
+    from outside.
 
-    @property
-    def active_flags(self) -> List[str]:
```
JuroOravec, line 171: `active_flags` was also moved to `BaseNode`, so now all our Node classes have this property

src/django_components/slots.py
```diff
@@ -368,99 +462,150 @@ def _resolve_slot_context(self, context: Context, slot_fill: "SlotFill") -> Cont
...
+    {% component "my_table" %}
+      {% fill "pagination" default="default_pag" %}
+        <div class="my-class">
+          {{ default_pag }}
+        </div>
+      {% endfill %}
+    {% endcomponent %}
+    ```
 
-    def resolve_kwargs(self, context: Context) -> "FillWithData":
```
JuroOravec, line 412: `FillNode.resolve_kwargs` and `FIllNode._process_kwarg` were removed. Now There is only `FillNode.render()`

src/django_components/util/template_tag.py
```diff
@@ -11,152 +16,36 @@
...
-        flags = self.flags.copy() if self.flags else None
-        return self.__class__(
-            signature=signature,
-            tag=self.tag,
-            end_tag=self.end_tag,
-            flags=flags,
-        )
-
-    # For details see https://github.com/EmilStenstrom/django-components/pull/902
-    def validate_params(self, params: List["TagParam"]) -> Tuple[List[Any], Dict[str, Any]]:
```
JuroOravec, line 50: `TagSpec.validate_params` is now a standalone function https://github.com/EmilStenstrom/django-components/pull/910/files#diff-56d60a8b1d11d8ff86b912ddde6d8ee539f71a1fae917bc39a1fd049ea22705cR21

src/django_components/util/template_tag.py
```diff
@@ -11,152 +16,36 @@
 from django_components.util.tag_parser import TagAttr, parse_tag
 
 
-@dataclass
-class TagSpec:
```
JuroOravec, line 15: `TagSpec` was removed. the `tag`, `end_tag`, `flags` and signature are now all defined by the `BaseNode` subclasses.

src/django_components/util/template_tag.py
```diff
@@ -180,64 +69,54 @@ class TagParam:
     value: Any
 
 
-class TagParams(NamedTuple):
```
JuroOravec, line 183: `TagParams` can be removed as a class. Previously it was meant only to ensure that the `params`, and `tag_spec` were held in a joint object. But now that's done by `BaseNode`.

So, similarly to validation, I refactored `TagParams.resolve()` into a standalone function `resolve_params`

src/django_components/util/template_tag.py
```diff
@@ -348,3 +227,99 @@ def merge_repeated_kwargs(params: List[TagParam]) -> List[TagParam]:
             params_by_key[param.key].value += " " + str(param.value)
 
     return resolved_params
+
+
+def apply_params_in_original_order(
```
JuroOravec, line 232: This logic was previously part of [`validate_params`](https://github.com/EmilStenstrom/django-components/pull/910/files#r1921049996).

The behaviour of this function relates to what we discussed in https://github.com/EmilStenstrom/django-components/discussions/900#discussioncomment-11857340.

In the end it's implemented so that template tags can have also kwargs like `data-id=123` or `class="pt-4".

`{% mytag data-id=123 class="pt-4" %}`

This would raise if these kwargs were passed like so:

`fn(data-id=123, class="pt-4")`

Because the first one is not a valid Python variable name, and the latter is a reserved keyword.

So instead, we pass those through the `extra_kwargs`, which makes these special kwargs to be applied as:

```py
extra_kwargs = {"data-id": 123, "class": "pt-4"}
fn(**extra_kwargs)
```

Which IS allowed by Python.

src/django_components/node.py
```diff
@@ -1,20 +1,460 @@
...
+
+        # Wrap cls.render() so we resolve the args and kwargs and pass them to the
+        # actual render method.
+        cls.render = wrapper_render  # type: ignore
+        cls.render._djc_wrapped = True  # type: ignore
+
+        return cls
+
+
+class BaseNode(Node, metaclass=NodeMeta):
```
JuroOravec, line 195: In this file is the core of this PR, which adds the `BaseNode` class and `@template_tag` decorator, for an easy creation of template tags that use the same features like other django-components' tags.

JuroOravec:
Note to self: After this, there's a bit of cleanup remaining:

1. I noticed that in the docs site, the page listing all Django signals is NOT shown. I think I need to push to source control the updated built reference pages in `docs/reference`.

2. Avoid duplicitly creating `ComponentNode` subclasses, as described in https://github.com/EmilStenstrom/django-components/pull/910#discussion_r1921047790

3. I still need to update the documentation regarding the use of literal dicts and lists in template tags, and other changes introduced in https://github.com/EmilStenstrom/django-components/pull/902. But after this PR things are clearer and I'll be able to do so.

src/django_components/component.py
```diff
@@ -1209,44 +1209,155 @@ def _validate_outputs(self, data: Any) -> None:
...
+        subcls: Type[ComponentNode] = type(subcls_name, (cls,), {"tag": start_tag, "end_tag": end_tag})
+
+        # Call `BaseNode.parse()` as if with the context of subcls.
+        node: ComponentNode = super().parse.__func__(  # type: ignore[attr-defined]
+            subcls,
+            parser,
+            token,
+            registry=registry,
+            name=name,
         )
```
EmilStenstrom, line 1350: I think this could be `node = super(cls, subcls).parse(parser, token, registry=registry, name=name)`, right?

tests/test_component_media.py
```diff
@@ -108,7 +108,7 @@ class TestComponent(AppLvlCompComponent):
 
         self.assertInHTML(
             """
-            <form data-djc-id-a1bc3f method="post">
+            <form data-djc-id-a1bc41 method="post">
```
EmilStenstrom, line 111: Why are these updating?

EmilStenstrom (approved):
Looks good to me. Really like the API!

tests/test_component_media.py
JuroOravec: Before, the `{% component_js_dependencies %}` tags only returned a string. Now they create instances of their own BaseNode subclasses, and return a string at `BaseNode.render()`. And those instances create one ID each [during initialization](https://github.com/EmilStenstrom/django-components/pull/910/files#diff-de799abd623d22009b507f957146f2e4851d4def29c7891fe979e04ce0aeb15dR329).

src/django_components/component.py
JuroOravec: Ah, nice one! I might have been calling `super()` with the wrong order of the the classes, because I couldn't get it to work, hence the `__func__`. Fixed now :)

PR #926, opened 2025-01-22
docs: update auto-gen'd docs to show signals reference page

JuroOravec:
Previously I've created a page where we document all Django signals (ours or otherwise) that we touch / emit / etc. While this page is showing up when running mkdocs locally, it doesn't  when the docs are released.I assume that it should show up when we add the `signals.md` page to the source control.

All the files in this PR were auto-generated when running `mkdocs serve`


JuroOravec:
cc @EmilStenstrom

JuroOravec:
Can confirm that the page now shows up in dev :)

Screenshot 2025-01-22 at 16 35 11

Issue #14, opened 2020-12-01
Out of curiosity - Performance


jonathan-s:
There is a similar library for rails -> https://viewcomponent.org/ (https://github.com/github/view_component) Apparently that library is 10x faster in rendering than partials, which is the equivalent of django's include. 

Do you have any idea what the performance characteristics are for this library?

EmilStenstrom:
@jonathan-s That is a great question. I haven't don't any benchmarking, but I would expect similar performance metrics as using a normal include. This would be easy to test I think. If you have a template that includes something in a loop, maybe several steps deep, and then at the bottom level, you put an include and a {% component %} call, and compare the loading time in a structured way.

The motivation for this library was not performance, but rather developer ergonomics in working with components. That said, I would happily accept patches that 1) adds a benchmark and 2) improved performance.

EmilStenstrom:
I think we should add a benchmark test in the form of a python script that renders a complex template 1000 times, first with include and then with component, and compares the results. If anyone is able to contribute such a script I'd gladly merge it.

mlissner:
Performance is one of the biggest concerns I have while evaluating this. My site has a number of pages with hundreds of links that we intend to convert to components. Right now, the links are generated in a Django template loop. 

If we move to components, and each link requires hitting disk to load the component, we're in a lot of trouble. I'm going to go dig into the documentation on this, but I thought I'd highlight how this is an important priority (at least for me).

Thank you!

EmilStenstrom:
@mlissner happy to include a benchmark in the library if you write one!

mlissner:
Fair! We'll look into it.

mlissner:
@EmilStenstrom What do you have in mind for the benchmark? Are you thinking of just some documentation here in the issue, a Github Action that runs on each PR, or what? I'm thinking for our purposes, we could just do a comparison of includes vs. components, and document it here. 

I think it's also worth mentioning performance on the main documentation page and linking to the caching docs. Performance was one of my biggest concerns.

mlissner:
We've done a Github Action for one of our repos that you can see here: 

https://github.com/freelawproject/reporters-db/pull/190#issuecomment-2518773885

It's nice b/c it runs on every PR and so in theory you can see the before and after of the PR. It's a lot of work to maintain though (Github is constantly changing permissions, etc). 

---

[...]

PR #927, opened 2025-01-23
refactor: replace bs4 and perf optimizations

JuroOravec:
Part of #14 

Changes:
- Removed dependency on BeautifulSoup4.
- Added custom HTML parser implementation.
- Allow to write self-closing HTML elements within component template (e.g. `<div />`
  - Plus added documentation section on that
- Couple of optimization refactors (see individual comments)

The plan of action is:
1. Add the pure Python implementation in this PR
2. Create the repo for the Rust impl and make a PR for that.
3. Lastly, add the Rust impl as a Python dependency for django-components, plus add setting to allow users to fall back onto the pure Python implementation, if, for whatever reason, they cannot use the Rust implementation. (as suggested [here](https://github.com/django-components/django-components/issues/14#issuecomment-2604176119))

### DIFF
--- a/tests/test_expression.py
+++ b/tests/test_expression.py
@@ -95,9 +95,7 @@ class DynamicExprTests(BaseTestCase):
                 bool_var="{{ is_active }}"
                 list_var="{{ list|slice:':-1' }}"
             / %}
-        """.replace(
-                "\n", " "
-            )
+            """
         )
 
         template = Template(template_str)
@@ -116,14 +114,14 @@ class DynamicExprTests(BaseTestCase):
         self.assertEqual(captured["bool_var"], True)
         self.assertEqual(captured["list_var"], [{"a": 1}, {"a": 2}])
 
-        self.assertEqual(
-            rendered.strip(),
-            (
-                "<!-- _RENDERED SimpleComponent_5b8d97,a1bc3f,, -->\n"
-                "                <div data-djc-id-a1bc3f>lorem</div>\n"
-                "                <div data-djc-id-a1bc3f>True</div>\n"
-                "                <div data-djc-id-a1bc3f>[{'a': 1}, {'a': 2}]</div>"
-            ),
+        self.assertHTMLEqual(
+            rendered,
+            """
+            <!-- _RENDERED SimpleComponent_5b8d97,a1bc3f,, -->
+            <div data-djc-id-a1bc3f>lorem</div>
+            <div data-djc-id-a1bc3f>True</div>
+            <div data-djc-id-a1bc3f>[{'a': 1}, {'a': 2}]</div>
+            """,
         )
 
     @parametrize_context_behavior(["django", "isolated"])
@@ -169,9 +167,7 @@ class DynamicExprTests(BaseTestCase):
                 list_var="{% noop list %}"
                 dict_var="{% noop dict %}"
             / %}
-        """.replace(
-                "\n", " "
-            )
+            """
         )
 
         template = Template(template_str)
@@ -191,15 +187,15 @@ class DynamicExprTests(BaseTestCase):
         self.assertEqual(captured["dict_var"], {"a": 3})
         self.assertEqual(captured["list_var"], [{"a": 1}, {"a": 2}])
 
-        self.assertEqual(
-            rendered.strip(),
-            (
-                "<!-- _RENDERED SimpleComponent_743413,a1bc3f,, -->\n"
-                "                <div data-djc-id-a1bc3f>lorem ipsum dolor</div>\n"
-                "                <div data-djc-id-a1bc3f>True</div>\n"
-                "                <div data-djc-id-a1bc3f>[{'a': 1}, {'a': 2}]</div>\n"
-                "                <div data-djc-id-a1bc3f>{'a': 3}</div>"
-            ),
+        self.assertHTMLEqual(
+            rendered,
+            """
+            <!-- _RENDERED SimpleComponent_743413,a1bc3f,, -->
+            <div data-djc-id-a1bc3f>lorem ipsum dolor</div>
+            <div data-djc-id-a1bc3f>True</div>
+            <div data-djc-id-a1bc3f>[{'a': 1}, {'a': 2}]</div>
+            <div data-djc-id-a1bc3f>{'a': 3}</div>
+            """,
         )
 
     @parametrize_context_behavior(["django", "isolated"])
@@ -245,9 +241,7 @@ class DynamicExprTests(BaseTestCase):
                 bool_var="{# noop is_active #}"
                 list_var=" {# noop list #} "
             / %}
-        """.replace(
-                "\n", " "
-            )
+            """
         )
 
         template = Template(template_str)
@@ -267,6 +261,7 @@ class DynamicExprTests(BaseTestCase):
         self.assertEqual(captured["bool_var"], "")
         self.assertEqual(captured["list_var"], "  ")
 
+        # NOTE: This is whitespace-sensitive test, so we check exact output
         self.assertEqual(
             rendered.strip(),
             (
@@ -325,9 +320,7 @@ class DynamicExprTests(BaseTestCase):
                 list_var=" {% noop list %} "
                 dict_var=" {% noop dict %} "
             / %}
-        """.replace(
-                "\n", " "
-            )
+            """
         )
 
         template = Template(template_str)
@@ -347,6 +340,7 @@ class DynamicExprTests(BaseTestCase):
         self.assertEqual(captured["dict_var"], " {'a': 3} ")
         self.assertEqual(captured["list_var"], " [{'a': 1}, {'a': 2}] ")
 
+        # NOTE: This is whitespace-sensitive test, so we check exact output
         self.assertEqual(
             rendered.strip(),
             (
@@ -388,9 +382,7 @@ class DynamicExprTests(BaseTestCase):
             """
             {% load component_tags %}
             {% component 'test' '"' "{%}" bool_var="{% noop is_active %}" / %}
-            """.replace(
-                "\n", " "
-            )
+            """
         )
 
         template = Template(template_str)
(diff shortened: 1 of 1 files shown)


src/django_components/component.py
```diff
@@ -1208,6 +1247,14 @@ def _validate_outputs(self, data: Any) -> None:
         validate_typed_dict(data, data_type, f"Component '{self.name}'", "data")
 
 
+# Perf
```
JuroOravec, line 1250: This section below relates to second point in https://github.com/django-components/django-components/pull/910#issuecomment-2599692113:

> Avoid duplicitly creating ComponentNode subclasses, as described in https://github.com/django-components/django-components/pull/910#discussion_r1921047790

src/django_components/component.py
```diff
@@ -1062,22 +1074,49 @@ def _render_impl(
...
-                    render_dependencies=render_dependencies,
-                )
-
         # After rendering is done, remove the current state from the stack, which means
         # properties like `self.context` will no longer return the current state.
         self._render_stack.pop()
         context.render_context.pop()
 
-        return output
+        # Internal component HTML post-processing:
```
JuroOravec, line 1082: This is part of the change to make it possible to parse the component HTML only once. It's achieved by deferring the postprocessing of the component output until its parent has been post-processed.

For context see ["Idea 2 - Render components top-down instead of bottom-up"](https://github.com/django-components/django-components/issues/14#issuecomment-2596096632)

tests/test_expression.py
```diff
@@ -118,7 +118,12 @@ def get_context_data(
 
         self.assertEqual(
             rendered.strip(),
-            "<!-- _RENDERED SimpleComponent_5b8d97,a1bc3f,, -->\n<div data-djc-id-a1bc3f>lorem</div>\n<div data-djc-id-a1bc3f>True</div>\n<div data-djc-id-a1bc3f>[{'a': 1}, {'a': 2}]</div>",  # noqa: E501
+            (
+                "<!-- _RENDERED SimpleComponent_5b8d97,a1bc3f,, -->\n"
```
JuroOravec, line 122: BS4 stripped extra whitespace, while our custom pure Python implementation leaves it as is, hence these changes.

src/django_components/util/html_parser.py
```diff
@@ -0,0 +1,1147 @@
...
+#
+# 5. Text
+#   5.1. Text allowed only as content (NOT inside tags or attributes)
+#   5.2. If in text, text WILL end at `<` (start of start tag), `</` (start of end tag),
+#        `<!--` (start of comment), `<![CDATA[` (start of CDATA)
+def _parse_html(
+    text: str,
+    on_tag: Callable[[HTMLTag, List[HTMLTag]], None],
+    *,
+    expand_shorthand_tags: bool,
```
JuroOravec, line 551: Ah, one more on the HTML parser - So I initially implemented it for the use case of converting Vue syntax to Django, I also added the support for expanding self-closing HTML attributes.

In other words, in standard HTML, this is invalid:

```html
<div />
```

Because `<div>` is NOT a [void element](https://developer.mozilla.org/en-US/docs/Glossary/Void_element) like `<input />`, ``, etc.

However, what Vue does is that when it sees a "self-closing" HTML element like that, which is NOT a void element, it expands it into

```html
<div></div>
```

For Vue and React I believe it's about normalizing the behaviour of components embedded inside the template, and the rest of HTML elements.

E.g. if one can do this in Vue / React:
```vue
<MyComp />
```

Then for convenence, they allow the same syntax also for standard HTML elements:

```html
<div />
```

So now we'd also allow this syntax for component templates.

tests/test_html_parser.py
```diff
@@ -0,0 +1,496 @@
...
+from django_components.util.html_parser import HTMLTag, _parse_html as parse_html, set_html_attributes
+
+from .django_test_setup import setup_test_config
+
+setup_test_config({"autodiscover": False})
+
+
+# This same set of tests is also found in djc_html_parser, to ensure that
+# this implementation can be replaced with the djc_html_parser's Rust-based implementation
+class TestHTMLParser(TestCase):
```
JuroOravec, line 13: This file contains 2 kinds of files

This `TestHTMLParser` is the contract that we expect from the HTML parser. So even if we switch to the Rust implementation, we will expect these tests to pass.

On the other hand, the tests defined in [TestHTMLParserInternal](https://github.com/django-components/django-components/pull/927/files#diff-09e1aa504d5622adf262aa999e73036eebeeace1a28cdf08ed01e85a80dcb27eR163) relate only to the pure-python implementation of the HTML parser.

Once I get to the PR that adds the Rust implementation, I might split this test file, so the pure python impl is separate from the expected "HTMLParser" API.

---

Going one step further, it might even make sense for the pure-python implementation to live as a separate package too, so one wouldn't have to load that if not needing it. But also there's no harm in the pure-python impl being here.

src/django_components/dependencies.py
```diff
@@ -734,6 +761,10 @@ def get_component_media(comp_cls_hash: str) -> Media:
     return (content, final_script_tags.encode("utf-8"), final_css_tags.encode("utf-8"))
 
 
+href_pattern = re.compile(r'href="([^"]+)"')
```
JuroOravec, line 764: The rest of the changes in this file is about removing dependency on BS4. For example here, to extract the URL from `<script>` or `<link>`, we'll use a regex instead of relying on BS4.

src/django_components/dependencies.py
```diff
@@ -917,37 +957,50 @@ def _insert_js_css_to_default_locations(
     This function tries to insert the JS and CSS content into the default locations.
 
     JS is inserted at the end of `<body>`, and CSS is inserted at the end of `<head>`.
-    """
-    elems = SoupNode.from_fragment(html_content)
 
-    if not elems:
+    We find these tags by looking for the first `</head>` and last `</body>` tags.
```
JuroOravec, line 961: And this section is about inserting JS and CSS into their default locations.

Now, to avoid having to parse the HTML to find the end of body and head, we use following approach:
  - We find the end of `<body>` by searching for **last** occurrence of `</body>`
  - And for the end of `<head>` we search for the **first** occurrence of `</head>`

src/django_components/perfutil/component.py
```diff
@@ -0,0 +1,169 @@
+import re
```
JuroOravec, line 1: In this file I implemented the top-down approach to rendering nested components, as desribed in[ Idea 2 - Render components top-down instead of bottom-up](https://github.com/django-components/django-components/issues/14#issuecomment-2596096632).

I've put this into a separate file, so I could proprly document and explain what's going on, and to keep the `component.py` file relatively navigatable, since it's already quite big.

Please see the comments throughout this file for details.

src/django_components/util/html_parser.py
EmilStenstrom: This is strange to me, but I don't quite see the harm other than code complexity. If this is common in Vue I'm fine with it.

tests/test_html_parser.py
EmilStenstrom: If we don't expect any problems with being able to install the rust version, I'm not sure why we should maintain two versions. Are we expecting problems?

CHANGELOG.md
```diff
@@ -1,5 +1,32 @@
...
+    ```html
+    <div style="background-image: url('...');" />
+    ```
+
+    will be expanded to:
+
+    ```html
+    <div style="background-image: url('...');">
+    </div>
+    ```
```
EmilStenstrom, line 20: This feels very random, don't think it has anything to do with django-components at all.

pyproject.toml
```diff
@@ -29,7 +29,6 @@ classifiers = [
 ]
 dependencies = [
     'Django>=4.2',
-    'beautifulsoup4>=4.12',
```
EmilStenstrom, line 32: Back to no dependencies! 🥳

tests/test_expression.py
EmilStenstrom: I think fixing the tests would be a more long-term bet in this case. We don't care about the whitespace, so lets skip it in the tests.

src/django_components/util/html_parser.py
```diff
@@ -0,0 +1,1147 @@
...
+            add_token(char)
+            continue
+
+        else:
+            raise ValueError(f"Invalid state '{state}' with character '{char}' at index {index} in text '{text}'")
+
+    return "".join(tokens)  # Join all tokens at the end
+
+
+def set_html_attributes(
```
EmilStenstrom, line 1077: This is a super-custom HTML parser, so i see what you mean now. Looks good!

EmilStenstrom (approved):
A couple of small nitpicks you can decide to ignore if you want, but I think this looks really good. Happy that you could solve this super-tricky problem so quickly.

CHANGELOG.md
JuroOravec: Oki, I'll move this out of this PR, and instead leave it for django-vue.

EmilStenstrom: That is a good solution imho! ⭐

tests/test_expression.py
JuroOravec: Updated in https://github.com/django-components/django-components/pull/927/commits/e756713adc78fd41dba35b808ae2e90af67f3943.

There's 2-3 tests that are whitespace-sensitive, so there I've kept using `assertEqual`. The rest of the tests I converted to use `assertHTMLEqual`, which ignores whitespace.

tests/test_html_parser.py
dalito: What I was thinking of:
- By looking at the python code one could understand what the hidden binary does. 
- You could plugin yet another solution easily. 
- If you don't want to have compiled code in your code base you can still use djc.

But keeping the code base small and less complex is a very strong counter argument. So doing that is absolutely fine by me.  My comment was purely based on discussion that I observed in the past.

EmilStenstrom: Good arguments dalito. This is my thoughts around those three points:
* I'm hoping that by keeping the other library close, you can easily go there and look at the code to understand the Rust code if you wanted.
* Since the above code defines an interface and has a good set of tests, I think that might be enough to understand enough about the project to build another plugin. I don't think we need two implementations to reach that goal.
* I think this is valid, but I don't understand how wide-spread this need is. A huge part of the Python ecosystem is based on python wrappers of C code, so I'm not sure this is even viable for a Django project. You are going to use a database for your website, and those python database drivers are likely C backed?

My hunch is that the maintenance overhead for this is larger that what it's worth, but I'm happy to hear your thoughts.

This doesn't mean this PR can't be merged, but maybe that we'll remove this code when we add the Rust-based parser?

EmilStenstrom: Sounds like a plan!

PR #929, opened 2025-01-26
feat: replacepython html parser djc-core-html-parser

JuroOravec:
This PR replaces the custom Python HTML parser implementation with our Rust impl. This brings down the initial loading time on one of my biggest pages from 10s down to 1.8s.

NOTE: On subsequent page requests, this gets further down to 1.1 s. And of that, about 50% is because there's bug in my work code where I fetch related fields one by one from within the template.

So overall, after I fix the glaring performance issues like in this PR, then we'll be at about 0.5s for rendering ~360 components (2MB to HTML).

Part of #14 

Before / after:

Screenshot 2025-01-26 at 15 35 07
Screenshot 2025-01-26 at 15 34 55


src/django_components/dependencies.py
```diff
@@ -438,9 +438,9 @@ def postprocess_component_html(
     rb"^(?P<comp_cls_hash>[\w\-\./]+?),(?P<id>[\w]+?),(?P<js>[0-9a-f]*?),(?P<css>[0-9a-f]*?)$"
 )
 # E.g. `data-djc-id-a1b2c3`
-MAYBE_COMP_ID = r"(?: data-djc-id-\w{6})?"
+MAYBE_COMP_ID = r'(?: data-djc-id-\w{6}="")?'
```
JuroOravec, line 441: When the Rust impl adds a boolean attribute to an HTML element, it does so as `attribute=""`, hence this change here, and change in some of the tests elsewhere in this PR.

tests/test_html_parser.py
```diff
@@ -138,340 +156,9 @@ def test_whitespace_preservation(self):
...
-            <span data-all> Text with spaces </span>
+        expected = """<div data-root="" data-all="">
+            <p data-all="">  Hello  World  </p>
+            <span data-all=""> Text with spaces </span>
         </div>"""
-        assert result == expected
-
-
-# This checks that the parser works irrespective of the main use case
-class TestHTMLParserInternal(TestCase):
```
JuroOravec, line 149: Removed the tests that were specifically for the Python implementation.

tests/test_html_parser.py
```diff
@@ -14,14 +13,27 @@ class TestHTMLParser(TestCase):
     def test_basic_transformation(self):
         html = "<div><p>Hello</p></div>"
         result, _ = set_html_attributes(html, root_attributes=["data-root"], all_attributes=["data-all"])
-        expected = "<div data-root data-all><p data-all>Hello</p></div>"
-        assert result == expected
+
```
JuroOravec, line 16: I've updated the tests in `TestHTMLParser` in this file, to use Django's `assertHTMLEqual` instead of plain `assert`.

EmilStenstrom (approved):
Annoying to have to add back the attr="" test fixes, but not worth the hassle to fix I guess. Excellent job on the perf here, as I've said before! :)

Issue #930, opened 2025-01-27
Media caching (possibly) broken when using non-local cache for template fragments


oliverhaas:
I've just observed some errors in production and had to revert, and I thought I'd share early and investigate further later.

My setup uses 3 web workers with Redis/Valkey as a default cache backend.

I get this error occasionally:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/django/core/handlers/exception.py", line 42, in inner
    response = await get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/sentry_sdk/integrations/django/asgi.py", line 243, in __acall__
    return await f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django_components/dependencies.py", line 1031, in __acall__
    response = self._process_response(response)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django_components/dependencies.py", line 1038, in _process_response
    response.content = render_dependencies(response.content, type="document")
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django_components/dependencies.py", line 475, in render_dependencies
    content_, js_dependencies, css_dependencies = _process_dep_declarations(content_, type)
                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django_components/dependencies.py", line 627, in _process_dep_declarations
    ) = _prepare_tags_and_urls(comp_data, type)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django_components/dependencies.py", line 793, in _prepare_tags_and_urls
    comp_cls = comp_hash_mapping[comp_cls_hash]
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/weakref.py", line 136, in __getitem__
    o = self.data[key]()
        ^^^^^^^^^^^^^^
```

I've invalidated/cleared my cache completely, but the issue pops up again.

[...]

PR #934, opened 2025-01-30
fix: implement safe retrieval of component classes from registry

oliverhaas:
To illustrate a possible fix of #930

EDIT:

I think this is decent by now.

I'm not sure how I could write a test for this, at least not in a way which would provoke the error from the issue above without a disproportionate amount of work, but all tests pass and many of them rely on this logic working.

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -51,9 +51,9 @@ from django_components.dependencies import (
     cache_component_css_vars,
     cache_component_js,
     cache_component_js_vars,
+    comp_hash_mapping,
     postprocess_component_html,
     set_component_attrs_for_js_and_css,
-    comp_hash_mapping,
 )
 from django_components.node import BaseNode
 from django_components.perfutil.component import component_post_render
@@ -72,7 +72,7 @@ from django_components.slots import (
 )
 from django_components.template import cached_template
 from django_components.util.django_monkeypatch import is_template_cls_patched
-from django_components.util.misc import hash_comp_cls, gen_id
+from django_components.util.misc import gen_id, hash_comp_cls
 from django_components.util.template_tag import TagAttr
 from django_components.util.validation import validate_typed_dict, validate_typed_tuple
 
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -1,4 +1,3 @@
-import inspect
 import types
 from collections import deque
 from contextlib import contextmanager
@@ -54,6 +53,7 @@ from django_components.dependencies import (
     cache_component_js_vars,
     postprocess_component_html,
     set_component_attrs_for_js_and_css,
+    comp_hash_mapping,
 )
 from django_components.node import BaseNode
 from django_components.perfutil.component import component_post_render
@@ -72,7 +72,7 @@ from django_components.slots import (
 )
 from django_components.template import cached_template
 from django_components.util.django_monkeypatch import is_template_cls_patched
-from django_components.util.misc import gen_id
+from django_components.util.misc import hash_comp_cls, gen_id
 from django_components.util.template_tag import TagAttr
 from django_components.util.validation import validate_typed_dict, validate_typed_tuple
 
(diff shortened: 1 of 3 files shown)


src/django_components/util/misc.py
```diff
@@ -71,3 +76,9 @@ def get_last_index(lst: List, key: Callable[[Any], bool]) -> Optional[int]:
 
 def is_nonempty_str(txt: Optional[str]) -> bool:
     return txt is not None and bool(txt.strip())
+
+
```
oliverhaas, line 80: Removed lru_cache from this function, since this function is now only ever called once per component class.

src/django_components/component.py
```diff
@@ -556,7 +556,7 @@ def on_render_after(self, context: Context, template: Template, content: str) ->
     # MISC
     # #####################################
 
-    _class_hash: ClassVar[int]
```
oliverhaas, line 559: This int hash was only used in the component_registry.py:319, and any kind of hash would work for that.

The hash used in dependencies.py is both human-readable and works just as well, so we just combine them to one hash per class.

src/django_components/util/misc.py
```diff
@@ -71,3 +76,9 @@ def get_last_index(lst: List, key: Callable[[Any], bool]) -> Optional[int]:
...
+def hash_comp_cls(comp_cls: Type["Component"]) -> str:
```
oliverhaas, line 81: Not sure if this is a good place for this function. It's only used in component.py, but it's a really generic utility function actually.

src/django_components/component.py
```diff
@@ -54,6 +53,7 @@
     cache_component_js_vars,
     postprocess_component_html,
     set_component_attrs_for_js_and_css,
+    comp_hash_mapping,
```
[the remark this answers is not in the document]
JuroOravec: Totally, agree, I was thinking too about moving it to `component.py` when I was looking into this issue yesterday.

src/django_components/component.py
JuroOravec: Again agree - this one predates me, so I never touched it. I digged and found that this was used introduced when the Component as View feature was added in https://github.com/django-components/django-components/pull/366. ([see source](https://github.com/django-components/django-components/commit/91b4accfebf74f3f7a8cf2ecab03789a1979d578#diff-a55586fd92ed66912ae3b950f873e4d2f8709509ddcc0e41ffeed0ab710b5ccbR17)).

And yeah, looks like we can use the other hash from dependencies.py instead

src/django_components/dependencies.py
```diff
@@ -1031,9 +1013,7 @@ def cached_script_view(
     if req.method != "GET":
         return HttpResponseNotAllowed(["GET"])
 
-    comp_cls = comp_hash_mapping.get(comp_cls_hash)
-    if comp_cls is None:
-        return HttpResponseNotFound()
```
JuroOravec, line 1036: Here I think it makes sense to keep `HttpResponseNotFound`, because it's an endpoint.

It could happen that a curious end user could try to use an endpoint like `/components/cache/table_0ab2c3.js`, but just trying out random numbers, so the value of `comp_cls_hash` would be actually invalid. And in that case I think it'd be better to return 404.

What do you think?

src/django_components/util/misc.py
JuroOravec: Yup, here is great! I have a lot of prototypes and half-baked features on my local branch, and there I was also thinking of moving this function here 😄 

I also plan to use `hash_comp_cls` for the CSS scoping feature, so this file totally makes sense.

JuroOravec:
@oliverhaas Looks great! Just one comment about the view function, the rest looks crisp!

Agree, I don't think this is something that can be meaningfully tested.

src/django_components/dependencies.py
oliverhaas: Yeah, makes sense.

PR #936, opened 2025-01-31
refactor: fix component recursion

### LINKED ISSUE #923: Fix potential recursion limitations
Btw, in terms of performance / reliability, I realized there might be one more limitation for large projects, in terms of recursion. 

I'd have to check, but when one renders a component from within the template tag using `{% component ... %}`, then a single call like this may generate 6-7 levels of calls. E.g. it might look like:

 `template.render()-> template._render() -> NodeList.render() -> ComponentNode.render() -> Component._render() -> Component._render_impl()`

Normally python allows recursion up to 1000 levels, which would allow components of depth of ~140 layers. So this should be fine for 99% of the projects. But I can imagine that that if there was a component like a [treeview](https://dribbble.com/tags/treeview), which would generate nested component depending on given data, then that limit could be reached.

So I want to check if this could be an issue.

With the changes I made with regards to optimistions (that I yet have to make a PR for), I think we could make further changes so that, instead of rendering components recursively, we'd have a render stack, rendering componenents depth-first and first-in-first-out. I think this would allow for practically infinite depth of components (limited only by the device's memory).

I think the trade-off for "infinite" depth would be that we'd have to a make "flattened" copies of the Context object for each component render instance ([`Context.flatten()`](https://docs.djangoproject.com/en/5.1/ref/templates/api/#django.template.Context.flatten)). This is so that we can take a "snapshot" of what the context looked like at the point when it reached `{% component %}`

This has lower priority than the performance issues. I made a ticket for that here https://github.com/django-components/django-components/issues/923

JuroOravec:
Closes #923 

## Background

The main goal of this PR was to work around the recusion limits when rendering deeply-nested (see https://github.com/django-components/django-components/issues/923).

I achieved this by:

1. If component is nested in another, instead of the inner component directly rendering its contents, it returns only a placeholder. After the root component is rendered, we check the HTML for any component placeholders, and replace them. It's at this point when we actually render the child components. This repeats recursively - if child component has subcomponents, those subcomponnt will first return only a placeholder, and will be rendered in next iteration. Thus, parent components get to be rendered before child components.

2. To avoid needing to use recursion, when a component's template is about to be rendered, I make a copy (a "snapshot") of the Context object, as it was at that point. Thus we isolate the data needed to render the components template, and we can render it at later time, simply by passing the context snapshot to the Template as `template.render(context)`.

   - I call "snapshot" a copy of the original Context, which can be moved freely. E.g. Django uses context managers to temporarily apply extra layers to the Context:
      ```py
      with context.push({"key": val}):
          print(context["key"])  # works
      print(context["key"])  # KeyError
      ```

      Instead, the context copies the Context layers, so they won't be affected when we leave the context
      ```py
      with context.push({"key": val}):
          snapshot = make_snapshot(context)
          print(snapshot["key"])  # works
      print(snapshot["key"])  # works
      ```

3. And with the above, we can avoid recursion when rendering the components by keeping a queue of components to render, and processing it depth-first, rendering components one at a time, as described in point 1).

So now the components can be nested at arbitrary depth, limited only by available memory.

---

## Changes

However, it was a real pain (it took me what, full 2.5 days?) to figure out what was wrong. Supporting all the different modes:
- Isolated vs django
- Slots/fills
- {% extends %} / {% block %}

and their interaction, wasn't easy to figure out what was going on wrong.

So to be able to debug what is going on, I really had to move things out of the way, so I could understand what is going on with the Context object at each step of the way.

Thus I made some improvements to debugging:

- When a component raises an error, the error message now also includes the slot path ([see here](https://github.com/django-components/django-components/pull/936/files#diff-b35cc8d48a3ea2ddde0c3bc9de0b0b768c567f9e8f532dc4ca1b260fe06e4ac8R428))
   
   E.g. `Root > parent > provider > provider(slot:content) > broken`

- Better tracing - I added tracing messages to when:
   1. Component "prep" starts (that's when we enter `Component.render()`
   2. Component "prep" ends
   3. Actual component render starts
   4. Actual component render end
   5. Slot render starts
   6. Slot render ends
   7. When rendering slot fill that was defined via `{% fill %}` tag

- Instead of storing parent component data directly on the Context, the Context object now contains only a key, and actual data is stored in a separate dictionary.

   The issue is that when there's nested components, then we store also the `outer_context` AKA the state of `Context`, as it was right outside of the Component
   ```django
                    <-- This is outer context
   {% component "table" %}
      {% fill headers %}
        {{ ... }}       <-- This should have access to vars
                                defined in the outer context
      {% endfill %}
   {% endcomponent %}
   ```
   
   This made it really hard to debug. Because when I tried to print out individual layers of the Context, it would also print out the `outer_context` Context object. And if this was a deeply-nested component, then that `outer_context` had an `outer_context`, and that had an `outer_context`, and so on...

   Instead, now there is only the ID there. See below, where the `_DJC_COMPONENT_CTX` is only a short string - the key is actually the component's ID.

   Screenshot 2025-01-28 at 09 09 13

   The added bonus is that it's now easy to access data of any layer. This could be potentially used also for testing, where it would be easy to mock or replace some provided or parent component data, simply by modifying the entries in the dictionary.

- As you can see in the screenshot from previous point, I did the same for data defined via `{% provide %}`. Before, they were defined directly on the Context, now its refered to only by its key.

   In both cases, the component data and the provide data, there is a tradeoff - Because we store the data in separate dictionaries, we have to manually manage their memory - keeping track when when corresponding components are finished rendering, and delete the data from the cache dictionaries.

   With the data from `{% provide %}` it's a bit more complex than with the component data, because for Provide we have to do reference counting - keep track of which components are being rendered while inside `{% provide %}`, and allow to delete the provided data only once all those components are done rendering.

   Because of this, I added the `perfutil/provide.py` file, to house this logic, as its related to optimization of the provide feature, but not to the "business" logic of provide itself.

- Better tracing for slots - I updated how the slot are displayed. So when we print them , the slot shows where it was originally defined - AKA which component it was passed and under which slot name:

   Screenshot 2025-01-28 at 09 52 00

   To achieve this, I had to pass extra info about the component name & id / slot name here and there.

- Visualizing slots - I also used a snippet to display where and how slots are displayed

   **NOTE: This is NOT included in this PR**, this was simple enough that it could be extracted without breaking everything else. So the plan is to make a separate PR with this, and add a page on troubleshooting, and mention it there.

   Screenshot 2025-01-27 at 22 53 55

## Next steps

1. As mentioned above, I want to add the diagnostics feature to highlight slots - actually, maybe this could also highlight components.

2. This PR breaks the `on_render_after` hook, so I want to fix it in separate PR.
    
   Because while before, the hook received a fully-rendered HTML, now it receives HTML, but with placeholders instead of rendered sub-components. So it may look like this:
   ```html
   <div>
     <template djc-render-id="a1b2c3"></template>
   </div>
   ```

   instead of e.g.
   ```html
   <div>
     <button class="btn btn-float" data-djc-id="a1b2c3">
       hello click me!
       ...
     </button>
   </div>
   ```

   the issue here is that, in the latest implementation, we build the final HTML only for the root component. For all subcomponents, we keep their HTML as a list of `[STATIC_TEXT, SUBCOMPONENT_OUTPUT, STATIC_TEXT, ...]

   This could be refactored to intermediately join up the HTML of the subcomponents, so that we could pass their HTML to the `on_render_after` hook. But I want to do so already in a separate PR.

### DIFF
--- a/sampleproject/components/recursive.py
+++ b/sampleproject/components/recursive.py
@@ -1,0 +1,31 @@
+from typing import Any, Dict
+
+from django_components import Component, register, types
+
+
+@register("recursive")
+class Recursive(Component):
+    def get(self, request):
+        import time
+        time_before = time.time()
+        output = self.render_to_response(
+            kwargs={
+                "depth": 0,
+            },
+        )
+        time_after = time.time()
+        print("TIME: ", time_after - time_before)
+        return output
+
+    def get_context_data(self, depth: int = 0) -> Dict[str, Any]:
+        return {"depth": depth + 1}
+
+    template: types.django_html = """
+        <div id="recursive">
+            depth: {{ depth }}
+            <hr/>
+            {% if depth <= 100 %}
+                {% component "recursive" depth=depth / %}
+            {% endif %}
+        </div>
+    """
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -1020,8 +1020,9 @@ class Component(
             is_filled=None,
         )
 
-        # If any slot fills were defined within the template, we want to scope them
-        # to the CSS of the parent component. Thus we keep track of the parent component.
+        # We pass down the components the info about the component's parent.
+        # This is used for correctly resolving slot fills, correct rendering order,
+        # or CSS scoping.
         if context.get(_COMPONENT_CONTEXT_KEY, None):
             parent_id = cast(str, context[_COMPONENT_CONTEXT_KEY])
             parent_comp_ctx = component_context_cache[parent_id]
@@ -1139,6 +1140,8 @@ class Component(
             del component_context_cache[component_id]  # type: ignore[arg-type]
             unregister_provide_reference(component_id)  # type: ignore[arg-type]
 
+        # After the component and all its children are rendered, we resolve
+        # all inserted HTML comments into <script> and <link> tags (if render_dependencies=True)
         def on_html_rendered(html: str) -> str:
             if render_dependencies:
                 html = _render_dependencies(html, type)
@@ -1180,12 +1183,6 @@ class Component(
     # ```html
     # <div djc-id-a1b3cf djc-id-f3d3cf>...</div>
     # ```
-    #
-    # ---
-    #
-    # After the component is rendered, we post-process the output:
-    # - Add the HTML attributes to work with JS and CSS variables
-    # - Resolve component's JS / CSS into <script> and <link> (if render_dependencies=True)
     def _gen_component_renderer(
         self,
         render_id: str,
@@ -1223,6 +1220,7 @@ class Component(
                 new_output = component.on_render_after(context, template, html_content)
                 html_content = new_output if new_output is not None else html_content
 
+            # Add necessary HTML attributes to work with JS and CSS variables
             updated_html, child_components = set_component_attrs_for_js_and_css(
                 html_content=html_content,
                 component_id=render_id,
@@ -1231,8 +1229,7 @@ class Component(
                 root_attributes=root_attributes,
             )
 
-            # Mark the generated HTML so that we will know which JS and CSS
-            # scripts are associated with it.
+            # Prepend an HTML comment to instructs how and what JS and CSS scripts are associated with it.
             updated_html = insert_component_dependencies_comment(
                 updated_html,
                 component_cls=component_cls,
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -27,10 +27,10 @@ from typing import (
 from django.core.exceptions import ImproperlyConfigured
 from django.forms.widgets import Media as MediaCls
 from django.http import HttpRequest, HttpResponse
-from django.template.base import NodeList, Parser, Template, TextNode, Token
+from django.template.base import NodeList, Origin, Parser, Template, TextNode, Token
 from django.template.context import Context, RequestContext
 from django.template.loader import get_template
-from django.template.loader_tags import BLOCK_CONTEXT_KEY
+from django.template.loader_tags import BLOCK_CONTEXT_KEY, BlockContext
 from django.test.signals import template_rendered
 from django.utils.html import conditional_escape
 from django.views import View
@@ -39,26 +39,24 @@ from django_components.app_settings import ContextBehavior
 from django_components.component_media import ComponentMediaInput, ComponentMediaMeta
 from django_components.component_registry import ComponentRegistry
 from django_components.component_registry import registry as registry_
-from django_components.context import (
-    _COMPONENT_SLOT_CTX_CONTEXT_KEY,
-    _REGISTRY_CONTEXT_KEY,
-    _ROOT_CTX_CONTEXT_KEY,
-    get_injected_context_var,
-    make_isolated_context_copy,
-)
+from django_components.context import _COMPONENT_CONTEXT_KEY, make_isolated_context_copy
 from django_components.dependencies import (
     RenderType,
     cache_component_css,
     cache_component_css_vars,
     cache_component_js,
     cache_component_js_vars,
-    postprocess_component_html,
+    insert_component_dependencies_comment,
+)
+from django_components.dependencies import render_dependencies as _render_dependencies
+from django_components.dependencies import (
     set_component_attrs_for_js_and_css,
 )
 from django_components.node import BaseNode
-from django_components.perfutil.component import component_post_render
+from django_components.perfutil.component import ComponentRenderer, component_context_cache, component_post_render
+from django_components.perfutil.provide import register_provide_reference, unregister_provide_reference
+from django_components.provide import get_injected_context_var
 from django_components.slots import (
-    ComponentSlotContext,
     Slot,
     SlotContent,
     SlotFunc,
@@ -71,8 +69,11 @@ from django_components.slots import (
     resolve_fills,
 )
 from django_components.template import cached_template
+from django_components.util.context import snapshot_context
 from django_components.util.django_monkeypatch import is_template_cls_patched
-from django_components.util.misc import gen_id
+from django_components.util.exception import component_error_message
+from django_components.util.logger import trace_component_msg
+from django_components.util.misc import gen_id, get_import_path
 from django_components.util.template_tag import TagAttr
 from django_components.util.validation import validate_typed_dict, validate_typed_tuple
 
@@ -99,7 +100,6 @@ CssDataType = TypeVar("CssDataType", bound=Mapping[str, Any])
 
 @dataclass(frozen=True)
 class RenderInput(Generic[ArgsType, KwargsType, SlotsType]):
-    id: str
     context: Context
     args: ArgsType
     kwargs: KwargsType
@@ -109,7 +109,8 @@ class RenderInput(Generic[ArgsType, KwargsType, SlotsType]):
 
 
 @dataclass()
-class RenderStackItem(Generic[ArgsType, KwargsType, SlotsType]):
+class MetadataItem(Generic[ArgsType, KwargsType, SlotsType]):
+    render_id: str
     input: RenderInput[ArgsType, KwargsType, SlotsType]
     is_filled: Optional[SlotIsFilled]
 
@@ -216,6 +217,21 @@ class ComponentView(View, metaclass=ComponentViewMeta):
     def __init__(self, component: "Component", **kwargs: Any) -> None:
         super().__init__(**kwargs)
         self.component = component
+
+
+# Internal data that are made available within the component's template
+@dataclass
+class ComponentContext:
+    component_name: str
+    component_id: str
+    component_class: Type["Component"]
+    component_path: List[str]
+    template_name: Optional[str]
+    is_dynamic_component: bool
+    default_slot: Optional[str]
+    fills: Dict[SlotName, Slot]
+    outer_context: Optional[Context]
+    registry: ComponentRegistry
 
 
 class Component(
@@ -580,14 +596,20 @@ class Component(
         self.as_view = types.MethodType(self.__class__.as_view.__func__, self)  # type: ignore
 
         self.registered_name: Optional[str] = registered_name
-        self.outer_context: Context = outer_context or Context()
+        self.outer_context: Optional[Context] = outer_context
         self.registry = registry or registry_
-        self._render_stack: Deque[RenderStackItem[ArgsType, KwargsType, SlotsType]] = deque()
+        self._metadata_stack: Deque[MetadataItem[ArgsType, KwargsType, SlotsType]] = deque()
         # None == uninitialized, False == No types, Tuple == types
         self._types: Optional[Union[Tuple[Any, Any, Any, Any, Any, Any], Literal[False]]] = None
 
     def __init_subclass__(cls, **kwargs: Any) -> None:
         cls._class_hash = hash(inspect.getfile(cls) + cls.__name__)
+
+    @contextmanager
+    def _with_metadata(self, item: MetadataItem) -> Generator[None, None, None]:
+        self._metadata_stack.append(item)
+        yield
+        self._metadata_stack.pop()
 
     @property
     def name(self) -> str:
@@ -622,7 +644,13 @@ class Component(
         # Rendering 'ab3c4d'
         ```
         """
-        return self.input.id
+        if not len(self._metadata_stack):
+            raise RuntimeError(
+                f"{self.name}: Tried to access Component's `id` attribute " "while outside of rendering execution"
+            )
+
+        ctx = self._metadata_stack[-1]
+        return ctx.render_id
 
     @property
     def input(self) -> RenderInput[ArgsType, KwargsType, SlotsType]:
@@ -630,12 +658,12 @@ class Component(
         Input holds the data (like arg, kwargs, slots) that were passsed to
         the current execution of the `render` method.
         """
-        if not len(self._render_stack):
+        if not len(self._metadata_stack):
             raise RuntimeError(f"{self.name}: Tried to access Component input while outside of rendering execution")
 
         # NOTE: Input is managed as a stack, so if `render` is called within another `render`,
         # the propertes below will return only the inner-most state.
-        return self._render_stack[-1].input
+        return self._metadata_stack[-1].input
 
     @property
     def is_filled(self) -> SlotIsFilled:
@@ -645,13 +673,13 @@ class Component(
         This attribute is available for use only within the template as `{{ component_vars.is_filled.slot_name }}`,
         and within `on_render_before` and `on_render_after` hooks.
         """
-        if not len(self._render_stack):
+        if not len(self._metadata_stack):
             raise RuntimeError(
                 f"{self.name}: Tried to access Component's `is_filled` attribute "
                 "while outside of rendering execution"
             )
 
-        ctx = self._render_stack[-1]
+        ctx = self._metadata_stack[-1]
         if ctx.is_filled is None:
             raise RuntimeError(
                 f"{self.name}: Tried to access Component's `is_filled` attribute " "before slots were resolved"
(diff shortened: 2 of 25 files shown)


tests/test_dependencies.py
```diff
@@ -508,7 +508,7 @@ def assert_dependencies(content: str):
 
         assert_dependencies(rendered1)
         self.assertEqual(
-            rendered1.count('Variable: <strong data-djc-id-a1bc41="" data-djc-id-a1bc42="">value</strong>'),
+            rendered1.count('Variable: <strong data-djc-id-a1bc42="" data-djc-id-a1bc41="">value</strong>'),
```
JuroOravec, line 511: When using `{% provide %}`, we also generate a unique ID for the provided data, hence the changes to the IDs in the tests

tests/test_template.py
```diff
@@ -38,8 +38,8 @@ def get_context_data(self, variable=None):
                 }
 
         comp = SimpleComponent()
-        template_1 = comp._get_template(Context({}))
+        template_1 = comp._get_template(Context({}), component_id="123")
```
JuroOravec, line 41: Passed `component_id` to `_get_template()` to pass it to tracing logs.

tests/test_templatetags_component.py
```diff
@@ -752,6 +752,32 @@ def get_context_data(self, *args, attrs, my_dict):
         )
 
 
+class RecursiveComponentTests(BaseTestCase):
+    @parametrize_context_behavior(["django", "isolated"])
+    def test_recursive_component(self):
```
JuroOravec, line 757: Here's the test for recursive renders.

tests/test_templatetags_templating.py
```diff
@@ -604,38 +604,10 @@ class DashboardComponent(Component):
             </div>
         """
 
-    class ComplexChildComponent(Component):
```
JuroOravec, line 607: Moved inside `test_component_nesting_deep_slot_inside_component_fill`

tests/test_templatetags_templating.py
```diff
@@ -792,6 +792,114 @@ def test_component_nesting_deep_slot_inside_component_fill(self):
         """
         self.assertHTMLEqual(rendered, expected)
 
+    # This test is based on real-life example.
+    # It ensures that deeply nested slots in fills with same names are resolved correctly.
+    # It also ensures that the component_vars.is_filled context is correctly populated.
+    @parametrize_context_behavior(["django", "isolated"])
+    def test_component_nesting_deep_slot_inside_component_fill_2(self):
```
JuroOravec, line 799: Test based on my work project layout.

Annoyingly, there was one more issue with making the slots/fills and blocks/extends all work correctly, that I experienced on my website, but I couldn't figure out how to isolate it into a test.

src/django_components/util/misc.py
```diff
@@ -20,13 +20,6 @@ def gen_id() -> str:
     )
 
 
-def find_last_index(lst: List, predicate: Callable[[Any], bool]) -> Any:
```
JuroOravec, line 23: `find_last_index` was extra, because there's already `get_last_index`

src/django_components/util/exception.py
```diff
@@ -0,0 +1,61 @@
+from contextlib import contextmanager
+from typing import Generator, List
```
JuroOravec, line 2: This file contains the logic for modifying error messages to include the component and slot path, e.g.

```
 KeyError: "An error occured while rendering components MyPage > MyComponent > MyComponent(slot:content) > Broken
```

This was previously defined in [`Component._render()`](https://github.com/django-components/django-components/pull/936/files#diff-35c84ebceae50f7149f03bbcaf2b4f1289bc2984bf7859ca521b5dbe7960c396L931).

sampleproject/components/recursive.py
```diff
@@ -0,0 +1,31 @@
+from typing import Any, Dict
+
+from django_components import Component, register, types
+
+
+@register("recursive")
+class Recursive(Component):
```
JuroOravec, line 7: I also included the recursive component in sampleproject, to showcase what the library can do.

src/django_components/component.py
```diff
@@ -109,7 +109,8 @@ class RenderInput(Generic[ArgsType, KwargsType, SlotsType]):
 
 
 @dataclass()
-class RenderStackItem(Generic[ArgsType, KwargsType, SlotsType]):
+class MetadataItem(Generic[ArgsType, KwargsType, SlotsType]):
+    render_id: str
```
JuroOravec, line 113: I renamed `RenderStackItem` to `MetadataItem`, hopefully that gives a better idea of what's going on. Also made a tiny change to move the component id from `RenderInput` to `MetadataItem`.

Officially, component's ID is accessed as `Component.id`. But before, one could theoretically access it as `Component.input.id`.

This was not documented, and it also doesn't feel right, because ID is NOT part of the input.

src/django_components/component.py
```diff
@@ -218,6 +219,21 @@ def __init__(self, component: "Component", **kwargs: Any) -> None:
         self.component = component
 
 
+# Internal data that are made available within the component's template
+@dataclass
+class ComponentContext:
```
JuroOravec, line 224: Moved this from `slots.py`

src/django_components/component.py
```diff
@@ -580,15 +596,21 @@ def __init__(
...
-        self._render_stack: Deque[RenderStackItem[ArgsType, KwargsType, SlotsType]] = deque()
+        self._metadata_stack: Deque[MetadataItem[ArgsType, KwargsType, SlotsType]] = deque()
         # None == uninitialized, False == No types, Tuple == types
         self._types: Optional[Union[Tuple[Any, Any, Any, Any, Any, Any], Literal[False]]] = None
 
     def __init_subclass__(cls, **kwargs: Any) -> None:
         cls._class_hash = hash(inspect.getfile(cls) + cls.__name__)
 
+    @contextmanager
+    def _with_metadata(self, item: MetadataItem) -> Generator[None, None, None]:
```
JuroOravec, line 609: To allow users to access data from within component methods as `self.id` or `self.input.slots`, that data is temporarily set to the stack.

This is done this way because user could theoretically render a Component while within `get_context_data` (or other method). And so in the nested method calls, `Component.id` should point to the correct data.

The change I made here is that, previously, we entered the "component context" by pushing an entry to `self._metadata_stack`- e.g. [see here](https://github.com/django-components/django-components/pull/936/files#diff-35c84ebceae50f7149f03bbcaf2b4f1289bc2984bf7859ca521b5dbe7960c396L998).

Now, I used a context manager to clearly show when we're inside the context and when not - [see example](https://github.com/django-components/django-components/pull/936/files#diff-35c84ebceae50f7149f03bbcaf2b4f1289bc2984bf7859ca521b5dbe7960c396R1072).

src/django_components/component.py
```diff
@@ -699,9 +727,11 @@ def _get_template(self, context: Context) -> Template:
         if template_body is not None:
             # We got template string, so we convert it to Template
             if isinstance(template_body, str):
+                trace_component_msg("COMP_LOAD", component_name=self.name, component_id=component_id, slot_name=None)
                 template: Template = cached_template(
                     template_string=template_body,
-                    name=self.template_file,
+                    name=self.template_file or self.name,
```
JuroOravec, line 733: Also added name and origin to the created `Template` instance.

src/django_components/component.py
```diff
@@ -989,28 +997,83 @@ def _render_impl(
...
+            "COMP_PREP_START",
+            component_name=self.name,
+            component_id=render_id,
+            slot_name=None,
+            component_path=component_path,
+            extra=f"Received {len(args)} args, {len(kwargs)} kwargs, {len(slots)} slots",
+        )
+
+        # Register the component to provide
+        register_provide_reference(context, render_id)
```
JuroOravec, line 1043: Here we register that this component is a descendant of any provide data that it has access to, so that we can do refence counting.

src/django_components/component.py
```diff
@@ -1022,101 +1085,173 @@ def _render_impl(
...
-                #   If multiple slots have the `default` attribute set, yet have different name, then
-                #   we will raise an error.
-                default_slot=None,
-            )
+            metadata.is_filled = is_filled
 
             with context.update(
                 {
                     # Private context fields
-                    _ROOT_CTX_CONTEXT_KEY: self.outer_context,
```
JuroOravec, line 1056: I also reduce the number of keys we need to pass down the Context object. Now it's just 2:
- internal `_COMPONENT_CONTEXT_KEY` (aka `"_DJC_COMPONENT_CTX"`)
- public `"component_vars"`

src/django_components/component.py
```diff
@@ -1022,101 +1085,173 @@ def _render_impl(
...
+                # Make a "snapshot" of the context as it was at the time of the render call.
+                #
+                # Previously, we recursively called `Template.render()` as this point, but due to recursion
+                # this was limiting the number of nested components to only about 60 levels deep.
+                #
+                # Now, we make a flat copy, so that the context copy is static and doesn't change even if
+                # we leave the `with context.update` blocks.
+                #
+                # This makes it possible to render nested components with a queue, avoiding recursion limits.
+                context_snapshot = snapshot_context(context)
```
JuroOravec, line 1117: This is where we previously called `Template.render(context)`, and this is where we now make only the copy or "snapshot" of the context.

src/django_components/component.py
```diff
@@ -1022,101 +1085,173 @@ def _render_impl(
...
+            context=context_snapshot,
+            metadata=metadata,
+            component_path=component_path,
+            css_input_hash=css_input_hash,
+            js_input_hash=js_input_hash,
+            css_scope_id=None,  # TODO - Implement CSS scoping
+        )
+
+        # Remove component from caches
+        def on_component_rendered(component_id: str) -> None:
```
JuroOravec, line 1138: These callbacks are passed to `component_post_render` so that this file remains focused on the business logic, while `perfutil/component.py` handles the underlying algo for rendering nested components.

src/django_components/component.py
```diff
@@ -1128,26 +1263,56 @@ def _normalize_slot_fills(
 
         # NOTE: `gen_escaped_content_func` is defined as a separate function, instead of being inlined within
         #       the forloop, because the value the forloop variable points to changes with each loop iteration.
-        def gen_escaped_content_func(content: SlotFunc) -> Slot:
-            def content_fn(ctx: Context, slot_data: Dict, slot_ref: SlotRef) -> SlotResult:
-                rendered = content(ctx, slot_data, slot_ref)
-                return conditional_escape(rendered) if escape_content else rendered
+        def gen_escaped_content_func(content: SlotFunc, slot_name: str) -> Slot:
```
JuroOravec, line 1266: This is part of the `_normalize_slot_fills` method, which ensures that the outputs of slot fills are escaped if configured so.

I updated this function to set the component and slot names onto the `Slot` objects. This is what makes it possible to print a user-friend representation when  printing out the slots, e.g.

Screenshot 2025-01-28 at 09 52 00

src/django_components/context.py
```diff
@@ -40,76 +35,14 @@ def make_isolated_context_copy(context: Context) -> Context:
...
     # sets a `my_val` key.
     # For this reason, instead of copying individual keys, we copy the whole stack layer
     # set by ForNode.
     if "forloop" in from_context:
-        forloop_dict_index = find_last_index(from_context.dicts, lambda d: "forloop" in d)
+        forloop_dict_index = get_last_index(from_context.dicts, lambda d: "forloop" in d) or -1
         to_context.update(from_context.dicts[forloop_dict_index])
-
-
-def get_injected_context_var(
```
JuroOravec, line 56: Moved the provide-related functions to `provide.py`.

src/django_components/perfutil/component.py
```diff
@@ -69,12 +100,15 @@
 def component_post_render(
     renderer: ComponentRenderer,
     render_id: str,
+    component_name: str,
     parent_id: Optional[str],
+    on_component_rendered: Callable[[str], None],
+    on_html_rendered: Callable[[str], str],
```
JuroOravec, line 106: This is the flow that smartly renders components top-down. This is where there's the queue that allows to render the components of arbitrary depth, avoiding recursion.

In this PR it was updated to also capture the component path, so show that path to the component when an error occurs during the actual rendering of the component.

src/django_components/perfutil/provide.py
```diff
@@ -0,0 +1,155 @@
...
+
+# Keep track of how many components are referencing each provided data.
+provide_references: Dict[str, Set[str]] = {}
+
+# Keep track of all the listeners that are referencing any provided data.
+all_reference_ids: Set[str] = set()
+
+
+@contextmanager
+def managed_provide_cache(provide_id: str) -> Generator[None, None, None]:
```
JuroOravec, line 90: This file is where we do the reference counting for provided data.

src/django_components/provide.py
```diff
@@ -87,8 +90,79 @@ def render(self, context: Context, name: str, **kwargs: Any) -> SafeString:
...
 
-            output = self.nodelist.render(context)
+            # `managed_provide_cache` will remove the cache entry at the end if no components reference it.
+            with managed_provide_cache(provide_id):
+                output = self.nodelist.render(context)
 
         return output
+
+
+def get_injected_context_var(
```
JuroOravec, line 102: `get_injected_context_var`  and `set_provided_context_var` were moved here from `context.py`.

src/django_components/util/context.py
```diff
@@ -0,0 +1,114 @@
...
+from django.template.loader_tags import BlockContext
+
+
+class CopiedDict(dict):
+    """Dict subclass to identify dictionaries that have been copied with `snapshot_context`"""
+
+    pass
+
+
+def snapshot_context(context: Context) -> Context:
```
JuroOravec, line 14: This is the implementation of logic to make a Context "snapshot".

Basically it copies the Context object, and all the dictionaries in it. But to avoid too much copying, the dictionaries are copied only once. And to detect, if a dict was already copied or not, they are instantiated as `CopiedDict`, which is a subclass of `dict`, and so behaves the same way.

I copy not the list at `Context.dicts`, but also individual dictionaries out of precaution, because I'm not 100% sure e.g. how nested forloops behave - whether they make copies of the parent loop's data, or mutate the existing object. And in case of the latter, we want to copy the dicts.

src/django_components/slots.py
```diff
@@ -137,16 +150,6 @@ def __missing__(self, key: Any) -> bool:
         return False
 
 
-@dataclass
-class ComponentSlotContext:
```
JuroOravec, line 141: This was moved to component.py

src/django_components/slots.py
```diff
@@ -335,23 +351,88 @@ def render(self, context: Context, name: str, **kwargs: Any) -> SafeString:
...
         # IF SUCH FILL EXISTS. Otherwise, we use the slot's name.
-        if is_default and DEFAULT_SLOT_KEY in component_ctx.fills:
+        if is_default and DEFAULT_SLOT_KEY in slot_fills:
             fill_name = DEFAULT_SLOT_KEY
         else:
             fill_name = slot_name
 
-        if fill_name in component_ctx.fills:
-            slot_fill_fn = component_ctx.fills[fill_name]
+        # NOTE: TBH not sure why this happens. But there's an edge case when:
```
JuroOravec, line 369: This section right here is what took me so long to debug. And after all this time, I'm still not sure why things happen this way. But at least things are working.

src/django_components/slots.py
```diff
@@ -432,36 +518,57 @@ def render(self, context: Context, name: str, **kwargs: Any) -> SafeString:
...
+            else:
+                # Otherwise we simply re-use the last layer, so that following logic uses `with` in either case
+                render_ctx_layer = used_ctx.render_context.dicts[-1]
+
             with used_ctx.render_context.push(render_ctx_layer):
-                # Render slot as a function
-                # NOTE: While `{% fill %}` tag has to opt in for the `default` and `data` variables,
-                #       the render function ALWAYS receives them.
-                output = slot_fill.slot(used_ctx, kwargs, slot_ref)
+                with add_slot_to_error_message(component_name, slot_name):
```
JuroOravec, line 534: By wrapping this in `add_slot_to_error_message`, we get to see the slot path in the error message too.

JuroOravec:
@EmilStenstrom This is now ready for review. Sorry it's again big and complex, changing the logic around the slot resolution isn't easy. No rush.

src/django_components/util/context.py
JuroOravec: I tested it just now, and we do indeed need to copy the dicts.

EmilStenstrom (approved):
This is way to large and too complex for me to get deep enough into in reasonable time. I've skimmed it, I think it looks good. Let's do it :)

JuroOravec:
@EmilStenstrom Appreciate that! 🚀

PR #937, opened 2025-01-31
docs: update development instructions to include directory change and…

oliverhaas:
… local installation

Very minor change, but honestly I was confused for 2 minutes there why django-components was not found.

I hope I'm not missing something obvious :confused:

### DIFF
--- a/docs/overview/development.md
+++ b/docs/overview/development.md
@@ -10,12 +10,19 @@ This makes a copy of the repository in your own name. Now you can clone this rep
 
 ```sh
 git clone https://github.com/<your GitHub username>/django-components.git
+cd django-components
 ```
 
 To quickly run the tests install the local dependencies by running:
 
 ```sh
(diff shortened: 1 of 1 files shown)


JuroOravec:
Thanks for the fix!

docs/overview/development.md
```diff
@@ -18,6 +19,12 @@ To quickly run the tests install the local dependencies by running:
 pip install -r requirements-dev.txt
 ```
 
+You also have to install this local django-components version, initially and after every change you want to test:
```
dalito, line 22: You can install in editable mode with `pip install -e .` or not? That should save you from running the install command again and again.

JuroOravec: Yup! I do that all the time

JuroOravec: I added the `-e` and a note on running only linters in https://github.com/django-components/django-components/pull/938

PR #941, opened 2025-02-01
refactor: fix on_render_after hook

JuroOravec:
Follow up to https://github.com/django-components/django-components/pull/936, this PR fixes the `on_render_after` hook, so that it:
1. Receives the *fully* rendered HTML. Because after #936, the hook actually receives an HTML that still contains internal placeholders for nested components (e.g. `<template djc-render-id="a1b2c3">`).
2. Also added tests to check that the Context that's passed to `on_render_before` and `on_render_after` contains the same info, which I think is important to check now, because the two hooks are being called from different places in the code.


JuroOravec:

PR #942, opened 2025-02-01
feat: allow to highlight slots and components for debugging

JuroOravec:
Another follow up from https://github.com/django-components/django-components/pull/936

For one, this PR documents various tools / tips users can use to debug page built using django-components.

Secondly, I added option to visualize the rendered components / slots for debugging.

I added 2 boolean settings `debug_highlight_components` and `debug_highlight_slots`.

When `debug_highlight_components` is set to `True`, the rendered components are wrapped in a blue border, and their name and ID is shown:

Screenshot 2025-02-01 at 14 26 46

Similarly, when `debug_highlight_slot` is set to `True`, the *slots* are wrapped in a red border, and component and slot name shown:

Screenshot 2025-02-01 at 14 17 45

I used the slot highlight when I was trying to figure out issues with https://github.com/django-components/django-components/pull/936.

When one turns ON both slots and components, it's a mess. So I was thinking of reducing it to a single setting name, which would be either `False` (OFF), `"component"`, or `"slot"`. But kept it like this for simplicity.


docs/guides/setup/logging_and_debugging.md
JuroOravec, line 1: This was merged into the troubleshooting.md file

src/django_components/util/component_highlight.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Wrap HTML (string) in a div with a border and a highlight color.
+
+    This is part of the component / slot highlighting feature. User can toggle on
+    to see the component / slot boundaries.
+    """
+    color = COLORS[type]
+
+    output = f"""
+        <div style="
```
JuroOravec, line 25: There could be potentially more clever ways to achieve the highlight. The issue with wrapping the rendered content in `<div>` as I did below is that it breaks any layouting styling like flex, `position: absolute`, etc.

But since this is only for debugging, I think this is fine for now.

src/django_components/util/component_highlight.py
```diff
@@ -0,0 +1,44 @@
...
+        ">
+            <p style="
+                background: white;
+                color: {color.text_color};
+                font-weight: 600;
+                border-radius: 6px;
+                padding: 2px;
+            ">{name}</p>
+            {output}
+        </div>
```
EmilStenstrom, line 41: I would suggest this is simplified (remove as many properties as possible to mess less with people's designs when enabling debugging. It will break people's designs, but this is more useful the less it breaks designs.

Maybe something like this?
```
<style>#id123::before { content: "{name}: "; font-weight: bold}</style>
<div id="id123" style="outline: 1px solid {color.border_color}">{output}</div>
```

tests/test_component_highlight.py
```diff
@@ -0,0 +1,52 @@
...
+    def test_highlight_styling(self):
+        """Test that the output contains expected styling elements"""
+        result = apply_component_highlight("component", "<p>Test</p>", "Test")
+
+        # Check for expected styling elements
+        self.assertIn("border-radius: 12px", result)
+        self.assertIn("padding: 4px", result)
+        self.assertIn("margin: 4px", result)
+        self.assertIn("transition: all 0.2s ease", result)
+        self.assertIn("box-shadow:", result)
```
EmilStenstrom, line 47: No need to check for all of these.

EmilStenstrom (approved):
I like this, good addition. One (optional) comment about simplifying the design of the debug view.

src/django_components/util/component_highlight.py
JuroOravec: Great idea with the `::before`! 👀 

And yeah, TBH I started with just a 1px border. The extra fluff was to give it a bit of a wow effect. After all, the extra padding, shadows, and round border DOES make it look better. This is how it looks just with border:

Screenshot 2025-02-02 at 09 53 56
Screenshot 2025-02-02 at 09 49 45

But yes, as for day-to-day development, it's better to avoid breaking the page layout.

I already got the pretty screenshots, which I think we can use for documentation / marketing (e.g. when showing what django-components can do). But the actual implementation would be bare-bone.

And just a small detail, but I'll use `border` instead of `outline`, because with `outline` the lines overlap, so it's not clear what ends and starts where.

tests/test_component_highlight.py
JuroOravec: Agree, removed!

src/django_components/util/component_highlight.py
EmilStenstrom: Still looks good I think! :)

PR #943, opened 2025-02-01
Documentation typo

mlissner:
One char PR. I'll add more if I find more, but so far docs have been great.


EmilStenstrom:
Thanks @mlissner! :)

Issue #935, opened 2025-01-30
[perf] Validation of args/kwargs seems to have a fairly large performance penalty (~35%) [v0.125]


oliverhaas:
The new changes for validation of `args` and `kwargs` order in [v0.125](https://github.com/django-components/django-components/releases/tag/0.125) slows down rendering by about ~35% in some of my test cases (lots of very simple components).

I assume the reason for this feature is to get better feedback on what exact syntax error the user made, but the problem is that the whole use of `eval()` is fairly slow.

Wouldn't it work to just check for `SyntaxError: positional argument follows keyword argument` manually and raise that one manually, and `TypeError: foo() got multiple values for argument 'whatever'` could just be raised by calling the render function with `*args` and `**kwargs`?

If I'm not missing anything or at least not much, then I'd be happy to provide a small PR including some benchmark for both v0.124 and my PR.

JuroOravec:
We can do so manually. I went with eval because I wanted the input to `{% component %}` template tags to behave similar to python functions, but I didn't want to bother with trying to replicate each error case separately. But I didn't realize eval would be this slow.

I used eval especially so that the generated function checks for multiple same keywords. Because there, one cannot simply do

```py
func(*args, **kwargs)
```

because any duplicate keys would already be overwritten when squashed into `kwargs`. So the generated function that's being evaled looks more like this:

```py
kwarg1 = {"key": 1}
kwarg2 = {"key": 2}
func(*args, **kwarg1, **kwarg2)
```

And it's that dynamic number of spreads that required `eval`.

If we do so manually, then the cases to cover are at least:
1. too few arguments (whether positional or kwargs)
2. too many argument (whether positional or kwargs)
3. multiple keyword arguments
4. value given by both positional and kwarg (That's the `got multiple values for argument 'whatever'` error)
5. positional arguments after keyword arguments.

After that, we can pass the data around as `*args` and `**kwargs`.

JuroOravec:
Btw, that ~35% you're seeing, is that specifically due to validation, or generally `eval()` as a black box? When I did some profiling I also saw `eval()` high up the chart, (71% in my case).

`eval()` is also used when calling [`BaseNode.render()` on line 197](https://github.com/django-components/django-components/blob/3d187d7abad80f8464bd70dfba847d27a8e173db/src/django_components/node.py#L197), for the same purpose - to pass the args and kwargs in the same order as they were defined on the `{% component %}` tag.

[...]

PR #944, opened 2025-02-02
refactor: remove use of eval for node validation

JuroOravec:
Part of #935.

As described [here](https://github.com/django-components/django-components/issues/935#issuecomment-2629165443), I tried to replace `eval()` with manual input validation, to see if it improves performance.

I does not. But at least we got rid of the eval, so that's still a win in my book.

No API changes.


JuroOravec:

PR #945, opened 2025-02-02
refactor: Do tag input validation with __code__ variables if available

JuroOravec:
Part of https://github.com/django-components/django-components/issues/935

This adds an alternative implementation of validating tag inputs, which relies on the `__code__` internals to get the function's  signature (see https://github.com/django-components/django-components/issues/935#issuecomment-2629317588).

This alternative is about 3x faster (46ms vs 16ms on my test page of 2 MB HTML and 360 components). If `__code__` is NOT available, we still fall back to the signature-based implementation, so that django-components can still be potentially used also on more exotic Python flavours.

### DIFF
--- a/src/django_components/util/template_tag.py
+++ b/src/django_components/util/template_tag.py
@@ -18,30 +18,28 @@ from django_components.util.tag_parser import TagAttr, parse_tag
 # For details see https://github.com/django-components/django-components/pull/902#discussion_r1913611633
 # and following comments
 def validate_params(
+    func: Callable[..., Any],
+    validation_signature: inspect.Signature,
     tag: str,
-    signature: inspect.Signature,
     params: List["TagParam"],
     extra_kwargs: Optional[Dict[str, Any]] = None,
 ) -> Tuple[Tuple[Any, ...], Dict[str, Any]]:
     """
     Validates a list of TagParam objects against this tag's function signature.
 
-    The validation preserves the order of parameters as they appeared in the template.
+    Raises `TypeError` if the parameters don't match tfuncsignature.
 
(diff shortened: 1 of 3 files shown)


src/django_components/util/template_tag.py
```diff
@@ -18,30 +18,28 @@
...
-
-    # Create a function that uses the given signature
-    def validator(*args: Any, **kwargs: Any) -> Tuple[Tuple[Any, ...], Dict[str, Any]]:
-        return args, kwargs
-
-    validator.__signature__ = signature  # type: ignore[attr-defined]
-
-    # Call the validator with our args and kwargs in the same order as they appeared
-    # in the template, to let the Python interpreter validate on repeated kwargs.
+    supports_code_objects = func is not None and hasattr(func, "__code__") and hasattr(func.__code__, "co_varnames")
```
JuroOravec, line 36: Here's where we decide which implementation to go with

tests/test_node.py
```diff
@@ -676,3 +678,383 @@ def render(node: BaseNode, context: Context, **kwargs) -> str:
...
+            # Return None for __code__ to force signature-based validation
+            if name == "__code__":
+                return None
+            # For all other attributes, delegate to the wrapped function
+            return getattr(self.__wrapped__, name)
+
+    return SignatureOnlyFunction(fn)
+
+
+class SignatureBasedValidationTests(BaseTestCase):
```
JuroOravec, line 707: Added tests for the signature-based validation, as the `__code__` base one is now the default - or at least I think it's fairly safe to assume that any one of us who will run tests will be using CPython / Pypy

PR #956, opened 2025-02-07
refactor: fix bug with complex slots and "django" mode + add docs on debugging with AI agents

JuroOravec:
Fix bug that I encountered when preparing the benchmarking test case (mentioned here https://github.com/django-components/django-components/issues/14#issuecomment-2642331396).

I didn't include a test to check for this bug, because, frankly, it's not clear to me under which conditions it occurs. I just know that it happens in my work project. So I'm hoping to implicitly test of it once I add the benchmark tests - as for those, I would also test the final output, and that assertion will test for this behaviour.

Also, I added a few more trace-level logs. 

As I faced another issue with this resolution of slots, I was annoyed, because I already spend ungodly amount of time on that. So I thought - what if we captured enough details in the trace logs so that AI could fix those issues? In other words, it's as if trying to fix the bug by just relying on the trace logs.

This got me close, tho there were two logical steps for which the AI was already out of its depth. But still, it got me fairly close fairly fast.

So I decided to document my approach. And I added a section to the Troubleshooting page about using AI agents for debugging django-components. And while I was at it, I update also README with a quick summary of the debugging features. 

Screenshot 2025-02-07 at 14 16 19

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,15 @@ Potential benefits:
 
 Django-components can be particularly useful for larger Django projects that require a more structured approach to UI development, without necessitating a shift to a separate frontend framework.
 
+### Debugging features
+
+- **Visual component inspection**: Highlight components and slots directly in your browser.
+- **Detailed tracing logs to supply AI-agents with context**: The logs include component and slot names and IDs, and their position in the tree.
+
+<div style="text-align: center;">
+<img src="https://github.com/django-components/django-components/blob/master/docs/images/debug-highlight-slots.png?raw=true" alt="Component debugging visualization showing slot highlighting" width="500" style="margin: auto;">
+</div>
+
 ## Quickstart
 
 django-components lets you create reusable blocks of code needed to generate the front end code you need for a modern app.
--- a/README.md
+++ b/README.md
@@ -16,6 +16,76 @@ in modern frontend frameworks like Vue or React.
 
 With `django-components` you can support Django projects small and large without leaving the Django ecosystem.
 
+## Quickstart
+
+A component in django-components can be as simple as a Django template and Python code to declare the component:
+
+```django
+{# components/calendar/calendar.html #}
+<div class="calendar">
+  Today's date is <span>{{ date }}</span>
+</div>
+```
+
+```py
+# components/calendar/calendar.html
+from django_components import Component
+
+class Calendar(Component):
+    template_file = "calendar.html"
+```
+
+Or a combination of Django template, Python, CSS, and Javascript:
+
+```django
+{# components/calendar/calendar.html #}
+<div class="calendar">
+  Today's date is <span>{{ date }}</span>
+</div>
+```
+
+```css
+/* components/calendar/calendar.css */
+.calendar {
+  width: 200px;
+  background: pink;
+}
+```
+
+```js
+/* components/calendar/calendar.js */
+document.querySelector(".calendar").onclick = () => {
+  alert("Clicked calendar!");
+};
+```
+
+```py
+# components/calendar/calendar.py
+from django_components import Component
+
+class Calendar(Component):
+    template_file = "calendar.html"
+    js_file = "calendar.js"
+    css_file = "calendar.css"
+
+    def get_context_data(self, date):
+        return {"date": date}
+```
+
+Use the component like this:
+
+```django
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -101,6 +101,15 @@ Potential benefits:
...
 
+### Debugging features
+
+- **Visual component inspection**: Highlight components and slots directly in your browser.
+- **Detailed tracing logs to supply AI-agents with context**: The logs include component and slot names and IDs, and their position in the tree.
+
+<div style="text-align: center;">
+<img src="https://github.com/django-components/django-components/blob/master/docs/images/debug-highlight-slots.png?raw=true" alt="Component debugging visualization showing slot highlighting" width="500" style="margin: auto;">
+</div>
+
```
EmilStenstrom, line 112: This is a great feature, but I don't think it warrants "front-page" like this.

docs/guides/other/troubleshooting.md
```diff
@@ -143,3 +143,82 @@ might print:
     'left_panel': <Slot component_name='layout' slot_name='left_panel'>,
 }
 ```
+
+## Agentic debugging
```
EmilStenstrom, line 147: This is awesome. I have this same workflow when working on another hobby-project, and it works great!

docs/overview/welcome.md
```diff
@@ -98,6 +98,15 @@ Potential benefits:
...
 
+### Debugging features
+
+- **Visual component inspection**: Highlight components and slots directly in your browser.
+- **Detailed tracing logs to supply AI-agents with context**: The logs include component and slot names and IDs, and their position in the tree.
+
+<div style="text-align: center;">
+<img src="https://github.com/django-components/django-components/blob/master/docs/images/debug-highlight-slots.png?raw=true" alt="Component debugging visualization showing slot highlighting" width="500" style="margin: auto;">
+</div>
+
```
EmilStenstrom, line 109: Not front-page stuff imho.

EmilStenstrom (approved):
Two quick comments. Can't say I understand the actual fix, but I trust you :)

docs/overview/welcome.md
JuroOravec: I'm thinking that on the README / front page we could have a rundown of django-components features like this. Basically like a typical SaaS landing page, where you show some nice screenshots and in short explain why your product is the best choice 😃

Other highlights would be (but not in this order):
- On performance (once we know how DJC compares with others) 
- on template tags - the stuff within the `{% ... %}` like the literal lists and dicts, spread, etc. 
- on slots and composition
- defining components (a code example with template / js / css inlined)
- on js / css variables (again as code example)
- that it's possible to publish and use 3rd party components

This would then replace [the current "features" and "potential benefits" sections](https://github.com/django-components/django-components/pull/956/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R87) in the README, because IMO currently it's too vague and it reads like it was AI generated.

EmilStenstrom: I like the features section, i think it makes the main question people might have well: "Why should I use a component library?". Looking at the readme again, I see that we again have `quickstart -> features -> quickstart`. I think we should go back to `features -> quickstart -> read all the details in the docs`

JuroOravec: I updated the README.  And also fixed the flow to `features -> quickstart -> read the docs`

Some screenshots:

Screenshot 2025-02-10 at 20 23 21

Screenshot 2025-02-10 at 20 23 30

Screenshot 2025-02-10 at 20 23 45

Screenshot 2025-02-10 at 20 23 53

Screenshot 2025-02-10 at 20 24 03

Screenshot 2025-02-10 at 20 24 11

Screenshot 2025-02-10 at 20 24 17

Screenshot 2025-02-10 at 20 24 25

Screenshot 2025-02-10 at 20 24 31

EmilStenstrom: @JuroOravec I didn't mean to expand the features like that, but reading through this it is undoubtedly **better**. Since the features list is so long now, I think we should switch order again to "short intro -> quickstart -> detailed features -> Read documentation".

This is awesome!

JuroOravec: Sounds good, updated!

PR #959, opened 2025-02-09
Update indenting typo in slots_and_blocks.md

bfrangi:
Fix extra indent that was making the whole text of the list item appear as a code block.

image

### DIFF
--- a/docs/guides/devguides/slots_and_blocks.md
+++ b/docs/guides/devguides/slots_and_blocks.md
@@ -5,44 +5,44 @@
     template was inlined. So if the "included" template contains `slot` tags, then the component
     uses those slots.
 
-        So if you have a template `abc.html`:
-        ```django
-        <div>
-          hello
-          {% slot "body" %}{% endslot %}
-        </div>
-        ```
+    So if you have a template `abc.html`:
+    ```django
+    <div>
+      hello
+      {% slot "body" %}{% endslot %}
+    </div>
+    ```
 
-        And components that make use of `abc.html` via `include` or `extends`:
-        ```py
-        from django_components import Component, register
+    And components that make use of `abc.html` via `include` or `extends`:
+    ```py
+    from django_components import Component, register
 
-        @register("my_comp_extends")
-        class MyCompWithExtends(Component):
-            template = """{% extends "abc.html" %}"""
+    @register("my_comp_extends")
+    class MyCompWithExtends(Component):
+        template = """{% extends "abc.html" %}"""
 
-        @register("my_comp_include")
-        class MyCompWithInclude(Component):
-            template = """{% include "abc.html" %}"""
-        ```
+    @register("my_comp_include")
+    class MyCompWithInclude(Component):
+        template = """{% include "abc.html" %}"""
+    ```
 
-        Then you can set slot fill for the slot imported via `include/extends`:
+    Then you can set slot fill for the slot imported via `include/extends`:
 
-        ```django
-        {% component "my_comp_extends" %}
-        	{% fill "body" %}
-        		123
-        	{% endfill %}
-        {% endcomponent %}
-        ```
+    ```django
+    {% component "my_comp_extends" %}
+        {% fill "body" %}
+            123
+        {% endfill %}
+    {% endcomponent %}
+    ```
 
-        And it will render:
-        ```html
-        <div>
-          hello
-          123
-        </div>
-        ```
+    And it will render:
+    ```html
+    <div>
+      hello
+      123
+    </div>
+    ```
 
 2.  Slot and block
 


JuroOravec:
@bfrangi thanks for the fix! 

Btw, did you find that page on slot and blocks useful? I wrote those "devguides" mainly to share the intuition on how things are implemented. But didn't get any feedback on it yet, so I don't know how many people are reading those.

bfrangi:
Hi @JuroOravec, I think those pages make a lot of sense. Blocks and slots are the typical thing that can easily be confusing and it helps to have those clear examples.

I've also found the 'JS and CSS rendering' guide very useful as I was fighting with using `django-components` inside html emails and converting all css dependencies to inline styles (using `css_inline`). In fact, that's another feature that would be really great to have in `django-components`, especially for people using it for html emails... not sure if you have considered this before (just curious, no feature request intended for now)?

JuroOravec:
> I've also found the 'JS and CSS rendering' guide very useful as I was fighting with using django-components inside html emails and converting all css dependencies to inline styles (using [css_inline](https://pypi.org/project/css-inline/)). In fact, that's another feature that would be really great to have in django-components, especially for people using it for html emails... not sure if you have considered this before (just curious, no feature request intended for now)?

@bfrangi that's an interesting idea 👀 Haven't heard before of django-components being used for generating HTML for emails, but agree that it should be possible, and it's true that with how JS / CSS is handled, that's not easy right now.

How do the `<style>` and `<link>` tags behave with emails?

1. Can you use `<link rel="spreadsheet">` that points to some URL? E.g.

   ```html
    <link rel="stylesheet" href="https://mysite.com/path/to/email/stylesheet.css">
   ```

2. Is it possible to include the CSS with the `<style>` tag? E.g.

   ```html
    <style>
    .p {
      color: red;
    }
    </style>
   ```

3. Or is it that the CSS is required to be inlined for emails, and that's why you are using `css-inline`? E.g.

   ```html
   <div style="width: 100px"> Hello </div>
   ```

---

I think there's two parts to what you described:

1. It'd make sense for django-components to have a "render mode" where the CSS and JS are explicitly included in the rendered HTML as `<script>` and `<style>` tags.

   Right now, django-components conceptually recognizes only 2 use cases in this regard (what I called "render modes" above):
   1. `"document"` - Rendering a full HTML document that is expected to be opened in a browser
   2. `"fragment"` - Rendering an HTML fragment that is expected to be inserted into the full HTML document from case 1).
 
   Truth is that CSS is included as a `<style>` tag in both modes, to avoid the [flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) (FOUC). However, if FOUC wasn't an issue, then the `<style>` would not be included. And there's some more nuances like:
   - JS is not included directly in the HTML the "fragment" mode, because we manage loading of `<script>` tags via custom JS logic
   - The "document" mode also addes extra JS script that defines our custom logic.

   So I think it makes sense to add a "render mode" for the use cases when the rendered HTML is not to expected be used in a browser. So:
   - Any JS / CSS rendered by the components would be directly included in the HTML as `<script>` and `<style>` tags,
   - And we would not add any extra JS logic.

   To select this render mode, you'd set `type="inline"` in `Component.render()`, e.g.:

   ```python
   class MyEmailHtml(Component):
       template = """
          <p>Hello</p>
       """

      css = """
        p {
          color: red;
        }
      """

   rendered = MyEmail.render(
       args=[ ... ],
       kwargs={ ... },
       ...
       type="inline",
   )
   ```

   Which would render:

   ```html
   <style>
    p {
      color: red;
    }
   </style>
   <p>Hello</p>
   ```

2. Now, I haven't looked it up yet, but if the situation is that for the email you actually MUST inline the styling like this:

   ```html
   <div style="width: 100px"> Hello </div>
   ```

   Then my intuition tells me that it could be implemented as a [plugin](https://github.com/django-components/django-components/issues/829) that would post-process the rendered components to insert the styling directly into the HTML as the `style="..."` attribute.

   I think some questions are:
    1. How would we insert the styling as `style` attribute?
       - A: Using `css_inline` as @bfrangi suggested - The package seems to focus solely on this issue, and it's implemented in Rust, so that sounds good to me.
    2. If I defined styling for `<p>` tag in parent component, should that affect `<p>` tags in child components?
       - A: I think yes. Unless the parent component had CSS scoping turned ON, in which case the the styling would be applied only to parent's own HTML. So it would practically work the same way when the CSS is not inlined as a `style` attribute.
    3. Should this inline the CSS defined as `Component.css` (or `Component.css_file`)?
       - A: Yes
    4. Should this inline CSS defined as `Component.Media.css`?
       - A: No. IMO the JS and CSS files defined on `Component.Media` should be treated as a black box - we know only the URL, but not the contents.
    5.  How would this work with CSS variables?
       - A: The plugin would have to apply the component's CSS variables as `style` attribute on all root HTML elements:
          ```html
            <div style="--my-var: 20px">
               Hi
            </div>
            <div style="--my-var: 20px">
               Hello
            </div>
          ```
 
---

@EmilStenstrom What do you think? I think it makes sense to make a separate ticket for this as feature request.

bfrangi:
Hi @JuroOravec, thanks for the quick and thorough reply. Let me give a few more details on the problem:

1. Some email clients accept `<style>` tags, but a lot don't, so most html email designers end up using inline styles for  wider compatibility. Of course, they don't want to go insane actually writing the inline styles, so they use libraries like `css_inline` to enable them to develop using stylesheets, converting them to inline automatically.
2. These libraries are quite good, and some (like `css_inline`) can even convert remote style sheets (such as `<link rel="stylesheet" href="https://mysite.com/path/to/email/stylesheet.css">`) to inline styles.
3. However, for these libraries to work, we need to have the stylesheets explicitly referenced in the html code that is given to them. Since `django-components` renders a placeholder and inserts the styles later, it is not possible to convert component css to inline styles.

**Current workaround**

The following example illustrates my current workaround in order to be able to use `django-components` in emails:

```py
from django_components import component

@component.register("button")
class Button(component.Component):
    template = """
    {% if css %}<style>{{css}}</style>{% endif %}
    <a href="{{url}}"><button>{{text}}</button></a>
    """
    css = """
    button {
      color: red;
    }
    """

    def get_context_data(self, text, url, inline=False):
        css = self.css if inline else None
        if inline:
            self.css = None
        return {
            "text": text,
            "url": url,
            "css": css,
        }
```

This is far from ideal, since the `<style>` tag is inserted once for each instance of the component, which is redundant. Luckily, libraries like `css_inline` enable you to remove all `<style>` tags and all references to stylesheets after they have inlined all the css.


**Comments on the solution proposed by @JuroOravec**

I think the suggestion to add a render mode that inserts the css directly (server side) instead of using the placeholder and inserting it later would definitely be a good option. That would already simplify things a lot for creating compatible html emails.

Adding the plugin for inlining the styles would be the cherry on top!

JuroOravec:
I've created https://github.com/django-components/django-components/issues/971 to track these ideas.

JuroOravec:
Hi @bfrangi, I'm coming around to implementing your idea. But re-reading it, it got me thinking:

> I think the suggestion to add a render mode that inserts the css directly (server side) instead of using the placeholder and inserting it later would definitely be a good option. That would already simplify things a lot for creating compatible html emails.

In your case, how is the overall template rendered?

- Is it rendered with Django's [`render()`](https://docs.djangoproject.com/en/5.2/topics/http/shortcuts/#render) helper?

   ```py
   from django.shortcuts import render

   html = render(request, "my_email.html", {"some": "args"})
   ```

- Or is it rendered with our `Component.render()`?

   ```py
   from components.my_email import MyEmail

   # MyEmail is a Component subclass
   html = MyEmail.render(kwargs={"some": "args"})
   ```

Because in the second case, all the dependencies should be inserted into the placeholders.

So if you're using Django's `render()` (or `Template.render()`) to render the outermost template, then if you convert that to  a component that should fix it.

---

Nevertheless, I think that for your case there's still a need to have something like I mentioned before with `type="inline"`:

```py
class MyEmailHtml(Component):
    template = """
       <p>Hello</p>
    """

   css = """
     p {
       color: red;
     }
   """

rendered = MyEmail.render(
    args=[ ... ],
    kwargs={ ... },
    ...
    type="inline",
)
```

But what this mode would do is that it would insert all JS and CSS, but WITHOUT injecting django-components' client-side script.

bfrangi:
Hi @JuroOravec, thanks for picking this up!

The overall template is rendered by django's `Template.render` function. I hadn't thought about making the email itself a component (that's what I understood from the second option you mentioned), although I guess it could be a possibility. The point is that I would need to be able to insert other components into the email, such as buttons.

---

Here is how I am currently rendering emails with inline styles:

* First of all I have a helper with methods to read and render templates:
    ```python
    from pathlib import Path
    from typing import TYPE_CHECKING

    from css_inline import inline
    from django import template as django_template
    from django.template import Context, RequestContext, Template

    if TYPE_CHECKING:
        from pathlib import PosixPath
        from typing import Optional

        from django.http import HttpRequest


    def complete_domain(request: 'HttpRequest') -> str:
        """Return the complete domain of the request."""
        dom = 'http://' if request.META['HTTP_HOST'] in [
            'localhost:8000',
            '127.0.0.1:8000'
        ] else 'https://'
        dom += request.META['HTTP_HOST']
        return dom


    def get_all_template_paths() -> 'list[PosixPath]':
        """Return all template paths in the project."""
        dirs = []
        for engine in django_template.loader.engines.all():
            dirs.extend(x for x in engine.template_dirs
                        if 'site-packages' not in str(x))
        files = []
        for dir in dirs:
            files.extend(x for x in Path(dir).glob('**/*.html') if x)
        return files


    def get_all_template_names(paths: 'Optional[list[str]]' = None) -> 'list[str]':
        """Return all template names in the project or in the given paths."""
        if paths is None:
            paths = get_all_template_paths()
        return [str(file).split('templates')[-1].strip('/') for file in paths]


    def render_html_template_string(string: str, context: 'Optional[dict]' = None,
                                    request: 'Optional[HttpRequest]' = None,
                                    inline_css: bool = False) -> str:
        """Render a template string with the given context and request."""
        context = {} if context is None else context
        context = RequestContext(request, context) if request else Context(context)
        template = Template(string)
        rendered = template.render(context)
        base_url = complete_domain(request) if request else 'http://localhost:8000'
        if inline_css:
            return inline(rendered, base_url=base_url)
        return rendered


    def render_html_template_file(template: str, context: 'Optional[dict]' = None,
                                request: 'Optional[HttpRequest]' = None,
                                inline_css: bool = False) -> str:
        """Render a template file with the given context and request."""
        template_paths = get_all_template_paths()
        template_names = get_all_template_names(template_paths)
        if template not in template_names:
            raise ValueError(f"Template {template} not found.")
        template_file = template_paths[template_names.index(template)]
        with open(template_file, 'r') as file:
            string = file.read()
        return render_html_template_string(string, context, request, inline_css)

    ```

* Then I have an email template `example-email.html` with a component, like this one:

    ```django
    {% component "example-button" text='Button' url='#' inline=True / %}
    ```
* And the component `example-button.py`:

    ```python
    from django_components import component

    @component.register("example-button")
    class ExampleButton(component.Component):
        template = """
        {% if css %}<style>{{css|safe}}</style>{% endif %}
        <a href="{{url}}"><button>{{text}}</button></a>
        """
        css = """
        button {
            color: red;
        }
        """

        def get_context_data(self, text, url, inline=False):
            return {
                "text": text,
                "url": url,
                "css": self.css if inline else None,
            }
    ```
* Then, opening a django shell `python manage.py shell`:
    ```python
    >>> from wherever import render_html_template_file
    >>> render_html_template_file('example-email.html')
    '<!-- _RENDERED Button_27bbe7,6k0W9V,, -->\n    <style data-djc-id-6k0W9V="">\n    button {\n      color: red;\n    }\n    </style>\n    <a href="#" data-djc-id-6k0W9V=""><button>Button</button></a>\n    '
    ```

    Since the `example-email.html` passed the parameter `inline=True` to the component, the CSS is inlined. However, since I did not pass `inline_css=True` to the `render_html_template_file`, the CSS is not inlined in the final email. That is the last step necessary to inline the CSS in the final email:

    ```python
    >>> render_html_template_file('example-email.html', inline_css=True)
    '<!-- _RENDERED Button_27bbe7,gIWVlk,, --><html><head>\n    </head><body><a href="#" data-djc-id-giwvlk=""><button style="color: red;">Button</button></a>\n    </body></html>'
    ```

    As you can see, it used `css_inline`'s `inline` method to inline the CSS in the final email.

---

So, in short:
1. The `inline` argument given to the component in the template is what I am using to force the css to  be inserted explicitly. 
2. Then, the `inline_css` parameter in the `render_html_template_file` function converts the explicitly inserted css into inline css.

It would be much cleaner to do this if there was some way of telling `django-components` to insert the css directly, instead of using placeholders (via some kind of render mode perhaps?). That way, step 1 would be much simpler. As a bonus, it would be nice to have also the inlining capabilities (step 2).

bfrangi:
Emails do contain `<head`> / `<body>` tags, and it should be easy to also include placeholders (I assume you mean `{% component_js_dependencies %}` and `{% component_css_dependencies %}`), so that should not be a problem. 

However, I think it would still be very useful to be able to render standalone components as you mention in the third point. Prepending JS / CSS when there are no `<head`> / `<body>` tags or placeholders seems like a good way to do that.

JuroOravec:
Hey @bfrangi, some of these things will be available in v0.140! See https://github.com/django-components/django-components/pull/1156

Issue #968, opened 2025-02-11
Component dependencies not included when using them in a template fragment


wofl:
When a component is rendered in a html fragment (e.g. for HTMX) that is a simple django template (using `{% component %}` no component dependencies are rendered or included in the resulting response.

_page.html - served by e.g. a TemplateView:_
```html
{% load component_tags %}
<!DOCTYPE html>
<html>
<head>
    {% component_css_dependencies %}
    <script src="https://unpkg.com/htmx.org@1.9.12"></script>
</head>
<body>
{% component 'inner' variable='foo' / %}

<div id="target">OLD</div>

<button id="loader" hx-get="/fragment/frag" hx-swap="outerHTML" hx-target="#target">
    Click me!
</button>

{% component_js_dependencies %}
</body>
</html>
```

_frag_comp.py - the component that will be added through htmx_
```python
@register("frag_comp")
class FragComp(Component):
    template: types.django_html = """
        <div class="frag">
            123
            <span id="frag-text"></span>
        </div>
    """

    js = """
        document.querySelector('#frag-text').textContent = 'xxx';
    """

    css = """
        .frag {
            background: blue;
        }
    """
```

_views.py - the view that serves the htmx request_
```python
def fragment_view(request):
    template_str: types.django_html = """
               {% load component_tags %}
               {% component 'frag_comp' / %}
           """
    template = Template(template_str)
    rendered = template.render(Context({}))
    return HttpResponse(rendered)
```

Neither the `css` nor the `js` of the component will be present after the swap following a click on the button.

This constellation is IMHO pretty obvious. Not everything has to be a component itself - nonetheless it should be possible to use components inside the template fragment and any dependencies should be obviously work.

I’m unsure if this is a configuration issue on my end or an actual bug in the library. I’ve created a [test case in a pull request](https://github.com/wofl/django-components/pull/1) for further investigation. Any insights would be appreciated.

JuroOravec:
Hi @wolf, have a read on the section about [rendering JS / CSS](https://django-components.github.io/django-components/latest/concepts/advanced/rendering_js_css/#setting-up-the-middleware).

Currently, django-components needs to internally post-process the rendered HTML string to find all the CSS and JS scripts that were referenced by the rendered components. We rely on three mechanism for that:
1. Using middleware
2. Using `Component.render()` or `Component.render_to_response()` as render entry point (instead of `Template.render()` that you've used)
3. Manually passing the rendered HTML to `render_dependencies()``

[...]

PR #972, opened 2025-02-12
feat: add setting to specify a render type provider for ComponentDependencyMiddleware

wofl:
Currently the render type will always be `document` in the ComponentDependencyMiddleware and this can not be altered by the developer without providing his own middleware.

This PR adds a new setting that allows the user to customize the render type that is used for a given request and response. The default will be `document` regardless of the request or response.

Ultimately this allows the developer to provide his own render type provider which could e.g. look like the following for htmx:

```python
def htmx_render_type(request: HttpRequest, response: HttpResponse) -> Literal["document", "fragment"]:
    assert hasattr(request, "htmx"), "The htmx middleware is not installed"
    if request.htmx:
        return "fragment"
    return "document"
```

### DIFF
--- a/src/django_components/__init__.py
+++ b/src/django_components/__init__.py
@@ -17,7 +17,11 @@ from django_components.component_registry import (
     registry,
 )
 from django_components.components import DynamicComponent
-from django_components.dependencies import render_dependencies
+from django_components.dependencies import (
+    render_dependencies,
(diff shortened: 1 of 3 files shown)


wofl:
Docs are currently missing because this is a draft for simplifying the discussion in #968

PR #975, opened 2025-02-14
feat:forward context processors variables in context in ISOLATED mode

ldurey:
Provide context_processors_data property to Component to access those variables in Component

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -339,6 +339,26 @@ class Component(
     def get_context_data(self, *args: Any, **kwargs: Any) -> DataType:
         return cast(DataType, {})
 
+    @property
+    def context_processors_data(self) -> DataType:
+        """Retrieve data injected in RequestContext via context_processors.
+
+        **Exemple:**
+
+        ```py
+        class MyComponent(Component):
+            get_context_data(self):
+            user = self.context_processors_data['user']
+            return {
+                'is_logged_in': user.is_authenticated,
+            }
+        ```
+        """
+        context_processors_data = {}
+        if hasattr(self, 'outer_context') and hasattr(self.outer_context, '_processors_index'):
+            context_processors_data = self.outer_context.dicts[self.outer_context._processors_index]
+        return context_processors_data
+
     js: Optional[str] = None
     """
     Main JS associated with this component inlined as string.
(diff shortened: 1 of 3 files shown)


ldurey:
I have done my best on this, i've struggled so much with the testing part even if there is not much tested...
I hope this will be ok for you for a start about this feature.

I'm starting some manual testing on my project to see if everything is ok.

src/django_components/component.py
```diff
@@ -339,6 +339,26 @@ def get_template(self, context: Context) -> Optional[Union[str, Template]]:
...
+        class MyComponent(Component):
+            get_context_data(self):
+            user = self.context_processors_data['user']
+            return {
+                'is_logged_in': user.is_authenticated,
+            }
+        ```
+        """
+        context_processors_data = {}
+        if hasattr(self, 'outer_context') and hasattr(self.outer_context, '_processors_index'):
```
ldurey, line 358: i'm testing if `outer_context` exist because it does not exist in case of Component.as_view().

JuroOravec:
@ldurey Thanks again for that! I was writing additional tests when I came across an issue. It turned out that this feature was more complex than expected, so I had to refactor most of the PR.

The problem was that:
1. I wanted to access the context processors data from within `get_context_data()`
1. Normally, Context object is "bound" to a template, and the context processors data generated
   only once Django "binds" the `RequestContext` to a Template.
2. But because we allow to dynamically resolve the template with `get_template_name()` and `get_template()`, we run these AFTER `get_context_data()`

So bit of a chicken-and-egg problem.

So I decided to take the logic that generates the context processors data, and make it into a standalone function, so we don't have to bind the context a template in order to generate the context processors data.

Moreover, since components accept a `request` kwarg, I made it such that the context processors data are generated when:
- The component is rendered with `RequestContext` (Regular Django behavior)
- The component is rendered with a regular `Context` (or none), but the `request` kwarg is set.
- The component is nested in another component that matches one of the two conditions above.

Next I dediced to expose the `request` object (whether from `RequestContext` or passed directly) on the Component insstance as `Component.request`.

And lastly, we now explicitly make the context processors data available from within component's templates (if `request` is present).

---

So to summarize, we now implement the same behaviour as RequestContext, if a request is passed to the Component.

JuroOravec:
I still need to update the docs, then this will be good to go.

ldurey:
I understand the issue you came across.
Thank you again for your work on this project !

JuroOravec:
@EmilStenstrom This is now ready for review, would you look at it?

EmilStenstrom:
@JuroOravec Don't need two reviewers! If you are happy, I am happy!

PR #979, opened 2025-02-15
fix: thread unsafe media resolve validation by moving it to ComponentMedia __post_init

### LINKED ISSUE #977: Error "Received non-null value from both 'template' and 'template_file' in Component Thumbnail. Only one of the two must be set." : _resolve_media in component_media.py maybe not thread-safe?
I'm currently investigation the above issue, here part of the stack trace:

```
  File "/usr/local/lib/python3.12/site-packages/django_components/component_media.py", line 333, in __get__
    return get_comp_media_attr(self._attr_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django_components/component_media.py", line 320, in get_comp_media_attr
    return _get_comp_cls_attr(comp_cls, attr)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django_components/component_media.py", line 349, in _get_comp_cls_attr
    _resolve_media(base, comp_media)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/django_components/component_media.py", line 522, in _resolve_media
    comp_media.template = _get_asset(

  File "/usr/local/lib/python3.12/site-packages/django_components/component_media.py", line 836, in _get_asset
    raise ImproperlyConfigured(
    ^

Exception Type: ImproperlyConfigured at /some_view
Exception Value: An error occured while rendering component some_component:
Received non-null value from both 'template' and 'template_file' in Component Thumbnail. Only one of the two must be set.
```

I definitely did not make the developer mistake this is trying to catch in my definitions.

The issue is hard to reproduce, and I did not really manage to get a small setup reproducing it. It reliably occurs in our production environment directly after deployment, though.

I think the culprit is the following (see comments)

```python
def _get_asset(
    comp_cls: Type["Component"],
    comp_media: ComponentMedia,
    inlined_attr: str,
    file_attr: str,
    comp_dirs: List[Path],
    type: Literal["template", "static"],
) -> Optional[str]:
    # ...
    asset_content = getattr(comp_media, inlined_attr, None)
    asset_file = getattr(comp_media, file_attr, None)

    # The following check might be true in multi-threaded environments,
    # since comp_media.template is set before comp_media.resolved
    if asset_file is not None and asset_content is not None:
        raise ImproperlyConfigured(
            f"Received non-null value from both '{inlined_attr}' and '{file_attr}' in"
            f" Component {comp_cls.__name__}. Only one of the two must be set."
        )

    # ...

    # The property
    return asset_content


def _resolve_media(comp_cls: Type["Component"], comp_media: ComponentMedia) -> None:
    # ....

    # comp_media.template is set here...
    comp_media.template = _get_asset(
        comp_cls,
        comp_media,
        inlined_attr="template",
        file_attr="template_file",
        comp_dirs=comp_dirs,
        type="template",
    )
    comp_media.js = _get_asset(
        comp_cls, comp_media, inlined_attr="js", file_attr="js_file", comp_dirs=comp_dirs, type="static"
    )
    comp_media.css = _get_asset(
        comp_cls, comp_media, inlined_attr="css", file_attr="css_file", comp_dirs=comp_dirs, type="static"
    )

    # ... but only here the resolved flag will be set, thus preventing the above check to run
    comp_media.resolved = True
```

So maybe it's just that the functions are not thread-safe, and django template tag rendering expects all nodes to be thread-safe (see https://docs.djangoproject.com/en/dev/howto/custom-template-tags/#template-tag-thread-safety).

 I think the quick fix might just to remove the check, with the minor downside that some code gets executed twice the first time assets are resolved.

I'm still a bit confused, so I hope the direction is roughly correct. Any ideas how to narrow this down more are welcome.

oliverhaas:
Closes #977

### DIFF
--- a/src/django_components/component_media.py
+++ b/src/django_components/component_media.py
@@ -4,9 +4,9 @@ from collections import deque
 from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Literal, Optional, Protocol, Tuple, Type, Union, cast
-from django.core.exceptions import ImproperlyConfigured
 
 from django.contrib.staticfiles import finders
+from django.core.exceptions import ImproperlyConfigured
 from django.forms.widgets import Media as MediaCls
 from django.template import Template, TemplateDoesNotExist
 from django.template.loader import get_template
--- a/src/django_components/component_media.py
+++ b/src/django_components/component_media.py
@@ -238,7 +238,6 @@ class ComponentMedia:
                     f"Received non-null value from both '{inlined_attr}' and '{file_attr}' in"
                     f" Component {self.comp_cls.__name__}. Only one of the two must be set."
                 )
-            
 
 
 # This metaclass is all about one thing - lazily resolving the media files.
--- a/src/django_components/component_media.py
+++ b/src/django_components/component_media.py
@@ -4,9 +4,9 @@ from collections import deque
 from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Literal, Optional, Protocol, Tuple, Type, Union, cast
+from django.core.exceptions import ImproperlyConfigured
 
 from django.contrib.staticfiles import finders
-from django.core.exceptions import ImproperlyConfigured
 from django.forms.widgets import Media as MediaCls
 from django.template import Template, TemplateDoesNotExist
 from django.template.loader import get_template
@@ -220,6 +220,7 @@ class ComponentMediaInput(Protocol):
 
 @dataclass
 class ComponentMedia:
+    comp_cls: Type["Component"]
     resolved: bool = False
     Media: Optional[Type[ComponentMediaInput]] = None
     template: Optional[str] = None
@@ -228,6 +229,16 @@ class ComponentMedia:
     js_file: Optional[str] = None
     css: Optional[str] = None
     css_file: Optional[str] = None
+
+    def __post_init__(self) -> None:
+        for inlined_attr in ("template", "js", "css"):
+            file_attr = f"{inlined_attr}_file"
+            if getattr(self, inlined_attr) is not None and getattr(self, file_attr) is not None:
+                raise ImproperlyConfigured(
+                    f"Received non-null value from both '{inlined_attr}' and '{file_attr}' in"
+                    f" Component {self.comp_cls.__name__}. Only one of the two must be set."
+                )
+            
 
 
 # This metaclass is all about one thing - lazily resolving the media files.
@@ -301,6 +312,7 @@ class ComponentMediaMeta(type):
 def _setup_lazy_media_resolve(comp_cls: Type["Component"], attrs: Dict[str, Any]) -> None:
     # Collect all the original values of the lazy attributes, so we can access them from the getter
     comp_cls._component_media = ComponentMedia(
+        comp_cls=comp_cls,
         resolved=False,
         # NOTE: We take the values from `attrs` so we consider only the values that were set on THIS class,
         #       and not the values that were inherited from the parent classes.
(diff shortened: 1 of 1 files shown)


src/django_components/component_media.py
```diff
@@ -220,6 +220,7 @@ class Media:
 
 @dataclass
 class ComponentMedia:
+    comp_cls: Type["Component"]
```
oliverhaas, line 223: This is only used for logging, but I thought having a reference here does not hurt.

src/django_components/component_media.py
```diff
@@ -832,12 +843,6 @@ class MyComp(Component):
     asset_content = getattr(comp_media, inlined_attr, None)
     asset_file = getattr(comp_media, file_attr, None)
 
-    if asset_file is not None and asset_content is not None:
-        raise ImproperlyConfigured(
-            f"Received non-null value from both '{inlined_attr}' and '{file_attr}' in"
-            f" Component {comp_cls.__name__}. Only one of the two must be set."
-        )
-
```
oliverhaas, line 840: The actual important change of this PR. This validation might fail if done here due to a race condition. Django template rendering is often multi-threaded...

src/django_components/component_media.py
```diff
@@ -229,6 +230,15 @@ class ComponentMedia:
...
 
+    def __post_init__(self) -> None:
+        for inlined_attr in ("template", "js", "css"):
+            file_attr = f"{inlined_attr}_file"
+            if getattr(self, inlined_attr) is not None and getattr(self, file_attr) is not None:
+                raise ImproperlyConfigured(
+                    f"Received non-null value from both '{inlined_attr}' and '{file_attr}' in"
+                    f" Component {self.comp_cls.__name__}. Only one of the two must be set."
+                )
+
```
oliverhaas, line 241: Validation of `inlined_attr` and `file_attr` are now here, checked once at the start.

JuroOravec: Sounds good to me.

I got the error message too couple of times, but it was low priority, so thanks for looking into this!

Tho I'm trying to understand *why* there was this race condition? Because the code does have [a check](https://github.com/django-components/django-components/pull/979/files#diff-bdb1917504a9c50a51e7fdf49498d5db3ce0233be9eb81b13938d4b3dcd5efc3R518) to ensure that the file resolution happens only once.

My best guess is that we set [`ComponentMedia.resolved = True`](https://github.com/django-components/django-components/pull/979/files#diff-bdb1917504a9c50a51e7fdf49498d5db3ce0233be9eb81b13938d4b3dcd5efc3R548) only AFTER reading the files from the filesystem. And so it could be that multiple functions entered the [`_resolve_media`](https://github.com/django-components/django-components/pull/979/files#diff-bdb1917504a9c50a51e7fdf49498d5db3ce0233be9eb81b13938d4b3dcd5efc3R478) function, and then the second one would see that e.g. `comp_media.js` is already set, and raise an error?

Anyway, `__post_init__` defo makes more sense 👀

oliverhaas: Yeah. As I've written somewhere else, it's fairly surprising that template rendering is multli-threaded. And I did not want to start with threading mutex/lock here.

EmilStenstrom: Is it possible to set up a test that triggers those bugs? Perhaps Django internally has tests for this we could be inspired by?

oliverhaas: I didn't know of a clean way from the top of my head. Here was the best I could come up with:

```python
from django_components import Component

from .django_test_setup import setup_test_config
from .testutils import BaseTestCase
import threading

class ThreadingTest(BaseTestCase):
    def test_many_parallel_renders(self):

        class SimpleComponent(Component):
            template_file = "simple_template.html"

            def get_context_data(self, variable=None):
                return {}
        
        exceptions = []
        def worker():
            try:
                SimpleComponent().render()
            except Exception as exc:
                exceptions.append(exc)

        threads = [threading.Thread(target=worker) for _ in range(1000)]
        for thread in threads:
            thread.start()
        for thread in threads:
            thread.join()

        if exceptions:
            raise exceptions[0]
```

So I've basically added threading from the outside, and just do a lot of renders, in the hopes that we will cause an error raising race condition. This is still flaky, and it wouldn't catch a race conditions which does not raise an error. In my eyes it's not a proper test for future code, but it was useful for debugging.

I don't have a suggestion which would give me confidence to catch future bugs related to the underlying issue of this PR specifically.

EmilStenstrom: Great idea, created a new issue around incorporating something like this.

PR #980, opened 2025-02-16
refactor: fix error when template cache is set to 0

JuroOravec:
Closes https://github.com/django-components/django-components/issues/974


JuroOravec:
cc @EmilStenstrom

PR #984, opened 2025-02-17
docs: Syntax highlighting for mkdocs

JuroOravec:
Update the docs to use the `djc_py` code block when the code block contains inlined HTML / CSS / JS definition.

Closes https://github.com/django-components/django-components/issues/966

Blocked by https://github.com/django-components/django-components/pull/975


JuroOravec:
cc @EmilStenstrom

PR #999, opened 2025-02-22
feat: benchmarking

JuroOravec:
This PR adds these benchmarking use cases:
- Benchmarking across releases
- Displaying performance results on the website
- Compare performance between commits on pull requests

I went with [`asv`](https://asv.readthedocs.io/en/latest/) as it covers the entire benchmarking workflow. We can:

1. Define benchmark tests similarly to writing pytest tests (supports both timing and memory benchmarks)
2. Run the benchmarks and generate results for individual git commits, tags, or entire branches
3. View results as an HTML report (dashboard with charts)
4. Compare performance between two commits / tags / branches for CI integration

The new use cases / workflows in a bit more depth:
- Benchmarking across releases:

  1.  When a git tag is created and pushed, this triggers a Github Action workflow (see `docs.yml`).
  2.  The workflow runs the benchmarks with the latest release, and commits the results to the repository.
      Thus, we can see how performance changes across releases.

- Displaying performance results on the website:

  1.  When a git tag is created and pushed, we also update the documentation website (see `docs.yml`).
  2.  Before we publish the docs website, we generate the HTML report for the benchmark results.
  3.  The generated report is placed in the `docs/benchmarks/` directory, and is thus
      published with the rest of the docs website and available under [`/benchmarks/`](https://django-components.github.io/django-components/benchmarks).
      - NOTE: The location where the report is placed is defined in `asv.conf.json`.

- Compare performance between commits on pull requests:
  1. When a pull request is made, this triggers a Github Action workflow (see `benchmark.yml`).
  2. The workflow compares performance between commits.
  3. The report is added to the PR as a comment made by a bot.

https://github.com/user-attachments/assets/d1c1acb5-3d4c-4ffc-8b56-8d5af337d592

And here is how the performance is reported on PRs ([example](https://github.com/JuroOravec/django-components/pull/6#issuecomment-2675771412)):

Screenshot 2025-02-22 at 09 56 45

Screenshot 2025-02-22 at 08 21 49

Part of #14


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,105 @@
+# Run benchmark report on pull requests to master.
+# The report is added to the PR as a comment.
```
JuroOravec, line 2: This file is the github action for running benchmarks on pull request, and comparing them with the latest commit on master.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,105 @@
...
+        cache: 'pip'
+
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install asv
+
+    - name: Run benchmarks
+      run: |
+        # TODO: REMOVE ONCE FIXED UPSTREAM
```
JuroOravec, line 37: Both `benchmark.yml` and `docs.yml` include this monkeypatch to fix https://github.com/airspeed-velocity/asv_runner/issues/45.

Unfortunately `asv` is not very polished, I had to fix 2-3 bugs to get the integration working (also see https://github.com/airspeed-velocity/asv/issues/1468). But once things are working, it's working really well.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,105 @@
...
+            cat "benchmarks/monkeypatch_asv_ci.txt" >> "$file"
+        done
+        echo "Monkeypatch applied."
+        # END OF MONKEYPATCH
+
+        # Prepare the profile under which the benchmarks will be saved.
+        # We assume that the CI machine has a name that is unique and stable.
+        # See https://github.com/airspeed-velocity/asv/issues/796#issuecomment-1188431794
+        echo "Preparing benchmarks profile..."
+        asv machine --yes
```
JuroOravec, line 65: asv is aware that different machines may generate different results. So asv fingerprints the machine to generate a unique name based on the number of available CPU cores, memory, and host name, and some other stuff.

What asv then does is that it groups the benchmarking results under these machine names. So in the resulting graphs, it's then clear that the results across two machines don't belong together.

In this file, we don't really care about the machine name, because this file is for creating the PR report, so the results are not saved anywhere.

But in the case of the flow that generates the benchmarking data across releases ([source](https://github.com/django-components/django-components/pull/999/files#diff-9cf2000c53760d837a449f874e53f792819108d3a4bf346336d0f7d082deae2cR97)), there we need to ensure that the name will remain the same every time an workflow is spun up.

- NOTE: This assumes that the instance on which the benchmarking runs won't change. Which is fairly safe to assume IMO

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,105 @@
...
+        #       With it, we run benchmarks FROM the latest commit (incl) TO ...
+        # - `!` means that we want to select range spanning a single commit.
+        #       Without it, we would run benchmarks for all commits FROM the latest commit
+        #       TO the start of the branch history.
+        #       With it, we run benchmarks ONLY FOR the latest commit.
+        echo "Running benchmarks for upstream/master..."
+        DJC_BENCHMARK_QUICK=1 asv run upstream/master^! -v
+        echo "Benchmarks for upstream/master DONE."
+        echo "Running benchmarks for HEAD..."
+        DJC_BENCHMARK_QUICK=1 asv run HEAD^! -v
```
JuroOravec, line 80: Here is how the benchmarks are run - with `asv run`. 

Since `asv` integrates with git, it accepts a range of commits to run the benchmarks against. asv then internally fetches the state of the project how it looked at those commits, runs tests in virtual environment, saves the results, and cleans up.

That's why we have to use this cryptic syntax for specifying which commits to run. Maybe there's a more human-readable way to define these, I'm not sure.

But as you can see above, there's:

- `upstream/master^!` - to specify the latest commit on master
- `HEAD^!` - to specify the latest commit on the PR branch

We have to run this command before we actually compare the results of these two commits below with `asv compare`.

---

Another topic is that I prefixed these tests with `DJC_BENCHMARK_QUICK=1`.

The thing is that the full set of benchmark tests that I've defined takes 5 minutes to run. It's because I've set to take 5 samples from each test. And when we're testing the "first render", AKA when the templates are not parsed yet and caches are not populated, I've set these to all run in separate Python processes.

There's about 16 tests, each which compare 2 states, and with 5 samples - So it's like if started `python` command line interpreter ~160 times, ran some code, and closed it.

So for the PRs, I decided to use just a single test. We get a hint of whether performance changed or not, but don't get bogged down by waiting 10 minutes after each push to the PR. With a single test, it takes only 1min 40s to generate the results, which is IMO ok.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,105 @@
...
+        echo "Benchmarks for HEAD DONE."
+
+        # Compare against master
+        echo "Comparing benchmarks..."
+        asv compare upstream/master HEAD --factor 1.1 --split > benchmark_results.md
```
JuroOravec, line 85: This compares the benchmarking results.
- `--factor 1.1` means that changes less than 10% (either way) are considered OK.
- `--split` just says format the results by splitting them into sections of what improved, what deteriorated, and what didn't change.

The command then prints out a markdown table with results. So we capture that and post it as a comment on the PR in the next section.

---

As such, you can see that we're not running a "check" similar to tests, where if a test failed, a PR would be blocked from merging.

Since django-components is still in active development, I think it makes sense to consider the performance report more as a guide than a rule, and so IMO the results shouldn't block merging.

Also because the way `asv compare` currently works, it doesn't care if the performance improved or worsened (+/- 10%), it checks only for absolute change (10%)

.github/workflows/docs.yml
```diff
@@ -9,11 +9,6 @@ on:
       - '[0-9]+.[0-9]+.[0-9]+'
     branches: 
       - master
-  pull_request:
-      branches:
-      - main
-  release:
```
JuroOravec, line 15: Removed the trigger on release, because it's practically the same as trigger on tag push above. So it was just confusing whether both need to be handled separately, or whether the "tag push" event is triggered also when one makes a release (it does).

I also removed the pull_request trigger. If I remember correctly, this was used for testing the documentation workflow. But for one, `django-components` uses `master`, not `main`, and two, for debugging workflows, people can fork this repo and add this trigger themselves.

.github/workflows/docs.yml
```diff
@@ -27,30 +22,116 @@ jobs:
...
 
       - name: Install Hatch
         run: |
             python -m pip install --upgrade pip wheel
-            python -m pip install -q hatch pre-commit
+            python -m pip install -q hatch pre-commit asv
             hatch --version
 
-      - name: Create Virtual Environment
-        run: hatch env create docs
```
JuroOravec, line 46: Didn't remove, just moved it further down

.github/workflows/docs.yml
```diff
@@ -27,30 +22,116 @@ jobs:
...
+
           # required for "mike deploy" command below which pushes to gh-pages
           git config user.name github-actions
           git config user.email github-actions@github.com
 
+      ###########################################
+      # RECORD BENCHMARK - ONLY ON PUSH TO MASTER
+      ###########################################
+
+      - name: Run benchmarks for tag
```
JuroOravec, line 62: This is the workflow to generate benchmark data on release (tag push), and include the data as part of the documentation site.

This is similar to the workflow in `benchmark.yml` (comparison on PR). Except here:
1. Instead of comparing 2 commits, we just run benchmarks for a single commit - the latest commit on master.
2. We then run `asv publish` to generate a directory with the web report
3. We push the benchmarking results AND the web report to master.

So this way, we actually "save" the results for tags, and make them available via the documnetaiton site.

.github/workflows/docs.yml
```diff
@@ -72,9 +153,3 @@ jobs:
         run: |
           hatch run docs:mike deploy --push --update-aliases ${{ github.ref_name }} latest
           hatch run docs:mike set-default latest --push
-
-      - name: Build & deploy docs for a new release
-        if: github.event_name == 'release'
```
JuroOravec, line 77: Since I removed the "on release" trigger, I could also remove this

.github/workflows/tests.yml
```diff
@@ -17,6 +17,16 @@ jobs:
         os: [ubuntu-20.04, windows-latest]
 
     steps:
+    # Configure git to handle long paths
+    # See https://stackoverflow.com/questions/22575662/filename-too-long-in-git-for-windows
+    #
+    # Long paths that are over the limit are because of the benchmarking data
+    # created by asv, as these may look like this:
+    # docs/benchmarks/graphs/arch-x86_64/branch-master/cpu-AMD EPYC 7763 64-Core Processor/django-5.1/djc-core-html-parser/machine-fv-az1693-854/num_cpu-4/os-Linux 6.8.0-1021-azure/python-3.13/ram-16373792/isolated vs django modes.timeraw_render_lg_subsequent.json
+    - name: Configure git
```
JuroOravec, line 26: As the comment says, the data saved by `asv` have actually file names too long for git on Windows. So git has to be configured to explicitly allow such long names.

asv.conf.json
```diff
@@ -0,0 +1,210 @@
+{
```
JuroOravec, line 1: Config file for asv. See https://asv.readthedocs.io/en/latest/asv.conf.json.html

benchmarks/component_rendering.py
JuroOravec, line 1: I removed the files that were previously in the `benchmarks/` directory.
- the `component_rendering.py` is superseded by the new benchmarking tests
- `test_lexer_performance.py` was just a "playground" file to test stuff around. But since the parsers are / will be implemented in Rust, then that's no longer needed.

benchmarks/monkeypatch_asv.py
```diff
@@ -0,0 +1,29 @@
...
+
+    #### Returns
+    **timer** (`_SeparateProcessTimer`)
+    : A timer that runs the function in a separate process.
+    """
+    if param:
+
+        def func():
+            # ---------- OUR CHANGES: ADDED RETURN STATEMENT ----------
+            return self.func(*param)
```
JuroOravec, line 21: Another bug I had to fix from outside was that this function was missing a return statement

docs/benchmarks/asv.css
JuroOravec, line 1: The files in `docs/benchmarks` were auto-generated by `asv publish`. Normally the updates to this folder will be pushed from within the workflow that generates docs on tag release. But I included the generated files here to show what to expect.

docs/overview/code_of_conduct.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: Code of Conduct
-weight: 8
+weight: 9
```
JuroOravec, line 3: Adjusted the ordering of the pages as I added a page on performance.

But TBH I'm thinking of using a different way to set the order of the pages, because currently it means that if I want to a page to the middle, I have to update  the weights for half of all the pages.

Ideally the order could be set on the file name, e.g. `1--code_of_conduct.md`. But file names are part of the final URL, so the URLs wouldn't look pretty.

So instead I'm thinking to set the ordering in `SUMMARY.md`, which defines the site navigation.

docs/overview/performance.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Performance
+weight: 8
+---
+
+We track the performance of `django-components` using [ASV](https://asv.readthedocs.io/en/stable/).
+
+See the [benchmarks dashboard](../../benchmarks).
```
JuroOravec, line 8: On the page on performance, we link to the benchmarks dashboard. And the dashboard links back to the doumentation website.

.github/workflows/docs.yml
dalito: Removing the pull_request trigger would remove the desired test-build of the docs for PRs or not?

Removing the trigger on release is fine. I also saw that recently.  You can then also remove the final section:

https://github.com/django-components/django-components/blob/dcd4203eeadafc5d500b4b72c5cf04f1fe7317e7/.github/workflows/docs.yml#L76-L80

JuroOravec: @dalito Thanks for reading through this!

> Removing the pull_request trigger would remove the desired test-build of the docs for PRs or not?

Nope, for testing docs there's [the "test_docs" step here](https://github.com/JuroOravec/django-components/blob/9bdd6155a3ea655edd23b632f3ac2714426dd62a/.github/workflows/tests.yml#L47) . And that still ran with other tests [here](https://github.com/JuroOravec/django-components/actions/runs/13471243317/job/37645017599)

docs/overview/code_of_conduct.md
EmilStenstrom: Is it possible to use deciment points? A simple way would then be to just use 7.5 to place a page between 7 and 8.

EmilStenstrom:
@JuroOravec Obviously will not read through 40k lines of code, but from skimming and reading through your comments, this looks awesome! A really mature way to look at performance, that I haven't seen on any other django project.

JuroOravec:
I had quite a problem getting the commenting to work. The problem was that when a PR is made from within the same repo, the permisions that a workflow / github action has are fine. But if the PR is being made from a fork, then the workflow didn't have permission to make comments on the PR.

At first I tried setting different permissions, and then using a [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens). But it still didn't work.

- NOTE: I created the personal access token as owned by django-components org, but since it didn't work I've already deleted it again.

So I tried [this approach](https://stackoverflow.com/a/71683208/9788634), which splits the workflow into two:
1. Benchmarking and saving results as artifacts ([pr-benchmark-generate.yml](https://github.com/django-components/django-components/pull/999/files#diff-a353f67aa9c6448d04796425fa419b0ff82527f31ca00ad108a9950371199874))
2. Downloading the results and commenting on the PR ([pr-benchmark-comment.yml](https://github.com/django-components/django-components/pull/999/files#diff-aa1db6353bffec358c44d9d0c58f29094d07db9932a775bdefae15eff4f93128))

The idea is that the second workflow is triggered when the first workflow uploads artifacts. But since it's not directly related to the PR, it has broader permissions than the first workflow and can make the comments.

You can see in [my fork](https://github.com/JuroOravec/django-components/pull/13), that once the workflow is merged, then  the commenting should work on future PRs.

Another problem with the commenting workflow was [this here](https://github.com/django-components/django-components/pull/999/files#diff-aa1db6353bffec358c44d9d0c58f29094d07db9932a775bdefae15eff4f93128R46). Basically, after splitting the workflow into two, I had to pass the benchmark results somehow. The StackOverflow thread suggested [@actions/upload-artifact](https://github.com/actions/upload-artifact) to handle file uploading, and [@actions/download-artifact](https://github.com/actions/download-artifact). But for some reason, the GIthub action was not detecting the artifact. HOWEVER, when using Github's REST directly, it DID find it ([source](https://github.com/django-components/django-components/pull/999/files#diff-aa1db6353bffec358c44d9d0c58f29094d07db9932a775bdefae15eff4f93128R55)).

tox.ini
```diff
@@ -54,15 +55,15 @@ commands = isort --check-only --diff src/django_components
...
+  syrupy  # snapshot testing
   # NOTE: Keep playwright in sync with the version in requirements-ci.txt
   playwright==1.48.0
   requests
   types-requests
   whitenoise
 commands =
-  coverage run --branch -m pytest
-  coverage report -m --fail-under=97
+  pytest --cov=django_components --cov-fail-under=87 --cov-branch
```
JuroOravec, line 66: Found this bug with testing coverage. It turned out that we weren't checking coverage of the django-components package, but of the `tests/` directory itself :D

I noticed that because the coverage percentage dropped below 97% with those huge benchmark test files. But as I was looking into it, I saw that the coverage was reporting on the wrong directory. That's why I updated the command and tuned it down to 87%, as the actual coverage is now 89%.

tox.ini
```diff
@@ -54,15 +55,15 @@ commands = isort --check-only --diff src/django_components
 
 [testenv:coverage]
 deps =
-  pytest-coverage
+  pytest-cov
+  syrupy  # snapshot testing
```
JuroOravec, line 59: I don't think I've already mentioned it, but for the benchmark, I added 4 test files:
- UI page rendered with Django templates
- UI page rendered with django-components
- single component (as template tag) with Django templates
- single component with django-components

One thing is the performance of these, but the other thing is whether these setups work at all!

So to ensure that these work correctly:
1. I placed them in `tests/`.
2. Added snapshot testing
3. and I checked the resulting HTML in the browser to ensure that everything's there.

Details of the setup are in `benchmarks/README.md`.

I also wanted to have the benchmarked cases tested, because that's currently the only way I know of to check for regressions with this bug https://github.com/django-components/django-components/pull/956

Overall, the page that is being rendered looks something like this (there's more stuff hidden away as tabs that are also included in the rendered HTML):

benchmark-page-ui

tests/testutils.py
```diff
@@ -19,14 +19,50 @@
 middleware = ComponentDependencyMiddleware(get_response=lambda _: response_stash)
 
 
+class GenIdPatcher:
```
JuroOravec, line 22: As a reminder, internally we use a "ID generator" which returns unique IDs, so that each component, template tag node, and even each time we render a component all have their unique IDs. But in tests this is problematic, because it's not deterministic. So we patch the ID generator, to make deterministic.

I moved this logic for the patching of the random ID generator out, so I could reuse it.

And the benchmarked code also used the CSRF token, so I also added the `CsrfTokenPatcher` below, so that the CSRF token returns `predictabletoken` in the tests.

tests/__snapshots__/test_benchmark_django.ambr
JuroOravec, line 1: These `.ambr` files store the snapshots.

src/django_components/cache.py
```diff
@@ -29,18 +29,19 @@ def get_template_cache() -> LRUCache:
 
 
 def get_component_media_cache() -> BaseCache:
+    if app_settings.CACHE is not None:
```
JuroOravec, line 32: Another bug I found - this relates to the [cache for storing the JS /  CSS scripts](https://github.com/django-components/django-components/pull/946). 

Previously the logic was that we checked only the first time whether to use user-provided cache or a default one, and then reused that for eternity.

Now, we check each time we want to access the cache, which one to use.

This was causing errors when all tests were run together.

tests/test_benchmark_django_small.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+#####################################
+#
+# IMPLEMENTATION END
+#
+#####################################
+
+
+# ----------- TESTS START ------------ #
```
JuroOravec, line 324: I hinted at it in the `benchmarks/README`, and it's described more in depth in `benchmarks/benchmark_templating.py`. But I wanted to have a single source of truth for the benchmarked code, while also having it tested to know it's working correctly.

But thus this file contains also the pytest tests. And those should not be part of the benchmark.

So the benchmarked files contain snippets like this `# ----------- TESTS START ------------ #`.

When preparing the code for benchmarking, we cut away anything after this comment, to remove the pytest code.

But when running these tests as part of pytest, the section below is included.

I made a feature request here https://github.com/airspeed-velocity/asv/issues/1471, as it would be cleaner if this could be configured e.g. as environmental variables, rather than with these custom comments.

tests/test_benchmark_django.py
```diff
@@ -0,0 +1,6447 @@
+# NOTE: This file is used for benchmarking. Before editing this file,
```
JuroOravec, line 1: I've updated the files with the benchmark code to refer back to this PR, so people have a chance to read up on all these nuances if they need to make any changes to the files in the future.

benchmarks/benchmark_templating.py
```diff
@@ -0,0 +1,447 @@
...
+#       it should be "What is the relative cost of using Components?".
+#
+#       As an example, the benchmarking for the startup time and memory usage is not comparing
+#       two independent approaches. Rather, the test is checking if defining Components classes
+#       is more expensive than vanilla Django templates.
+class DjangoComponentsVsDjangoTests:
+    # Testing startup time (e.g. defining classes and templates)
+    @benchmark(
+        pretty_name="startup - large",
+        group_name=DJC_VS_DJ_GROUP,
```
JuroOravec, line 187: `asv` allows to give human-readable names to the tests with the [`pretty_name` attribute](https://asv.readthedocs.io/en/latest/benchmarks.html#benchmark-attributes).

However, it doesn't allow to modify the group names. ...But I found a way 😄 

So our dashboard can have group names like "Components vs Django":

Screenshot 2025-02-20 at 11 03 46

By default, the group names are the module names - so the files where the benchmarks are defined.

What happens is that, by default, asv assigns an attribute `benchmark_name` to each test function. The value of `benchmark_name` is:
- `<module_name>.<func_name>`
- or `<module_name>.<class_name>.<func_name>`

And it's actually this `<module_name>`, taken from the `benchmark_name` that is used as a group name.

So I was able to kinda "hack" this, by overwriting the `benchmark_name` attribute with strings like `"Components vs Django.timeraw_startup_lg"`. It still has the format `<module_name>.<func_name>`, but now the groups in the dashboard have a human-readable names.

benchmarks/benchmark_templating.py
```diff
@@ -0,0 +1,447 @@
...
+        number=1,
+        rounds=5,
```
JuroOravec, line 189: From asv's documentation:

> How ASV runs benchmarks is as follows (pseudocode for main idea):
>
> ```txt
> for round in range(`rounds`):
>    for benchmark in benchmarks:
>        with new process:
>            <calibrate `number` if not manually set>
>            for j in range(`repeat`):
>                <setup `benchmark`>
>                sample = timing_function(<run benchmark `number` times>) / `number`
>                <teardown `benchmark`>
> ```

For the tests that measure the "first render", I went with having each iteration run in a separate process, so that I don't have to manually guard against a state spilling over, and checking if all caches are reset.

For simplicity, I did the same also for other tests - so even the tests that measure render time on subsequent render (when caches are already populated), I've set them up to run each iteration in a separate process.

Here, I think that the "subsequent render" tests could actually run in the same process, which would make the overall benchmarking a bit faster (right now its 5 minutes).

But I didn't want to fiddle too much when a baseline isn't yet established.

benchmarks/benchmark_templating.py
```diff
@@ -0,0 +1,447 @@
...
+        params={
+            "renderer": ["django", "django-components"],
+        },
+    )
+    def timeraw_startup_lg(self, renderer: TemplatingRenderer):
+        return prepare_templating_benchmark(renderer, "lg", "startup", "isolated")
```
JuroOravec, line 195: And I've described this in more detail in `benchmarks/README`, but I used two kinds of tests:
- `timeraw_` for measuring time
- `peakmem_` for measuring memory

How `timeraw_` works is that the test function returns a script (as a string), and then this string is executed as python in a separate process. And so, the `prepare_templating_benchmark()` actually returns a string.

On the other hand, `peakmem_` run in the current python process [as seen here](https://github.com/django-components/django-components/pull/999/files#diff-b87efd051d493f69af971f3b0769da48a1458749292b9604793d37218b883929R257). So there, the approach is a bit different - during the test setup, we load the benchmarking code as a python module and prepare the data. And in the actual test body, we just execute the render function with given data.

docs/overview/code_of_conduct.md
JuroOravec: That's a good point, but it still requires you to know which files have weight 7 and weight 8, so you still need to look through all the files to find them 😄

JuroOravec:
> @JuroOravec Obviously will not read through 40k lines of code, but from skimming and reading through your comments, this looks awesome! A really mature way to look at performance, that I haven't seen on any other django project.

Hopefully this could serve as an inspiration to other projects! Ngl, all this info could be a nice article / tutorial for a wider Python community, but I'm not one for writing blogs (or at least not by myself).

@EmilStenstrom I've added more comments since your last. I'm finished now. After this PR is merged, the following PRs should already have the benchmark reports.

I'll probably also update the main README with a quick section on performance, but I'll do so in a next PR, but I'll do int in a separate PR.

EmilStenstrom:
@JuroOravec I've read all the comments, looks really good! :) 👍from me.

HaoZeke:
This is excellent, please also consider using the benchmarked by asv button for more visibility:

```
[![asv](http://img.shields.io/badge/benchmarked%20by-asv-blue.svg?style=flat)](http://your-url-here/)
```

PR #1000, opened 2025-02-23
refactor: use .nav.yml to define page order instead of nav weights

JuroOravec:
This addresses [this comment](https://github.com/django-components/django-components/pull/999#discussion_r1966487163) about the ordering of docs pages.

I switched from `mkdocs-nav-weight` and `mkdocs-literate-nav` to use only [`mkdocs-awesome-nav`](https://lukasgeiter.github.io/mkdocs-awesome-nav/).

`mkdocs-awesome-nav` allows to specify the navigation per directory by placing a `.nav.yml` inside that directory. So inside each docs directory there can be `.nav.yml` which specifies the order of pages.

No feature / documentation changes.


docs/overview/.nav.yml
```diff
@@ -0,0 +1,12 @@
+# `.nav.yml` is provided by https://lukasgeiter.github.io/mkdocs-awesome-nav
+nav:
+  - Welcome to Django Components: welcome.md
```
JuroOravec, line 3: I've moved the titles into the new nav files. Doesn't matter much, IMO, but at least the title in the navigation will match the title at the top of the page.

E.g. in this screenshot below, the main title is using `code blocks`, whereas the title on the side does not.

Screenshot 2025-02-23 at 22 39 23

JuroOravec:
cc @EmilStenstrom

PR #1007, opened 2025-02-27
Temp dependency rendering fix for HTMX fragments

maximbelyayev:
My personal fork of django-components 0.123 that has:

1) Ability to hot refresh to quickly view changes to component templates (lost with 0.124)

2) Temp fix to middleware for component dependency rendering from fragments (e.g., HTMX); https://github.com/django-components/django-components/issues/968


maximbelyayev:
Delete this im an idiot

PR #1008, opened 2025-03-01
feat: add decorator for writing component tests

### LINKED ISSUE #895: Testing utilities
I was writing up tests for the icon component package, and I realized that I had to address the same issues with testing components as when I'm writing tests in django-components.

So IMO it'd be helpful if django-components had a way to simplify writing tests.

Some issues with testing components are:

- Component IDs are random, so assertions don't work well
- Components may be already registered
- Our cache of `Template` instances may have leftover entries

So what I did in the icon package was to wrap that all in a `TestEnv` class, and then call `test_env.setup()` and `test_env.teardown()` in respective methods:

```py
test_env = TestEnv()

class BaseTestCase(SimpleTestCase):
    def setUp(self):
        super().setUp()
        test_env.setup()

    def tearDown(self):
        super().tearDown()
        test_env.teardown()
```

The implementation looks like this, it's taken from django-components from [here](https://github.com/EmilStenstrom/django-components/blob/master/tests/testutils.py#L22):

```py
from typing import List, Optional
from unittest.mock import patch

from django.test import SimpleTestCase

from django_components.component_registry import ComponentRegistry, all_registries

class TestEnv:
    def __init__(self, registries: Optional[List[ComponentRegistry]] = None):
        self.registries = registries or all_registries

    def setup(self):
        self._start_gen_id_patch()

    def teardown(self):
        self._stop_gen_id_patch()

        for registry in self.registries:
            registry.clear()

        from django_components.template import template_cache

        # NOTE: There are 1-2 tests which check Templates, so we need to clear the cache
        if template_cache:
            template_cache.clear()

    # Mock the `generate` function used inside `gen_id` so it returns deterministic IDs
    def _start_gen_id_patch(self):
        # Random number so that the generated IDs are "hex-looking", e.g. a1bc3d
        self._gen_id_count = 10599485

        def mock_gen_id(*args, **kwargs):
            self._gen_id_count += 1
            return hex(self._gen_id_count)[2:]

        self._gen_id_patch = patch("django_components.util.misc.generate", side_effect=mock_gen_id)
        self._gen_id_patch.start()

    def _stop_gen_id_patch(self):
        self._gen_id_patch.stop()
        self._gen_id_count = 10599485
```

### Questions

- Should this be part of the main package, or separate?
- How should it be imported?

I'm thinking that, since it's just one small class right now, it makes sense to include it in the `django_components` distribution.

So it could be used e.g. like this:

```py
from django_components.testing import TestEnv

test_env = TestEnv()

...
```

If it ever gets too large, it could be moved into a separate package, and then exposed via django-components as an [optional dependency](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies):

```
pip install django-components[testing]
```

### Notes

As part of this issues we should also document how to test components, and maybe also point to the icon package as a live example.

JuroOravec:
I was implementing the extensions feature. But when it came to writing tests, I found that I had issues with leaky global state between tests. So instead I switched to addressing testing components.

This PR unifies various helper functions used for testing, and defines a single API, the `@djc_test` decorator, for writing tests for components:

```python
from django_components.testing import djc_test

@djc_test
def test_my_component():
    @register("my_component")
    class MyComponent(Component):
        template = "..."
    ...
```

The decorator covers all our uses cases:
- Allowing to override Django settings
- Allowing to override Components settings
- Test parametrization with pytest

The decorator sets up / cleans up various global state that the tests may modify:
- Components and registries created or registered within the tests are unregistered and removed
- Imports caused by autodiscovery are reset, so that the next test can run autodiscovery, and the module will still be executed
- When components media (JS / CSS /  template) are loaded during the test, this is cleared at the end of the test
- Various caches are cleared.
- The decorator handles settings Django or Components settings per-test

```python
from django_components.testing import djc_test

@djc_test(
    components_settings={
        # Settings shared by all tests
        "app_dirs": ["custom_dir"],
    },
    parametrize=(
        # Parametrized settings
        ["components_settings"],
        [
            [{"context_behavior": "django"}],
            [{"context_behavior": "isolated"}],
        ],
        ["django", "isolated"],
    )
)
def test_context_behavior(components_settings):
    rendered = MyComponent().render()
    ...
```

The decorator is also being exposed as public API. But since it's meant for testing, it's exported from `django_components.testing`:

```py
from django_components.testing import djc_test
```

---

NOTE: I had to move away from the class-based definition as I initially suggested in #895, because it turns out that `unittest`-style tests (where the test class subclasses from `TestCase`) wasn't compatible with pytest's parametrization.

Closes #895 

Screenshot 2025-03-01 at 12 53 42

Screenshot 2025-03-01 at 12 53 52


docs/concepts/advanced/testing.md
JuroOravec, line 1: This file is the docs entry.

docs/reference/testing_api.md
```diff
@@ -0,0 +1,9 @@
+<!-- Autogenerated by reference.py -->
+
+# Testing API
```
JuroOravec, line 3: I also added a separate section to the API reference. So that the public API meant for testing is separate from the rest of the API reference.

Screenshot 2025-03-01 at 13 28 31

pyproject.toml
```diff
@@ -109,6 +109,7 @@ disallow_untyped_defs = true
 testpaths = [
     "tests"
 ]
+asyncio_mode = "auto"
```
JuroOravec, line 112: The E2E tests use Playwright, and that requires those tests to be written with async (`async def`). When these tests are managed by pytest and not Django, then pytest needs [`pytest-asyncio`](https://pytest-asyncio.readthedocs.io/en/latest/concepts.html) to handle those async tests.

And by default , `pytest-asyncio` expects us to explicitly manually set which tests are meant to be async and which not, by tagging the async functions with their decorator.

This setting configures `pytest-asyncio` to automatically identify async tests by whether they are written with `async def` or plain `def`.

src/django_components/app_settings.py
```diff
@@ -671,94 +671,140 @@ class Dynamic(Generic[T]):
...
 
+# Interface through which we access the settings.
+#
+# This is the only place where we actually access the settings.
+# The settings are merged with defaults, and then validated.
+#
+# The settings are then available through the `app_settings` object.
+#
+# Settings are loaded from Django settings only once, at `apps.py` in `ready()`.
 class InternalSettings:
```
JuroOravec, line 682: This relates to not just testing utils, but also the extensions fetures, and the [decoupling from Django](https://github.com/django-components/django-components/issues/1004):

Previously, we accessed Django's settings (`django.conf.settings`) every time we needed to access a field. I remember I implemented it this way so that tests could be parametrized with the custom parametrization logic.

When a field needed to be post-processed somehow, this was done at the point when we were accessing the settings field.

However, this had a couple of issues:
- With tests, it was harder to keep track where and how the settings are set
- The settings will need to do one-time initialization of the extensions. This clashed with the possibility that the list of extensions may change any time.
- In the context of decoupling django-components from Django, the logic that reads Components' settings from Django settings could be theoretically replaced with anything else. And that anything else could be fairly expensive (e.g. reading a file). So it makes sense to load the settings only once.

So now, instead, there's a single function that loads the settings from Django's `settings.COMPONENTS`, and does all the post-processing all at once.

This also means that the properties that were defined before, like below, could be now removed, because they are no longer needed. But I didn't want to remove them in this PR as this one is already big.

```py
@property
def APP_DIRS(self) -> Sequence[str]:
    return self._settings.app_dirs  # type: ignore[return-value]
```

src/django_components/autodiscovery.py
```diff
@@ -1,9 +1,15 @@
...
+import os
 from typing import Callable, List, Optional
 
 from django_components.util.loader import get_component_files
 from django_components.util.logger import logger
 
+# In tests, we want to capture which modules have been loaded, so we can
+# clean them up between tests. But there's no need to track this in
+# production.
+LOADED_MODULES: List[str] = []
```
JuroOravec, line 11: Added logic here so that, while a test is running within `djc_test`, we capture which modules were imported with autodiscover.

This way we can remove the loaded modules after the test, so that autodiscovery works consistently across tests.

If we did not do this, then the second test that runs autodiscovery would still find the same files, but their code would not be executed the second time, because they were already loaded.

src/django_components/component.py
```diff
@@ -99,6 +101,10 @@
 CssDataType = TypeVar("CssDataType", bound=Mapping[str, Any])
 
 
+# Keep track of all the Component classes created, so we can clean up after tests
+ALL_COMPONENTS: List[ReferenceType[Type["Component"]]] = []
```
JuroOravec, line 105: When keeping track of globally which component classes were defined, these are not stored as weak references, to allow the classes to be garbage collected if user decides to delete them.

src/django_components/component_media.py
```diff
@@ -238,6 +239,13 @@ def __post_init__(self) -> None:
                     f"Received non-null value from both '{inlined_attr}' and '{file_attr}' in"
                     f" Component {self.comp_cls.__name__}. Only one of the two must be set."
                 )
+        # Make a copy of the original state, so we can reset it in tests
+        self._original = copy(self)
+
+    # Return ComponentMedia to its original state before the media was resolved
+    def reset(self) -> None:
+        self.__dict__.update(self._original.__dict__)
```
JuroOravec, line 247: When it comes to component media (template / JS / CSS), these are loaded and populated on access (so when user e.g. tries to access `MyComponent.js`.

To ensure consistency (and because it was triggering error otherwise), the loaded media are discarded after each test.

To be able to restore the "initial" state, the initial state is saved as a copy of the `ComponentMedia` object. Through it we can access the initial state unchanged, whenever needed.

src/django_components/component_registry.py
```diff
@@ -130,7 +132,7 @@ class InternalRegistrySettings(NamedTuple):
 # We keep track of all registries that exist so that, when users want to
 # dynamically resolve component name to component class, they would be able
 # to search across all registries.
-all_registries: List["ComponentRegistry"] = []
+ALL_REGISTRIES: List[ReferenceType["ComponentRegistry"]] = []
```
JuroOravec, line 135: Also the global list of all created ComponentRegistries is storing now weakrefs instead of strong refenrences.

src/django_components/component_registry.py
```diff
@@ -223,10 +225,19 @@ def __init__(
...
         self._tags: Dict[str, Set[str]] = {}  # tag -> list[component names]
         self._library = library
-        self._settings_input = settings
-        self._settings: Optional[Callable[[], InternalRegistrySettings]] = None
+        self._settings = settings
 
-        all_registries.append(self)
+        ALL_REGISTRIES.append(cached_ref(self))
+
+    def __del__(self) -> None:
```
JuroOravec, line 232: I also added some lifecycle mgmt hooks to ComponentRegistries:
- When a registry is being deleted, we now first unregister all components, so that the associated `Library` instance is properly cleaned up.
- When a Component class is deleted, then it will be automatically unregistered from all Registries it was registered in ([see here](https://github.com/django-components/django-components/pull/1008/files#diff-a5dcf86b1b1bd690092e8ed60318835a1ce9ed2c62befcc05d43dfd4459393b6R329)). Again to ensure that the associated `Library` instance is cleaned up properly.

src/django_components/component_registry.py
```diff
@@ -254,40 +265,24 @@ def settings(self) -> InternalRegistrySettings:
...
-        # First-time initialization
         # NOTE: We allow the settings to be given as a getter function
         # so the settings can respond to changes.
-        # So we wrapp that in our getter, which assigns default values from the settings.
+        if callable(self._settings):
+            settings_input: Optional[RegistrySettings] = self._settings(self)
         else:
+            settings_input = self._settings
 
-            def get_settings() -> InternalRegistrySettings:
```
JuroOravec, line 269: I realized that this function was needlessly complex - Previously the function to return the registry settings was set to a variable on first access, and then that function was used directly to generate the settings.

Instead, the whole function can be regular method, which is called when needed.

I think I implemented the previous logic, but don't remember if there was a specific reason for that (And I don't think so).

src/django_components/component_registry.py
```diff
@@ -360,22 +358,27 @@ def unregister(self, name: str) -> None:
         entry = self._registry[name]
         tag = entry.tag
 
-        # Unregister the tag from library if this was the last component using this tag
-        # Unlink component from tag
-        self._tags[tag].remove(name)
+        # Unregister the tag from library.
+        # If this was the last component using this tag, unlink component from tag.
+        if tag in self._tags:
```
JuroOravec, line 363: Changes in this block were simply to handle when the template tag / component names were already unregistered by.

This is now needed, because this method may now be triggered automatically as part of tear down. And we'd run into an index / key error when user explicitly unregisters the component, and then we try to do it again.

src/django_components/expression.py
```diff
@@ -106,9 +106,16 @@ def render(self, context: Context) -> str:
 
 
 def is_aggregate_key(key: str) -> bool:
+    key = key.strip()
     # NOTE: If we get a key that starts with `:`, like `:class`, we do not split it.
     # This syntax is used by Vue and AlpineJS.
-    return ":" in key and not key.startswith(":")
+    return (
```
JuroOravec, line 112: Turns out that there was a test for `is_aggregate_key`, but that test was not running. So after fixing it, I realized that this function had a few bugs.

src/django_components/testing.py
```diff
@@ -0,0 +1,9 @@
+# Public API for test-related functionality
```
JuroOravec, line 1: This file is the entrypoint for test-related public API

src/django_components/util/testing.py
```diff
@@ -0,0 +1,497 @@
...
+        self._csrf_token_patch = patch("django.middleware.csrf.get_token", return_value=self._csrf_token)
+        self._csrf_token_patch.start()
+
+    def stop(self):
+        if self._csrf_token_patch:
+            self._csrf_token_patch.stop()
+            self._csrf_token_patch = None
+
+
+def djc_test(
```
JuroOravec, line 59: Here is the actual definition of the decorator. Much of the logic was previously in `tests/testutils.py`.

tests/django_test_setup.py
JuroOravec, line 1: `tests/django_test_setup.py` contained a function that sets up Django in tests with common settings. I moved it to `tests/testutils.py`. So now:

- `django_components.testing` contains public testing helpers
- Internal testing helpers are in `tests/testutils.py`

tests/test_attributes.py
```diff
@@ -1,86 +1,66 @@
...
-from .django_test_setup import setup_test_config
-from .testutils import BaseTestCase, parametrize_context_behavior
+from .testutils import PARAMETRIZE_CONTEXT_BEHAVIOR, setup_test_config
 
 setup_test_config({"autodiscover": False})
 
 
-class AttributesToStringTest(BaseTestCase):
+@djc_test
+class TestAttributesToString:
```
JuroOravec, line 18: Majority of the changes comes from migrating tests from using Django (`unittest`-style) to pytest for testing. This consists of:

1. Previously test methods were identified by being on classes that subclass `TestCase` / `SimpleTestCase`. Now, instead, for pytest to identiy test classes, the class names must start with `Test...`.

2. Previously, to manage setup / teardown of tests, there was the `BaseTestCase` class, which was our subclass of `SimpleTestCase`. Now, since test classes don't need to subclass anything for pytest, the tests are instead decorated with `@djc_test`

3. Also regarding the setup / teardown, the `setUp` / `tearDown` methods were Djagno (or `unittest`) specific. pytest instead offers [fixtures](https://docs.pytest.org/en/stable/explanation/fixtures.html).

   However, I wanted to keep things simple if possible, so instead of converting `setUp` to fixtures, I just copy-pasted and inlined the setup code. The setup code was only about registering components anyway.

4. Django's `TestCase` class offered a lot of extra methods for assertions. This is no longer available, and so a lot of changes is simply from using the assertion helpers to using plain `assert` statements, e.g. from

   ```py
    self.assertEqual(
        attributes_to_string({"foo": "bar"}),
        'foo="bar"',
    )
   ```
   To
   ```py
    assert attributes_to_string({"foo": "bar"}) == 'foo="bar"'
   ```

tests/test_attributes.py
```diff
@@ -1,86 +1,66 @@
...
+class TestHtmlAttrs:
     template_str: types.django_html = """
         {% load component_tags %}
         {% component "test" attrs:@click.stop="dispatch('click_event')" attrs:x-data="{hello: 'world'}" attrs:class=class_var %}
         {% endcomponent %}
     """  # noqa: E501
 
-    @parametrize_context_behavior(["django", "isolated"])
-    def test_tag_positional_args(self):
+    @djc_test(parametrize=PARAMETRIZE_CONTEXT_BEHAVIOR)
```
JuroOravec, line 62: The `djc_test` decorator also handles parametrization. This is available only when pytest is installed. In theory, the decorator should work also when pytest is not installed, as long as the `parametrize` kwarg is not used in the non-pytest environment.

And so I was also able replace and remove the `@parametrize_context_behavior` decorator. `@parametrize_context_behavior` was an attempt at having parametrized tests. But because Django / unittest didn't properly support that, then in the reports the parametrized tests actually showed up as one.

So one side effect is that then number of shown tests might up to double, as the parametrized tests now show up as separate tests in the report.

tests/test_component.py
```diff
@@ -23,13 +24,14 @@
 from django.test import Client
 from django.urls import path
 from django.utils.safestring import SafeString
+from pytest_django.asserts import assertHTMLEqual, assertInHTML
```
JuroOravec, line 27: So I mentioned in one of the other comments that I moved the assertion from Django's assertion helpers to plain assertion statements. However, there's two assertions `assertHTMLEqual` and `assertInHTML`, which have very specific behaviour, and so I kept them.

But to keep them while still using pytest for collecting tests, I had to add `pytest-django`, which also adds [all the Django's assertions](https://pytest-django.readthedocs.io/en/latest/helpers.html#assertions).

So as I've noted in https://github.com/django-components/django-components/issues/1004, once we get to working on decoupling django-components from Django, these assertions will have to be replaced. But until then it's fine.

tests/test_extension_view.py
JuroOravec, line 1: The renaming of this test file is actually a spillover from the work on extensions feature.

Also, I've kept this file to use the Django / unittest-style tests. Because my thinking is that, once it comes to v3 (decoupling Components from Django), then this "Component  as View" feature is a Django-specific feature, and will become a standalone installable extension. And since it's specific to Django, then we can keep using the `SimpleTestCase` for tests.

tests/test_templatetags_component.py
```diff
@@ -581,16 +581,8 @@ def test_both_components_render_correctly_with_slots(self):
             """,
         )
 
-    @parametrize_context_behavior(
-        # TODO: Why is this the only place where this needs to be parametrized?
```
JuroOravec, line 585: One nice thing with the introduction of `djc_test` is that I was able to solve this puzzle, where in some tests, the IDs didn't match.

The root cause of that was Django's caching of Template instances. 

1. When a component defined a template through it's name (`MyComponent.template_file`), we use Django's `get_template()` to load the template.
2. Django has a cache behind the scenes, so if `get_template()` is called multiple times with the same path and engine, it returns the same Template instance.
3. When a Template is being parsed, each of our template tags like `{% component %}` or `{% slot %}` receive a unique ID generated by us.
4. In tests, this ID generating helper is increasing incrementally.

And so what was happening was that:
1. The first test loaded the template, parsing the template tags and assigning IDs. So overall the test used maybe 10 calls to the ID generator.
2. On the second test, the template instance was cached. And so the template tags didn't need to be parsed again. And so the ID generator received e.g. 8 calls.
3. And thus, there was the mismatch in the results, because the results also relied on calling the ID generator.

This was fixed with the [snippet here](https://github.com/django-components/django-components/pull/1008/files#diff-e26f6801b55405a66694b4dc75a76f03d2c7935e92b20186dd1d1f912b91e487R426) in the teardown phase:

```py
for engine in engines.all():
    engine.engine.template_loaders[0].reset()
```

JuroOravec:
@EmilStenstrom this is now ready for review

src/django_components/autodiscovery.py
```diff
@@ -94,4 +100,10 @@ def _import_modules(
         logger.debug(f'Importing module "{module_name}"')
         importlib.import_module(module_name)
         imported_modules.append(module_name)
+
+        # In tests tagged with `@djc_test`, we want to capture the modules that
+        # are imported so we can clean them up between tests.
+        if os.environ.get("DJC_TESTING"):
```
EmilStenstrom, line 106: So you have to set this env variable when you run with the decorator? I don't see this mentioned in the docs?

EmilStenstrom (approved):
Nice refactoring and simplifications of many of the tests!

src/django_components/autodiscovery.py
JuroOravec: It's the other way around, the decorator sets this env variable to make django-components start recording which libraries were imported with autodiscovery.

Tbh Not sure if an env var is the right approach. Actually might better to just use a plain module-level variable to keep track of the state. It's an implementation detail, not something that people should rely on. 👀

JuroOravec: Updated in https://github.com/django-components/django-components/pull/1008/commits/070ba7cb93b89a2613ab8b8fa1466cd99e2d823e

PR #1009, opened 2025-03-03
feat: extensions

JuroOravec:
Basic support for extensions (plugins). Part of https://github.com/django-components/django-components/issues/829

This PR adds just enough features to:
- Be able to implement the "Component as View" feature as an extensions
- So that the [component-level caching](https://github.com/django-components/django-components/issues/992) could be implemented as an extension
- So that input validation could be moved out into a third-party extension

Changes:
- Extensions are set via the `COMPONENTS.extensions` setting.
- Extensions can tap into lifecycle events like Component creation / deletion, Registry creation / deletion, registering component, or for processing component input at render.
- Extensions can be configured on component level by setting a nested class, e.g. that for the caching, there would be a nested `Cache` class:
   ```py
	class Icon(Component):
	    class Cache:
	        enabled = True
	        cache = {
	            "expire_time": 3600,
	            # ... more, but only whatever django template fragment caching offers.
	        }
	
	       def get_cache_key(self, *args, **kwargs):
	           # Use `self.component` to access the owner component
	           pass
   ```

Note:

- The original issue (https://github.com/django-components/django-components/issues/829) mentioned more hooks. These will be added in due time, and we can keep the issue open till then. The ones I've added in this PR, the code around them is unlikely to be changed much (at least not by me).

   But on the other hand, for example when it comes to the hook to modify the inputs from template tags, that hook might be affected by https://github.com/django-components/django-components/issues/935, so I want to first take care of #935.

Docs article on extensions:

https://github.com/user-attachments/assets/d6bc84c4-cd67-441f-9a37-057a0ae408ce

Auto-generated reference of what hooks are available to extensions:

Screenshot 2025-03-03 at 13 53 39

Settings:

Screenshot 2025-03-03 at 13 53 58


src/django_components/app_settings.py
```diff
@@ -718,6 +746,35 @@ def _load_settings(self) -> None:
             tag_formatter=default(components_settings.tag_formatter, defaults.tag_formatter),  # type: ignore[arg-type]
         )
 
+    def _prepare_extensions(self, new_settings: ComponentsSettings) -> List["ComponentExtension"]:
```
JuroOravec, line 749: In the previous PR, I made the change to load the Components settings from Django's settings only once, instead of doing so every time we need to access Components settings.

The reason for that is here - Because I made it such that the `extensions` setting can accept import path, class, or an instance, then this needs to be normalized to Extension instances.

And if this normalization was to be done each time we access settings, there could be weird behaviour if we instantiated the instances each time, because we'd get something like:

```py
app_settings.EXTENSIONS[0] is not app_settings.EXTENSIONS[0]
```

Because behind the scenes, the two `app_settings.EXTENSIONS` would each generate a new instance on access.

Alternatively, this could have been approached with some smart caching...

...Or simply by relying on the fact that we load the settings only once.

src/django_components/apps.py
```diff
@@ -57,6 +58,9 @@ def ready(self) -> None:
         # Register the dynamic component under the name as given in settings
         registry.register(app_settings.DYNAMIC_COMPONENT_NAME, DynamicComponent)
 
+        # Let extensions process any components which may have been created before the app was ready
+        extensions._init_app()
```
JuroOravec, line 62: One thing I had to add that I didn't anticipate is that some Component classes may be defined BEFORE the Django settings are ready.

And our extensions are sourced from Django settings, so extensions don't work BEFORE Django settings are ready.

BUT, in case of the Storybook integration, the extension will need to get a list of ALL defined components. Internally there is already such a variable, but if I used that, then extensions wouldn't be able to react to NEWLY added Components, only those that were defined at the time when the extension checked.

So instead of using the internal variable, I made it into extension hooks, so extensions can be notified when Component classes are created or deleted.

And so, an extension can easily create the list of all known components by itself by collecting all Components it gets notified with.

With this approach, however, it can still happen that some Component classes get defined BEFORE the Django settings are ready. So I used an approach that I've seen in some telemetry tools (Google analytics, Mixpanel, etc):
1. The extension manager has two states - before and after setup
2. Code that may trigger hooks can be called in either states
3. E.g. when a Component class is created BEFORE the extensions manager is set up, this is stored as an "event" in a temporary list. This applies also to any other events / hooks.
4. When the extension manager is being set up, it checks if there are any stored "events", and "flushes" them  - we call the respective hooks on each extension in the order of the events.
5. After the extension manager is set up, and future events (calls to the hooks) will directly call the hooks.

And this is why there is the `extensions._init_app()`

src/django_components/component.py
```diff
@@ -199,40 +202,10 @@ def __new__(mcs, name: Any, bases: Tuple, attrs: Dict) -> Any:
 
         return super().__new__(mcs, name, bases, attrs)
 
-
-# NOTE: We use metaclass to automatically define the HTTP methods as defined
-# in `View.http_method_names`.
-class ComponentViewMeta(type):
```
JuroOravec, line 205: I've moved the logic for the Component as View feature to its own file in `extensions/view.py`

src/django_components/component.py
```diff
@@ -1152,6 +1130,19 @@ def _render_impl(
             request=request,
         )
 
+        # Allow plugins to modify or validate the inputs
+        extensions.on_component_input(
```
JuroOravec, line 1134: Here's an example how the hooks are triggered from within the codebase

src/django_components/component_registry.py
```diff
@@ -237,7 +244,19 @@ def __init__(
...
         ALL_REGISTRIES.append(cached_ref(self))
 
+        extensions.on_registry_created(
+            OnRegistryCreatedContext(
+                registry=self,
+            )
+        )
+
     def __del__(self) -> None:
+        extensions.on_registry_deleted(
```
JuroOravec, line 254: Since there are now the hooks that are triggered on Component / Registry creation or deletion, I had to also make use of the `__del__` method here ([and here](https://github.com/django-components/django-components/pull/1009/files#diff-35c84ebceae50f7149f03bbcaf2b4f1289bc2984bf7859ca521b5dbe7960c396R206))

src/django_components/extensions/view.py
```diff
@@ -0,0 +1,80 @@
...
+        BaseExtensionClass.__init__(self, component)
+        View.__init__(self, **kwargs)
+
+    # NOTE: The methods below are defined to satisfy the `View` class. All supported methods
+    # are defined in `View.http_method_names`.
+    #
+    # Each method actually delegates to the component's method of the same name.
+    # E.g. When `get()` is called, it delegates to `component.get()`.
+
+    def get(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
```
JuroOravec, line 36: Previously, these methods like `get`, `post`, etc were defined dynamically by using the `View.http_method_names` variable.

But it wasn't very readable and intuitive what's going on, so instead I expanded.

src/django_components/extension.py
```diff
@@ -0,0 +1,580 @@
...
+
+################################################
+# HOOK TYPES
+#
+# This is the source of truth for what data is available in each hook.
+# NOTE: These types are also used in docs generation, see `docs/scripts/reference.py`.
+################################################
+
+
+class OnComponentClassCreatedContext(NamedTuple):
```
JuroOravec, line 25: Within the hooks, various data can be accessed on the given "Context" object.

All these data classes are also used for generating from the docstrings on these properties.

src/django_components/extension.py
```diff
@@ -0,0 +1,580 @@
...
+            self._events.append((fn_name, ctx))
+            return
+
+        return func(self, ctx)
+
+    return wrapper  # type: ignore[return-value]
+
+
+# Manage all extensions from a single place
+class ExtensionManager:
```
JuroOravec, line 402: And the `ExtensionManager` is used for managing the extensions internally.

JuroOravec:
Mental note - I'm using the Storybook integration as a complex enough use case to tease out requirements for the extensions API.

As I was writing this PR, I realized that there will also have to be some ways for the extensions to add custom CLI subcommands.

E.g. for the Storybook integration there needs to be a command to generate Storybook JSON files out of the Components.

At first I was thinking that each extension could handle this individually by specifying the commands via Django's interface.

But it'd be better to define the commands through django-components, so that the extensions would be portable to non-Django environments.

So there would be some interface maybe like this:

```py
class ExtensionCommand:
    name: list[str]
    type: Type = str
    help: str = ""
    default: Any = None
    action: Callable
```

So that the commands added by the extension could be defined declaratively:

```py
gen_cmd = ExtensIonCommand(
    name=["arg_name"],
    action=generate_files
)

cleanup_cmd = ExtensIonCommand(
    name=["--verbose", "-v"],
    action=cleanup_files
)
```

And these could then be defined on the extension like so:

```py
    name = "my_extension"

    commands = {
        "gen": gen_cmd,
        "cleanup": cleanup_cmd,
    }
```

Django-components would then define a single command that would delegate the rest to the extensions
So there could be a command like this:

```bash
python manage.py components ext my_extension gen --all
```

Breaking it down:

```bash
1. python manage.py components
2.                             ext my_extension
3.                                            gen --all
```

1. Entrypoint to django-components' commands via Django
2. Specify to call the commands defined by the Extension called `my_extension`
3. Specify to call the `gen` command of `my_extension`, passing in `--all`.

In the future, in non-Django env, this could work the same way, but simply replacing the CLI entrypoint:
```bash
components ext my_extension gen --all
```

I'll have to play around with it to see what a good interface for `ExtensIonCommand` would be.

JuroOravec:
@EmilStenstrom This is ready for review!

EmilStenstrom (approved):
Really like the way this turned out. Clean and readable, great API design!

JuroOravec:
Thanks Emil! And good point with passing in extensions as classes. I'm used to configs mostly from JS, and since there isn't really a strong difference between a class, instance or a dictionary, then things like plugins are often passed in as plain objects.

PR #1017, opened 2025-03-08
feat: allow extensions to add commands

JuroOravec:
This PR makes it possible for extensions to define their own commands. 

```py

class DummyCommand(ComponentCommand):
    name = "dummy_cmd"
    help = "Dummy command description."

    def handle(self, *args, **kwargs):
        print(f"DummyCommand.handle: args={args}, kwargs={sorted_kwargs}")

class DummyExtension(ComponentExtension):
    name = "dummy"

    commands = [
        DummyCommand,
    ]
```

Plus these commands can have subcommands. E.g. this is an implementation of a new `components ext` command, which has two subcommands:
- `components ext run`
- `components ext list`

```py
from django_components.commands.ext_list import ExtListCommand
from django_components.commands.ext_run import ExtRunCommand
from django_components.util.command import ComponentCommand

class ExtCommand(ComponentCommand):
    name = "ext"
    help = "Run extension commands."

    subcommands = [
        ExtListCommand,
        ExtRunCommand,
    ]
```

Changes:

- Commands:
  - Refactored how commands are defined. Now the definition is independent of Django, and instead uses the underlying [`argparse`](https://docs.python.org/3/library/argparse.html).
    - This way, there is only an adaptor for Django's Commands API. But the commands definitions could be easily adapted to be used standalone.
  - All django-components commands are scoped under `components` top-level command, e.g. instead of `startcomponent`:
    ```bash
    python manage.py startcomponent
    ```

    there is now:
    ```bash
    python manage.py components create
    ```
  - `upgradecomponent` and `startcomponent` commands are deprecated, and marked for removal in v1.
    - Instead of `upgradecomponent`, one should use `components upgrade`
    - Instead of `startcomponent`, one should use `components create`
  - I also updated the docs generation, so that this new kind of defining commands still has the documentation generated.
  - Our definition of commands can have subcommands. This is how it's possible to have to all the components scoped under `components <cmd>`

- Extensions
  - Extensions can now specify a list of commands that they introduce
  - Extension's command can be executed with `components ext run`:
     ```bash
     python manage.py components ext run <my_ext> <ext_cmd>
     ```
    - This is so that, for the Storybook integration, there is a way to have a command that generates Storybook JSON files from the defined components.
  - This was really more a proof of concept to check that things are working, but I also added a command `components ext list`, which lists all the currently installed extensions for django-components.

Screenshot 2025-03-04 at 16 31 49

Screenshot 2025-03-05 at 12 02 54

https://github.com/user-attachments/assets/e31740a9-7f07-4ab9-a4bd-e481bdfbd6d7


docs/concepts/advanced/extensions.md
```diff
@@ -328,3 +329,286 @@ class MyComponent(Component):
...
+        ),
+        # Optional argument
+        CommandArg(
+            name_or_flags=["--shout", "-s"],
+            action="store_true",
+            help="Shout the hello",
+        ),
+        # When printing the command help message, `--bar` and `--baz`
+        # will be grouped under "group bar".
+        CommandArgGroup(
```
JuroOravec, line 467: These `CommandArg` and `CommandArgGroup` objects are passed to `argparse` functions. But those argparse functions don't provide a standalone typing for their input. So I created types like `CommandARg` and `CommandArgGroup`.

docs/scripts/reference.py
```diff
@@ -477,13 +471,72 @@ def gen_reference_commands():
     with out_file.open("w", encoding="utf-8") as f:
         f.write(preface + "\n\n")
 
-        for cmd_name, cmd_path in command_modules:
-            cmd_module = import_module(cmd_path)
+        # Document all commands defined by django-components
```
JuroOravec, line 474: Previously, the reference docs for the commands were generated by searching for all commands in `management/commands/` dir.

This now applies only for the deprecated `startcomponent` and `upgradecomponent` commands.

The rest is now all scoped under the `components` top-level command.

So the logic to discover the commands defined by django-components was updated to:
1. Get the class defining the root `components` command.
2. Add documentation entry on this command.
3. Get all subcommands, and repeat.

src/django_components/__init__.py
```diff
@@ -6,6 +6,15 @@
...
 from django_components.app_settings import ContextBehavior, ComponentsSettings
 from django_components.autodiscovery import autodiscover, import_libraries
+from django_components.util.command import (
+    CommandArg,
+    CommandArgGroup,
+    CommandHandler,
+    CommandLiteralAction,
+    CommandParserInput,
+    CommandSubcommand,
+    ComponentCommand,
```
JuroOravec, line 16: Exported the command-related entities. `ComponentCommand` is the command class. The rest of the `CommandXxx` variables are for typing.

src/django_components/commands/components.py
```diff
@@ -0,0 +1,29 @@
+from django_components.commands.create import CreateCommand
+from django_components.commands.ext import ExtCommand
+from django_components.commands.upgrade import UpgradeCommand
+from django_components.util.command import ComponentCommand
+
+
+# TODO_V3 - This command should be called standalone as "components":
+#           `python -m components start <name>`
+#           `components start <name>`
+class ComponentsRootCommand(ComponentCommand):
```
JuroOravec, line 10: This is the definition of the `python manage.py components` command.

src/django_components/commands/create.py
```diff
@@ -0,0 +1,220 @@
...
+from textwrap import dedent
+from typing import Any
+
+from django.conf import settings
+from django.core.management.base import CommandError
+
+from django_components.util.command import CommandArg, ComponentCommand, style_success, style_warning
+
+
+class CreateCommand(ComponentCommand):
```
JuroOravec, line 12: This is what was previously `python manage.py startcomponent`. The `startcomponent` command is still available to the users, but the source of truth for the command logic is now defined indepedently of Django.

src/django_components/commands/ext_list.py
```diff
@@ -0,0 +1,55 @@
+from typing import Any
+
+from django_components.extension import extensions
+from django_components.util.command import CommandArg, ComponentCommand
+
+
+class ExtListCommand(ComponentCommand):
```
JuroOravec, line 7: Definition for `python manage.py components ext list`, which lists the names of all installed extensions. Mainly for debugging

src/django_components/commands/ext_run.py
```diff
@@ -0,0 +1,113 @@
...
+#
+# NOTE: This is possible, because Django sets up the project and settings BEFORE the commands are loaded.
+class SubcommandsDescriptor:
+    def __get__(self, obj: Any, objtype: Optional[Type] = None) -> List[Type[ComponentCommand]]:
+        # This will be called when accessing ExtRunCommand.subcommands
+        # or instance.subcommands
+        return _gen_subcommands()
+
+
+class ExtRunCommand(ComponentCommand):
```
JuroOravec, line 49: Definition for `python manage.py components ext run`. This is the command that one can use to run commands defined by extensions. E.g. the idea is that the storybook integration could use something like:

`python manage.py components ext storybook generate`

src/django_components/commands/startcomponent.py
```diff
@@ -0,0 +1,11 @@
+from django_components.commands.create import CreateCommand
+
+
+# TODO_REMOVE_IN_V1 - Superseded by `components create`
+class StartComponentCommand(CreateCommand):
```
JuroOravec, line 5: The old `startcomponent` command, defined for backwards compatibility.

src/django_components/commands/upgrade.py
```diff
@@ -0,0 +1,79 @@
...
+
+from django.conf import settings
+from django.template.engine import Engine
+
+from django_components.template_loader import Loader
+from django_components.util.command import CommandArg, ComponentCommand
+
+
+# TODO_V1 - Remove, no longer needed?
+class UpgradeCommand(ComponentCommand):
```
JuroOravec, line 14: `upgradecomponent` is the same as `startcomponent`, that the original API was kept for backwards compatibility. But the command itself is now defined as a `ComponentCommand` subclass, so it's independent of Django. And the new way to call this command is `components upgrade`.

src/django_components/compat/django.py
```diff
@@ -0,0 +1,116 @@
+from argparse import ArgumentParser
+from typing import Any, Optional, Type
+
+import django
+from django.core.management.base import BaseCommand as DjangoCommand
+
+from django_components.util.command import (
```
JuroOravec, line 7: In this file is logic that converts the `ComponentCommand` into Django's `BaseCommand`.

src/django_components/management/commands/components.py
```diff
@@ -0,0 +1,5 @@
+from django_components.commands.components import ComponentsRootCommand
+from django_components.compat.django import load_as_django_command
+
+# TODO_V3
+Command = load_as_django_command(ComponentsRootCommand)
```
JuroOravec, line 5: This file ensures that we can use our commands through Django as

```
python manage.py components
```

src/django_components/util/command.py
```diff
@@ -0,0 +1,407 @@
+import sys
+from argparse import Action, ArgumentParser
+from dataclasses import asdict, dataclass
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Literal, Optional, Protocol, Sequence, Type, Union
+
+if TYPE_CHECKING:
+    from argparse import _ArgumentGroup, _FormatterClass
+
+#############################
+# Argparse typing
```
JuroOravec, line 10: This is the file with the main logic around managing and creating commands.

The API for defining our commands is build on top of `argparse`. Django also uses `argparse`, so at least the behaviour is similar.

I wanted the API to be declarative, but the way `ArgumentParser` is defined is that you add arguments by calling [`ArgumentParser.add_argument()`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_argument).

So I added these dataclasses to be able to define the inputs to functions like `add_argument()` declaratively.

src/django_components/util/command.py
```diff
@@ -0,0 +1,407 @@
...
+    exit_on_error: Optional[bool] = None
+    """Determines whether or not ArgumentParser exits with error info when an error occurs. (default: `True`)"""
+
+    def asdict(self) -> dict:
+        """Convert the dataclass to a dictionary, stripping out fields with `None` values"""
+        return _remove_none_values(asdict(self))
+
+
+#############################
+# Command logic
```
JuroOravec, line 202: The rest of the file the is about:
- the `ComponentCommand` class, the class for defining commands for extensions
- helpers for transforming the declarative command definition into what `argparse` expects.

src/django_components/util/command.py
```diff
@@ -0,0 +1,407 @@
...
+    The function that is called when the command is run. If `None`, the command will
+    print the help message.
+    """
+
+    parser_input: Optional[CommandParserInput] = None
+    """
+    The input to use when creating the `ArgumentParser` for this command. If `None`,
+    the default values will be used.
+    """
+    subparser_input: Optional[CommandSubcommand] = None
```
JuroOravec, line 304: Note: ArgumentParser supports defining subcommands (they call it [subparsers](https://docs.python.org/3/library/argparse.html#sub-commands)). However, the fields that can be set for the subparser are slightly different than what can be set when initializing `ArgumentParser`.

Wasn't sure how to approach this, so for simplicity I kept the two separate:
- If a command class is being used as a top-level command, we're creating an instance of `ArgumentParser`, and in that case `setup_parser_from_command()` uses the settings from the `parser_input` field
- But if the command class is being used as a subcommand, we're not instantiating `ArgumentParser` directly, but instead call `ArgumentParser.add_subparsers().add_parser()`. And in this case we'll pass the settings in `subparser_input`.

JuroOravec:
@EmilStenstrom

EmilStenstrom:
> @EmilStenstrom

Sorry for the delay, I had reviewed and even double-checked that I had. But had not hit submit on the comments. Apologies.

docs/concepts/advanced/extensions.md
```diff
@@ -328,3 +329,286 @@ class MyComponent(Component):
...
 This will log the component name and color when the component is created, deleted, or rendered.
+
+## Extension Commands
+
+Extensions in django-components can define custom commands that can be executed via the Django management command interface. This allows for powerful automation and customization capabilities.
+
+For example, if you have an extension that defines a command that prints "Hello world", you can run the command with:
+
+```bash
+python manage.py components ext run my_ext hello
```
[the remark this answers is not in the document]
JuroOravec: Agree, but with shorter formats there could be conflicts:

- `python manage.py components <ext_name> <ext_command>`

   In this case the extension names could conflict with normal commands. Theoretically someone could name an extension `create`, and then it would conflict with the `components create` command.

   Ofc this could be countered with checking and disallowing some extension names, but it'd then be more problematic to add more commands in the future.

- `python manage.py components ext <ext_name> <ext_command>`

   In this case, only the `run` word is omitted. While this is not that bad and IMO could work, I can also imagine that there could be more extension commands we might want to add. Beside the `ext run`, there could be `ext info`, which would print out some info / stats. Or `ext create` to create a boilerplate for a new extension.

   So again to avoid conflicts, the simplest is to scope user extensions / commands under `ext run`.

In the long term, the `python manage.py` could be dropped, and then we'd have just:

`components ext run my_ext hello`

src/django_components/commands/components.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    The entrypoint for the "components" commands.
+
+    ```bash
+    python manage.py components start <name>
+    python manage.py components upgrade <name>
+    python manage.py components ext list
```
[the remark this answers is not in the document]
JuroOravec: Love that idea! I outlined it here https://github.com/django-components/django-components/issues/1024

CHANGELOG.md
```diff
@@ -6,8 +6,19 @@
...
 
+    - Hook into lifecycle events of django-components
+    - Pre-/post-process component inputs, outputs, and templates
+    - Add extra methods or attributes to Components
+    - Add custom CLI commands to django-components
+
     Read more on [Extensions](https://django-components.github.io/django-components/0.131/concepts/advanced/extensions/).
 
+- New CLI commands:
+    - `components create <name>` - Create a new component (supersedes `startcomponent`)
```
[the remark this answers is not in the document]
JuroOravec: Create. I think it makes more sense to rename it to `create` since also the description and docs talk about "creating" a new component.

JuroOravec:
@EmilStenstrom No worries, thanks for the review!

PR #1025, opened 2025-03-16
feat: allow extensions to add url views

JuroOravec:
Another PR for the extensions feature.

### DIFF
--- a/docs/concepts/advanced/extensions.md
+++ b/docs/concepts/advanced/extensions.md
@@ -15,10 +15,10 @@ Extensions can be set by either as an import string or by passing in a class:
 ```python
 # settings.py
 
-class MyExtension(ComponentsExtension):
+class MyExtension(ComponentExtension):
     name = "my_extension"
 
-    class ExtensionClass(BaseExtensionClass):
+    class ExtensionClass(ComponentExtension.ExtensionClass):
         ...
 
 COMPONENTS = ComponentsSettings(
@@ -188,13 +188,13 @@ This is how extensions define the "default" behavior of their nested extension c
 For example, the `View` base extension class defines the handlers for GET, POST, etc:
 
 ```python
-from django_components.extension import ComponentExtension, BaseExtensionClass
+from django_components.extension import ComponentExtension
 
 class ViewExtension(ComponentExtension):
     name = "view"
 
     # The default behavior of the `View` extension class.
-    class ExtensionClass(BaseExtensionClass):
+    class ExtensionClass(ComponentExtension.ExtensionClass):
         def get(self, request):
             return self.component.get(request)
 
@@ -229,7 +229,7 @@ class MyTable(Component):
 
 !!! warning
 
-    When writing an extension, the `ExtensionClass` MUST subclass the base class [`BaseExtensionClass`](../../../reference/api/#django_components.ComponentExtension.BaseExtensionClass).
+    When writing an extension, the `ExtensionClass` MUST subclass the base class [`ComponentExtension.ExtensionClass`](../../../reference/api/#django_components.ComponentExtension.ExtensionClass).
 
     This base class ensures that the extension class will have access to the component instance.
 
@@ -276,7 +276,7 @@ from django_components.extension import (
     OnComponentInputContext,
 )
 
-class ColorLoggerExtensionClass(BaseExtensionClass):
+class ColorLoggerExtensionClass(ComponentExtension.ExtensionClass):
     color: str
 
 
@@ -356,7 +356,7 @@ This subclass should define:
 - `handle` - the logic to execute when the command is run
 
 ```python
-from django_components import ComponentCommand, ComponentsExtension
+from django_components import ComponentCommand, ComponentExtension
 
 class HelloCommand(ComponentCommand):
     name = "hello"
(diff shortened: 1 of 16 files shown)


docs/concepts/advanced/extensions.md
```diff
@@ -15,10 +15,10 @@ Extensions can be set by either as an import string or by passing in a class:
 ```python
 # settings.py
 
-class MyExtension(ComponentsExtension):
+class MyExtension(ComponentExtension):
     name = "my_extension"
 
-    class ExtensionClass(BaseExtensionClass):
```
JuroOravec, line 21: In the docs we ask people to subclass from `BaseExtensionClass` in order to define the "extension class" which would add some nested class to components, e.g.:

```py
class Table(Component):
    class MyExtension:
        ...
```

But `BaseExtensionClass` wasn't publically exposed. Which made me think that maybe it could be easier to access and remember that base class as `ComponentExtension.ExtensionClass`.

src/django_components/commands/ext_run.py
```diff
@@ -55,15 +55,15 @@ class ExtRunCommand(ComponentCommand):
     For example, if you define and install the following extension:
 
     ```python
-    from django_components ComponentCommand, ComponentsExtension
+    from django_components import ComponentCommand, ComponentExtension
```
JuroOravec, line 58: Typos from previous PRs.

src/django_components/compat/django.py
```diff
@@ -114,3 +121,44 @@ def handle(self, *args: Any, **options: Any) -> None:
...
 
     return Command
+
+
+################################################
+# ROUTING
+################################################
+
+
+def routes_to_django(routes: List[URLRoute]) -> List[URLPattern]:
```
JuroOravec, line 131: Same way as there is a function that converts our own CLI command format into Django's, here is a fuction that converts our own URL route format into Django's.

src/django_components/component.py
```diff
@@ -111,7 +111,7 @@
 
 # NOTE: `ReferenceType` is NOT a generic pre-3.9
 if sys.version_info >= (3, 9):
-    AllComponents = List[ReferenceType["ComponentRegistry"]]
+    AllComponents = List[ReferenceType["Component"]]
```
JuroOravec, line 114: Typo from previous PRs.

src/django_components/extension.py
```diff
@@ -562,33 +573,51 @@ def _init_component_instance(self, component: "Component") -> None:
...
             if hook == "on_component_class_created":
                 on_component_created_data: OnComponentClassCreatedContext = data
                 self._init_component_class(on_component_created_data.component_cls)
             getattr(self, hook)(data)
         self._events = []
 
+        # Populate the `urlpatterns` with URLs specified by the extensions
+        # TODO_V3 - Django-specific logic - replace with hook
+        urls: List[URLResolver] = []
+        for extension in self.extensions:
```
JuroOravec, line 606: This is where the custom `URLRoute` objects are converted to Django's `URLResolver` objects.

src/django_components/extension.py
```diff
@@ -651,3 +680,25 @@ def on_component_data(self, ctx: OnComponentDataContext) -> None:
 
 # NOTE: This is a singleton which is takes the extensions from `app_settings.EXTENSIONS`
 extensions = ExtensionManager()
+
+
+################################
+# VIEW
+################################
+
+# Extensions can define their own URLs, which will be added to the `urlpatterns` list.
```
JuroOravec, line 689: There's a chicken-and-egg problem, as the URLs are collected before the Django is set up (so before the `AppConfig.ready()` hook in `apps.py` runs). As such, at the point where the URLs are collected, we don't know yet about any extensions.

One approach I tried out a bit was to subclass Django's `URLResolver`, or define similar class that could be instantiated in its place. `URLResolver` is a class that does what it says - when resolving a URL, the `URLResolver` is given the URL path, and the resolver decides if any of its subpaths match that.

But then I decided to go with the approach as seen below, as this way Django takes care of creating and configuring up the `URLResolver` instance:

1. Use `django.urls.include()` with empty list. `django.urls.include()` simply formats the given data into a correct shape (a tuple) for the `django.urls.path()`. It may be given a python import string or a list of subpaths.
2. Use `django.urls.path()` for the root route. If `django.urls.path()` receives a callable, it creates an instance of `URLPattern`. But if it reiceves a tuple (as created by `include()`, it will create an instance of `URLResolver` with the list of routes as defined in `include()`.

So by passing in an empty list to `include()` and then passing that to `path()`, we get a valid instance of Django's `URLResolver` with no subpaths, without having to touch Django's internals.

And so then what happens is that, when Django is initialized, we check the extensions we've been given, and populate the dummy `URLResolver` with the routes, as seen on [line 618](https://github.com/django-components/django-components/pull/1025/files#diff-db08f1fb2ef2c3d9f128ad726be584b57ce9ff226060506d1093c3f87031fb6fR618).

JuroOravec:
@EmilStenstrom This is now ready for review! This is the last one for the extensions API.

Altho to be exact:
- There's one more PR after this with some helpers (e.g. to get all components) to make writing extensions easier.
- After that the extension API can be released as part of v0.131
- And even after that it still makes sense to keep the issue with extensions API open, as there's still plenty of plugin hooks to implement. But those will be done in due time.

EmilStenstrom (approved):
Looks good to me!

PR #1026, opened 2025-03-17
docs: fix links in getting started section + few other

JuroOravec:
There were a lot of broken links in the "Getting started" section. This might have been caused when I moved those pages into its own top-level section out from the "concepts" section.


JuroOravec:
@EmilStenstrom Tagging for visibility

PR #1027, opened 2025-03-17
refactor: refactor tests to create components inside test functions

JuroOravec:
Another maintenance / cleanup PR. As I was writing the helper function to get all components ([mentioned here](https://github.com/django-components/django-components/pull/1025#issuecomment-2727405187)), I found that inside the test, the helper function returned like 60 extra components.

Turns out that this is because there's still a lot of tests where the component class is created outside of the test function, and so they are not cleaned up. On the other hand, the components that are defined within test functions wrapped with `@djc_test` are cleaned up properly.

My first approach was to rewrite the tests, so that the component classes are defined within the scope of `@djc_test`. After refactoring two test files, I gave up, because it would be still a lot of work, and there are more pressing tasks. And instead, I rewrote the test such that it doesn't check for the absolute number of returned components, but only checks that the total was increased by 1 when a component is created (NOTE: This test is not part of this PR).

So this PR includes those two tests to show how it can be done.

### DIFF
--- a/tests/test_templatetags_slot_fill.py
+++ b/tests/test_templatetags_slot_fill.py
@@ -13,21 +13,17 @@ from .testutils import PARAMETRIZE_CONTEXT_BEHAVIOR, setup_test_config
 setup_test_config({"autodiscover": False})
 
 
-class SlottedComponent(Component):
-    template: types.django_html = """
-        {% load component_tags %}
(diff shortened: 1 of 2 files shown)


JuroOravec:
cc @EmilStenstrom

PR #1029, opened 2025-03-17
refactor: save PR number to a file to access it when posting a benchmark comment

JuroOravec:
The workflow for benchmarking the PR and posting the results consists of two Github actions:
1. Check out to PR, install deps, run the benchmarks, save them as files ([workflow artifacts](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow))
2. Load the saved file, and post it to the PR as a comment.

The second part is currently not working, because the second workflow is unable to get the metadata about which PR to write the comment to.

Don't know why that metadata is missing, there were some cases where it worked, and I wasn't able to figure out why it's not available.

So in this PR I'm trying a different approach to pass the PR number:
- Before, I tried accessing the metadata from within the second workflow with github script action with `context.payload.workflow_run.pull_requests[0].number`
- Now, instead, the first workflow pushes that info in an extra file pushed as a workflow artifact

### DIFF
--- a/.github/workflows/pr-benchmark-comment.yml
+++ b/.github/workflows/pr-benchmark-comment.yml
@@ -86,6 +86,7 @@ jobs:
           script: |
             const fs = require('fs');
             const results = fs.readFileSync('./benchmark_results.md', 'utf8');
+            const issue_number = Number.parseInt(fs.readFileSync('./pr_number.txt', 'utf8'));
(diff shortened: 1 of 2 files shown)


JuroOravec:
cc @EmilStenstrom

PR #1030, opened 2025-03-17
feat: registry.has(); helpers to get all components and registries; access component from ext class

JuroOravec:
PR with various small changes related to writing extensions, as mentioned here https://github.com/django-components/django-components/pull/1025#issuecomment-2727405187

- Add `ComponentRegistry.has()` to check if a component is registered without raising an error.
- Get all created `Component` classes with `all_components()`.
- Get all created `ComponentRegistry` instances with `all_registries()`.

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -111,13 +111,23 @@ CssDataType = TypeVar("CssDataType", bound=Mapping[str, Any])
 
 # NOTE: `ReferenceType` is NOT a generic pre-3.9
 if sys.version_info >= (3, 9):
-    AllComponents = List[ReferenceType["Component"]]
+    AllComponents = List[ReferenceType[Type["Component"]]]
 else:
     AllComponents = List[ReferenceType]
 
 
 # Keep track of all the Component classes created, so we can clean up after tests
 ALL_COMPONENTS: AllComponents = []
+
+
+def all_components() -> List[Type["Component"]]:
+    """Get a list of all created [`Component`](../api#django_components.Component) classes."""
+    components: List[Type["Component"]] = []
+    for comp_ref in ALL_COMPONENTS:
+        comp = comp_ref()
+        if comp is not None:
+            components.append(comp)
+    return components
 
 
 @dataclass(frozen=True)
--- a/src/django_components/component_media.py
+++ b/src/django_components/component_media.py
@@ -398,7 +398,11 @@ def _get_comp_cls_attr(comp_cls: Type["Component"], attr: str) -> Any:
     return None
 
 
-media_cache: Dict[Type["Component"], MediaCls] = {}
+# NOTE: We use weakref to avoid issues with lingering references.
+if sys.version_info >= (3, 9):
+    media_cache: WeakKeyDictionary[Type["Component"], MediaCls] = WeakKeyDictionary()
+else:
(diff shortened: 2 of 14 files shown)


src/django_components/component.py
```diff
@@ -111,7 +111,7 @@
 
 # NOTE: `ReferenceType` is NOT a generic pre-3.9
 if sys.version_info >= (3, 9):
-    AllComponents = List[ReferenceType["Component"]]
+    AllComponents = List[ReferenceType[Type["Component"]]]
```
JuroOravec, line 114: Fixed a typo

src/django_components/slots.py
```diff
@@ -56,9 +56,6 @@ class SlotFunc(Protocol, Generic[TSlotData]):
     def __call__(self, ctx: Context, slot_data: TSlotData, slot_ref: "SlotRef") -> SlotResult: ...  # noqa E704
 
 
-SlotContent = Union[SlotResult, SlotFunc[TSlotData], "Slot[TSlotData]"]
```
JuroOravec, line 59: Needed to move this to avoid using the forward reference (AKA type inside a string like `"Slot[TSlotData]"`), as that was causing some issues with the weakrefs. Don't remember 100% what it was.

src/django_components/util/testing.py
```diff
@@ -265,7 +265,7 @@ def test_context_behavior(components_settings):
     """  # noqa: E501
 
     def decorator(func: Callable) -> Callable:
-        if isinstance(func, type):
+        if isinstance(func, type) and func.__name__.lower().startswith("test"):
```
JuroOravec, line 268: The `@djc_test` decorator can be applied to a test class, so one does not have to apply it to all test methods one by one.

So this:

```py
@djc_test
class TestSomething:
    def test_something(self):
        ...
    def test_other_thing(self):
        ...
```

Instead of this:

```py
class TestSomething:
    @djc_test
    def test_something(self):
        ...
    @djc_test
    def test_other_thing(self):
        ...
```

I also made it such that it is recursively applied to nested test classes like:

```py
@djc_test
class TestSomething:
    class TestNested:
        def test_other_thing(self):
            ...
```

Now, I had to update the `@djc_test` decorator, so that it recursively applies itself to the nested classes only if they start with `test`.

The issue was that, in some tests, the component classes are defined as nested classes:

```py
@djc_test
class TestSomething:
    class MyComponent(Component):
        ...
```

And `@djc_test` tried to walk down the `MyComponent` class.

src/django_components/component_media.py
```diff
@@ -398,7 +399,11 @@ def check_pair_empty(inline_attr: str, file_attr: str) -> bool:
     return None
 
 
-media_cache: Dict[Type["Component"], MediaCls] = {}
+# NOTE: We use weakref to avoid issues with lingering references.
+if sys.version_info >= (3, 9):
```
JuroOravec, line 403: Updated so that `media_cache` does not hold a strong reference to the Component classes

EmilStenstrom (approved):
Looks good to me!

JuroOravec:
@EmilStenstrom How did you do the Copilot review? :D

EmilStenstrom:
@JuroOravec it just showed up as a possible reviewer in the web ui list, and when I requested I got it a few minutes later :D

PR #1041, opened 2025-03-18
feat: add a command to list all components

JuroOravec:
Add the `components list` command that lists all components like so:

```txt
full_name                                                     path
==================================================================================================
project.pages.project.ProjectPage                             project/pages/project
project.components.dashboard.ProjectDashboard                 project/components/dashboard
project.components.dashboard_action.ProjectDashboardAction    project/components/dashboard_action
```

Also the `components ext list` command was updated to use the same underlying logic as this new list command.

I thought I'd get this one in before releasing the extensions feature.

Closes https://github.com/django-components/django-components/issues/1024

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,14 +4,9 @@
 
 ### <table><td>[Read the full documentation](https://django-components.github.io/django-components/latest/)</td></table>
 
-<!-- TODO - Remove this banner after a month(?), so March 2025 -->
-> ⚠️ Attention ⚠️ - We migrated from `EmilStenstrom/django-components` to `django-components/django-components`.
->
-> **Repo name and documentation URL changed. Package name remains the same.**
->
-> Report any broken links links in [#922](https://github.com/django-components/django-components/issues/922).
+`django-components` is a modular and extensible UI framework for Django.
 
-`django-components` combines Django's templating system with the modularity seen
+It combines Django's templating system with the modularity seen
 in modern frontend frameworks like Vue or React.
 
 With `django-components` you can support Django projects small and large without leaving the Django ecosystem.
--- a/src/django_components/commands/list.py
+++ b/src/django_components/commands/list.py
@@ -1,0 +1,221 @@
+import os
+from typing import Any, Dict, List, Optional, Type
+
+from django_components.component import all_components
+from django_components.util.command import CommandArg, ComponentCommand
+from django_components.util.misc import get_import_path, get_module_info
+
+
+# This descriptor generates the list of command arguments (e.g. `--all`), such that
+# their descriptions are dynamically generated based on the Command class.
+class ListArgumentsDescriptor:
+    # This will be called when accessing `ListCommand.arguments`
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -4,14 +4,9 @@
 
 ### <table><td>[Read the full documentation](https://django-components.github.io/django-components/latest/)</td></table>
 
-<!-- TODO - Remove this banner after a month(?), so March 2025 -->
```
JuroOravec, line 7: Removed the warning since it's been up for around 1.5 months already.

src/django_components/commands/list.py
```diff
@@ -0,0 +1,221 @@
...
+                ["-s", "--simple"],
+                action="store_true",
+                help="Only show table data, without headers. Use this option for generating machine-readable output.",
+            ),
+        ]
+
+
+# Common base class for all `list` commands, so they all have the same arguments
+# and same formatting.
+class ListCommand(ComponentCommand):
```
JuroOravec, line 46: Both the `components list` and `components ext list` use the same underlying logic, defined by this `ListCommand`. 

The two commands differ only in 1. What data is displayed, and 2. What columns are shown. This is defined on the respective subclasses with `columns`, `default_columns`, and `get_data()`.

src/django_components/commands/list.py
```diff
@@ -0,0 +1,221 @@
...
+            ),
+            CommandArg(
+                ["--columns"],
+                help=(
+                    f"Comma-separated list of columns to show. Available columns: {all_cols_readable}. "
+                    f"Defaults to `--columns {default_cols_input}`."
+                ),
+            ),
+            CommandArg(
+                ["-s", "--simple"],
```
JuroOravec, line 37: One difference from what I described in https://github.com/django-components/django-components/issues/1024 is that I actually used `-s/--simple` instead of `-q/--quiet` to hide the table headers. Because I realized that `-q` is more commonly used when one wants to hide all outputs, which is not the case here.

JuroOravec:
@EmilStenstrom Another day, another PR ready 😄

tests/test_command_ext.py
```diff
@@ -91,40 +91,57 @@ def test_root_command(self):
...
+    )
+    def test_list_extra_extensions(self):
         out = StringIO()
         with patch("sys.stdout", new=out):
-            call_command("components", "ext", "list", "-v", "0")
+            call_command("components", "ext", "list")
         output = out.getvalue()
 
-        assert "view\n" == output
+        assert output.strip() == "name \n=====\nview \nempty\ndummy"
```
EmilStenstrom, line 111: Nit: Why is there a space after name and view here?

EmilStenstrom (approved):
Looks really good! <3

tests/test_command_ext.py
JuroOravec: The rows are formatted to have the same width as the longest row.

PR #1043, opened 2025-03-19
feat: allow globs when specifynigg additionall JS and CSS

JuroOravec:
Allow to use glob patterns in the `Media.js` and `Media.css`, e.g.

```py
class MyTable(Component):
    class Media:
        js = "*.js"
        css = "*.css"
```

This works only for the JS / CSS in the Media class. `Component.css_file` and `Component.js_file` still need to be exact paths.

Closes https://github.com/django-components/django-components/issues/995


src/django_components/component_media.py
```diff
@@ -730,96 +753,157 @@ def _resolve_component_relative_files(
...
+        comp_media.js_file = resolve_relative_media_file(comp_media.js_file, False)[0]
     if getattr(comp_media, "css_file", None):
-        comp_media.css_file = resolve_media_file(comp_media.css_file)
+        comp_media.css_file = resolve_relative_media_file(comp_media.css_file, False)[0]
 
     if hasattr(comp_media, "Media") and comp_media.Media:
-        _map_media_filepaths(comp_media.Media, resolve_media_file)
-
-
-def _get_dir_path_from_component_path(
```
JuroOravec, line 796: Renamed `_get_dir_path_from_component_path` to `_find_component_dir_containing_file`, and moved some logic out, so it's clearer what's happening.

src/django_components/component_media.py
```diff
@@ -683,26 +699,33 @@ def _is_media_filepath(filepath: Any) -> bool:
     return False
 
 
-def _normalize_media_filepath(filepath: ComponentMediaInputPath) -> Union[str, SafeData]:
-    if callable(filepath):
-        filepath = filepath()
+def _normalize_media_filepath(filepaths: Sequence[ComponentMediaInputPath]) -> List[Union[str, SafeData]]:
```
JuroOravec, line 702: Before, we went file path by file path, resolving it relative to the component file if possible.

Some functions like this `_normalize_media_filepath` received only a single path, and returned it mapped.

But with support of globs, a single glob path may be expanded to multiple file paths. So some changes are to the mapping logic. And e.g. this `_normalize_media_filepath` now receives a whole list, and returns a whole list.

src/django_components/component_media.py
```diff
@@ -730,96 +753,157 @@ def _resolve_component_relative_files(
...
-        logger.debug(
-            f"Interpreting template '{repr(filepath)}' of component '{module_name}'"
-            " relatively to components directory"
+        return resolved_filepaths
+
+    # Check if filepath is a glob pattern that points to files relative to the components directory
+    # (as defined by `COMPONENTS.dirs` and `COMPONENTS.app_dirs` settings) in which the component is defined.
+    # If yes, modify the path to refer to the globbed files.
+    # If not, don't modify anything.
+    def resolve_static_media_file(
```
JuroOravec, line 814: This bit is new, it's called from [here](https://github.com/django-components/django-components/pull/1043/files#diff-bdb1917504a9c50a51e7fdf49498d5db3ce0233be9eb81b13938d4b3dcd5efc3R849).

What it boils down to is that the glob patttern may or may not be relative to the component file. E.g. if I have a component defined in `components/table.py`, both approaches below would point to the same file:

Relative to the component:

```py
class MyTable(Component):
    class Media:
        js = "*.js"
```

Relative to the `components` directory:

```py
class MyTable(Component):
    class Media:
        js = "table/*.js"
```

But to figure out which one it is, we have to test the globs. So for globs the flow is like so:
1. Try to resolve the glob as path relative to the component file.
2. If there are any matches, glob is assumed to be relative to the component.
3. Otherwise, check if the glob is relative to the root dictionary that encompases the components, like in this case `components/`
4. If match, use those files.
5. If still no match, leave the path string as is.

JuroOravec:
@EmilStenstrom

EmilStenstrom (approved):
I think the tradeoffs you descibe makes sense. Looks good to me!

DNIIBOY:
I think this might have broken something.. Version 0.132 breaks something in the Media class of a component. In my case:
```python
    class Media:
        js = ["https://cdnjs.cloudflare.com/ajax/libs/Chart.js/3.0.2/chart.min.js"]
```
This works completely as intended up to 0.131, but in 0.132 it tries to request the resource from local staticfiles instead (`localhost:8000/static/dashboards/entrance_overview/https%3A/cdnjs.cloudflare.com/ajax/libs/Chart.js/3.0.2/chart.min.js`)

PR #1045, opened 2025-03-19
ci: Make a separate PR when pushing release benchmark data

JuroOravec:
Fix (hopefully) for https://github.com/django-components/django-components/issues/1044.

The issue is that `master` is a protected branch, and the CI bot doesn't have permissions to push to that branch.

One potential workaround is to use the `gh` CLI tool to create a PR with the changes, and also push it, before updating the docs.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -46,11 +46,6 @@ jobs:
 
       - name: Configure git
         run: |
-          # Get the master branch so we can run benchmarks on it
-          git remote add upstream https://github.com/${{ github.repository }}.git
-          git fetch origin master:master
-          git checkout master
-
           # required for "mike deploy" command below which pushes to gh-pages
           git config user.name github-actions
           git config user.email github-actions@github.com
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -46,11 +46,6 @@ jobs:
 
       - name: Configure git
         run: |
-          # Get the master branch so we can run benchmarks on it
```
JuroOravec, line 49: Moved this so that we check out to `master` only when generating the benchmarks. Otherwise there was an error popping up, see https://github.com/django-components/django-components/actions/runs/13942214025/job/39021265205

.github/workflows/docs.yml
```diff
@@ -117,14 +117,30 @@ jobs:
...
+          # 2. Push to the new branch
+          echo "Pushing benchmark results to new branch..."
+          git push -u origin "benchmark-${TAG}"
+
+          # 3. Create a PR
+          PR_URL="$(gh pr create --title "chore: Add benchmark results for ${TAG}" --body "Add benchmark results for ${TAG}" --base master)"
+          echo "PR URL: ${PR_URL}"
+
+          # 4. Merge the PR
+          gh pr merge "${PR_URL}" --admin --squash --delete-branch
```
JuroOravec, line 137: Using `gh pr create` and `gh pr merge` to push the changes to `master`, as seen [here](https://github.com/krlmlr/pr-bots/blob/main/.github/workflows/write-file-branch-merge.yaml)

PR #1054, opened 2025-03-21
refactor: display verbose output for benchmark comparison

JuroOravec:
[This workflow](https://github.com/django-components/django-components/actions/runs/13979989243/job/39160267799?pr=1043) failed, and I can't see why because there's no output to the terminal from the command that does the comparison (`asv compare`)*.

So this PR makes the comparison command run twice - once with `--verbose` flag, which will be printed out to the terminal, and once without, which will print out the actual results and save them as before.

\* Note: the `No module named 'libmambapy'` in the workflow output is a red herring

### DIFF
--- a/.github/workflows/pr-benchmark-generate.yml
+++ b/.github/workflows/pr-benchmark-generate.yml
@@ -89,14 +89,17 @@ jobs:
         DJC_BENCHMARK_QUICK=1 asv run HEAD^! -v
         echo "Benchmarks for HEAD DONE."
 
-        # Compare against master
-        echo "Comparing benchmarks..."
+        echo "Creating pr directory..."
         mkdir -p pr
-        asv compare upstream/master HEAD --factor 1.1 --split > ./pr/benchmark_results.md
-
         # Save the PR number to a file, so that it can be used by the next step.
         echo "${{ github.event.pull_request.number }}" > ./pr/pr_number.txt
 
+        # Compare against master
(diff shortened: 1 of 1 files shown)


JuroOravec:
CC @EmilStenstrom This fixes the benchmark on PR workflow.

.github/workflows/pr-benchmark-generate.yml
```diff
@@ -83,20 +84,23 @@ jobs:
...
         mkdir -p pr
-        asv compare upstream/master HEAD --factor 1.1 --split > ./pr/benchmark_results.md
-
         # Save the PR number to a file, so that it can be used by the next step.
         echo "${{ github.event.pull_request.number }}" > ./pr/pr_number.txt
 
+        # Compare against master
+        # NOTE: The command is run twice, once so we can see the debug output, and once to save the results.
+        echo "Comparing benchmarks... (debug)"
+        asv compare upstream/master HEAD --factor 1.1 --split --machine ${MACHINE} --verbose
```
JuroOravec, line 101: The bug was that the `asv compare` command threw error as it needed to be specified which data set to use. I fixed this by specifying the `--machine`.

This problem popped up after the release of v0.131, which is the first release that is actually benchmarked. So after v0.131 was released, we saved the benchmark data for that release, and it means that, when afterwards the PR benchmarking workflow ran, it all of a sudden found new benchmarking datasets in the `.asv` directory. And so the `asv compare` command didn't know which one to use, and needed to be explicitly said.

Issue #1062, opened 2025-03-23
CDN / content not hosted in staticfiles not working since 0.132


DNIIBOY:
I have a component like this, which has been working fine in all version up until 0.132:
```python
@register("my_component")
class MyComponent(Component):
    template_name = "my_component.html"
    js_file = "my_component.js"

    class Media:
        js = ["https://cdnjs.cloudflare.com/ajax/libs/Chart.js/3.0.2/chart.min.js"]
```

After version 0.132, the `<script>` tag now tries to the content from a staticfile, instead of using the provided cdn, as it thinks the path is a local path:
```html
<script src="/static/components/my_component/https%3A/cdnjs.cloudflare.com/ajax/libs/Chart.js/3.0.2/chart.min.js"></script>
```
This seems to be related to #1043

JuroOravec:
@DNIIBOY Thanks for reporting! This is now fixed in [0.133](https://github.com/django-components/django-components/releases/tag/0.133).

PR #1063, opened 2025-03-23
Simple fix for #1062

DNIIBOY:
I have not had the chance to get the tests running etc, but this fix does solve the issue, even if it may require some tweaking.


JuroOravec:
Sorry I didn't noticed your PR, I also wrote a fix in https://github.com/django-components/django-components/pull/1064 😅 That one's now merged and released as part of 0.133, so we can close this one.

PR #1064, opened 2025-03-23
refactor: fix use of URLs in Component.Media

JuroOravec:
Fixes https://github.com/django-components/django-components/issues/1062

### DIFF
--- a/src/django_components/component_media.py
+++ b/src/django_components/component_media.py
@@ -865,9 +865,14 @@ def resolve_media_file(
 
     filepath_abs_or_glob = os.path.join(media_root_dir, filepath)
 
+    # If the path is a URL, don't resolve it
+    # (e.g. https://cdnjs.cloudflare.com/ajax/libs/Chart.js/3.0.2/chart.min.js)
+    # This is defined based on Django's `Media.absolute_path()` method.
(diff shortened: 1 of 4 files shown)


src/django_components/component_media.py
```diff
@@ -865,9 +865,14 @@ def resolve_media_file(
 
     filepath_abs_or_glob = os.path.join(media_root_dir, filepath)
 
+    # If the path is a URL, don't resolve it
+    # (e.g. https://cdnjs.cloudflare.com/ajax/libs/Chart.js/3.0.2/chart.min.js)
+    # This is defined based on Django's `Media.absolute_path()` method.
+    is_url_path = filepath.startswith(("http://", "https", "/"))
```
JuroOravec, line 871: The fix is simply to check if a path is URL, and skip the glob resolution if so.

JuroOravec:
CC @EmilStenstrom

DNIIBOY:
Hi guys, thanks pushing this so quickly, the problem does however still exist, and this does not fix it. This change only disables glob searching of the url, but doesn't actually treat it as a url.

DNIIBOY:
I have opened a pr which solves this: https://github.com/django-components/django-components/pull/1065

PR #1065, opened 2025-03-23
fix fix use of URLs in Component.Media 

### LINKED ISSUE #1062: CDN / content not hosted in staticfiles not working since 0.132

DNIIBOY:
#1064 Did not correctly fix #1062, this PR correctly does that

### DIFF
--- a/src/django_components/component_media.py
+++ b/src/django_components/component_media.py
@@ -839,7 +839,7 @@ def _resolve_component_relative_files(
                 # If the path is a URL, don't resolve it
                 # (e.g. https://cdnjs.cloudflare.com/ajax/libs/Chart.js/3.0.2/chart.min.js)
                 # This is defined based on Django's `Media.absolute_path()` method.
-                is_url_path = isinstance(f, str) and f.startswith(("http://", "https", "/"))
+                is_url_path = isinstance(f, str) and f.startswith(("http://", "https://", "/"))
                 if is_url_path:
                     return [f]
                 return resolve_static_media_file(f, True)
--- a/src/django_components/component_media.py
+++ b/src/django_components/component_media.py
@@ -832,11 +832,19 @@ def _resolve_component_relative_files(
         comp_media.css_file = resolve_relative_media_file(comp_media.css_file, False)[0]
 
     if hasattr(comp_media, "Media") and comp_media.Media:
+        def _resolve_media_file(
+            filepaths: Sequence[Union[str, SafeData]],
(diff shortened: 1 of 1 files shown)


src/django_components/component_media.py
```diff
@@ -865,14 +878,9 @@ def resolve_media_file(
...
     filepath_abs_or_glob = os.path.join(media_root_dir, filepath)
 
-    # If the path is a URL, don't resolve it
-    # (e.g. https://cdnjs.cloudflare.com/ajax/libs/Chart.js/3.0.2/chart.min.js)
-    # This is defined based on Django's `Media.absolute_path()` method.
-    is_url_path = filepath.startswith(("http://", "https", "/"))
-
     # The path may be a glob. So before we check if the file exists,
     # we need to resolve the glob.
-    if allow_glob and is_glob(filepath_abs_or_glob) and not is_url_path:
```
DNIIBOY, line 875: The old fix didn't work, no need to keep it

EmilStenstrom:
@DNIIBOY That seems to break some of the existing tests, mind taking a look and see if you can make it work?

DNIIBOY:
I fixed it, albeit with a relatively messy solution, feel free to change it if needed

EmilStenstrom:
@DNIIBOY Can you add a testcase for your fix as well? Since Juro's tests passed but your case wasn't handled, I'm worried that we'll break it again without noticing.

JuroOravec:
Hey, sorry for not getting to both of you sooner, I'm on a train with patchy network signal, and didn't have signal for the last hour 💀 But I wrote fix + updated tests + other boilerplate in https://github.com/django-components/django-components/pull/1067, as I was trying to understand why the old solution didn't work right away.

DNIIBOY:
#1067 Seems to be the better fix for this

PR #1066, opened 2025-03-23
feat: granular handling of class and style in {% html_attrs %}

JuroOravec:
I don't remember in which issue / discussion I've mentioned this, but some time ago I described the idea of having a cleaner way to define the `class` HTML attribute.

I based it on [Vue's implementation](https://vuejs.org/guide/essentials/class-and-style), where:
- `class` can be given as a list to merge class names from multiple sources
- Class names in `class` can be given as:
   - plain string `class="class-a class-b"`,
   - an object `:class={"class-a": true, "class-b": false}`,
   - or a list of the former `:class=[{"class-a": my_var}, "extra-class"]`

   If it's an object, the keys are the class names, and the values determine whether the individual class names will be rendered. The values that are falsy are omitted.

   This makes it easy for template variables to be used to toggle the classes.

- The `style` HTML attribute has almost the same behaviour - the value can be:
   - a string `style="color: red; width: 200px;"`,
   - an object `:style={"color": "red", "width": "200px"}`,
   - or a list of the former `:style=[{"color": "red"}, "width: 200px;"]`

   The only difference is that the object contains styling properties, not booleans.

So, bringing that feature to Django, it looks like this:

`class` HTML attribute:

```django
{% html_attrs
    class="foo"
    class=[
        "bar",
        {"baz": True, "foo": False},
        ["extra"],
    ]
%}
```

Renders:

```html
<div class="bar baz extra"></div>
```

`style` HTML attribute:

```django
{% html_attrs
    style="text-align: center;",
    style=[
        "background-color: blue; width: 200px;",
        {"background-color": "green", "color": None, "width": False},
        ["position: absolute", {"height": "12px"}],
    ]
%}
```

Renders:

```html
<div style="text-align: center; background-color: green; position: absolute; height: 12px;"></div>
```


EmilStenstrom (approved):
This looks good to me. Might be that this is a little too "Vue:y" for django-components, but I trust your judgement that this is a nice convenience when writing real projects. Go head!

I've also requested a review by Copilot.

JuroOravec:
@EmilStenstrom On the Vue-ness, we could simplify this by removing the support for lists, e.g.:

```django
{% html_attrs
    class="foo"
    class=[
        "bar",
        {"baz": True, "foo": False},
        ["extra"],
    ]
%}
```

Could be rewritten without lists as:

```django
{% html_attrs
    class="foo"
    class="bar"
    class={"baz": True, "foo": False},
    class="extra",
%}
```

What do you think?

EmilStenstrom:
@JuroOravec you should decide on what is useful in practice, when using the library. Agree that the updated version is less Vue:y but if that makes it less useful I’m not sure…

Your call.

JuroOravec:
@EmilStenstrom Since we allow to pass the same attribute (e.g. `class`) multiple times, it's still possible to achieve the same even without explictly supporting lists.

But the implementation is simpler if it handles also lists, so I went  with a middle ground - documentation for the `{% html_attrs %}` doesn't mention lists, and instead suggests to define the `class` / `style` attribute multiple times. But inernally the logic handles also lists.

src/django_components/attributes.py
```diff
@@ -100,19 +121,321 @@ def attributes_to_string(attributes: Mapping[str, Any]) -> str:
...
+
+# Match CSS comments `/* ... */`
+style_comment_re = re.compile(r"/\*.*?\*/", re.DOTALL)
+# Split CSS properties by semicolon, but not inside parentheses
+list_delimiter_re = re.compile(r";(?![^(]*\))", re.DOTALL)
+# Split CSS property name and value
+property_delimiter_re = re.compile(r":(.+)", re.DOTALL)
+
+
+def parse_string_style(css_text: str) -> StyleDict:
```
JuroOravec, line 414: When users defines a `class` or a `style` attribute to the `{% html_attrs %}`, the strings are actually first parsed into objects.

E.g. in

```django
{% html_attrs
    class="foo bar"
    class={"baz": True, "faz": False}
    style="color: red; width: 120px"
    style={"height": "10px", "display": "flex"}
%}
```

The `class="foo bar"` and `style="color: red; width: 120px"` are parsed, so it's  as if we were given:

```django
{% html_attrs
    class={"foo": True, "bar": True}
    class={"baz": True, "faz": False}
    style={"color": "red", "width": "120px"}
    style={"height": "10px", "display": "flex"}
%}
```

This is what makes it possible to merge these from multiple instances of `class` / `style`, because we can just merge the dictionaries, with later fields overriding the earlier.

Parsing the class is simple, it's just splitting the string at whitespace. For the style definition, the string is split at semicolons (`;`) to get individual properties, and then at colon (`:`) to split that into the property name and value.

All of this is based on Vue's implementation here https://github.com/vuejs/core/blob/d65b25cdda4c0e7fe8b51e000ecc3696baad0492/packages/shared/src/normalizeProp.ts

EmilStenstrom:
@JuroOravec Great solution! :)

PR #1067, opened 2025-03-23
refactor: fix component media URLs

JuroOravec:
Further fix for https://github.com/django-components/django-components/issues/1062, as raised in https://github.com/django-components/django-components/pull/1065.

@DNIIBOY Could you test this PR locally to see if it works for your case? IMO it should work, but better to double check.

And sorry for making yet another PR, I don't want to modify other people's PRs (unless it's tiny things) to avoid confusion.

Supersedes https://github.com/django-components/django-components/pull/1065.

### DIFF
--- a/src/django_components/component_media.py
+++ b/src/django_components/component_media.py
@@ -868,7 +868,11 @@ def resolve_media_file(
     # If the path is a URL, don't resolve it
     # (e.g. https://cdnjs.cloudflare.com/ajax/libs/Chart.js/3.0.2/chart.min.js)
     # This is defined based on Django's `Media.absolute_path()` method.
-    is_url_path = filepath.startswith(("http://", "https", "/"))
+    is_url_path = filepath.startswith(("http://", "https://", "://", "/"))
+
+    # If the path is a URL, don't resolve it
+    if is_url_path:
+        return [filepath], False
 
     # The path may be a glob. So before we check if the file exists,
     # we need to resolve the glob.
--- a/src/django_components/component_media.py
+++ b/src/django_components/component_media.py
@@ -876,7 +876,7 @@ def resolve_media_file(
 
     # The path may be a glob. So before we check if the file exists,
     # we need to resolve the glob.
-    if allow_glob and is_glob(filepath_abs_or_glob) and not is_url_path:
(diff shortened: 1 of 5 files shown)


DNIIBOY:
This works in my case, and is a better solution than #1065

src/django_components/component_media.py
```diff
@@ -868,7 +868,11 @@ def resolve_media_file(
     # If the path is a URL, don't resolve it
     # (e.g. https://cdnjs.cloudflare.com/ajax/libs/Chart.js/3.0.2/chart.min.js)
     # This is defined based on Django's `Media.absolute_path()` method.
-    is_url_path = filepath.startswith(("http://", "https", "/"))
+    is_url_path = filepath.startswith(("http://", "https://", "://", "/"))
+
+    # If the path is a URL, don't resolve it
+    if is_url_path:
```
DNIIBOY, line 874:
```python
if allow_glob and is_glob(filepath_abs_or_glob) and not is_url_path:
```
We can remove the `and not is_url_path` from the following if statement for clarity

tests/test_component_media.py
```diff
@@ -415,41 +415,19 @@ def test_glob_pattern_relative_to_root_dir(self):
         assertInHTML('<script src="glob/glob_2.js"></script>', rendered)
 
     def test_glob_pattern_does_not_break_urls(self):
-        class MyComponent(Component):
```
JuroOravec, line 418: This is why the edge case slipped the test - When a component is defined within a test file, the test file is not within the directories in `COMPONENTS.dirs`, and so the file paths are not even attempted to be resolved.

So after I moved the component class to be defined in a separate file, which is within `COMPONENTS.dirs`, then that triggered to try to resolve the file paths relative to the component file, which revealed the bug.

src/django_components/component_media.py
JuroOravec: Good catch, thanks!

JuroOravec:
cc @EmilStenstrom

JuroOravec:
@DNIIBOY Thanks again for raising this and your help! This is now released in [0.134](https://github.com/django-components/django-components/releases/tag/0.134)

PR #1072, opened 2025-03-24
feat: allow to set defaults

JuroOravec:
This PR adds the option to set defaults on Components, as described in https://github.com/django-components/django-
components/issues/881

Closes https://github.com/django-components/django-components/issues/881

Sreenshots:

Screenshot 2025-03-24 at 20 18 30

Screenshot 2025-03-24 at 22 18 56

### DIFF
--- a/src/django_components/app_settings.py
+++ b/src/django_components/app_settings.py
@@ -750,9 +750,10 @@ class InternalSettings:
         )
 
         # Prepend built-in extensions
+        from django_components.extensions.defaults import DefaultsExtension
         from django_components.extensions.view import ViewExtension
 
-        extensions = [ViewExtension] + list(extensions)
+        extensions = [DefaultsExtension, ViewExtension] + list(extensions)
 
         # Extensions may be passed in either as classes or import strings.
         extension_instances: List["ComponentExtension"] = []
--- a/src/django_components/extensions/defaults.py
+++ b/src/django_components/extensions/defaults.py
@@ -1,0 +1,157 @@
+import sys
+from dataclasses import MISSING, Field, dataclass
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, NamedTuple, Optional, Type
+from weakref import WeakKeyDictionary
+
+from django_components.extension import ComponentExtension, OnComponentClassCreatedContext, OnComponentInputContext
+
+if TYPE_CHECKING:
+    from django_components.component import Component
+
+
+# NOTE: `WeakKeyDictionary` is NOT a generic pre-3.9
+if sys.version_info >= (3, 9):
+    ComponentDefaultsCache = WeakKeyDictionary[Type["Component"], List["ComponentDefaultField"]]
+else:
+    ComponentDefaultsCache = WeakKeyDictionary
+
+
+defaults_by_component: ComponentDefaultsCache = WeakKeyDictionary()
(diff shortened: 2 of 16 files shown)


src/django_components/app_settings.py
```diff
@@ -750,9 +750,10 @@ def _prepare_extensions(self, new_settings: ComponentsSettings) -> List["Compone
         )
 
         # Prepend built-in extensions
+        from django_components.extensions.defaults import DefaultsExtension
         from django_components.extensions.view import ViewExtension
 
-        extensions = [ViewExtension] + list(extensions)
+        extensions = [DefaultsExtension, ViewExtension] + list(extensions)
```
JuroOravec, line 756: I've implemented the feature as an extension that modifies the given kwargs. Ngl, it's very satisfying that this extensions API can be used even for features like this :D

src/django_components/extensions/defaults.py
```diff
@@ -0,0 +1,157 @@
...
+    ComponentDefaultsCache = WeakKeyDictionary[Type["Component"], List["ComponentDefaultField"]]
+else:
+    ComponentDefaultsCache = WeakKeyDictionary
+
+
+defaults_by_component: ComponentDefaultsCache = WeakKeyDictionary()
+
+
+@dataclass
+class Default:
```
JuroOravec, line 23: In the convo on https://github.com/django-components/django-components/issues/881, there was also the question of how to handle factory functions - e.g. if I wanted to set the default to some object, like a list or a dictionary instance.

One option was to treat all functions as factories. But I think this would be more error-prone - user would have to first learn about how the defaults are handled before they could realize that, if they provided a function, that they need to "escape" it in another function:

```py
class MyComponent(Component):
    class Defaults:
        selected_items = lambda: some_callback
```

Also, if there was some simple inlined function, it wouldn't be clear if the function is there because it's a factory, or because the function is meant to be passed as is.

So I went for the approach that the dataclass module has with the `field()` function.

In other words, all defaults are treated as plain values, even functions, and a field is understood to be a factory function only when user opts into that with `Default()` helper class:

```py
from django_components import Component, Default

class MyComponent(Component):
    class Defaults:
        # Plain value doesn't need a factory
        position = "left"
        # Lists and dicts need to be wrapped in `Default`
        # Otherwise all instances will share the same value
        selected_items = Default(lambda: [1, 2, 3])
```

---

I was also considering using the dataclass' `field()` directly. But the `field()` has extra attributes that are unrelated to our use case, for us only the `default_factory` attribute is relevant.

So that's why I went with the custom `Default()` class.

But also, since the `Default()` and `field()` are so similar, I added logic [here](https://github.com/django-components/django-components/pull/1072/files#diff-388d2feafaadbdc537b53563a3a72dc3f570999a84f2595f9f7517f2566447c5R71) to also handle if user decides to (re)use the dataclass `field()`.

JuroOravec:
@EmilStenstrom Here comes another one! 😄

JuroOravec:
@EmilStenstrom I'll go ahead and merge this, but feel free to leave any comments!

EmilStenstrom:
@JuroOravec Was away on vacation, sorry for the delay. No comments, looks good!

JuroOravec:
@EmilStenstrom Welcome back :)

PR #1082, opened 2025-03-31
refactor: remove input validation and link to it

JuroOravec:
This PR removes the component input validation from this core codebase. Instead, input validation will be an optional extension people can install with `djc-ext-pydantic` (PR for that is here https://github.com/django-components/djc-ext-pydantic/pull/6).

The problem with the old implementation was that it didn't support validating nested structures (e.g. TypedDict inside another, or inside a Tuple). Instead, it's simpler to delegate that to Pydantic. And to keep the core library independent of Pydantic, it was all moved out as an extension.


JuroOravec:
@EmilStenstrom Please see this and https://github.com/django-components/djc-ext-pydantic/pull/6, they go hand-in-hand to move the input validation out of django-components, and into a separate package.

EmilStenstrom (approved):
Really like this kind of change, moving things out in favour of plugins.

PR #1086, opened 2025-04-05
refactor: fix ext URLs lookup

JuroOravec:
Small fixes to be part of v0.136:

- Renamed the module import path of djc-ext-pydantic from `djc_ext_pydantic` to `djc_pydantic`
- Trigger `URLResolver._populate()` when we are registering extensions to allow the URLs introduced by extensions to be looked up with `reverse()`


JuroOravec:
cc @EmilStenstrom

PR #1088, opened 2025-04-05
feat: component URL

JuroOravec:
This PR allows components to be assigned their own unique URL (as [first mentioned here](https://github.com/django-components/django-components/issues/897#issuecomment-2750511976)) by setting `Component.Url.public = True`. The component URL can then be obtained with `get_component_url(Component)`.

```py
from django_components import Component, get_component_url

class Table(Component):
    class Url
        public = True

url = get_component_url(Table)
```

Also as part of this PR I'm suggesting to deprecate `Component.get()`, `Component.post()`, etc, and instead encourage people to use `Component.View.get()` etc ([see explanation](https://github.com/django-components/django-components/pull/1088#discussion_r2029852488))

---

Off topic, but just this Friday I was implementing an autocomplete field for my work project, and this feature would fit it really nicely 🔥


CHANGELOG.md
```diff
@@ -1,5 +1,56 @@
...
+    url = get_component_url(MyComponent)
+    ```
+
+    This way you don't have to mix your app URLs with component URLs.
+
+    Read more on [Component views and URLs](https://django-components.github.io/django-components/0.135/concepts/fundamentals/component_views_urls/).
+
+#### Deprecation
+
+- Currently, view request handlers such as `get()` and `post()` methods can be defined
```
JuroOravec, line 34: As mentioned here, as part of this PR I also suggest to softly depracate the use of `Component.get()` (and similar), and instead prefer to use `Component.View.get`, etc.

`Component.get()` was fine previously (although it was still just a shorthand for `Component.View.get()`). But with the extensions pattern added to the mix, I think the view-specific methods should be scoped within `Component.View`, to stick to the idea that extensions can do whatever they want within their class, but don't pollute the shared scope.

docs/concepts/fundamentals/component_views_urls.md
```diff
@@ -0,0 +1,128 @@
+_New in version 0.34_
```
JuroOravec, line 1: I renamed the old `components_as_views` to `component_views_urls.md`, but git didn't pick it up right. But I've updated links, the section on using components as views, and then added section on getting component URLs.

src/django_components/app_settings.py
```diff
@@ -751,9 +751,10 @@ def _prepare_extensions(self, new_settings: ComponentsSettings) -> List["Compone
 
         # Prepend built-in extensions
         from django_components.extensions.defaults import DefaultsExtension
+        from django_components.extensions.url import UrlExtension
         from django_components.extensions.view import ViewExtension
 
-        extensions = [DefaultsExtension, ViewExtension] + list(extensions)
+        extensions = [DefaultsExtension, ViewExtension, UrlExtension] + list(extensions)
```
JuroOravec, line 757: Here the URL feature is added as a built-in extension.

src/django_components/component.py
```diff
@@ -570,6 +572,19 @@ def on_render_after(self, context: Context, template: Template, content: str) ->
...
+    # #####################################
+
+    # NOTE: These are the classes and instances added by defaults extensions. These fields
+    # are actually set at runtime, and so here they are only marked for typing.
+    Defaults: Type[ComponentDefaults]
+    defaults: ComponentDefaults
+    View: Type[ComponentView]
+    view: ComponentView
+    Url: Type[ComponentUrl]
+    url: ComponentUrl
```
JuroOravec, line 586: I've added these so that the attributes / classes of the built-in extensions are recognized by type checkers.

Ideally the typing for the Python class could be dynamically extended with extra attributes defined by third parties, similar how interfaces can be extended in TypeScript.

But Python's typing system isn't that advanced. So the next best thing I can do is to explicitly type the known extensions here.

src/django_components/extension.py
```diff
@@ -652,6 +661,56 @@ def get_extension_command(self, name: str, command_name: str) -> Type[ComponentC
                 return command
         raise ValueError(f"Command {command_name} not found in extension {name}")
 
+    def add_extension_urls(self, name: str, urls: List[URLRoute]) -> None:
```
JuroOravec, line 664: The basic idea is that when a component class is defined, this extension picks that up and registers a unique endpoint for this component.

But a component may be defined before or after Django apps were initialized.

- If a component was defined BEFORE the Django calls `AppsConfig.ready()`, then everything is all right and the component's URL is configured properly.

- But if the component is defined AFTER `AppsConfig.ready()`, then, even if we manually add the URL to `urlpatterns`, we still need to trigger `URLResolver._populate()`  in order to force Django to update its lookups, so things like `reverse()` work

And to make working with that easier, I defined the `Extensions.add_extension_urls()` and `Extensions.remove_extension_urls()`.

src/django_components/extensions/url.py
```diff
@@ -0,0 +1,180 @@
...
+    def url(self) -> str:
+        """
+        The URL for the component.
+
+        Raises `RuntimeError` if the component is not public.
+        """
+        return get_component_url(self.component.__class__)
+
+
+class UrlExtension(ComponentExtension):
```
JuroOravec, line 118: The main logic of this PR - the extension listens to when a component class is created or deleted, and creates / deletes a URL route that has the `Component.as_view()`. as the handler.

tests/test_component_url.py
```diff
@@ -0,0 +1,87 @@
...
+from django_components import Component, get_component_url
+from django_components.testing import djc_test
+
+from .testutils import setup_test_config
+
+setup_test_config({"autodiscover": False})
+
+
+@djc_test
+class TestComponentUrl:
```
JuroOravec, line 16: Main tests for this PR

docs/concepts/fundamentals/slots.md
```diff
@@ -663,3 +663,42 @@ class MyTable(Component):
     </div>
     """
 ```
+
+### Escaping slots content
```
JuroOravec, line 667: This bit about escaping slots was in the "Componens as views" page. I moved it here instead, and I've mentioned it in "Getting started" (`docs/getting_started/rendering_components.md`) too.

docs/getting_started/rendering_components.md
```diff
@@ -159,3 +159,133 @@ def my_view(request):
...
+    If you disable escaping, you should make sure that any content you pass to the slots is safe,
+    especially if it comes from user input!
+
+!!! info
+
+    If you're planning on passing an HTML string, check Django's use of
+    [`format_html`](https://docs.djangoproject.com/en/5.0/ref/utils/#django.utils.html.format_html)
+    and [`mark_safe`](https://docs.djangoproject.com/en/5.0/ref/utils/#django.utils.safestring.mark_safe).
+
+### Component views and URLs
```
JuroOravec, line 202: Added a section to getting started on how to use the component views and URLs.

JuroOravec:
@EmilStenstrom Ready for review!

JuroOravec:
@EmilStenstrom I'll go ahead and merge this, as there's few more PRs waiting in the pipeline, blocking me from working on the core. Feel free to still leave comments when you have the time!

EmilStenstrom:
@JuroOravec Happy to see you are moving quickly. I'm not sure about the public API of this. Does an URL even make sense without a View? I think that this should probably be a setting on the View class rather than a separate extensions, don't you think?

JuroOravec:
@EmilStenstrom Yeah, I'm 60/30 for/against what you're suggesting.

- On one hand, the implementation of the  `public` field depends on View class, so it would make sense for it to be within `View`.
- On the other hand, the `View` class already has the API of Django's View. So to err on the safe side I made them distinct.

But ultimately, it was that by the time I realized that it would make sense to move the `Url.public` under `View.public`, I had already written all the docs and tests, and wanted to publish it so I could use it at work.

Now that the functionality is there, moving it to `View.public` can be done as a separate step later.

EmilStenstrom:
@JuroOravec Yeah, no rush! :)

PR #1092, opened 2025-04-06
chore: add support for Django 5.2 and drop for 5.0

JuroOravec:
[Django 5.2](https://www.djangoproject.com/weblog/2025/apr/02/django-52-released/) was released on April 2nd. At the same time, Django dropped support for 5.0


JuroOravec:
cc @EmilStenstrom

PR #1093, opened 2025-04-06
docs: move extension command- and url-related API to own API pages

JuroOravec:
A bit of maintenance in the auto-generated section of the docs, because the [main API page](https://django-components.github.io/django-components/latest/reference/api/) felt too crowded:

- Moved the API reference that's related to defining extension commands and extension URLs to their own pages.
- Move the API reference for data available within extension hooks to the "Extension hooks" section

Screenshot 2025-04-06 at 13 43 21

Screenshot 2025-04-06 at 12 59 50


docs/scripts/reference.py
```diff
@@ -95,6 +95,9 @@ def gen_reference_api():
                 or _is_error_cls(obj)
                 or _is_tag_formatter_instance(obj)
                 or _is_tag_formatter_cls(obj)
+                or _is_extension_command_api(obj)
```
JuroOravec, line 98: This is added so that the entries related to extension hooks, urls, and commands are excluded from the main reference page.

src/django_components/extension.py
```diff
@@ -27,26 +27,38 @@
 ################################################
 
 
+# Mark a class as an extension hook context so we can place these in
+# a separate documentation section
+def mark_extension_hook_api(cls: Type[Any]) -> Type[Any]:
```
JuroOravec, line 32: To be able to mark which objects should be considered as related to extension hooks / urls / commands, I added little decorators like these.

JuroOravec:
cc @EmilStenstrom

PR #1094, opened 2025-04-07
feat: expose _class_hash as class_id

JuroOravec:
For the Storybook integration, I needed to uniquely identify components from the Storybook UI. So I wanted to reuse the `_class_hash` that's already available. But `_class_hash` was an internal variable. So this PR exposes it publicly as `Component.class_id`.

Together with this goes accessing the component classes based on their new class IDs. So I've also added `get_component_by_class_id()` for that. It's a bit mouthful, but since components also have the instance IDs under `Component.id`, I went with unambiguous name.

```py
from django_components import Component, get_component_by_class_id

class MyComp(Component):
    pass

assert MyComp is get_component_by_class_id(MyComp.class_id)
```

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -22,7 +22,7 @@ from typing import (
     Union,
     cast,
 )
-from weakref import ReferenceType, finalize
+from weakref import ReferenceType, WeakValueDictionary, finalize
 
 from django.core.exceptions import ImproperlyConfigured
 from django.forms.widgets import Media as MediaCls
@@ -46,7 +46,6 @@ from django_components.dependencies import (
     cache_component_css_vars,
     cache_component_js,
     cache_component_js_vars,
-    comp_hash_mapping,
     insert_component_dependencies_comment,
 )
 from django_components.dependencies import render_dependencies as _render_dependencies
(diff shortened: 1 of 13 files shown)


src/django_components/component.py
```diff
@@ -589,7 +631,18 @@ def on_render_after(self, context: Context, template: Template, content: str) ->
...
+    class_id: ClassVar[str]
+    """
+    Unique ID of the component class, e.g. `MyComponent_ab01f2`.
+
+    This is derived from the component class' module import path, e.g. `path.to.my.MyComponent`.
+    """
+
+    # TODO_V1 - Remove this in v1
+    @property
+    def _class_hash(self) -> str:
```
JuroOravec, line 643: I've kept `_class_hash` in case someone relied on it, but marked it as deprecated.

src/django_components/component.py
```diff
@@ -131,6 +132,47 @@ def all_components() -> List[Type["Component"]]:
     return components
 
 
+# NOTE: Initially, we fetched components by their registered name, but that didn't work
```
JuroOravec, line 135: This comment, as well as the `comp_cls_id_mapping` dictionary, were previously in `dependencies.py`.

JuroOravec:
cc @EmilStenstrom

EmilStenstrom:
Awesome!

PR #1097, opened 2025-04-07
feat: component caching

JuroOravec:
This makes it possible to configure caching on the component:

```py
class TestComponent(Component):
    template = "Hello"

    class Cache:
        enabled = True
        ttl = 0.1  # .1 seconds TTL
        cache_name = "custom_cache"

        def hash_args(self, args):
            # Custom hash method for args
            return json.dumps(args)

        def hash_kwargs(self, kwargs):
            # Custom hash method for kwargs
            return json.dumps(kwargs)
```

Closes https://github.com/django-components/django-components/issues/992


src/django_components/component.py
```diff
@@ -1234,6 +1238,11 @@ def _render_impl(
             )
         )
 
+        # The component rendering was short-circuited by an extension, skipping
+        # the rest of the rendering process. This may be for example a cached content.
+        if result_override is not None:
```
JuroOravec, line 1243: One of the changes to make this work is that, if an extension returns a value from `on_component_input` hook, then we short-circuit the rendering, and return this  value instead.

CHANGELOG.md
```diff
@@ -33,7 +33,32 @@
...
+            ttl = 0.1  # .1 seconds TTL
+            cache_name = "custom_cache"
+
+            def hash_args(self, args):
+                # Custom hash method for args
+                return json.dumps(args)
+
+            def hash_kwargs(self, kwargs):
+                # Custom hash method for kwargs
+                return json.dumps(kwargs)
```
EmilStenstrom, line 58: I'm not sure I like this API. Couldn't there be a `def hash(self, args, kwargs`) method to override instead?

Maybe also change the comment to some reason why you would want to override that method.

EmilStenstrom (approved):
Looks great. One small comment about the API, if we should do hash() instead of hash_args, and hash_kwargs?

CHANGELOG.md
JuroOravec: Good point, updated :)

JuroOravec:
Note: Tests are failing because PyPI is having some [outage](https://status.python.org/#day).

Screenshot 2025-04-08 at 11 04 18

oliverhaas:
Cool! Will check it out hopefully soon.

PR #1098, opened 2025-04-07
refactor: fix link in docs + call django.setup() in @djc_test if not done so

JuroOravec:
Two minor fixes:
- Fix the link to benchmarks in README.
- Call `django.setup()` from within `@djc_test` if not done so alrady

The latter was because in one of the extensions I forgot to call `django.setup()` first, and then I got an error. So for a bit better user experience we can call `django.setup()` for the user instead.


JuroOravec:
cc @EmilStenstrom

PR #1099, opened 2025-04-07
docs: mention URL auto-gen in README

JuroOravec:
Tiny docs update to mention the URL endpoint auto generation in the README in the HTML fragment section. What prompted me was that there was a positive feedback from one guy on Reddit on this feature, but I realized that it's not mentioned in the overview.


JuroOravec:
cc @EmilStenstrom

PR #1109, opened 2025-04-08
docs: update testing, single file comp, and syntax highlight

JuroOravec:
Bit of cleanup in the docs:

- Merged the page on "setting up syntax highlight for the inlined HTML / JS / CSS code" back with the page on single file component. I think I was the one who originally split it, but it feels now that it fits better together.
- Updated links and the "single file component" feels done now.
- Updated the page on testing, mentioning that users can omit `django.setup()`, in which case we run it for them automatically.

Screenshot 2025-04-07 at 21 45 39

Screenshot 2025-04-07 at 21 45 24


JuroOravec:
cc @EmilStenstrom

PR #1112, opened 2025-04-09
refactor: fix - allow components with Url.public to be defined before django.setup()

JuroOravec:


src/django_components/extension.py
```diff
@@ -707,6 +686,29 @@ def _init_app(self) -> None:
         resolver = get_resolver(urlconf)
         resolver._populate()
 
+        # Flush stored events
```
JuroOravec, line 689: Had to change the order, so the flushing of events happens after the URLResolver was updated

JuroOravec:
cc @EmilStenstrom

PR #1115, opened 2025-04-10
docs: document how to access component instance from within View

JuroOravec:
Closes https://github.com/django-components/django-components/issues/1114


JuroOravec:
cc @EmilStenstrom

PR #1121, opened 2025-04-12
refactor: fix compat with Django 5.2 Finder.find()

JuroOravec:
In Django 5.2 the `Finder.find()` attribute receives a `find_all` kwarg instead of `all` kwarg ([see changelog](https://docs.djangoproject.com/en/5.2/releases/5.2/#features-deprecated-in-5-2)). So, depending on the Django version, our `ComponentsFileSystemFinder` will use either one or another.

Closes https://github.com/django-components/django-components/issues/1119

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Release notes
 
+## v0.139
+
+#### Fix
+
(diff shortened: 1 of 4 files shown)


JuroOravec:
cc @EmilStenstrom

PR #1123, opened 2025-04-12
build: Change versioning to major.minor.patch

### LINKED ISSUE #1120: Change versioning to major.minor.patch
@EmilStenstrom Just a quick question - I'm thinking of already introducing the `major.minor.patch` versioning for django-components.

Initially I was thinking of doing that only at v1.0.

But lately it feels like there was a lot of releases that were purely to fix a bug in the previous release. So it'd be clearer if these were versioned as 0.135.1 instead of going jumping to 0.136.

JuroOravec:
Closes #1120

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,13 +4,14 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "django_components"
-version = "0.139"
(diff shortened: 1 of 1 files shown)


JuroOravec:
cc @EmilStenstrom

PR #1125, opened 2025-04-14
refactor: use typevar defaults + raise on conflicting extensions

### LINKED ISSUE #1113: Make dynamic typing declaration with TypeVar from typings_extensions
In I think either 3.12 or 3.13, it should be possible to define generics (`TypeVar`) with defaults ([docs](https://typing.python.org/en/latest/spec/generics.html#defaults-for-type-parameters)). In those case, the type parameter can be omitted from the type declaration.

Currently, when we define the type for the Component class, we have to specify all 6 types at once:

```py
Component[ArgsType, KwargsType, SlotsType, DataType, JsDataType, CssDataType]
```

which is a mouthful, because when I want to only e.g. args and kwargs, I still have to set the rest to `Any`:

```py
Component[MyArgs, MyKwargs, Any, Any, Any, Any]
```

If we added defaults, this would make the declarations shorter. Any of the following would be allowed:

```py
Component
Component[ArgsType]
Component[ArgsType, KwargsType]
Component[ArgsType, KwargsType, SlotsType]
Component[ArgsType, KwargsType, SlotsType, DataType]
Component[ArgsType, KwargsType, SlotsType, DataType, JsDataType]
Component[ArgsType, KwargsType, SlotsType, DataType, JsDataType, CssDataType]
```

Since we support Python versions all the way down to 3.8*, simply waiting until we support only Python 3.12 would take many years.

Instead, I just learnt that it's possible to install [typing-extensions](https://pypi.org/project/typing-extensions/) on older versions to back-port the new typing-related features.

I haven't checked yet, but if works also on Py3.8, that'd be really nice.

The only thing is that this would have to be declared as a package dependency. But since this is a Python's own package, I don't see an issue with that.

---

\* We should support Python 3.8 until Apr 2026 as that's when support for [Django 4.2 ends](https://endoflife.date/django).

JuroOravec:
When typing a Component, one can now specify as few or as many parameters as you want.

```py
Component[Args]
Component[Args, Kwargs]
Component[Args, Kwargs, Slots]
Component[Args, Kwargs, Slots, Data]
Component[Args, Kwargs, Slots, Data, JsData]
Component[Args, Kwargs, Slots, Data, JsData, CssData]
```

All omitted parameters will default to `Any`.

Closes #1113

### DIFF
--- a/README.md
+++ b/README.md
@@ -396,7 +396,7 @@ Components API is fully typed, and supports [static type hints](https://django-c
 To opt-in to static type hints, define types for component's args, kwargs, slots, and more:
 
 ```py
-from typing import NotRequired, Tuple, TypedDict, SlotContent, SlotFunc
+from typing import NotRequired, Tuple, TypedDict, SlotContent, Slot
 
 from django_components import Component
 
@@ -407,22 +407,19 @@ class ButtonKwargs(TypedDict):
     another: int
     maybe_var: NotRequired[int] # May be omitted
 
-class ButtonData(TypedDict):
-    variable: str
-
 class ButtonSlots(TypedDict):
-    my_slot: NotRequired[SlotFunc]
+    # Use `Slot` for slot functions.
+    my_slot: NotRequired[Slot]
+    # Use `SlotContent` when you want to allow either `Slot` instance or plain string
     another_slot: SlotContent
 
-ButtonType = Component[ButtonArgs, ButtonKwargs, ButtonSlots, ButtonData, JsData, CssData]
+ButtonType = Component[ButtonArgs, ButtonKwargs, ButtonSlots]
 
 class Button(ButtonType):
     def get_context_data(self, *args, **kwargs):
         self.input.args[0]  # int
         self.input.kwargs["variable"]  # str
-        self.input.slots["my_slot"]  # SlotFunc[MySlotData]
-
-        return {}  # Error: Key "variable" is missing
+        self.input.slots["my_slot"]  # Slot[MySlotData]
 ```
 
 When you then call
--- a/docs/concepts/advanced/extensions.md
+++ b/docs/concepts/advanced/extensions.md
@@ -169,6 +169,14 @@ class MyExtension(ComponentExtension):
         ctx.component_cls.my_attr = "my_value"
 ```
 
+!!! warning
+
+    The `name` attribute MUST be unique across all extensions.
+
+    Moreover, the `name` attribute MUST NOT conflict with existing Component class API.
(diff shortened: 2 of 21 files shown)


README.md
```diff
@@ -407,22 +407,19 @@ class ButtonKwargs(TypedDict):
...
-
 class ButtonSlots(TypedDict):
-    my_slot: NotRequired[SlotFunc]
+    # Use `Slot` for slot functions.
+    my_slot: NotRequired[Slot]
+    # Use `SlotContent` when you want to allow either `Slot` instance or plain string
     another_slot: SlotContent
 
-ButtonType = Component[ButtonArgs, ButtonKwargs, ButtonSlots, ButtonData, JsData, CssData]
+ButtonType = Component[ButtonArgs, ButtonKwargs, ButtonSlots]
```
JuroOravec, line 416: While at it I also updated the docs on typing and validation (note: that page still has work to do). And in the README, to simplify things, I changed it to show only input validation (using only the first 3 type params).

docs/concepts/advanced/extensions.md
```diff
@@ -169,6 +169,14 @@ class MyExtension(ComponentExtension):
         ctx.component_cls.my_attr = "my_value"
 ```
 
+!!! warning
+
+    The `name` attribute MUST be unique across all extensions.
```
JuroOravec, line 174: Unrelated to the linked issue, but added validation to extensions so that if there's duplicate name, or the name conflicts with any attributes or methods on the Component class, then we raise error.

src/django_components/component.py
```diff
@@ -103,12 +103,12 @@
 COMP_ONLY_FLAG = "only"
 
 # Define TypeVars for args and kwargs
-ArgsType = TypeVar("ArgsType", bound=tuple, contravariant=True)
-KwargsType = TypeVar("KwargsType", bound=Mapping[str, Any], contravariant=True)
-SlotsType = TypeVar("SlotsType", bound=Mapping[SlotName, SlotContent])
-DataType = TypeVar("DataType", bound=Mapping[str, Any], covariant=True)
-JsDataType = TypeVar("JsDataType", bound=Mapping[str, Any])
-CssDataType = TypeVar("CssDataType", bound=Mapping[str, Any])
+ArgsType = TypeVar("ArgsType", bound=tuple, contravariant=True, default=Any)
```
JuroOravec, line 106: This is what allows us to omit some of the types when configuring `Component[...]`

JuroOravec:
cc @EmilStenstrom

PR #1127, opened 2025-04-14
refactor: prefix component ID with `c`

JuroOravec:
This prefixes the Component IDs (accessible as `Component.id`) with a `c`, so that the ID always starts with a alphabet character.

See https://github.com/django-components/django-components/issues/1084#issuecomment-2779764174

I opted to add only `c` (for Components), instead of `djc`, because in HTML, the component ID is already used as `data-djc-id-xxxxxx` attribute. So adding `djc` would end up with having the `djc` in there twice.


src/django_components/constants.py
```diff
@@ -0,0 +1,3 @@
+UID_LENGTH = 6
```
JuroOravec, line 1: Defined the ID length and the prefix here in `constants.py`, so that, if we ever need to change the prefix or the length, it would be only a matter of changing these and updating the tests.

JuroOravec:
cc @EmilStenstrom

docs/concepts/fundamentals/render_api.md
```diff
@@ -51,6 +51,26 @@ multiple times, the ID will be different for each call.
...
+
+Thus, there is currently a soft-cap of ~30K components rendered on a single page.
+
+If you need to expand this limit, please open an issue on GitHub.
+
+```python
+class Table(Component):
+    def get_context_data(self, *args, **attrs):
+        # Access component's ID
+        assert self.id == "djc1A2b3c"
```
EmilStenstrom, line 69: @JuroOravec I think this should be without the "dj" prefix?

PR #1135, opened 2025-04-19
test: test to illustrate bug in component caching when using include tag

oliverhaas:
Will create an issue in a second...

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -1306,7 +1306,7 @@ class Component(
         # The component rendering was short-circuited by an extension, skipping
         # the rest of the rendering process. This may be for example a cached content.
         if result_override is not None:
-            # Cleanup here as well
+            # Cleanup needs to be done even if short-circuited
             context.render_context.pop()
             return result_override
 
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -1306,6 +1306,8 @@ class Component(
         # The component rendering was short-circuited by an extension, skipping
         # the rest of the rendering process. This may be for example a cached content.
         if result_override is not None:
+            # Cleanup here as well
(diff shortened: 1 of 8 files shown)


src/django_components/component.py
```diff
@@ -1306,6 +1306,8 @@ def _render_impl(
         # The component rendering was short-circuited by an extension, skipping
         # the rest of the rendering process. This may be for example a cached content.
         if result_override is not None:
+            # Cleanup needs to be done even if short-circuited
+            context.render_context.pop()
```
JuroOravec, line 1310: @oliverhaas Looks good to me, nice find!

I wonder if we could move the calling of `context.render_context.push()` from [line 1269](https://github.com/django-components/django-components/pull/1135/files#diff-35c84ebceae50f7149f03bbcaf2b4f1289bc2984bf7859ca521b5dbe7960c396R1269) to below this point. That way we wouldn't have to call `context.render_context.pop()`.

But also the compatibility with Django's `{% block %}` tags is a bit of a black magic, so tbh I'd leave things as they are, and use this fix.

JuroOravec (approved):

JuroOravec:
@oliverhaas Sorry forgot to comment last night, but I guess you've already seen - I pushed commits to fix the linter errors and merged main, to fix the tests.

Anyway, let me know if there's anything you want to add to the PR. Otherwise I think it's good to merge.

oliverhaas:
All good! I would be super happy about a merge, and also a release. I actually found this bug while trying to enable caching for my projects, so I'm heading to production with this :).

JuroOravec:
Alright, released in [v0.139.1](https://github.com/django-components/django-components/releases/tag/0.139.1)! 🎉

PR #1138, opened 2025-04-20
refactor: change component typing from generics to class attributes

JuroOravec:
Closes https://github.com/django-components/django-components/issues/1122, closes https://github.com/django-components/django-components/issues/1133

Changes:
- The component typing and validation will now use nested class attributes instead of generics.
- Introduced `get_template_data()` that supersedes `get_context_data()`.
- `get_js_data()` and `get_css_data()` have had their function signature updated to match `get_template_data()`
- `EmptyTuple` and `EmptyDict` helper types were removed. Instead, there's a single `Empty` class that can be used with the new typing system.
- Renamed `SlotContent` to `SlotInput` to clarify its purpose. The old name `SlotContent` will stay until v1.

See full list of changes in the [changelog](https://github.com/django-components/django-components/pull/1138/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed).

## Typing input

Before:

```py
from typing import NotRequired, Tuple, TypedDict, SlotContent, Slot
from django_components import Component

ButtonArgs = Tuple[int, str]

class ButtonKwargs(TypedDict):
    variable: str
    another: int
    maybe_var: NotRequired[int] # May be omitted

class ButtonSlots(TypedDict):
    my_slot: NotRequired[Slot]
    another_slot: SlotContent

ButtonType = Component[ButtonArgs, ButtonKwargs, ButtonSlots]

class Button(ButtonType):
    def get_context_data(self, *args, **kwargs):
        self.input.args[0]  # int
        self.input.kwargs["variable"]  # str
        self.input.slots["my_slot"]  # Slot[MySlotData]


Button.render(
    # Error: First arg must be `int`, got `float`
    args=(1.25, "abc"),
    # Error: Key "another" is missing
    kwargs={
        "variable": "text",
    },
)
```

After:

```py
from typing import NamedTuple, Optional
from django.template import Context
from django_components import Component, Slot, SlotInput

class Button(Component):
    class Args(NamedTuple):
        size: int
        text: str

    class Kwargs(NamedTuple):
        variable: str
        another: int
        maybe_var: Optional[int] = None  # May be omitted

    class Slots(NamedTuple):
        my_slot: Optional[SlotInput] = None
        another_slot: SlotInput

    def get_template_data(self, args: Args, kwargs: Kwargs, slots: Slots, context: Context):
        args.size  # int
        kwargs.variable  # str
        slots.my_slot  # Slot[MySlotData]


Button.render(
    # Error: First arg must be `int`, got `float`
    args=Button.Args(
        size=1.25,
        text="abc",
    ),
    # Error: Key "another" is missing
    kwargs=Button.Kwargs(
        variable="text",
    ),
)
```

## Typing data

SImilarly as before, it's still possible to type / validate also the data returned from `get_template_data()` and others.

The way this works is that:
1. One can specify the data types the same way they can specify the `Args` / `Kwargs` / `Slots` classes.
2. In the respective method, one can return either a plain dictionary (as before), or an instance of the matching type

```py
class Button(Component):
    class TemplateData(NamedTuple):
        size: int
        text: str

        return Button.TemplateData(
            size=kwargs["size"],
            text=kwargs["text"],
        )
```

So this adds both type hints and validation.

Alternatively, one can still return a plain dictionary. In such case, we instantiate `TemplateData` with the returned data behind the scenes to validate that returned data:

```py
class Button(Component):
    class TemplateData(NamedTuple):
        size: int
        text: str

        return {
            "size": kwargs["size"],
            "text": kwargs["text"],
        )
```

Some screenshots from the docs:

Screenshot 2025-04-19 at 09 18 52
Screenshot 2025-04-18 at 22 52 49
Screenshot 2025-04-18 at 22 53 07
Screenshot 2025-04-18 at 23 32 47


docs/concepts/fundamentals/render_api.md
```diff
@@ -15,7 +15,7 @@ Example:
 class Table(Component):
     def get_context_data(self, *args, **attrs):
         # Access component's ID
-        assert self.id == "djc1A2b3c"
+        assert self.id == "c1A2b3c"
```
JuroOravec, line 18: Leftover from the PR that prefixed the component IDs with `c`

src/django_components/util/misc.py
```diff
@@ -130,3 +131,19 @@ def is_glob(filepath: str) -> bool:
 
 def flatten(lst: Iterable[Iterable[T]]) -> List[T]:
     return list(chain.from_iterable(lst))
+
+
+def to_dict(data: Any) -> dict:
```
JuroOravec, line 136: The `Args`, `Kwargs`, etc, classes can be NamedTuple or dataclasses (among other). This utility makes sure that we're able to handle these different classes by normalizing their data to a plain dictionary.

src/django_components/util/types.py
```diff
@@ -1,144 +1,30 @@
...
-    from typing import TypedDict
-else:
-    from typing_extensions import TypedDict as TypedDict
 
-try:
-    from typing import Annotated  # type: ignore
-except ImportError:
-
-    @typing.no_type_check
-    class Annotated:  # type: ignore
```
JuroOravec, line 17: No need to redefine the `Annotated` class, since we use the `typing_extensions` package.

And no need to import `TypedDict`, since we now use `NamedTuple` or `dataclass` to define the component typing.

tests/test_component.py
```diff
@@ -385,100 +382,6 @@ def get_context_data(self):
 
         assert SimpleComponent.render() == "Hello"
 
-    def test_typing(self):
```
JuroOravec, line 388: Moved the typing to its own test file.

src/django_components/component.py
```diff
@@ -275,6 +261,10 @@ def __new__(mcs, name: Any, bases: Tuple, attrs: Dict) -> Any:
 
     # This runs when a Component class is being deleted
     def __del__(cls) -> None:
+        # Skip if `extensions` was deleted before this registry
+        if not extensions:
```
JuroOravec, line 265: This, together with [this change in `component.registry.py`](https://github.com/django-components/django-components/pull/1138/files#diff-a5dcf86b1b1bd690092e8ed60318835a1ce9ed2c62befcc05d43dfd4459393b6R263) fixes https://github.com/django-components/django-components/issues/1133

src/django_components/component.py
```diff
@@ -1023,9 +1917,9 @@ def get_context_data(self):
 
         As the `{{ data.hello }}` is taken from the "provider".
         """
-        if self.input is None:
```
JuroOravec, line 1026: This was wrong, `self.input` cannot be `None`.

src/django_components/component.py
```diff
@@ -1362,28 +2260,20 @@ def _render_impl(
         # or inspect the Context object, your screen is filled with the deeply nested ComponentContext objects.
         component_context_cache[render_id] = component_ctx
 
-        # Allow to access component input and metadata like component ID from within these hook
+        # Allow to access component input and metadata like component ID from within data methods
         with self._with_metadata(metadata):
+            template_data, js_data, css_data = self._call_data_methods(args_list, kwargs_dict, slots_dict, context)
```
JuroOravec, line 2265: Moved the calling of `get_template_data()`, `get_js_data()`, etc into `_call_data_methods()`.

src/django_components/component.py
```diff
@@ -1579,9 +2469,48 @@ def renderer(root_attributes: Optional[List[str]] = None) -> Tuple[str, Dict[str
 
         return renderer
 
+    def _call_data_methods(self, args: Any, kwargs: Any, slots: Any, context: Context) -> Tuple[Dict, Dict, Dict]:
+        # If user doesn't specify `Args`, `Kwargs`, `Slots` types, then we pass them in as plain
+        # dicts / lists.
+        args_inst = self.Args(*args) if self.Args is not None else args
```
JuroOravec, line 2475: Here's the logic that does the trick that, when one defines `Component.Args`, so that the `get_template_data()` receives an instance of the `Args` class, instead of plain dicts / lists.

JuroOravec:
Course of action:

1. https://github.com/django-components/django-components/pull/1135 should be merged first, so we can release it still as v0.139.1.
2. Then merge this one. Since there's a lot of breaking changes here, I'd hold off releasing v0.140.0, and instead get in other issues that will introduce some breaking changes:
  - https://github.com/django-components/django-components/issues/1116
  - https://github.com/django-components/django-components/issues/1096
  - https://github.com/django-components/django-components/issues/1085

3. After this PR, there will be also a separate one focusing just on documenting the `get_template_data()`, `get_js_data()`, and `get_css_data()`.

4. And there will be last one to update the tests from using `get_context_data()` to `get_template_data()`.

5. After that, we would release v0.140.0.

6. And after the release, I will also need to update the Pydantic integration, because it relied on using the Component generics.

docs/reference/commands.md
```diff
@@ -237,7 +238,7 @@ List all extensions.
 - `--columns COLUMNS`
     - Comma-separated list of columns to show. Available columns: name. Defaults to `--columns name`.
 - `-s`, `--simple`
-    - Only show table data, without headers. Use this option for generating machine-readable output.
+    - Only show table data, without headers. Use this option for generating machine- readable output.
```
EmilStenstrom, line 241: This looks like an error.

docs/reference/commands.md
```diff
@@ -400,7 +401,7 @@ List all components created in this project.
 - `--columns COLUMNS`
     - Comma-separated list of columns to show. Available columns: name, full_name, path. Defaults to `--columns full_name,path`.
 - `-s`, `--simple`
-    - Only show table data, without headers. Use this option for generating machine-readable output.
+    - Only show table data, without headers. Use this option for generating machine- readable output.
```
EmilStenstrom, line 404: This looks like an error.

src/django_components/component.py
```diff
@@ -103,14 +101,6 @@
...
-# Define TypeVars for args and kwargs
-ArgsType = TypeVar("ArgsType", bound=tuple, contravariant=True, default=Any)
-KwargsType = TypeVar("KwargsType", bound=Mapping[str, Any], contravariant=True, default=Any)
-SlotsType = TypeVar("SlotsType", bound=Mapping[SlotName, SlotContent], default=Any)
-DataType = TypeVar("DataType", bound=Mapping[str, Any], covariant=True, default=Any)
-JsDataType = TypeVar("JsDataType", bound=Mapping[str, Any], default=Any)
-CssDataType = TypeVar("CssDataType", bound=Mapping[str, Any], default=Any)
-
 
 # NOTE: `ReferenceType` is NOT a generic pre-3.9
```
EmilStenstrom, line 105: While this is not part of this PR, I think this is solved by typing_extensions?

EmilStenstrom (approved):
This looks **really** good. So nice to have the types nested inside the Component class instead of laying sprinkled around the files like before.

I'm assuming you've already considered this, but it's not possible to automatically wrap args to Button.render in Button.Args (et al) without having to be explicit about it?

So that you could do Button.render(args=[1, 2]) and that would automatically be interpreted as Button.render(args=Button.Args([1, 2]))?

src/django_components/component.py
JuroOravec: Unfortunately not, the `ReferenceType` comes from the `weakref` module.

docs/reference/commands.md
JuroOravec: Yeah, these files in `docs/reference/` are auto-generated and are overwritten when generating docs anyway, so it's kinda just noise. But if these files are totally omitted from the source code, then these pages will be missing in the documentation site. So I just include the changes after running `mkdocs serve` locally :D

JuroOravec:
> This looks really good. So nice to have the types nested inside the Component class instead of laying sprinkled around the files like before.

Yeah it actually feels really nice to write it out! Simple, no need to import too many things, no awkward syntax.

> I'm assuming you've already considered this, but it's not possible to automatically wrap args to Button.render in Button.Args (et al) without having to be explicit about it?
>
> So that you could do Button.render(args=[1, 2]) and that would automatically be interpreted as Button.render(args=Button.Args([1, 2]))?

Yeah, no, not possible. That would require generics, and those are problematic to subclass.

PR #1145, opened 2025-04-22
docs: docstrings, fundamentals, and minor changes

JuroOravec:


CHANGELOG.md
```diff
@@ -124,6 +130,29 @@
                 return self.render_to_response()
     ```
 
+- Component name in the `{% component %}` tag can no longer be set as a kwarg.
```
JuroOravec, line 133: Before it was possible to specify the component name like this:

```django
{% component "my_table"  / %}
```

or like this

```django
{% component name="my_table"  / %}
```

and the `name` kwarg could be anywhere in the declaration:

```django
{% component foo="bar" name="my_table"  / %}
```

I realized that the `name` kwarg for the component name is problematic.

The reason for that is because it *looks* like a regular kwarg, but the component name is processed at a different stage than the rest of the kwargs.

Also because if one specifies the component name without the `name` kwarg, then they can use it as a regular kwarg:

```django
{% component "button" name="John" job="Developer" / %}
```

So it was inconsistent / buggy, and would permit to something like this, where the first `name` kwarg is extracted at node parse time, while the second `name` kwarg is a normal kwarg:

```django
{% component foo="bar" name="my_table" name=my_var  / %}
```

And if we had just flipped the two, it would raise an error, because djc would think that the component name is a variable when it expects a string:

```django
{% component foo="bar" name=my_var name="my_table"  / %}
```

So IMO this is more of a bug than feature.

So the change is to allow only setting the component name as positional argument:

```django
{% component "my_table"  / %}
```

CHANGELOG.md
```diff
@@ -90,7 +97,6 @@
         type: RenderType = "document",
         render_dependencies: bool = True,
         request: Optional[HttpRequest] = None,
-        *response_args: Any,
```
JuroOravec, line 93: Also, since I've already made change to the `render_to_response()` method previously, I changed it so that it accepts only addional kwargs, but not args.

For context it's that any additional args / kwargs passed to `render_to_response()` are then forwarded to the `HttpResponse` constructor.

The reason for removing the forwarding of args is that one would have to specify all other 8 (!) arguments in order to pass the additional args to `HtppResponse`.

So it makes a lot more sense to just pass the kwargs.

docs/concepts/advanced/.nav.yml
```diff
@@ -6,9 +6,9 @@ nav:
   - Lifecycle hooks: hooks.md
   - Registering components: component_registry.md
   - Component caching: component_caching.md
-  - Typing and validation: typing_and_validation.md
+  - Component context and scope: component_context_scope.md
```
JuroOravec, line 9: I've moved the page on component context to the "advanced" section, and the typing to the "fundamentals".

docs/concepts/fundamentals/.nav.yml
```diff
@@ -1,15 +1,17 @@
...
   - Single-file components: single_file_components.md
-  - Components in Python: components_in_python.md
-  - Render API: render_api.md
+  - HTML / JS / CSS files: html_js_css_files.md
+  - HTML / JS / CSS variables: html_js_css_variables.md
+  - Secondary JS / CSS files: secondary_js_css_files.md
   - Component defaults: component_defaults.md
-  - Component context and scope: component_context_scope.md
+  - Render API: render_api.md
+  - Rendering components: rendering_components.md
```
JuroOravec, line 9: What was previously "Components in Python" was renamed to "Rendering components", and talks about all ways of rendering components.

What was previously "Defining HTML / JS / CSS files" was split to "HTML / JS / CSS files" and "Secondary JS / CSS files".

The page on "secondary JS / CSS" basically just talks about the `Media` class and how to define paths on it. I've split it off because it's quite info-dense topic.

Added page on variables / data methods like `get_template_data()`.

docs/concepts/fundamentals/render_api.md
```diff
@@ -138,3 +151,23 @@ rendered = Table.render(
     request=HttpRequest(),
 )
 ```
+
+## Provide / Inject
```
JuroOravec, line 155: Forgot to document provide / inject when I previously added this page on "Render API"

docs/concepts/fundamentals/typing_and_validation.md
```diff
@@ -447,6 +447,75 @@ class Button(Component):
     Slots = Empty
 ```
 
+## Subclassing
```
JuroOravec, line 450: Added section on subclassing for typing.

src/django_components/component.py
```diff
@@ -1960,61 +1999,60 @@ def render_to_response(
...
-        - `request` - The request object. This is only required when needing to use RequestContext,
-                      e.g. to enable template `context_processors`.
-
-        Any additional args and kwargs are passed to the `response_class`.
-
-        Example:
-        ```py
-        MyComponent.render_to_response(
-            args=[1, "two", {}],
+        Render the component and wrap the content in an HTTP response class.
```
JuroOravec, line 2007: The `render()` and `render_to_response()` methods are finally documented, as is now most of the Component class API 🚀

JuroOravec:
IMO with this PR the core of the documentation is in really good shape.

The pages with `+`  are pages which I think could still do with more work:

Screenshot 2025-04-22 at 18 37 54

Screenshot 2025-04-22 at 18 37 57

For the "fundamentals" section:

- The pages on variables and main files, there I need to update these pages once I enable and document JS / CSS variables
- "Rendering components" page - this one I will consider done once this one is in https://github.com/django-components/django-components/issues/971 (which will be the next PR 😄 )
- "Template tag syntax" - I want to review that page after the stuff around template tag parsing is fixed and finished (https://github.com/django-components/django-components/issues/935)
- "Slots" - I want to review this after https://github.com/django-components/django-components/issues/1096

But overall, I'm happy that with this update, it's starting to feel like the docs are catching up to what's implemented.

CHANGELOG.md
EmilStenstrom: This makes a lot of sense I think.

EmilStenstrom (approved):
⭐

PR #1156, opened 2025-04-28
feat: add "simple", "prepend", and "append" render types

JuroOravec:
Closes https://github.com/django-components/django-components/issues/971

The actual convo is here https://github.com/django-components/django-components/pull/959#issuecomment-2824018780

---

Currently in django-components we assumed that components would be rendered only in browsers. So even if one wanted to use djc to generate HTML for emails, we would still insert the `<script>` tag with our client-side dependency manager.

There were two "modes" for deciding how to render the JS and CSS, configured by the `type` kwarg that we pass to `Component.render()` - `"document"` and `"fragment"`.

This PR adds 3 more options - `"simple"`, `"prepend"`, `"append"`

```py
Calendar.render_to_response(
    request=request,
    kwargs={
        "date": request.GET.get("date", ""),
    },
    type="append",
)
```

Overview

- `"document"` (default)
    - Smartly inserts JS / CSS into placeholders or into `<head>` and `<body>` tags.
    - Inserts extra script to allow `fragment` types to work.
    - Assumes the HTML will be rendered in a JS-enabled browser.
- `"fragment"`
    - A lightweight HTML fragment to be inserted into a document.
    - No JS / CSS included.
- `"simple"`
    - Smartly insert JS / CSS into placeholders or into `<head>` and `<body>` tags.
    - No extra script loaded.
- `"prepend"`
    - Insert JS / CSS before the rendered HTML.
    - No extra script loaded.
- `"append"`
    - Insert JS / CSS after the rendered HTML.
    - No extra script loaded.

Other way to think about these is that:
- `"simple"` is the same as `"document"`, except we don't insert our dependency manager JS.
- `"prepend"` and `"append"` are the same as `"simple"`, except we ignore the placeholders and `<head>`/`<body>`, and always prepend/append all JS/CSS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,12 +29,14 @@
 - The interface of the not-yet-released `get_js_data()` and `get_css_data()` methods has changed to
   match `get_template_data()`.
 
+    Before:
+
     ```py
     def get_js_data(self, *args, **kwargs):
     def get_css_data(self, *args, **kwargs):
     ```
 
-    to:
+    After:
 
     ```py
     def get_js_data(self, args, kwargs, slots, context):
@@ -78,7 +80,7 @@
           escape_slots_content: bool = True,
           args: Optional[ArgsType] = None,
           kwargs: Optional[KwargsType] = None,
-          type: RenderType = "document",
+          deps_strategy: DependenciesStrategy = "document",
           request: Optional[HttpRequest] = None,
           *response_args: Any,
           **response_kwargs: Any,
@@ -94,7 +96,7 @@
         kwargs: Optional[Mapping] = None,
         slots: Optional[Mapping] = None,
         escape_slots_content: bool = True,
-        type: RenderType = "document",
+        deps_strategy: DependenciesStrategy = "document",
         render_dependencies: bool = True,
         request: Optional[HttpRequest] = None,
         **response_kwargs: Any,
@@ -153,6 +155,20 @@
     {% component "profile" name="John" job="Developer" / %}
     ```
 
+- The second argument to `render_dependencies()` is now `strategy` instead of `type`.
+
+    Before:
+
+    ```py
+    render_dependencies(content, type="document")
+    ```
+
+    After:
+
+    ```py
+    render_dependencies(content, strategy="document")
+    ```
+
 #### 🚨📢 Deprecation
 
 - `get_context_data()` is now deprecated. Use `get_template_data()` instead.
@@ -162,9 +178,55 @@
 
     Since `get_context_data()` is widely used, it will remain available until v2.
 
+- The `type` kwarg in `Component.render()` and `Component.render_to_response()` is now deprecated. Use `deps_strategy` instead. The `type` kwarg will be removed in v1.
+
+    Before:
+
+    ```py
+    Calendar.render_to_response(type="fragment")
+    ```
+
+    After:
+
+    ```py
+    Calendar.render_to_response(deps_strategy="fragment")
+    ```
+
 - `SlotContent` was renamed to `SlotInput`. The old name is deprecated and will be removed in v1.
 
 #### Feat
+
+- New method to render template variables - `get_template_data()`
+
+    `get_template_data()` behaves the same way as `get_context_data()`, but has
+    a different function signature to accept also slots and context.
+
+    ```py
+    class Button(Component):
+        def get_template_data(self, args, kwargs, slots, context):
+            return {
+                "val1": args[0],
+                "val2": kwargs["field"],
+            }
+    ```
+
+    If you define `Component.Args`, `Component.Kwargs`, `Component.Slots`, then
+    the `args`, `kwargs`, `slots` arguments will be instances of these classes:
+
+    ```py
+    class Button(Component):
+        class Args(NamedTuple):
+            field1: str
+
+        class Kwargs(NamedTuple):
+            field2: int
+
+        def get_template_data(self, args: Args, kwargs: Kwargs, slots, context):
+            return {
+                "val1": args.field1,
+                "val2": kwargs.field2,
+            }
+    ```
 
 - Input validation is now part of the render process.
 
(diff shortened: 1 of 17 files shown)


sampleproject/components/calendar/calendar.py
```diff
@@ -25,6 +25,7 @@ def get(self, request, *args, **kwargs):
                 kwargs={
                     "date": request.GET.get("date", ""),
                 },
+                type="append",
```
JuroOravec, line 28: WIth the new `"prepend"` and `"append"` options, it's possible to render a component ad hoc, without using neither the `{% component_dependencies %}` placeholders, nor requiring `<head>` and `<body>` HTML tags.

These types simply prepend or append all associated JS / CSS.

docs/concepts/advanced/rendering_js_css.md
```diff
@@ -220,3 +220,201 @@ with [`Component.render()`](#TODO) and inserting them into larger structures.
     - Directly passing rendered HTML to [`render_dependencies()`](#TODO)
 3. If you pre-render one component to pass it into another, the pre-rendered component must be rendered with
    [`render_dependencies=False`](#TODO).
+
+## Render types
```
JuroOravec, line 224: Moved this here from `rendering_components.md`, and instead in `rendering_components.md` there is only a short overview of what each mode does.

Screenshot 2025-04-28 at 21 47 23

CHANGELOG.md
```diff
@@ -166,6 +168,38 @@
 
 #### Feat
 
+- New method to render template variables - `get_template_data()`
```
JuroOravec, line 171: Not related to this PR, but just realized that `get_template_data()` was not mentioned on its own

src/django_components/dependencies.py
```diff
@@ -545,27 +551,27 @@ def on_replace_match(match: "re.Match[bytes]") -> bytes:
...
 
     (
-        to_load_input_js_urls,
-        to_load_input_css_urls,
-        inlined_input_js_tags,
-        inlined_input_css_tags,
-        loaded_input_js_urls,
-        loaded_input_css_urls,
-    ) = _prepare_tags_and_urls(inputs_data, type)
+        js_variables_urls_to_load,
```
JuroOravec, line 560: Unrelated to the issue, but I renamed these internal variables so it's clearer what they do.

JuroOravec:
Btw with the addition of these extra options, the `type` kwarg feels like THE main interface for configuring how a component should handle JS / CSS during rendering.

Right now there's also one more kwarg `render_dependencies`, which, if set to `False`, will skip the resolution of JS / CSS.

```py
Calendar.render_to_response(
    request=request,
    kwargs={
        "date": request.GET.get("date", ""),
    },
    render_dependencies=False,
)
```

I'm thinking that we could deprecate `render_dependencies` kwarg, and instead have `type="none"`:

```py
Calendar.render_to_response(
    request=request,
    kwargs={
        "date": request.GET.get("date", ""),
    },
    type="none",
)
```

The `render_dependencies=False` and the `type` kwarg are already mutually exclusive, so IMO it would make sense to merge them.

EmilStenstrom:
@JuroOravec Not that fond of the API here. So if we're changing things around, why not rethink?

"Render with type append" does not make sense to me. "Render with dependencies append" makes more sense. Or "Render with dependencies set to none". Can we change the name of the parameter to something clearer?

JuroOravec:
@EmilStenstrom Totally agree, just was low priority for me. I'd go with something like `dependencies_mode` or `dependencies_strategy` (or `deps_strategy`). I'd go with something that has the word `dependencies` in it, since this affects only the JS / CSS. What do you think?

EmilStenstrom:
@JuroOravec Like all of them! So deprecate the old param, add a new one with these new settings?

docs/concepts/advanced/html_fragments.md
```diff
@@ -1,5 +1,5 @@
-Django-components provides a seamless integration with HTML fragments ([HTML over the wire](https://hotwired.dev/)),
-whether you're using HTMX, AlpineJS, or vanilla JavaScript.
+Django-components provides a seamless integration with HTML fragments with AJAX ([HTML over the wire](https://hotwired.dev/)),
```
JuroOravec, line 1: I plan to review / update this page some time in the future, so I just updated the links / names

docs/concepts/advanced/rendering_js_css.md
```diff
@@ -220,3 +220,203 @@ with [`Component.render()`](#TODO) and inserting them into larger structures.
     - Directly passing rendered HTML to [`render_dependencies()`](#TODO)
 3. If you pre-render one component to pass it into another, the pre-rendered component must be rendered with
```
JuroOravec, line 221: Rest of the page needs cleaning up, but I will do so in 3 PR from now.

There, I implemented [this](https://github.com/django-components/django-components/issues/968#issuecomment-2653965260) and was able to remove the middleware, and also replaced `render_dependcies=False` with a separate value for `deps_strategy`.

docs/reference/api.md
```diff
@@ -67,6 +67,10 @@
     options:
       show_if_no_docstring: true
 
+::: django_components.DependenciesStrategy
```
JuroOravec, line 70: Also made public the type that describes available strategies.

tests/test_dependencies.py
```diff
@@ -49,6 +49,24 @@ class Media:
         js = "script.js"
 
 
+@djc_test
+class TestDependenciesLegacy:
```
JuroOravec, line 53: Added test to ensure backwards compatibility.

JuroOravec:
@EmilStenstrom Updated! I went with `deps_strategy` 🚀

docs/concepts/advanced/rendering_js_css.md
```diff
@@ -220,3 +220,203 @@ with [`Component.render()`](#TODO) and inserting them into larger structures.
...
+- [`prepend`](../../advanced/rendering_js_css#prepend)
+    - Insert JS / CSS before the rendered HTML.
+    - No extra script loaded.
+- [`append`](../../advanced/rendering_js_css#append)
+    - Insert JS / CSS after the rendered HTML.
+    - No extra script loaded.
+
+### `document`
+
+`deps_strategy="document"` is the default. Use this if you are rendering a whole page, or if no other option suits better.
```
EmilStenstrom, line 260: I wonder if this should be a setting too? Easier to disable the dependency manager once than for every component.

tests/test_dependencies.py
```diff
@@ -472,6 +393,581 @@ class ComponentWithScript(SimpleComponent):
...
+            <!DOCTYPE html>
+            <html>
+                <head></head>
+                <body>
+                    {% component "test" variable="foo" / %}
+                </body>
+            </html>
+        """
+        rendered_raw = Template(template_str).render(Context({}))
+        rendered = render_dependencies(rendered_raw)
```
EmilStenstrom, line 412: I think I would set dep_strategy="document" here just to be explicit

EmilStenstrom (approved):
Looks good to me! Nice little cleanup/clarification of the API.

docs/concepts/advanced/rendering_js_css.md
JuroOravec: Sounds good, make a follow up ticket https://github.com/django-components/django-components/issues/1159

tests/test_dependencies.py
JuroOravec: Thanks, fixed!

PR #1160, opened 2025-05-03
feat: allow to set query and fragment on get_component_url

JuroOravec:
Updates `get_component_url()` to now optionally accept `query` and `fragment` arguments.

```py
from django_components import get_component_url

url = get_component_url(
    MyComponent,
    query={"foo": "bar"},
    fragment="baz",
)
# /components/ext/view/components/c1ab2c3?foo=bar#baz
```

Closes https://github.com/django-components/django-components/issues/1149

### DIFF
--- a/src/django_components/util/misc.py
+++ b/src/django_components/util/misc.py
@@ -5,7 +5,8 @@ from hashlib import md5
 from importlib import import_module
 from itertools import chain
 from types import ModuleType
-from typing import TYPE_CHECKING, Any, Callable, Iterable, List, Optional, Tuple, Type, TypeVar, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Tuple, Type, TypeVar, Union
+from urllib import parse
 
 from django_components.constants import UID_LENGTH
 from django_components.util.nanoid import generate
@@ -147,3 +148,32 @@ def to_dict(data: Any) -> dict:
         return asdict(data)  # type: ignore[arg-type]
 
     return dict(data)
+
+
+def format_url(url: str, query: Optional[Dict] = None, fragment: Optional[str] = None) -> str:
+    """
+    Given a URL, add to it query parameters and a fragment, returning an updated URL.
+
+    ```py
+    url = format_url(url="https://example.com", query={"foo": "bar"}, fragment="baz")
+    # https://example.com?foo=bar#baz
+    ```
+
+    `query` and `fragment` are optional, and not applied if `None`.
+    """
+    url_parts = parse.urlsplit(url)
+
+    escaped_fragment = parse.quote(str(fragment)) if fragment is not None else url_parts.fragment
+
+    # NOTE: parse_qsl returns a list of tuples. For simpicity we support only one
+    # value per key, so we need to unpack the first element.
+    query_from_url = {key: val[0] for key, val in parse.parse_qsl(url_parts.query)}
+    all_params = {**query_from_url, **(query or {})}
+
+    encoded_params = parse.urlencode(all_params)
+    updated_parts = url_parts._replace(
+        query=encoded_params,
+        fragment=escaped_fragment,
+    )
+    new_url = parse.urlunsplit(updated_parts)
+    return new_url
--- a/tests/test_component.py
+++ b/tests/test_component.py
@@ -373,6 +373,16 @@ class TestComponent:
         with pytest.raises(KeyError):
             get_component_by_class_id("nonexistent")
 
+    def test_component_render_id(self):
+        class SimpleComponent(Component):
+            template = "render_id: {{ render_id }}"
+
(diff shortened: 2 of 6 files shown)


tests/test_component.py
```diff
@@ -373,6 +373,16 @@ def test_get_component_by_id_raises_on_missing_component(self):
         with pytest.raises(KeyError):
             get_component_by_class_id("nonexistent")
 
+    def test_component_render_id(self):
```
JuroOravec, line 376: Unrelated to this PR, but I realized there wasn't a test to check the accessing of `Component.id`.

tests/test_component_view.py
```diff
@@ -297,6 +297,24 @@ def get(self, request, *args, **kwargs) -> HttpResponse:
...
+        # Check if the URL is correctly generated
+        component_url = get_component_url(TestComponent)
+        assert component_url == f"/components/ext/view/components/{TestComponent.class_id}/"
+
+        component_url2 = get_component_url(TestComponent, query={"foo": "bar"}, fragment="baz")
+        assert component_url2 == f"/components/ext/view/components/{TestComponent.class_id}/?foo=bar#baz"
+
+        # Check that the query and fragment are correctly escaped
+        component_url3 = get_component_url(TestComponent, query={"f'oo": "b ar&ba'z"}, fragment='q u"x')
+        assert component_url3 == f"/components/ext/view/components/{TestComponent.class_id}/?f%27oo=b+ar%26ba%27z#q%20u%22x"  # noqa: E501
```
JuroOravec, line 316: Here we check that any quotes and similar are properly escaped.

JuroOravec:
cc @EmilStenstrom

src/django_components/util/misc.py
```diff
@@ -147,3 +148,32 @@ def to_dict(data: Any) -> dict:
...
+    query_from_url = {key: val[0] for key, val in parse.parse_qsl(url_parts.query)}
+    all_params = {**query_from_url, **(query or {})}
+
+    encoded_params = parse.urlencode(all_params)
+    updated_parts = url_parts._replace(
+        query=encoded_params,
+        fragment=escaped_fragment,
+    )
+    new_url = parse.urlunsplit(updated_parts)
+    return new_url
```
EmilStenstrom, line 179: I passed this method to OpenAI o3 for a review, since it's security related. Does it make sense to harden? https://chatgpt.com/share/681692e6-f698-8006-9553-5a4bfacffd56

EmilStenstrom (commented):
I'm a little worried of format_url, mind reading through the ChatGPT review?

src/django_components/util/misc.py
JuroOravec: - **Functional bug** - Yup, this really is a bug in `format_url()`. But not an issue in the context of `get_component_url()`, because the URL string passed to `format_url()` doesn't have any query params.

- **Duplicate‑key handling** - Risk identified correctly. But again not an issue in `get_component_url()`, because the URL string doesn't contain any query params, and the dictionary of query params passed to `get_component_url()` is expected to have only singular values, not lists.

- **Parameter precedence** - IMO not relevant - If an attacker can modify the query params dict passed to `get_component_url()`, then the system is already compromised, and user has bigger problems than this.

   Basically this is the same as saying that Python's [`parse.urlunsplit`](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlunsplit) is insecure because you can construct unsafe URLs with it. But [Python docs say](https://docs.python.org/3/library/urllib.parse.html#url-parsing-security) it's the user's responsibility to ensure that safe inputs are used.

- **Scheme / host validation** - Again, not relevant in context of `get_component_url()`, because the resulting URL is just a relative path (e.g. `/components/ext/view/component/a1b2c3?foo=bar`)

- **Fragment encoding** - ChatGPT is right that the fragment from the URL string is not encoded. Again here I assumed that the URL string provided to `format_url()` should be already safe. If not, then again, there's bigger problems than this.

- **Unicode & norm‑confusion** - No opinion on that, not familiar with NFC / IDNA, etc.

- **No output‑context encoding** - Again IMO not applicable, because we control how the URL string is constructed in `get_component_url()` before passed to `format_url()` (by delegating to `django.urls.reverse()`). So the URL should never have a stray `"` or similar.

So apart from the bug, then I don't think `format_url()` is an issue **assuming it's used only from `get_component_url()`**. But the implementation given by Chatgpt still passes all the tests, so better to address any ambiguous parts now, rather than finding out too late if we ever used `format_url()` in a different part of the codebase.

JuroOravec: Updated in https://github.com/django-components/django-components/pull/1163

EmilStenstrom: Happy to hear that this is not an issue with the current usage. Probably a good idea to have the safer version, if we where to expand its usage going forward. Great work as always.

PR #1161, opened 2025-05-03
refactor: Update docs and tests to use get_template_data()

JuroOravec:
As the title says, this is just a renaming update so examples and tests use the new `get_template_data()`, as originally mention in https://github.com/django-components/django-components/pull/1138#issuecomment-2817077487.

As a reminder, `get_context_data()` is still mentioned / documented, but only in API reference and this section:

Screenshot 2025-05-03 at 11 19 07

### DIFF
--- a/docs/concepts/fundamentals/template_tag_syntax.md
+++ b/docs/concepts/fundamentals/template_tag_syntax.md
@@ -30,14 +30,12 @@ so are still valid:
 </body>
 ```
 
-These can then be accessed inside `get_context_data` so:
+These can then be accessed inside `get_template_data` so:
 
 ```py
 @register("calendar")
 class Calendar(Component):
-    # Since # . @ - are not valid identifiers, we have to
-    # use `**kwargs` so the method can accept these args.
-    def get_context_data(self, **kwargs):
+    def get_template_data(self, args, kwargs, slots, context):
         return {
             "date": kwargs["my-date"],
             "id": kwargs["#some_id"],
@@ -92,17 +90,17 @@ _New in version 0.93_
 When passing data around, sometimes you may need to do light transformations, like negating booleans or filtering lists.
 
 Normally, what you would have to do is to define ALL the variables
-inside `get_context_data()`. But this can get messy if your components contain a lot of logic.
+inside `get_template_data()`. But this can get messy if your components contain a lot of logic.
 
 ```py
 @register("calendar")
 class Calendar(Component):
-    def get_context_data(self, id: str, editable: bool):
+    def get_template_data(self, args, kwargs, slots, context):
         return {
-            "editable": editable,
-            "readonly": not editable,
-            "input_id": f"input-{id}",
-            "icon_id": f"icon-{id}",
+            "editable": kwargs["editable"],
+            "readonly": not kwargs["editable"],
+            "input_id": f"input-{kwargs['id']}",
+            "icon_id": f"icon-{kwargs['id']}",
             ...
         }
 ```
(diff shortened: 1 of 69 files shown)


docs/concepts/fundamentals/template_tag_syntax.md
```diff
@@ -30,14 +30,12 @@ so are still valid:
 </body>
 ```
 
-These can then be accessed inside `get_context_data` so:
+These can then be accessed inside `get_template_data` so:
 
 ```py
 @register("calendar")
 class Calendar(Component):
-    # Since # . @ - are not valid identifiers, we have to
```
JuroOravec, line 38: Removed this comment on kwargs with special characters, since that's not an issue with `get_template_data()`

docs/getting_started/parametrising_components.md
```diff
@@ -68,46 +74,40 @@ from django_components import Component, register
...
+    def get_template_data(self, args, kwargs, slots, context):
         return {
-            "date": date,
-            "extra_class": extra_class,
+            "date": kwargs["date"],
+            "extra_class": kwargs.get("extra_class", "text-blue"),
         }
 ```
 
-!!! info
```
JuroOravec, line 78: These info / warning notes no longer relevant.

docs/getting_started/parametrising_components.md
```diff
@@ -68,46 +74,40 @@ from django_components import Component, register
...
-    Since [`get_context_data()`](../../reference/api#django_components.Component.get_context_data)
-    is just a regular Python function, it will raise TypeError if it receives incorrect parameters.
-
 Since `extra_class` is optional in the function signature, it's optional also in the template.
 So both following calls are valid:
 
 ```htmldjango
-{% component "calendar" "2024-12-13" / %}
-{% component "calendar" "2024-12-13" extra_class="text-red" / %}
+{% component "calendar" date="2024-12-13" / %}
```
JuroOravec, line 88: Slight difference is that with `get_context_data()`, one could pass a value (e.g. a `date`) EITHER as arg or kwarg.

But with `get_template_data()`, the args and kwargs are strictly separated.

...Actually, I've just updated the docs to mention this in more detail:

Screenshot 2025-05-03 at 11 41 47

sampleproject/components/greeting.py
```diff
@@ -1,24 +1,8 @@
-from typing import Any, Dict
-
 from django_components import Component, register, types
 
 
 @register("greeting")
 class Greeting(Component):
-    class View:
```
JuroOravec, line 8: Moved the View class below the component definition

src/django_components/commands/create.py
```diff
@@ -199,16 +199,20 @@ def handle(self, *args: Any, **kwargs: Any) -> None:
 
                     @register("{name}")
                     class {name.capitalize()}(Component):
-                        template_file = "{name}/{template_filename}"
+                        template_file = "{template_filename}"
```
JuroOravec, line 202: Updated the boilerplate that's generated by the command that creates component files.

tests/test_component.py
```diff
@@ -60,6 +60,31 @@ def get_template_string(self, context):
...
+        assertHTMLEqual(
+            rendered,
+            """
+            Variable: <strong data-djc-id-ca1bc3e>test</strong>
+            """,
+        )
+
+    # TODO_REMOVE_IN_V2 - `get_context_data()` was superseded by `self.get_template_data`
+    @djc_test(parametrize=PARAMETRIZE_CONTEXT_BEHAVIOR)
+    def test_get_context_data(self, components_settings):
```
JuroOravec, line 82: Added a test specifically for `get_context_data()` to ensure we remain backwards compatible.

JuroOravec:
cc @EmilStenstrom

PR #1162, opened 2025-05-03
refactor: rename `context_data` field to `template_data`

JuroOravec:
Made this a separate PR from https://github.com/django-components/django-components/pull/1161 so that https://github.com/django-components/django-components/pull/1161 contained no new changes / suggestions.

As part of renaming `get_context_data` to `get_template_data`, there is also one object that had field `context_data`. This PR renames it to `template_data`.

The object was available from within the `on_component_data` extension hook.

Again the old field is kept for backwards compatibility and tagged to be removed in v1.

Screenshot 2025-05-03 at 12 13 55

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -194,6 +194,26 @@
 
 - `SlotContent` was renamed to `SlotInput`. The old name is deprecated and will be removed in v1.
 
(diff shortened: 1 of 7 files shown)


JuroOravec:
@EmilStenstrom This one is tiny, but deserves a review, because we didn't discuss this.

EmilStenstrom (approved):
Makes sense to me!

PR #1166, opened 2025-05-05
refactor: remove middleware, add strategy "raw", and call render_deps() from within Template.render()

JuroOravec:
Closes https://github.com/django-components/django-components/issues/968


docs/concepts/fundamentals/rendering_components.md
```diff
@@ -515,17 +518,18 @@ Button.render(
 )
 ```
 
-## Nesting components
```
JuroOravec, line 518: Renamed, because I realized that people may confuse this "nesting" with action of placing one `{% component %}` tag into another.

docs/overview/installation.md
```diff
@@ -16,43 +19,13 @@
 3. `BASE_DIR` setting is required. Ensure that it is defined:
 
     ```python
+    # app/settings.py
     from pathlib import Path
 
     BASE_DIR = Path(__file__).resolve().parent.parent
     ```
 
-4. _Optional._ Set [`COMPONENTS.dirs`](../reference/settings.md#django_components.app_settings.ComponentsSettings.dirs)
```
JuroOravec, line 24: Since this installation step is optional, I've moved it down [here](https://github.com/django-components/django-components/pull/1166/files#diff-54caa5c9093caf057439e1e8d510a8d2131298fb28516a36a551e220a7bfb636R139) into the "Optional" section.

docs/overview/installation.md
```diff
@@ -82,6 +56,20 @@
     ]
     ```
 
+5. Add django-component's URL paths to your `urlpatterns`:
```
JuroOravec, line 59: Previously the installation step to add django-component's urlpatterns was in section for adding support for JS and CSS. Moved it up here into the "core" installation steps.

Back then it made sense for it to be up there.

But since we have the feature to automatically create endpoints for components with `View.public = True`, it means that we need to set `urlpatterns` not just when using JS / CSS, but all the time.

docs/scripts/reference.py
```diff
@@ -338,49 +338,6 @@ def _gen_default_settings_section(app_settings_filepath: str) -> str:
     )
 
 
-def gen_reference_middlewares():
```
JuroOravec, line 341: Removed the script that generated reference documentation for the middleware.

tests/test_dependencies.py
```diff
@@ -110,19 +109,64 @@ def test_middleware_renders_dependencies(self):
...
+        registry.register(name="test", component=SimpleComponent)
+
+        template_str: types.django_html = """
+            {% load component_tags %}
+            {% component_js_dependencies %}
+            {% component_css_dependencies %}
+            {% component 'test' variable='foo' / %}
+        """
+        template = Template(template_str)
+        rendered: str = template.render(Context({"DJC_DEPS_STRATEGY": "append"}))
```
JuroOravec, line 141: Added tests to ensure that the DJC_DEPS_STRATEGY approach works.

tests/test_dependencies.py
```diff
@@ -212,6 +257,103 @@ class SimpleComponentWithDeps(SimpleComponent):
         assert rendered.count("<link") == 1
         assert rendered.count("<style") == 1
 
+    def test_inserts_styles_and_script_to_default_places_if_not_overriden(self):
```
JuroOravec, line 260: Had these two tests stashed aside for some time already

src/django_components/util/django_monkeypatch.py
```diff
@@ -99,9 +101,32 @@ def _template_render(self: Template, context: Context, *args: Any, **kwargs: Any
...
+        # they render a Template directly with `Template.render()` or Django's `django.shortcuts.render()`.
+        #
+        # Example:
+        # ```
+        # result = render_dependencies(
+        #     result,
+        #     Context({ "DJC_DEPS_STRATEGY": "fragment" }),
+        # )
+        # ```
+        if _STRATEGY_CONTEXT_KEY in context and context[_STRATEGY_CONTEXT_KEY] is not None:
```
JuroOravec, line 124: This is the new logic added inside `Template.render()`

After removing the middleware, there was a question of how to configure which rendering strategy should be used when one renders templates with `Template.render()` or `django.shortcuts.render()`.

What I went with is that you can set the strategy by setting a context variable `DJC_DEPS_STRATEGY`.

If `DJC_DEPS_STRATEGY` is omitted, then the default strategy ("document") is used, in which case the behaviour is the same as it was when using the middleware.

src/django_components/slots.py
```diff
@@ -1102,7 +1102,11 @@ def render_func(ctx: Context, slot_data: Dict[str, Any], slot_ref: SlotRef) -> S
 
         trace_component_msg("RENDER_NODELIST", component_name, component_id=None, slot_name=slot_name)
 
-        rendered = template.render(ctx)
+        # We wrap the slot nodelist in Template. However, we also override Django's `Template.render()`
+        # to call `render_dependencies()` on the results. So we need to set the strategy to `raw`
+        # so that the dependencies are processed only once the whole component tree is rendered.
+        with ctx.push({"DJC_DEPS_STRATEGY": "raw"}):
```
JuroOravec, line 1108: Had to wrap slots in `DJC_DEPS_STRATEGY="raw"`, because we render slots via `Template.render()`.

docs/concepts/advanced/rendering_js_css.md
```diff
@@ -1,24 +1,29 @@
-## JS and CSS output locations
```
JuroOravec, line 1: Rewrote the first half of the page. Some screen grabs:

Screenshot 2025-05-05 at 14 12 33

Screenshot 2025-05-05 at 14 12 39

Screenshot 2025-05-05 at 14 12 48

Screenshot 2025-05-05 at 14 13 02

JuroOravec:
@EmilStenstrom Look at that, looks like we've got a [first catch](https://github.com/django-components/django-components/pull/1166#issuecomment-2850899765) for the benchmarking bot!

After seeing that, I've updated the logic inside `Template.render()` to call `render_dependencies()` only if there is AT LEAST ONE component rendered inside the HTML - [See here](https://github.com/django-components/django-components/pull/1166/files#diff-6cad9d5565786c00d355ba20ae181e9ebc658eca28f7d762dd588172c0d6fdc6R114).

Also, this is a nitty-gritty thing, but this cleared up one small inconsistency:
- Before, when the middleware was used, then the middleware inserted our dependency manager JS script (`django_components.min.js`) even if there were no components rendered.
- Now, the `django_components.min.js` is inserted only if the rendered HTML contains AT LEAST ONE component.

The new behaviour is now consistent with that we say in the docs - that users need to render a template as a component and with the "document" strategy in order to be able to insert fragments.

CHANGELOG.md
```diff
@@ -192,6 +230,20 @@
...
+
+    ```py
+    Calendar.render_to_response(render_dependencies=False)
+    ```
+
+    After:
+
+    ```py
+    Calendar.render_to_response(deps_strategy="raw")
+    ```
```
EmilStenstrom, line 245: I'm not super fond of saying "I want to render this with a _raw_ dependency strategy". I agree that "none" isn't great either, because it sounds like you have no strategy. Maybe "ignore" could work? "The strategy is to ignore the dependencies" makes sense I think!

CHANGELOG.md
```diff
@@ -26,6 +26,44 @@
...
+
+    - NOTE: If you rendered HTML with `Component.render()` or `Component.render_to_response()`, the JS and CSS were already rendered.
+
+    To disable this behavior, set the `DJC_DEPS_STRATEGY` context key to `"raw"`
+    when rendering the template:
+
+    ```py
+    # With `Template.render()`:
+    template = Template(template_str)
+    rendered = template.render(Context({"DJC_DEPS_STRATEGY": "raw"}))
```
EmilStenstrom, line 45: This feels a bit dirty to expose to users. A magical variable that you are passing. I wonder if it would make more sense if we provided our own version of Template that you can import if you want the change the deps_strategy? Feels a lot cleaner I think.

EmilStenstrom (approved):
Code looks great! I have two suggested changes to the API, take it or leave it! :)

CHANGELOG.md
JuroOravec: Yeah, I like "ignore"!

CHANGELOG.md
JuroOravec: Yeah, but unfortunately this is the best I could come up with.

The issue with having for example a custom `Template` class is that it might not work with other part of Django or even 3rd party packages that render templates.

Take for example just the `django.shortcuts.render()` method - you give it only the template name, and it creates an instance of `Template` for you. So you can't override that.

On the other hand, the context variables will be passed through no matter how the template is created and called.

EmilStenstrom: Is this something that we could change with our own template engine?

JuroOravec: Yeah, then we could have own reimplementations of e.g. `django.shortcuts.render()`, which could accept extra argument to set strategy.

PR #1172, opened 2025-05-06
Fix context processors

### LINKED ISSUE #1165: Context processors run for every component
Basically title. I'm not sure if this is intended or what exactly the behavior of normal Django is, but in any case, this can really be an issue if one is using Django components a lot for smaller components.

With **normal Django templates** I would expect that if I have 100 smaller templates with `include` inside a main template, that the context processors only run once. Same if I implement a simple template tag or inclusion template tag.

With **Django components** the context processors seem to run for every component, which was actually quite a performance problem for me, since I have views with hundreds of components, and my context processors do up to a few ms of work.

I haven't had the time yet to look at the code, but I thought I get this out of the door to keep track of the issue.

JuroOravec:
See #1165 

When generating context processors data, in `gen_context_processors_data()`, we meant to cache the results, so we would reuse the same data for the same request object. But the line of code to save the generated data was missing.

Closes #1165

### DIFF
--- a/src/django_components/util/context.py
+++ b/src/django_components/util/context.py
@@ -131,7 +131,7 @@ def _copy_block_context(block_context: BlockContext) -> BlockContext:
 # See https://github.com/django/django/blame/2d34ebe49a25d0974392583d5bbd954baf742a32/django/template/context.py#L255
 def gen_context_processors_data(context: BaseContext, request: HttpRequest) -> Dict[str, Any]:
     if request in context_processors_data:
-        return context_processors_data[request]
+        return context_processors_data[request].copy()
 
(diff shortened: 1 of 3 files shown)


src/django_components/util/context.py
```diff
@@ -153,4 +153,6 @@ def gen_context_processors_data(context: BaseContext, request: HttpRequest) -> D
         except TypeError as e:
             raise TypeError(f"Context processor {processor.__qualname__} didn't return a " "dictionary.") from e
 
+    context_processors_data[request] = processors_data
```
JuroOravec, line 156: This was the missing line

JuroOravec:
cc @EmilStenstrom

PR #1177, opened 2025-05-09
feat: add BaseNode.contents

JuroOravec:
Another part towards https://github.com/django-components/django-components/issues/1164.

This modifies our implementation of template tags so that the Node (well, `BaseNode`) class remembers what was the content of the body (the string between `{% tag %}` / `{% endtag %}`), and makes it accessible via `BaseNode.contents`.

### DIFF
--- a/src/django_components/util/template_tag.py
+++ b/src/django_components/util/template_tag.py
@@ -104,7 +104,7 @@ def resolve_params(
 class ParsedTag(NamedTuple):
     flags: Dict[str, bool]
     params: List[TagAttr]
-    parse_body: Callable[[], NodeList]
+    parse_body: Callable[[], Tuple[NodeList, Optional[str]]]
 
 
 def parse_template_tag(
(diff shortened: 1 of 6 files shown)


tests/test_node.py
```diff
@@ -836,6 +839,96 @@ def render(self, context: Context, name: str, **kwargs) -> str:
 
         TestNode.unregister(component_tags.register)
 
+    def test_node_class_attributes(self):
```
JuroOravec, line 842: Added test to check also for other attributes on `BaseNode` as we don't check for that yet.

src/django_components/util/template_tag.py
```diff
@@ -155,10 +157,53 @@ def _parse_tag_body(parser: Parser, end_tag: str, inline: bool) -> NodeList:
...
 
+# Similar to `parser.parse(parse_until=[end_tag])`, except:
+# 1. Does not remove the token it goes over (unlike `parser.parse()`, which mutates the parser state)
+# 2. Returns a string, instead of a NodeList
+#
+# This is used so we can access the contents of the tag body as strings, for example
+# to be used for caching slots.
+#
+# See https://github.com/django/django/blob/1fb3f57e81239a75eb8f873b392e11534c041fdc/django/template/base.py#L471
+def _extract_contents_until(parser: Parser, until_blocks: List[str]) -> str:
```
JuroOravec, line 172: Basically, before we pass template parsing to Django's `Parser.parse()`, which mutates the state, we use similar logic to obtain the raw string of the template tag body (string between `{% tag %}` / `{% endtag %}`

JuroOravec:
@EmilStenstrom Another rather small one.

EmilStenstrom (approved):

PR #1179, opened 2025-05-11
chore: util to manage URLs in the codebase

JuroOravec:
I noticed there were old links in the codebase pointing to Django 4.0 docs.

I updated the links and checked that they are still valid even when updated to Django 5.2.

This then prompted me to write a script so I wouldn't have to manually manually check / update links again. How the script works is that it:

1. Collects all URLs in the codebase (minus those files in .gitignore and some extra)
2. Makes GET requests to those URLs, checking if the links are still alive
3. If the URL has a fragment (`#abc`), then use BeautifulSoup to check that the fragment is present in the HTML response.
4. Output errors to terminal, or optionally to separate file.
5. And to be able to manage easily the links to Django docs, I also added an option to overwrite the URLs

Screenshot 2025-05-11 at 11 46 46


.gitignore
```diff
@@ -1,3 +1,6 @@
+# Project-specific files
+sampleproject/staticfiles/
```
JuroOravec, line 2: `sampleproject/staticfiles/` is already in .gitignore of nested `sampleproject/.gitignore`. But for simplicity the script doesn't handle nested `.gitignores`, so added this pattern here.

.gitignore
```diff
@@ -43,6 +46,7 @@ htmlcov/
 nosetests.xml
 coverage.xml
 *,cover
+.pytest_cache/
```
JuroOravec, line 49: Seems these are already ignored from git  in VSCode somehow, but yet they weren't specified in .gitignore.

CHANGELOG.md
```diff
@@ -718,7 +718,7 @@ where each class name or style property can be managed separately.
     - If using `pytest`, the decorator allows you to parametrize Django or Components settings.
     - The decorator also serves as a stand-in for Django's `@override_settings`.
 
-    See the API reference for [`@djc_test`](https://django-components.github.io/django-components/0.131/reference/testing_api/#djc_test) for more details.
+    See the API reference for [`@djc_test`](https://django-components.github.io/django-components/0.131/reference/testing_api/#django_components.testing.djc_test) for more details.
```
JuroOravec, line 721: It might be hard to see, so unless mentioned otherwise, these changes are just URL updates. Like here the fragment is changed from `#djc_test` to `#django_components.testing.djc_test`.

Sidenote - It would be nicer to have the fragment here just `#djc_test`, but this  would need to be configured in mkdocs somehow, and I didn't bother looking into that.

docs/concepts/advanced/provide_inject.md
```diff
@@ -82,7 +82,6 @@ class ChildComponent(Component):
         my_data = self.inject("my_data")
         print(my_data.hello)    # hi
         print(my_data.another)  # 123
-        return {}
```
JuroOravec, line 85: Unrelated to the URL managing script, but there were also a lot of instances of `get_template_data()` where we return empty dict. That's no longer required, so I removed them.

docs/getting_started/rendering_components.md
```diff
@@ -39,7 +39,7 @@ If you have embedded the component in a Django template using the
 
 You can simply render the template with the Django tooling:
 
-#### With [`django.shortcuts.render()`](https://docs.djangoproject.com/en/5.1/topics/http/shortcuts/#render)
+#### With [`django.shortcuts.render()`](https://docs.djangoproject.com/en/5.2/topics/http/shortcuts/#render)
```
JuroOravec, line 42: Btw so Django docs also have a URL format with `stable` instead of specific version number. So alternatively we could have set all links just to `stable`.

However, then if / once a link breaks, it would be harder to pinpoint when that occured, because there would not be any associated git commit.

This way, when all docs links are pinned, then if we next update the links version and run the link validation script, we will know exactly between which versions the error occured.

One example (tho not on this particular line of code) was that the Django setting `STORAGES` was called differently in Django 4.2. And so the URL fragment was correct on 4.2, but broken on 5.0, 5.1, 5.2.

src/django_components/commands/list.py
```diff
@@ -87,58 +87,6 @@ def handle(self, *args: Any, **kwargs: Any) -> None:
         print(table)
 
 
-def format_as_ascii_table(data: List[Dict[str, Any]], headers: List[str], include_headers: bool = True) -> str:
```
JuroOravec, line 90: Moved to `misc.py`.

scripts/validate_links.py
```diff
@@ -0,0 +1,406 @@
...
+Features:
+- Finds all URLs in code, markdown, and docstrings.
+- Validates URLs by making GET requests (with caching and rate limiting).
+- Uses BeautifulSoup to check for HTML fragments (e.g., #section) in the target page.
+- Outputs a summary table of all issues (invalid, broken, missing fragment, etc).
+- Can output the summary table to a file with `-o`/`--output`.
+- Can rewrite URLs in-place using URL_REWRITE_MAP (supports both prefix and regex mapping).
+- Supports dry-run mode for rewrites with `--dry-run`.
+
+Usage:
```
JuroOravec, line 17: Here below are examples how to run this script

scripts/validate_links.py
```diff
@@ -0,0 +1,406 @@
...
+
+    # Output summary to file if specified
+    if output:
+        output_path = Path(output)
+        output_path.write_text(table + "\n", encoding="utf-8")
+    else:
+        print(table + "\n")
+
+
+# TODO: Run this as a test in CI?
```
JuroOravec, line 303: Altho it would be nice if this ran as part of CI, there's a lot of false positives. So this script makes more sense to ran once in couple of months to check that the URLs are still correct.

JuroOravec:
cc @EmilStenstrom.

PR #1180, opened 2025-05-11
feat: Expose slot input as Slot.contents

JuroOravec:
Part of https://github.com/django-components/django-components/issues/1164

This updates the `Slot` class and its constructor such that it exposes `Slot.contents`, which contains the original contents:

  - If `Slot` was created from `{% fill %}` tag, `Slot.contents` will contain the body (string) of the `{% fill %}` tag.
  - If `Slot` was created from string via `Slot("...")`, `Slot.contents` will contain that string.
  - If `Slot` was created from a function, `Slot.contents` will contain that function.
  - If `Slot` was created from another Slot `Slot(slot)`, `Slot.contents` will contain the inner slot's `Slot.contents`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 🚨📢 v0.140.0
 
+⚠️ Major release ⚠️ - Please test thoroughly before / after upgrading.
+
 #### 🚨📢 BREAKING CHANGES
 
 **Middleware**
@@ -221,6 +223,28 @@
     {% component "profile" name="John" job="Developer" / %}
     ```
 
+**Slots**
+
+- If you instantiated `Slot` class with kwargs, you should now use `contents` instead of `content_func`.
+
+    Before:
+
+    ```py
+    slot = Slot(content_func=lambda *a, **kw: "CONTENT")
+    ```
+
+    After:
+
+    ```py
+    slot = Slot(contents=lambda *a, **kw: "CONTENT")
+    ```
+
+    Alternatively, pass the function / content as first positional argument:
+
+    ```py
+    slot = Slot(lambda *a, **kw: "CONTENT")
+    ```
+
 **Miscellaneous**
 
 - The second argument to `render_dependencies()` is now `strategy` instead of `type`.
@@ -435,6 +459,12 @@
     ```
 
     Then, the `contents` attribute of the `BaseNode` instance will contain the string `"Hello, world!"`.
+
+- `Slot` class now has a `Slot.contents` attribute, which contains the original contents:
+
+    - If `Slot` was created from `{% fill %}` tag, `Slot.contents` will contain the body of the `{% fill %}` tag.
+    - If `Slot` was created from string via `Slot("...")`, `Slot.contents` will contain that string.
+    - If `Slot` was created from a function, `Slot.contents` will contain that function.
 
 #### Fix
 
--- a/src/django_components/slots.py
+++ b/src/django_components/slots.py
@@ -125,7 +125,7 @@ class Slot(Generic[TSlotData]):
                 nodelist=NodeList([TextNode(contents)]),
                 contents=contents,
                 data_var=None,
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 🚨📢 v0.140.0
 
+⚠️ Major release ⚠️ - Please test thoroughly before / after upgrading.
```
JuroOravec, line 5: At this amount of small but breaking changes, it's probably better to warn people to treat this as a major release.

src/django_components/slots.py
```diff
@@ -70,17 +77,30 @@ class Slot(Generic[TSlotData]):
     slot_name: Optional[str] = None
     """Name of the slot that originally defined or accepted this slot fill."""
     nodelist: Optional[NodeList] = None
-    """Nodelist of the slot content."""
+    """If the slot was defined with `{% fill %} tag, this will be the Nodelist of the slot content."""
 
     def __post_init__(self) -> None:
+        # Since the `Slot` instance is treated as a function, it may be passed as `contents`
```
JuroOravec, line 83: This is the main part of the PR 

- Before, the slot input was normalized when calling `Component.render()`, inside [`_normalize_slot_fills()`](https://github.com/django-components/django-components/pull/1180/files#diff-35c84ebceae50f7149f03bbcaf2b4f1289bc2984bf7859ca521b5dbe7960c396R2707)

- Now the main normalization logic is `Slot()` constructor, so people will be able to call `Slot(...)` from wherever and still get the same behaviour.

tests/test_component.py
```diff
@@ -325,98 +356,6 @@ def get_template(self, context):
             """,
         )
 
-    @djc_test(parametrize=PARAMETRIZE_CONTEXT_BEHAVIOR)
-    def test_prepends_exceptions_with_component_path(self, components_settings):
```
JuroOravec, line 329: The diff in Github is not great, but:
- `test_prepends_exceptions_with_component_path` was moved at the end of `TestComponentRender`
- `test_get_component_by_id` and `test_get_component_by_id_raises_on_missing_component` actually remained here unde `TestComponent`.
- `test_component_render_id` and [`test_input`](https://github.com/django-components/django-components/pull/1180/files#diff-b35cc8d48a3ea2ddde0c3bc9de0b0b768c567f9e8f532dc4ca1b260fe06e4ac8L286) (the latter not highlighted by the diff) were actually moved to own [TestComponentRenderAPI](https://github.com/django-components/django-components/pull/1180/files#diff-b35cc8d48a3ea2ddde0c3bc9de0b0b768c567f9e8f532dc4ca1b260fe06e4ac8R306)

tests/test_component.py
```diff
@@ -586,92 +525,6 @@ class SimpleComponent(Component):
...
-        parametrize=(
-            ["components_settings", "is_isolated"],
-            [
-                [{"context_behavior": "django"}, False],
-                [{"context_behavior": "isolated"}, True],
-            ],
-            ["django", "isolated"],
-        )
-    )
-    def test_render_slot_as_func(self, components_settings, is_isolated):
```
JuroOravec, line 599: And these following tests were moved to `test_slots.py`.

tests/test_slots.py
```diff
@@ -0,0 +1,351 @@
...
+        parametrize=(
+            ["components_settings", "is_isolated"],
+            [
+                [{"context_behavior": "django"}, False],
+                [{"context_behavior": "isolated"}, True],
+            ],
+            ["django", "isolated"],
+        )
+    )
+    def test_render_slot_as_func(self, components_settings, is_isolated):
```
JuroOravec, line 36: These new tests check that the `Slot` instance has its fields correctly populated when instantiated in different ways

JuroOravec:
@EmilStenstrom I saw the thumb up, so just pinging in case you forgot to submit comments (because that actually happens to me a lot 😄)

PR #1190, opened 2025-05-14
refactor: Rename `{% fill default=... %}` to `{% fill fallback=... %}`

JuroOravec:
Renamed to avoid confusion between `{% slot default %}` and `{% fill default="..." %}`

The old kwarg `{% fill default="..." %}` is still avaialble with the plan to remove it in v1.

I didn't update the documentation page in ["concepts/fundamentals/slots"](https://django-components.github.io/django-components/0.139.1/concepts/fundamentals/slots/) in this PR, because there's could more PRs after this one, so then there is one PR that solely just updates the docs page.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -322,9 +322,29 @@
             ctx.template_data["my_template_var"] = "my_value"
     ```
 
-**Miscellaneous**
+**Slots**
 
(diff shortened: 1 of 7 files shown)


EmilStenstrom (approved):
Looks good to me!

PR #1196, opened 2025-05-19
feat: slot caching

### LINKED ISSUE #1164: ComponentCache hash() does not account for slot content
I think the default behavior should be that the content of a component slot (i.e. the html inside a slot) should also be hashed and included in the ComponentCache hash.

For example, I have a badge component where I want to be able to define a slot with html instead of just a variable:

```html
{% component "badge" type="info" %}
  <i class="fas fa-band-aid"></i> <span>Some Text</span>
{% endcomponent %}
```

I think adding "slot hashing" would enable ComponentCache to cover like >99% of the use cases. I'm aware that I can still shot myself in the foot since the component context is not "isolated" by default.

I'll probably get around to looking at an implementation for this in the next month.

JuroOravec:
This makes component caching consider also the given slots.

The slot caching is not perfect - it doesn't consider when context variables in the template change. As such, it's opt-in with `include_slots=True`.

```python
class MyComponent(Component):
    class Cache:
        enabled = True
        include_slots = True
```

Closes #1164

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -486,6 +486,42 @@
     - If `Slot` was created from string via `Slot("...")`, `Slot.contents` will contain that string.
     - If `Slot` was created from a function, `Slot.contents` will contain that function.
 
+- Component caching can now take slots into account, by setting `Component.Cache.include_slots` to `True`.
+
+    ```py
+    class MyComponent(Component):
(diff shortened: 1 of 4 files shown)


JuroOravec:
cc @EmilStenstrom

PR #1199, opened 2025-05-19
refactor: slots cleanup

JuroOravec:
Small cleanup PR

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -29,7 +29,6 @@ from django.template.context import Context, RequestContext
 from django.template.loader import get_template
 from django.template.loader_tags import BLOCK_CONTEXT_KEY, BlockContext
 from django.test.signals import template_rendered
-from django.utils.html import conditional_escape
 from django.views import View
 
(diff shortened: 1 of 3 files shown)


src/django_components/component.py
```diff
@@ -2704,75 +2701,6 @@ def _call_data_methods(self, args: Any, kwargs: Any, slots: Any, context: Contex
 
         return template_data, js_data, css_data
 
-    def _normalize_slot_fills(
```
JuroOravec, line 2707: Moved `_normalize_slot_fills` to `slots.py`

src/django_components/slots.py
```diff
@@ -176,21 +199,6 @@ class Slots(NamedTuple):
 SlotName = str
 
 
-@dataclass(frozen=True)
-class SlotFill(Generic[TSlotData]):
```
JuroOravec, line 180: Refactored the code so that we don't have to pass this internal object around.

JuroOravec:
cc @EmilStenstrom

PR #1202, opened 2025-05-19
refactor: simplify slot API

### LINKED ISSUE #1096: Simplify Slot function
> Note to self, currently the functions for rendering the slots from within the Python are a bit crude - they accept the inputs that we need to pass internally.

> But from user perspective, the function interface `slot(Context, Data, SlotRef)` is needlessly complex. To make the slots more intuitive, we could simplify it to `slot(data)`, so basically the same as we know from React and Vue.

_Originally posted by @JuroOravec in https://github.com/django-components/django-components/discussions/1091#discussioncomment-12739441_

---
---

⚠️⚠️ **UPDATE: SEE THIS COMMENT FOR LATEST API https://github.com/django-components/django-components/issues/1096#issuecomment-2864543017** ⚠️⚠️ 

----
----

I wonder how to approach this:

## Background

Currently, the API for calling a slot function is:

```
slot(Context, Data, SlotRef)
```

> NOTE: `SlotRef` could be maybe more accurately called `SlotDefault`, because it's whole purpose is that when the object is coerced to string, it will render its template (the slot default). However, it's hard to tell if this object will need any extra fields until we have scoped CSS implemented.

So when one defines a `{% fill %}` like so:

```django
{% component "table" %}
  {% fill "content" data="data" default="default" %}
    <div>
      {{ my_var }}
      {{ data.var_123 }}
    </div>
  {% endfill %}
{% endcomponent %}
```

Then:
- `Context` is the Context that's used for rendering the contents of the slot fill. In other words, in the example above, we would use `Context` to render the `my_var`.
- `Data` is the object that holds the data that was passed to the `{% slot %}` tag:

   ```django
   {% slot "content" var_123=123 / %}
   ```
- `SlotRef` is an object holding the default slot value. So if we had
   ```django
   {% slot "content" %}
      I AM DEFAULT CONTENT
   {% endslot %}
   ```

   then the `I AM DEFAULT CONTENT` would be stored by the SlotRef as a template (I think)

## Challenge

Now, as I said in the linked convo, I think that, when accessing the slots from Python code, user should NOT need to know about the `Context` and `SlotRef`, and so they should be able to render the slot fill as a plain function, passing in only the  slot data.

```
slot({ "key": value })
```

So following should be possible:

```py
class MyTable(Component):
    def get_context_data(self):
        footer_content = self.input.slots["footer"]({"key": value })
        return {
            "footer": footer_content,
        }

    template = """
        {{ footer }}
    """
```

### `Context`

In order to make the `Context` optional, the Slot object should internally hold the Context within which the slot contents should be rendered.

So again if I have an example like this:
```django
{{ my_var }}     <-----
{% component "table" %}
  {% fill "content" data="data" default="default" %}
    <div>
      {{ my_var }}     <-----
      {{ data.var_123 }}
    </div>
  {% endfill %}
{% endcomponent %}
```

Then:
- If using the "isolated" mode, then `{{ my_var }}` both outside and inside the component should use the same context.
  - To achieve this, at render time, component would pass its `Component.outer_context` to the slots.
- In "django" mode, the inner `{{ my_var }}` would use a Context that is enriched with the Component's data.
  - To achieve this, slots should be given the `Context` after it has been applied Component stuff [here](https://github.com/django-components/django-components/blob/bb5de86b6992400d9539b4dedcd6fd9821c3d132/src/django_components/component.py#L1348)

Thus, the slot object could be called by the user without needing them to pass in the `Context`, beause the right context would already be prepped.

### `SlotRef`

With the SlotRef, it's a bit different. Because while the fill context can be passed down inside `Component._render_imp()`, we don't know the slot default until we come across a `{% slot %}` tag within the component.

But without the `Context` in the function signature, calling a slot would already look like this:

```
slot(data, default)
```

So we could simply allow the `default` to be `None` and treat it as an internal input to achieve

```
slot(data)
```

> NOTE: Also `Slot.content_func()` should be changed to accept only the two args.

---

## Allow slots to accept 1, 2, or 3 args?

Building on top of the above, users should be also allowed to define the slots as functions by defining functions that accept only the data:

```py
def footer(data):
    return "HELLO " + data["name"]

MyTable.render(
    slots={
        "footer": footer,
    }
)
```

For comparison, currently when defining a slot function it needs to receive all 3 inputs:

```py
def footer(ctx, slot_data, slot_ref):
    return "HELLO " + data["name"]
```

So, I'm thinking that, to make things compatible with the short format, we should:
1. Change the order of args from `slot(ctx, data, default)` to `slot(data, default, ctx)`
3. The place where we actually render the slot function in the code ([source](https://github.com/django-components/django-components/blob/bb5de86b6992400d9539b4dedcd6fd9821c3d132/src/django_components/slots.py#L576)) there we should expect that the slot function could be defined in any of the three formats:
   - `slot(data)`
   - `slot(data, default)`
   - `slot(data, default, context)`
   So we should pass in only the expected args to avoid raising a TypeError.

## [❌ REJECTED] Going beyond

With the setup above, I wonder if we should go to completion with the slot features and allow to dynamically render slots in the template too.

Currently, when we have a `{% slot %}` like so:

 ```django
 {% slot "content" key=val key2=val2 %}
    I AM DEFAULT CONTENT
 {% endslot %}
 ```

The slot fill is set by its *name*. So the corresponding slot fill has to be defined as `{% fill %}` or passed as slot to `Component.render`.

However, maybe instead of specifying the name, we could also allow to pass in a function:

 ```django
 {% slot my_slot_func key=val key2=val2 %}
    I AM DEFAULT CONTENT
 {% endslot %}
 ```

Or like so:

```py
class MyTable(Component):
    def get_context_data(self):
        footer_slot = lambda data: "HELLO " + data["name"]
        return {
            "footer_slot": footer_slot,
        }

    template = """
        {% with a=1 %}
          {% slot footer_slot a=a b=2 %}
            I'M DEFAULT
          {% endslot %}
        {% endwith %}
    """
```

In this case:
- `footer_slot` would receive an object `{ "a": 1, "b": 2 }`
- If `footer_slot` accepted a second argument, it would receive the `SlotRef` object that would hold `I'M DEFAULT`.
- If `footer_slot` accepted a third argument, it would be the Context within which the slot is being rendered.

JuroOravec:
Breaking change to simplify the slot functions API:

- Defining slot functions:
	
    Before:
	
	```py
	def slot_fn(context: Context, data: Dict, slot_ref: SlotRef):
	    isinstance(context, Context)
	    isinstance(data, Dict)
	    isinstance(slot_ref, SlotRef)

	    return "CONTENT"
	```
	
	After:
	
	```py
	def slot_fn(ctx: SlotContext):
	    assert isinstance(ctx.context, Context) # May be None
	    assert isinstance(ctx.data, Dict)
        # Renamed `SlotRef`/`slot_ref` to `SlotFallback`/`fallack`
	    assert isinstance(ctx.fallback, SlotFallback) # May be None
	
	    return "CONTENT"
	```

- Calling slot functions:

    Before:

    ```py
    def slot_fn(context: Context, data: Dict, slot_ref: SlotRef):
        return "CONTENT"

    html = slot_fn(Context(), None, None)  # All inputs were required and all positional
	html = slot(Context(), {"data1": "abc", "data2": "hello"}, None)
	html = slot(Context(), {"data1": "abc", "data2": "hello"}, "FALLBACK")
    ```

    After:

    ```py
    def slot_fn(ctx: SlotContext):
        return "CONTENT"

    # Must be wrapped in Slot instance. We auto-wrap those slots passed to components
    # So users won't have to do do `Slot(slot_fn)` unless they are doing something custom
    slot = Slot(slot_fn)
    # Can be called with 0-3 inputs
    html = slot()
    html = slot({"data1": "abc", "data2": "hello"})
    html = slot({"data1": "abc", "data2": "hello"}, fallback="FALLBACK")
    ```

Closes #1096

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -236,14 +236,102 @@
     After:
 
     ```py
-    slot = Slot(contents=lambda *a, **kw: "CONTENT")
+    slot = Slot(contents=lambda ctx: "CONTENT")
     ```
 
     Alternatively, pass the function / content as first positional argument:
 
     ```py
-    slot = Slot(lambda *a, **kw: "CONTENT")
+    slot = Slot(lambda ctx: "CONTENT")
     ```
+
+- Slot functions behavior has changed. See the new [Slots](https://django-components.github.io/django-components/latest/concepts/fundamentals/slots/) docs for more info.
+
+    - Function signature:
+
+        1. All parameters are now passed under a single `ctx` argument.
+
+            You can still access all the same parameters via `ctx.context`, `ctx.data`, and `ctx.fallback`.
+
+        2. `context` and `fallback` now may be `None` if the slot function was called outside of `{% slot %}` tag.
+
+        Before:
+
+        ```py
+        def slot_fn(context: Context, data: Dict, slot_ref: SlotRef):
+            isinstance(context, Context)
+            isinstance(data, Dict)
+            isinstance(slot_ref, SlotRef)
+
+            return "CONTENT"
+        ```
+
+        After:
+
+        ```py
+        def slot_fn(ctx: SlotContext):
+            assert isinstance(ctx.context, Context) # May be None
+            assert isinstance(ctx.data, Dict)
+            assert isinstance(ctx.fallback, SlotFallback) # May be None
+
+            return "CONTENT"
+        ```
+
+    - Calling slot functions:
+
+        1. Rather than calling the slot functions directly, you should now call the `Slot` instances.
+
+        2. All parameters are now optional.
+
+        3. The order of parameters has changed.
+
+        Before:
+
+        ```py
+        def slot_fn(context: Context, data: Dict, slot_ref: SlotRef):
+            return "CONTENT"
+
+        html = slot_fn(context, data, slot_ref)
+        ```
+
+        After:
+
+        ```py
+        def slot_fn(ctx: SlotContext):
+            return "CONTENT"
+
+        slot = Slot(slot_fn)
+        html = slot()
+        html = slot({"data1": "abc", "data2": "hello"})
+        html = slot({"data1": "abc", "data2": "hello"}, fallback="FALLBACK")
+        ```
+
+    - Usage in components:
+
+        Before:
+
+        ```python
+        class MyComponent(Component):
+            def get_context_data(self, *args, **kwargs):
+                slots = self.input.slots
+                slot_fn = slots["my_slot"]
+                html = slot_fn(context, data, slot_ref)
+                return {
+                    "html": html,
+                }
+        ```
+
+        After:
+
+        ```python
+        class MyComponent(Component):
+            def get_template_data(self, args, kwargs, slots, context):
+                slot_fn = slots["my_slot"]
+                html = slot_fn(data)
+                return {
+                    "html": html,
+                }
+        ```
 
 **Miscellaneous**
 
--- a/docs/concepts/advanced/component_caching.md
+++ b/docs/concepts/advanced/component_caching.md
@@ -131,7 +131,7 @@ MyComponent.render(
     ```py
     MyComponent.render(
         kwargs={"position": "left"},
-        slots={"content": lambda *a, **kwa: "Hello, Alice"}
+        slots={"content": lambda ctx: "Hello, Alice"}
     )
     ```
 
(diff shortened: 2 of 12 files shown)


JuroOravec:
@EmilStenstrom Would you look at this one? Thanks!

EmilStenstrom (approved):
Looks good to me! Nice little simplification.

PR #1203, opened 2025-05-20
feat: Pass Slots to {% fill %} with 'body' kwarg

JuroOravec:
This PR allows to pass `Slot` instances to `{% fill %}` tags. Basically a mechanism to render the Python-defined Slots in the template.

Closes https://github.com/django-components/django-components/issues/1171

In the discussion there were options for the kwarg name like `slot`, `contents`, or `with`.
- `slot` was confusing, and could not use `with`, because it's a Python keyword. And `contents` feels too long.
- So I set the kwarg name to `body`, as in "set the body of this fill tag"

So the final API looks like this:

```python
from django_components import component, Slot

class Table(Component):
    return {
        "my_slot": Slot(lambda ctx: "Hello, world!"),
    }
```

Then pass the slot to the `{% fill %}` tag:

```django
{% component "table" %}
  {% fill "pagination" body=my_slot / %}
{% endcomponent %}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -574,6 +574,30 @@
     - If `Slot` was created from string via `Slot("...")`, `Slot.contents` will contain that string.
     - If `Slot` was created from a function, `Slot.contents` will contain that function.
 
+- `{% fill %}` tag now accepts `body` kwarg to pass a Slot instance to fill.
+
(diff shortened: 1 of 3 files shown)


JuroOravec:
@EmilStenstrom Another slot PR 🚀

JuroOravec:
I'll merge this one as we've talked about it, and there's more (and more important) PRs to go through.

PR #1205, opened 2025-05-22
feat: Component.args/kwargs/slots and {{ component_vars.args/kwargs/s…

JuroOravec:
This is a preparation for https://github.com/django-components/django-components/issues/1186.

Before, to access the args / kwargs / slots that were passed to a component, one could either get them in `get_template_data()`, or access them via `self.input.args`, `self.input.kwargs`, etc.

```py
class MyTable(Component):
        ...

    def on_render_before(self, context, template):
        self.input.args[0] == "Hello"
```

This PR adds `self.args`, `self.kwargs`, `self.slots`, which hold the same value as the `args`, `kwargs`, `slots` passed to `get_template_data()`.

```py
class MyTable(Component):
    def on_render_before(self, context, template):
        self.args[0] == "Hello"
```

This means that if user set the typing for args / kwargs / slots, then these `self.args` etc will be also of the corresponding instance:

```py
class MyTable(Component):
    class Args(NamedTuple):
        name: str

    def on_render_before(self, context, template):
        self.args.name == "John"
```

In this PR, the `Component.args` etc are handled as dynamic attributes with `@property`.

In https://github.com/django-components/django-components/issues/1186, these will become regular attributes of the component instance.

---

Next, I also exposed these `args`, `kwargs`, and `slots` to the template as:
- `{{ component_vars.args }}`,
- `{{ component_vars.kwargs }}`
- `{{ component_vars.slots }}`

This is to again bring our API closer to the ease of use of django-cotton. In other words, the `get_context_data()` / `get_template_data()` are now optional:

```py
@register("my_table")
class MyTable(Component):
    template = """
        Table: {{ component_vars.args.0 }}
	    <table>
          {% for item in component_vars.kwargs.items %}
             ...
          {% endfor %}
	    </table>
    """
```

The tradeoff is that one then has to prefix everything with lenghty `component_vars.kwargs` to access the inputs. But I think that's fine for v1.

---

Next, since both the component and the template can access slots (as `Component.slots` and `{{ component_vars.slots }}`, I marked the `is_filled` API as deprecated and to be removed in v1.

Background:
- Originally (AKA when I joined), there was a special template tag for checking if a slot is filled.
- I introduced the `{{ component_vars.is_filled }}` API as an alternative to that, so that we don't needlessly recreate the behaviour of `{% if %}` tag. Back then I also introduced `Component.is_filled` so that the same behaviour can be accessed from Python.
- So now, with a separate `Component.slots` and `{{ component_vars.slots }}` attributes, the `is_filled` attribute is no longer needed.

### DIFF
--- a/docs/concepts/fundamentals/html_js_css_variables.md
+++ b/docs/concepts/fundamentals/html_js_css_variables.md
@@ -152,75 +152,141 @@ The difference is that:
 
 ## Accessing component inputs
 
-The component inputs are available in two ways:
+The component inputs are available in 3 ways:
 
-1. **Function arguments (recommended)**
+### Function arguments
 
-    The data methods receive the inputs as parameters, which you can access directly.
+The data methods receive the inputs as parameters directly.
 
-    ```python
+```python
+class ProfileCard(Component):
+    # Access inputs directly as parameters
+    def get_template_data(self, args, kwargs, slots, context):
+        return {
+            "user_id": args[0],
+            "show_details": kwargs["show_details"],
+        }
+```
+
+!!! info
+
+    By default, the `args` parameter is a list, while `kwargs` and `slots` are dictionaries.
+
+    If you add typing to your component with
+    [`Args`](../../../reference/api/#django_components.Component.Args),
+    [`Kwargs`](../../../reference/api/#django_components.Component.Kwargs),
+    or [`Slots`](../../../reference/api/#django_components.Component.Slots) classes,
+    the respective inputs will be given as instances of these classes.
+
+    Learn more about [Component typing](../../fundamentals/typing_and_validation).
+
+    ```py
     class ProfileCard(Component):
-        def get_template_data(self, args, kwargs, slots, context):
-            # Access inputs directly as parameters
+        class Args(NamedTuple):
+            user_id: int
+
+        class Kwargs(NamedTuple):
+            show_details: bool
+
+        # Access inputs directly as parameters
+        def get_template_data(self, args: Args, kwargs: Kwargs, slots, context):
             return {
-                "user_id": user_id,
-                "show_details": show_details,
+                "user_id": args.user_id,
+                "show_details": kwargs.show_details,
             }
     ```
 
-    !!! info
+### `args`, `kwargs`, `slots` properties
 
-        By default, the `args` parameter is a list, while `kwargs` and `slots` are dictionaries.
+In other methods, you can access the inputs via
+[`self.args`](../../../reference/api/#django_components.Component.args),
+[`self.kwargs`](../../../reference/api/#django_components.Component.kwargs),
+and [`self.slots`](../../../reference/api/#django_components.Component.slots) properties:
 
-        If you add typing to your component with
-        [`Args`](../../../reference/api/#django_components.Component.Args),
-        [`Kwargs`](../../../reference/api/#django_components.Component.Kwargs),
-        or [`Slots`](../../../reference/api/#django_components.Component.Slots) classes,
-        the respective inputs will be given as instances of these classes.
+```py
+class ProfileCard(Component):
+    def on_render_before(self, *args, **kwargs):
+        # Access inputs via self.args, self.kwargs, self.slots
+        self.args[0]
+        self.kwargs.get("show_details", False)
+        self.slots["footer"]
+```
 
-        Learn more about [Component typing](../../advanced/typing_and_validation).
+!!! info
 
-2. **`self.input` property**
+    These properties work the same way as `args`, `kwargs`, and `slots` parameters in the data methods:
 
-    The data methods receive only the main inputs. There are additional settings that may be passed
-    to components. If you need to access these, you can do so via the [`self.input`](../../../reference/api/#django_components.Component.input) property.
+    By default, the `args` property is a list, while `kwargs` and `slots` are dictionaries.
 
-    The `input` property contains all the inputs passed to the component (instance of [`ComponentInput`](../../../reference/api/#django_components.ComponentInput)).
+    If you add typing to your component with
+    [`Args`](../../../reference/api/#django_components.Component.Args),
+    [`Kwargs`](../../../reference/api/#django_components.Component.Kwargs),
+    or [`Slots`](../../../reference/api/#django_components.Component.Slots) classes,
+    the respective inputs will be given as instances of these classes.
 
-    This includes:
+    Learn more about [Component typing](../../fundamentals/typing_and_validation).
 
-    - [`input.args`](../../../reference/api/#django_components.ComponentInput.args) - List of positional arguments
-    - [`input.kwargs`](../../../reference/api/#django_components.ComponentInput.kwargs) - Dictionary of keyword arguments
-    - [`input.slots`](../../../reference/api/#django_components.ComponentInput.slots) - Dictionary of slots. Values are normalized to [`Slot`](../../../reference/api/#django_components.Slot) instances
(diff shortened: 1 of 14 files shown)


src/django_components/components/dynamic.py
```diff
@@ -99,35 +99,24 @@ class DynamicComponent(Component):
 
     _is_dynamic_component = True
 
-    def get_template_data(
```
JuroOravec, line 102: Slightly change the implementation of the dynamic component. I realized that instead of:
1. Returning the `comp_class`, `args`, and `kwargs` from `get_template_data()`, 
2. Accessing it in `on_render_before()`.

The same data can be already prepared in `on_render_before()`, so there's no need to pass it through.

docs/concepts/fundamentals/render_api.md
```diff
@@ -49,42 +46,159 @@ rendered = Table.render(
...
+- [`self.args`](../render_api/#args) - The positional arguments for the current render call
+- [`self.kwargs`](../render_api/#kwargs) - The keyword arguments for the current render call
+- [`self.slots`](../render_api/#slots) - The slots for the current render call
 - [`self.input`](../render_api/#component-inputs) - All the component inputs
+- [`self.id`](../render_api/#component-id) - The unique ID for the current render call
 - [`self.request`](../render_api/#request-object-and-context-processors) - The request object (if available)
 - [`self.context_processors_data`](../render_api/#request-object-and-context-processors) - Data from Django's context processors (if request is available)
 - [`self.inject()`](../render_api/#provide-inject) - Inject data into the component
 
-## Component ID
```
JuroOravec, line 58: The section on component ID wasn't deleted, only moved to [line 231](https://github.com/django-components/django-components/pull/1205/files#diff-9fbdf12fc67b61f0e13b4b521ee3375a107cdf4a87b9b5403a453540ec06bdd5R231).

docs/concepts/fundamentals/render_api.md
```diff
@@ -49,42 +46,159 @@ rendered = Table.render(
...
+class Table(Component):
+    def on_render_before(self, context: Context, template: Template) -> None:
+        assert isinstance(self.slots["header"], Slot)
+        assert isinstance(self.slots["footer"], Slot)
 ```
 
 ## Component inputs
 
-All the component inputs are captured and available as [`self.input`](../../../reference/api/#django_components.Component.input).
+You can access the most important inputs via [`self.args`](../render_api/#args),
```
JuroOravec, line 195: Added the `Component.args`, `Component.kwargs`, and `Component.slots` to the docs. And also started referring to the old `Component.input` as "low-level" access, because it contains the args/kwargs/slots as plain lists/dicts.

docs/concepts/fundamentals/html_js_css_variables.md
```diff
@@ -152,75 +152,141 @@ The difference is that:
...
-                "show_details": show_details,
+                "user_id": args.user_id,
+                "show_details": kwargs.show_details,
             }
     ```
 
-    !!! info
-
-        By default, the `args` parameter is a list, while `kwargs` and `slots` are dictionaries.
+### `args`, `kwargs`, `slots` properties
```
JuroOravec, line 199: Updated the docs section on how to access the args / kwargs / slots.

src/django_components/component.py
```diff
@@ -199,15 +214,174 @@ class ComponentVars(NamedTuple):
     Type for the variables available inside the component templates.
 
     All variables here are scoped under `component_vars.`, so e.g. attribute
-    `is_filled` on this class is accessible inside the template as:
+    `kwargs` on this class is accessible inside the template as:
 
     ```django
-    {{ component_vars.is_filled }}
+    {{ component_vars.kwargs }}
```
JuroOravec, line 220: These are the fields and docs for `{{ component_vars.kwargs }}` and such.

src/django_components/component.py
```diff
@@ -2346,34 +2706,7 @@ def _render_impl(
         if not isinstance(context, Context):
             context = RequestContext(request, context) if request else Context(context)
 
-        # Required for compatibility with Django's {% extends %} tag
```
JuroOravec, line 2349: Moved this further down

src/django_components/component.py
```diff
@@ -2478,15 +2862,17 @@ def _render_impl(
...
+        #############################################################################
+
         with _prepare_template(self, context, template_data, metadata) as template:
             component_ctx.template_name = template.name
 
-            # For users, we expose boolean variables that they may check
-            # to see if given slot was filled, e.g.:
-            # `{% if variable > 8 and component_vars.is_filled.header %}`
-            is_filled = SlotIsFilled(slots_dict)
-            metadata.is_filled = is_filled
```
JuroOravec, line 2488: Moved this to [line 2828](https://github.com/django-components/django-components/pull/1205/files#diff-35c84ebceae50f7149f03bbcaf2b4f1289bc2984bf7859ca521b5dbe7960c396R2828)

src/django_components/component.py
```diff
@@ -2496,7 +2882,15 @@ def _render_impl(
                     # NOTE: Public API for variables accessible from within a component's template
                     # See https://github.com/django-components/django-components/issues/280#issuecomment-2081180940
                     "component_vars": ComponentVars(
-                        is_filled=is_filled,
+                        args=args_inst,
```
JuroOravec, line 2885: This exposes the `{{ component_vars.args }}` etc.

src/django_components/component.py
```diff
@@ -199,15 +214,174 @@ class ComponentVars(NamedTuple):
...
+            per_page: int
+
+        template = '''
+            <div>
+                <h1>Table</h1>
+                <p>Page: {{ component_vars.args.page }}</p>
+                <p>Per page: {{ component_vars.args.per_page }}</p>
+            </div>
+        '''
+    ```
```
EmilStenstrom, line 257: This is a really nice shortcut. You need to do some mental gymnastics to understand that `Table.Args` is filled by `{% component "table" 1 2 3 %}` but that should be implied by the name. 

Maybe we should push this format more?

EmilStenstrom (approved):

docs/concepts/fundamentals/html_js_css_variables.md
```diff
@@ -152,75 +152,141 @@ The difference is that:
...
-    !!! info
+### `input` property (low-level)
+
+The previous two approaches allow you to access only the most important inputs.
+
+There are additional settings that may be passed to components.
+If you need to access these, you can use [`self.input`](../../../reference/api/#django_components.Component.input) property
+for a low-level access to all the inputs.
+
+The `input` property contains all the inputs passed to the component (instance of [`ComponentInput`](../../../reference/api/#django_components.ComponentInput)).
```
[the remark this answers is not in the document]
JuroOravec: I think we can keep it for now, it ensures that there's a way to access all the inputs that a component received when rendered.

Without `input`, we would either have to:
- Make some render inputs inaccessible 
- Or set all inputs as Component attributes

Altho, looking at my latest code, all the inputs to `Component.render()` already are set on the Component instance as attributes, except for `deps_strategy`...

On the other hand, the `input.args`, `input.kwargs`, `input.slots` are handy when writing extensions. Because these are plain dicts / lists. Whereas the `args` / `kwargs` / `slots` on the Component instance may be instances of any exotic class that might not be possible to revert back to dict / lists.

One more for removing `input` - I see that it has already drifted - that I forgot to add some fields to the `input` object that are being passed to `Component.render()`.

So maybe, we could remove `input`, and instead:
1. Add `deps_strategy` as Component instance attribute
2. Add `raw_args`, `raw_kwargs`, `raw_slots` instance attributes, which would how the plain lists / dicts.

src/django_components/component.py
JuroOravec: Good idea, captured it here https://github.com/django-components/django-components/issues/1207

PR #1208, opened 2025-05-24
docs: update docs on slots

JuroOravec:
Documents the recent changes to slots.

Here's preview of it. I've moved the video to Youtube, because Github has file limit 10mb, and the last time I tried to compress similar video, it ended up choppy as hell 😄 

https://youtu.be/x6yCEwt0rUk

- All the ways how to use slots / fills is now documented in consice overview.
- Added section on the Slot class
- `{{ component_vars.is_filled }}` was moved to the bottom as legacy section.

Closes https://github.com/django-components/django-components/issues/1095


JuroOravec:
cc @EmilStenstrom

PR #1209, opened 2025-05-25
feat: on_slot_rendered extension hook + refactor debug highlight as extension

JuroOravec:
Added an extension hook to be able to post-process when a `{% slot %}` is rendered. This will be required for CSS scoping, AlpineJS integration, and maybe more.

To test that the hook is working I've refactored the highlighting of components / slots into an extension.


src/django_components/extensions/debug_highlight.py
```diff
@@ -0,0 +1,131 @@
...
+        </style>
+        <div class="{type}-highlight-{highlight_id}" style="border: 1px solid {color.border_color}">
+            {output}
+        </div>
+    """
+
+    return output
+
+
+# TODO - Deprecate `DEBUG_HIGHLIGHT_SLOTS` and `DEBUG_HIGHLIGHT_COMPONENTS` (with removal in v1)
```
JuroOravec, line 48: This is important comment. After this PR, the highlighting is still enabled by the two settings `DEBUG_HIGHLIGHT_SLOTS` and `DEBUG_HIGHLIGHT_COMPONENTS`.

However, since the highlighting is now implemented as an extension, it measn that users can now theoretically override the behavior on component level by setting
- `Component.DebugHighlight.highlight_components`
- `Component.DebugHighlight.highlight_slots`

```py
class Table(Component):
    class DebugHighlight:
        highlight_components = True
        highlight_components = False
```

But then, once https://github.com/django-components/django-components/issues/1159 is merged, then instead of configuring the settings, people could enable / disable highlighting by setting the extension defaults:

```py
# settings.py

COMPONENTS = {
    "extensions_defaults": {
        "debug_highlight": {
            "highlight_components": True,
            "highlight_slots": True,
        }
    }
}
```

src/django_components/extensions/debug_highlight.py
```diff
@@ -0,0 +1,131 @@
...
+
+    # TODO_v1 - Remove `DEBUG_HIGHLIGHT_COMPONENTS` and `DEBUG_HIGHLIGHT_SLOTS`
+    #           Instead set this as plain boolean fields.
+    highlight_components = HighlightComponentsDescriptor()
+    """Whether to highlight this component in the rendered output."""
+    highlight_slots = HighlightSlotsDescriptor()
+    """Whether to highlight slots of this component in the rendered output."""
+
+
+# TODO_v1 - Move into standalone extension (own repo?) and ask people to manually add this extension in settings.
```
JuroOravec, line 100: Also, as pointed out here, for v1 it might make sense to move this to it's own repo / project.

PR #1210, opened 2025-05-25
fix: KeyError on component_context_cache when slot rendered outside

JuroOravec:
Fixes https://github.com/django-components/django-components/issues/1189


JuroOravec:
cc @EmilStenstrom

PR #1212, opened 2025-05-25
refactor: Instantiate component when rendering, and remove metadata stack

JuroOravec:
Changes (taken from changelog)

- When a component is being rendered, a proper `Component` instance is now created.

    Previously, the `Component` state was managed as half-instance, half-stack.

- Component's "Render API" (args, kwargs, slots, context, inputs, request, context data, etc)
  can now be accessed also outside of the render call. So now its possible to take the component
  instance out of `get_template_data()`.

    ```py
    comp = None

    class TestComponent(Component):
        template = "..."

            nonlocal comp
            comp = self

    TestComponent.render(
        args=[1],
        kwargs={"foo": "bar"},
        slots={"content": "Hello, world!"},
    )

    assert comp.args == [1]
    assert comp.kwargs == {"foo": "bar"}
    assert comp.slots == {"content": "Hello, world!"}
    assert comp.context == Context()
    ```

- The kwargs that the `Component` constructor  accepted - `registered_name`, `outer_context`, and `registry` - should no longer be passed to the Component constructor, and instead should be passed to `Component.render()`.

    Before, you could instantiate a standalone component,
    and then call `render()` on the instance:

    ```py
    comp = MyComponent(
        registered_name="my_component",
        outer_context=my_context,
        registry=my_registry,
    )
    comp.render(
        args=[1, 2, 3],
        kwargs={"a": 1, "b": 2},
        slots={"my_slot": "CONTENT"},
    )
    ```

    Now you should instead pass all that data to `Component.render()` / `Component.render_to_response()`:

    ```py
    MyComponent.render(
        args=[1, 2, 3],
        kwargs={"a": 1, "b": 2},
        slots={"my_slot": "CONTENT"},
        # NEW
        registered_name="my_component",
        outer_context=my_context,
        registry=my_registry,
    )
    ```

- The way to access the component class from within component views is now different.

    Instead of `self.component`, use `self.component_cls`. `self.component` will be removed in v1.

    Before:

    ```py
    class MyView(View):
        def get(self, request):
            return self.component.render_to_response(request=request)
    ```

    After:

    ```py
    class MyView(View):
        def get(self, request):
            return self.component_cls.render_to_response(request=request)
    ```

Closes https://github.com/django-components/django-components/pull/1212

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -867,28 +867,6 @@ Summary:
   can now be accessed also outside of the render call. So now its possible to take the component
   instance out of `get_template_data()` (although this is not recommended).
 
-    ```py
-    comp = None
-
-    class TestComponent(Component):
-        template = "..."
-
-        def get_template_data(self, args, kwargs, slots, context):
-            nonlocal comp
-            comp = self
-
-    TestComponent.render(
-        args=[1],
-        kwargs={"foo": "bar"},
-        slots={"content": "Hello, world!"},
-    )
-
-    assert comp.args == [1]
-    assert comp.kwargs == {"foo": "bar"}
-    assert comp.slots == {"content": "Hello, world!"}
-    assert comp.context == Context()
-    ```
-
 #### Fix
 
 - Fix bug: Context processors data was being generated anew for each component. Now the data is correctly created once and reused across components with the same request ([#1165](https://github.com/django-components/django-components/issues/1165)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -430,6 +430,58 @@ Summary:
     Calendar.render_to_response(deps_strategy="ignore")
     ```
 
+- Support for `Component` constructor kwargs `registered_name`, `outer_context`, and `registry` is deprecated, and will be removed in v1.
+
+    Before, you could instantiate a standalone component,
+    and then call `render()` on the instance:
+
+    ```py
+    comp = MyComponent(
+        registered_name="my_component",
+        outer_context=my_context,
+        registry=my_registry,
+    )
+    comp.render(
+        args=[1, 2, 3],
+        kwargs={"a": 1, "b": 2},
+        slots={"my_slot": "CONTENT"},
+    )
+    ```
+
+    Now you should instead pass all that data to `Component.render()` / `Component.render_to_response()`:
+
+    ```py
+    MyComponent.render(
+        args=[1, 2, 3],
+        kwargs={"a": 1, "b": 2},
+        slots={"my_slot": "CONTENT"},
+        # NEW
+        registered_name="my_component",
+        outer_context=my_context,
+        registry=my_registry,
+    )
+    ```
+
+- If you are using the Components as views, the way to access the component class is now different.
+
+    Instead of `self.component`, use `self.component_cls`. `self.component` will be removed in v1.
+
+    Before:
+
+    ```py
+    class MyView(View):
+        def get(self, request):
+            return self.component.render_to_response(request=request)
+    ```
+
+    After:
+
+    ```py
+    class MyView(View):
+        def get(self, request):
+            return self.component_cls.render_to_response(request=request)
+    ```
+
 **Extensions**
 
 - In the `on_component_data()` extension hook, the `context_data` field of the context object was superseded by `template_data`.
@@ -805,11 +857,45 @@ Summary:
 
     See all [Extension hooks](https://django-components.github.io/django-components/0.140/reference/extension_hooks/).
 
+#### Refactor
+
+- When a component is being rendered, a proper `Component` instance is now created.
+
+    Previously, the `Component` state was managed as half-instance, half-stack.
+
+- Component's "Render API" (args, kwargs, slots, context, inputs, request, context data, etc)
+  can now be accessed also outside of the render call. So now its possible to take the component
+  instance out of `get_template_data()` (although this is not recommended).
+
+    ```py
+    comp = None
+
+    class TestComponent(Component):
+        template = "..."
+
+        def get_template_data(self, args, kwargs, slots, context):
+            nonlocal comp
+            comp = self
+
+    TestComponent.render(
+        args=[1],
+        kwargs={"foo": "bar"},
+        slots={"content": "Hello, world!"},
+    )
+
+    assert comp.args == [1]
+    assert comp.kwargs == {"foo": "bar"}
+    assert comp.slots == {"content": "Hello, world!"}
+    assert comp.context == Context()
+    ```
+
 #### Fix
 
 - Fix bug: Context processors data was being generated anew for each component. Now the data is correctly created once and reused across components with the same request ([#1165](https://github.com/django-components/django-components/issues/1165)).
 
 - Fix KeyError on `component_context_cache` when slots are rendered outside of the component's render context. ([#1189](https://github.com/django-components/django-components/issues/1189))
+
+- Component classes now have `do_not_call_in_templates=True` to prevent them from being called as functions in templates.
 
 ## v0.139.1
 
--- a/docs/concepts/advanced/html_fragments.md
+++ b/docs/concepts/advanced/html_fragments.md
@@ -136,7 +136,7 @@ class MyPage(Component):
 
     class View:
         def get(self, request):
-            return self.component.render_to_response(request=request)
+            return self.component_cls.render_to_response(request=request)
 ```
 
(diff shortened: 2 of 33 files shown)


docs/concepts/advanced/html_fragments.md
```diff
@@ -136,7 +136,7 @@ class MyPage(Component):
 
     class View:
         def get(self, request):
-            return self.component.render_to_response(request=request)
+            return self.component_cls.render_to_response(request=request)
```
JuroOravec, line 139: As I was going through all places where we instantiate Components, I realized that we also do so when executing view functions like `Component.View.get()`.

There, we make a dummy instance and assign it to `View.component`.

But with this PR, it no longer makes sense to manually instantiate Components.

It's also a bit misleading, because `View.get()` is called BEFORE any component is rendered.

So instead of having the access to the dummy component instance, I added this `View.component_cls` field that points just to component CLASS, which better reflects what's actually avilable.

docs/concepts/fundamentals/slots.md
```diff
@@ -770,10 +770,10 @@ print(slot.nodelist)  # <django.template.Nodelist: ['Hello!']>
 
 Slots content are automatically escaped by default to prevent XSS attacks.
 
-In other words, it's as if you would be using Django's [`mark_safe()`](https://docs.djangoproject.com/en/5.2/ref/utils/#django.utils.safestring.mark_safe) function on the slot content:
+In other words, it's as if you would be using Django's [`escape()`](https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#std-templatefilter-escape) on the slot contents / result:
```
JuroOravec, line 773: Changees in this file is leftover from last PR.

docs/getting_started/rendering_components.md
```diff
@@ -164,10 +164,10 @@ def my_view(request):
 
 Slots content are automatically escaped by default to prevent XSS attacks.
 
-In other words, it's as if you would be using Django's [`mark_safe()`](https://docs.djangoproject.com/en/5.2/ref/utils/#django.utils.safestring.mark_safe) function on the slot content:
+In other words, it's as if you would be using Django's [`escape()`](https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#std-templatefilter-escape) on the slot contents / result:
```
JuroOravec, line 167: Same, leftover from last PR.

docs/overview/migrating.md
```diff
@@ -0,0 +1,25 @@
+Django-components is still in active development.
```
JuroOravec, line 1: Not related to this PR, but prepared a page for documenting large migrations.

tests/test_component_media.py
```diff
@@ -1047,6 +1046,13 @@ class Media:
         assertInHTML('<script src="child.js"></script>', rendered)
         assertInHTML('<script src="parent.js"></script>', rendered)
 
+        assert str(ChildComponent.media) == (
```
JuroOravec, line 1049: Unrelated to this PR, but added tests to check that coercing to string the `Component.media` attribute behaves same as in Django, where it prints out all JS / CSS as `<link>` / `<script>` tags.

tests/test_context.py
```diff
@@ -916,25 +914,24 @@ def get_template_data(self, args, kwargs, slots, context):
         assert child_data["dummy"] == "a1bc3f"
         assert parent_data["csrf_token"] == child_data["csrf_token"]
 
-    def test_raises_on_accessing_context_processors_data_outside_of_rendering(self):
+    def test_context_processors_data_outside_of_rendering(self):
```
JuroOravec, line 917: This test previously checked that we can't use Component instance outside of the component methods. Now it tests that we can 🎉

src/django_components/slots.py
```diff
@@ -677,9 +685,9 @@ def render(self, context: Context, name: str, **kwargs: Any) -> SafeString:
         # In this case, we need to find the context that was used to render the component,
         # and use the fills from that context.
         if (
-            component_ctx.registry.settings.context_behavior == ContextBehavior.DJANGO
-            and component_ctx.outer_context is None
-            and (slot_name not in component_ctx.fills)
+            component.registry.settings.context_behavior == ContextBehavior.DJANGO
```
JuroOravec, line 688: One nice side effect of this change is that now inside the `{% slot %}` nodes we can access the `Component` instance that owns those slots.

src/django_components/extensions/defaults.py
```diff
@@ -171,11 +171,3 @@ class Defaults:
     def on_component_class_created(self, ctx: OnComponentClassCreatedContext) -> None:
         defaults_cls = getattr(ctx.component_cls, "Defaults", None)
         defaults_by_component[ctx.component_cls] = _extract_defaults(defaults_cls)
-
-    # Apply defaults to missing or `None` values in `kwargs`
-    def on_component_input(self, ctx: OnComponentInputContext) -> None:
```
JuroOravec, line 176: The changes in this file will be actually reverted in the next PR.

As I was trying to figure out how to prepare and pass around the data, I came across a chicken-and-egg situation where:

1. I wanted to apply the defaults from `Component.Defaults` to the component input before calling `extensions.on_component_input`.
2. But component defaults are themselves an extension, which applied the defaults when `extensions.on_component_input` was called.

So for this PR I called `apply_defaults()` from this file directly in the `component.py`.

Only later I realized that I can avoid that if I just move things around a bit more.

And since there's still a couple of more PRs prepared after this, I didn't want to revert the change in just this one PR, because it could then make a mess for all the downstream changes.

tests/test_component.py
```diff
@@ -103,6 +103,55 @@ class Media:
             """,
         )
 
+    # TODO_REMOVE_IN_V1 - Registry and registered name should be passed to `Component.render()`,
+    #                     not to the constructor.
+    @djc_test(parametrize=PARAMETRIZE_CONTEXT_BEHAVIOR)
+    def test_component_instantiation(self, components_settings):
```
JuroOravec, line 109: This test checks the old way of passing data to the component via its instance:

```py
MyComponent("my_comp").render(...)
```

src/django_components/component.py
```diff
@@ -1798,344 +1846,373 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
         extensions._init_component_class(cls)
         extensions.on_component_class_created(OnComponentClassCreatedContext(cls))
 
-    @contextmanager
-    def _with_metadata(self, item: MetadataItem) -> Generator[None, None, None]:
-        self._metadata_stack.append(item)
-        yield
-        self._metadata_stack.pop()
+    ########################################
+    # INSTANCE PROPERTIES
```
JuroOravec, line 1850: In this section below are the newly-assigned instance properties. For most part, these were previously property getters that took the value from the stack, and now they are plain attributes.

src/django_components/component.py
```diff
@@ -2692,13 +2796,38 @@ def _render_impl(
...
+        render_id = _gen_component_id()
+
+        # Apply defaults to missing or `None` values in `kwargs`
+        defaults = defaults_by_component.get(comp_cls, None)
+        if defaults is not None:
+            apply_defaults(kwargs_dict, defaults)
+
+        # If user doesn't specify `Args`, `Kwargs`, `Slots` types, then we pass them in as plain
+        # dicts / lists.
+        args_inst = comp_cls.Args(*args_list) if comp_cls.Args is not None else args_list
```
JuroOravec, line 2808: This was previously on the [line 2782](https://github.com/django-components/django-components/pull/1212/files#diff-35c84ebceae50f7149f03bbcaf2b4f1289bc2984bf7859ca521b5dbe7960c396L2782)

JuroOravec:
@EmilStenstrom Another one to review, this one's a bit bigger, so ofc no rush :)

CHANGELOG.md
```diff
@@ -805,12 +857,46 @@ Summary:
...
+        kwargs={"foo": "bar"},
+        slots={"content": "Hello, world!"},
+    )
+
+    assert comp.args == [1]
+    assert comp.kwargs == {"foo": "bar"}
+    assert comp.slots == {"content": "Hello, world!"}
+    assert comp.context == Context()
+    ```
+
```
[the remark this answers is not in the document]
JuroOravec: Fair point :D

JuroOravec:
@EmilStenstrom Thansk for quick feedback! And yes, it's very satisfying to watch how the Python side of components is  coming together into a simple yet powerful shape :)

PR #1215, opened 2025-05-26
feat: extension defaults + docs + API cleanup

JuroOravec:
Closes https://github.com/django-components/django-components/issues/1159

### DIFF
--- a/src/django_components/extension.py
+++ b/src/django_components/extension.py
@@ -1,5 +1,19 @@
 from functools import wraps
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, NamedTuple, Optional, Set, Tuple, Type, TypeVar, Union
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    List,
+    NamedTuple,
+    Optional,
+    Set,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+)
 
 import django.urls
 from django.template import Context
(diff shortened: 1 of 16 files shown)


src/django_components/extension.py
```diff
@@ -154,43 +168,159 @@ class OnSlotRenderedContext(NamedTuple):
...
+        if "ExtensionClass" in attrs:
+            attrs["ComponentConfig"] = attrs.pop("ExtensionClass")
+
+        return super().__new__(mcs, name, bases, attrs)
+
+
 # NOTE: This class is used for generating documentation for the extension hooks API.
 #       To be recognized, all hooks must start with `on_` prefix.
-class ComponentExtension:
+class ComponentExtension(metaclass=ExtensionMeta):
```
JuroOravec, line 251: I updated the name of the class inside the Extension that defines how the component-level classes should behave.  Using `ExtensionClass` was confusing. So now instead it's called `ComponentConfig`.

I added a metaclass here to keep things backwards compatible till v1.
The metaclass checks if `ExtensionClass` was defined, and renames it to `ComponentConfig`.

src/django_components/extensions/cache.py
```diff
@@ -15,7 +16,7 @@
 CACHE_KEY_PREFIX = "components:cache:"
 
 
-class ComponentCache(ComponentExtension.ExtensionClass):  # type: ignore
```
JuroOravec, line 18: Another minor issue with the Extension classes was that to define the inner class (now "ComponentConfig"), it had to subclass from

`ComponentExtension.ExtensionClass`

But Mypy doesn't like that, and it doesn't allow to use attributes when specifying bases for subclassing.

So now the base class of the inner class is part of public API as `ExtensionComponentConfig`, so that extensions can be created without this annoying error.

src/django_components/extension.py
```diff
@@ -629,52 +788,86 @@ def _init_component_class(self, component_cls: Type["Component"]) -> None:
...
             # ```
             ext_class_override_attr = extension.name + "_class"  # "my_extension_class"
-            ext_base_class = getattr(component_cls, ext_class_override_attr, extension.ExtensionClass)
+            ext_base_class = getattr(component_cls, ext_class_override_attr, extension.ComponentConfig)
 
-            if component_ext_subclass:
-                bases: tuple[Type, ...] = (component_ext_subclass, ext_base_class)
-            else:
-                bases = (ext_base_class,)
+            # Extensions have 3 levels of configuration:
```
JuroOravec, line 829: Here below is the actual implementation of the feature.

JuroOravec:
Here's how new section on extensions defaults looks like:

Screenshot 2025-05-26 at 12 39 29

Screenshot 2025-05-26 at 12 41 46

JuroOravec:
@EmilStenstrom Another one ready for review 🚀

EmilStenstrom (approved):
Looks good to me! These PRs always sounds so innocent, but then I click in and it's a 1000 line diff instead :)

PR #1221, opened 2025-05-27
feat: Slot.extra and Slot.source metadata

JuroOravec:


JuroOravec:
@EmilStenstrom Another one for review!

EmilStenstrom (approved):
Makes sense, looks good!

PR #1222, opened 2025-05-31
refactor: deprecate template caching, get_template_name, get_template, assoc template with Comp cls

JuroOravec:
Sorry for big PR, I already removed one part of it, but what remains is all too closely related.

Changes:

## Component template API

Changes that relate to how the component templates are loaded / created:

- `Component.template` no longer accepts a Template instance*, only plain string.
- `Component.get_template_name()` and `Component.get_template()` are now deprecated. 

In other words, the Template should now be specified as plain string (`Component.template`) or file (`Component.template_file`).

With these changes, we will have control over how the template string is compiled into `Template` instance, and will allow us (or our extensions) to pre-process the template. Thus we can implement Vue-like or django-cotton-like syntax as an extension.

As a side effect, we can also deprecate `template_cache_size` setting and `cached_template()` helper, since these were related to caching `Template` instances from those deprecated source above.

\* NOTE: This was something I introduced long time ago, but I don't think it was every properly documented.

## Associating Templates with Components

There's two use cases for which I need to be able to know if a loaded `Template` belongs to a component or not:

1. Pre-processing component templates
2. Know in which Component a `{% slot %}` tag was defined, so we can apply CSS scoping for that Component class.

Implementation:
 - To make it possible to write extensions for using Vue / django-cotton / markdown syntax instead of vanilla HTML + Django, I need to be able to modify the raw template string before Django compiles it into Template / NodeList.

 - When a template is inlined within Component as `Component.template = "<div>...<div>"`, this is easy, because we directly have the string available.
 - But when the template is loaded from file as `Component.template_file = "my_file.html"`, this is a problem, because it's Django who does both the file reading and template compiling.
 - I found a way to intercept this process by overriding the `Template.__init__()` method because this is when Django compiles the template (https://github.com/django-components/django-components/pull/1222/files#diff-6cad9d5565786c00d355ba20ae181e9ebc658eca28f7d762dd588172c0d6fdc6R26)
 - The problem, still was - even if I intercept calls to `Template.__init__()`, how do I know if the template that is being instantiated belongs to a Component?
 - The solution was hacky, but, well, it works (and most of the complexity is hidden away in the `template.py` file):
   1. When a Component class is defined, we remember if the component specified a file under `Component.template_file` (see https://github.com/django-components/django-components/pull/1222/files#diff-35c84ebceae50f7149f03bbcaf2b4f1289bc2984bf7859ca521b5dbe7960c396R423)
   3. When we get around to resolving and loading component's template (it's done lazily), we hand over to Django's `get_template()`, which uses template loaders to eventually instantiate `Template`.
   4. When `Template` is instantiated like this, it receives an `Origin` object, which describes the file name from which the string was read. This means that inside `Template.__init__()`, we can access this Origin to know where the template came from.
   5. We can thus compare this file path from the Origin against the `template_file` entries we remembered from the first step (see https://github.com/django-components/django-components/pull/1222/files#diff-6cad9d5565786c00d355ba20ae181e9ebc658eca28f7d762dd588172c0d6fdc6R46 and https://github.com/django-components/django-components/pull/1222/files#diff-bfd0b555c48797bfe54ed731d42c3ef507cfc992ed1f49359f806849c20f7a21R378).
   6. Thus we are able to say that if we have found a Component class for given template file path, then the Template belongs to that Component.
 - NOTE: There is more nuance to this implementation:
    - Multiple components may use the same `template_file` (e.g. when subclassing from same base). So there was a challenge of "how do I know to which Component does this `Template` instance belong to?"
      - Trick to that was to use a global state to store the Component class that we are currently loading (see https://github.com/django-components/django-components/pull/1222/files#diff-bfd0b555c48797bfe54ed731d42c3ef507cfc992ed1f49359f806849c20f7a21R169)
        - -> If it is set, that's the one
        - -> If not set, then the `Template` instance is NOT being created by us, and so we DON'T associate the Component class to it.
      - Another tricky part of multiple components with same `template_file` is that when you use cached loader, then the `Template` instance is reused. So we have to check for that and make `Template` copies. (see https://github.com/django-components/django-components/pull/1222/commits/0e6965d4dd75a4c025a62ed6bb6982acf4cad2ae)
    - This all has to be done lazily. Because if `Component.template_file` is defined relative to the Component's file, we need to resolve that first.

 - Maybe the best way to summarize this is through code (taken from [this test](https://github.com/django-components/django-components/pull/1222/files#diff-b35cc8d48a3ea2ddde0c3bc9de0b0b768c567f9e8f532dc4ca1b260fe06e4ac8R340)):

    ```py
    # Two unrelated components point to the same template
    class SimpleComponent1(Component):
        template_file = "simple_template.html"

    class SimpleComponent2(Component):
        template_file = "simple_template.html"

    # Each component has its own Template instance
    assert isinstance(SimpleComponent1._template, Template)
    assert isinstance(SimpleComponent2._template, Template)
    assert SimpleComponent1._template is not SimpleComponent2._template
    # But still point to the same file
    assert SimpleComponent1._template.source == SimpleComponent2._template.source

    # The Templates have different Origin instances
    assert SimpleComponent1._template.origin is not SimpleComponent2._template.origin

    # And these Origins point to the respective Component classes
    assert SimpleComponent1._template.origin.component_cls == SimpleComponent1
    assert SimpleComponent2._template.origin.component_cls == SimpleComponent2
    ```

Closes https://github.com/django-components/django-components/issues/901, closes https://github.com/django-components/django-components/issues/709

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -186,6 +186,40 @@ Summary:
     )
     ```
 
+- `Component.template` no longer accepts a Template instance, only plain string.
+
+    Before:
+
+    ```py
+    class MyComponent(Component):
+        template = Template("{{ my_var }}")
+    ```
+
+    Instead, either:
+
+    1. Set `Component.template` to a plain string.
+
+        ```py
+        class MyComponent(Component):
+            template = "{{ my_var }}"
+        ```
+
+    2. Move the template to it's own HTML file and set `Component.template_file`.
+
+        ```py
+        class MyComponent(Component):
+            template_file = "my_template.html"
+        ```
+
+    3. Or, if you dynamically created the template, render the template inside `Component.on_render()`.
+
+        ```py
+        class MyComponent(Component):
+            def on_render(self, context, template):
+                dynamic_template = do_something_dynamic()
+                return dynamic_template.render(context)
+        ```
+
 - The `Component.Url` class was merged with `Component.View`.
 
     Instead of `Component.Url.public`, use `Component.View.public`.
@@ -403,6 +437,51 @@ Summary:
     to accept also slots and context.
 
     Since `get_context_data()` is widely used, it will remain available until v2.
+
+- `Component.get_template_name()` and `Component.get_template()` are now deprecated. Use `Component.template`,
+`Component.template_file` or `Component.on_render()` instead.
+
+    `Component.get_template_name()` and `Component.get_template()` will be removed in v1.
+
+    In v1, each Component will have at most one static template.
+    This is needed to enable support for Markdown, Pug, or other pre-processing of templates by extensions.
+
+    If you are using the deprecated methods to point to different templates, there's 2 ways to migrate:
+
+    1. Split the single Component into multiple Components, each with its own template. Then switch between them in `Component.on_render()`:
+
+        ```py
+        class MyComponentA(Component):
+            template_file = "a.html"
+
+        class MyComponentB(Component):
+            template_file = "b.html"
+
+        class MyComponent(Component):
+            def on_render(self, context, template):
+                if context["a"]:
+                    return MyComponentA.render(context)
+                else:
+                    return MyComponentB.render(context)
+        ```
+
+    2. Alternatively, use `Component.on_render()` with Django's `get_template()` to dynamically render different templates:
+
+        ```py
+        from django.template.loader import get_template
+
+        class MyComponent(Component):
+            def on_render(self, context, template):
+                if context["a"]:
+                    template_name = "a.html"
+                else:
+                    template_name = "b.html"
+
+                actual_template = get_template(template_name)
+                return actual_template.render(context)
+        ```
+
+    Read more in [django-components#1204](https://github.com/django-components/django-components/discussions/1204).
 
 - The `type` kwarg in `Component.render()` and `Component.render_to_response()` is now deprecated. Use `deps_strategy` instead. The `type` kwarg will be removed in v1.
 
@@ -651,6 +730,22 @@ Summary:
 
 **Miscellaneous**
 
+- Template caching with `cached_template()` helper and `template_cache_size` setting is deprecated.
+    These will be removed in v1.
+
+    This feature made sense if you were dynamically generating templates for components using
+    `Component.get_template_string()` and `Component.get_template()`.
+
+    However, in v1, each Component will have at most one static template. This static template
+    is cached internally per component class, and reused across renders.
+
+    This makes the template caching feature obsolete.
+
+    If you relied on `cached_template()`, you should either:
+
+    1. Wrap the templates as Components.
+    2. Manage the cache of Templates yourself.
+
 - The `debug_highlight_components` and `debug_highlight_slots` settings are deprecated.
     These will be removed in v1.
 
@@ -875,6 +970,13 @@ Summary:
 
     Then, the `contents` attribute of the `BaseNode` instance will contain the string `"Hello, world!"`.
 
+- The `BaseNode` class also has two new metadata attributes:
+
+    - `template_name` - the name of the template that rendered the node.
+    - `template_component` - the component class that the template belongs to.
+
+    This is useful for debugging purposes.
+
 - `Slot` class now has 3 new metadata fields:
 
     1. `Slot.contents` attribute contains the original contents:
@@ -1003,6 +1105,35 @@ Summary:
 - Component's "Render API" (args, kwargs, slots, context, inputs, request, context data, etc)
   can now be accessed also outside of the render call. So now its possible to take the component
   instance out of `get_template_data()` (although this is not recommended).
+
+- Components can now be defined without a template.
+
+    Previously, the following would raise an error:
+
+    ```py
+    class MyComponent(Component):
+        pass
+    ```
+
+    "Template-less" components can be used together with `Component.on_render()` to dynamically
+    pick what to render:
+
+    ```py
+    class TableNew(Component):
+        template_file = "table_new.html"
+
+    class TableOld(Component):
+        template_file = "table_old.html"
+
+    class Table(Component):
+        def on_render(self, context, template):
+            if self.kwargs.get("feat_table_new_ui"):
+                return TableNew.render(args=self.args, kwargs=self.kwargs, slots=self.slots)
+            else:
+                return TableOld.render(args=self.args, kwargs=self.kwargs, slots=self.slots)
+    ```
+
+    "Template-less" components can be also used as a base class for other components, or as mixins.
 
 - Passing `Slot` instance to `Slot` constructor raises an error.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -970,13 +970,6 @@ Summary:
 
     Then, the `contents` attribute of the `BaseNode` instance will contain the string `"Hello, world!"`.
 
-- The `BaseNode` class also has two new metadata attributes:
-
-    - `template_name` - the name of the template that rendered the node.
-    - `template_component` - the component class that the template belongs to.
-
-    This is useful for debugging purposes.
-
 - `Slot` class now has 3 new metadata fields:
 
     1. `Slot.contents` attribute contains the original contents:
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -1,5 +1,4 @@
 import sys
-from contextlib import contextmanager
 from dataclasses import dataclass
 from types import MethodType
 from typing import (
@@ -7,7 +6,6 @@ from typing import (
     Callable,
     ClassVar,
     Dict,
-    Generator,
     List,
     Mapping,
     NamedTuple,
@@ -19,12 +17,10 @@ from typing import (
 )
 from weakref import ReferenceType, WeakValueDictionary, finalize
 
-from django.core.exceptions import ImproperlyConfigured
 from django.forms.widgets import Media as MediaCls
 from django.http import HttpRequest, HttpResponse
-from django.template.base import NodeList, Origin, Parser, Template, Token
+from django.template.base import NodeList, Parser, Template, Token
 from django.template.context import Context, RequestContext
-from django.template.loader import get_template
 from django.template.loader_tags import BLOCK_CONTEXT_KEY, BlockContext
 from django.test.signals import template_rendered
 from django.views import View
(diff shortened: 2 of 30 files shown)


tests/test_templatetags_component.py
```diff
@@ -12,22 +12,28 @@
 setup_test_config({"autodiscover": False})
 
 
-class SlottedComponent(Component):
-    template_file = "slotted_template.html"
+def gen_slotted_component():
```
JuroOravec, line 15: I wrapped the components in functions so that the component creation is done within a function that is wrapped in `@djc_test` decorator to fix some global state issue with tests.

src/django_components/component_media.py
```diff
@@ -953,27 +954,47 @@ class MyComp(Component):
...
+        return asset_content
+
+    # The rest of the code assumes that we were given only a file name
+    asset_file = cast(str, asset_file)
+
+    if type == "template":
+        # NOTE: While we return on the "source" (plain string) of the template,
+        #       by calling `load_component_template()`, we also cache the Template instance.
+        #       So later in Component's `render_impl()`, we don't have to re-compile the Template.
+        template = load_component_template(comp_cls, asset_file)
```
JuroOravec, line 979: Previously, we did use Django's `get_template()` to load the template, but then we discarded the actual `Template` instance, and read the file contents as plain string.

Now, one difference is that when calling `load_component_template()` here, we save the loaded `Template` instance to `Component._template`.

So we use the Component classes to cache the Template instances.

And because we now use the `Template` instance returned from Django's `get_template()`, we now play nicely with template loaders (https://github.com/django-components/django-components/issues/901)

src/django_components/template_loader.py
```diff
@@ -26,3 +26,10 @@ def get_dirs(self, include_apps: bool = True) -> List[Path]:
         `BASE_DIR` setting is required.
         """
         return get_component_dirs(include_apps)
+
+
+# NOTE: Django's template loaders have the pattern of using the `Loader` class name.
+#       However, this then makes it harder to track and distinguish between different loaders.
+#       So internally we use the name `DjcLoader` instead.
+#       But for public API we use the name `Loader` to match Django.
+Loader = DjcLoader
```
JuroOravec, line 35: My initial implementation relied on passing extra data through our template loader, which forced me to debug how template loaders work and whatnot.

As part of that I renamed it from `Loader` to `DjcLoader`, so it's easier to distinguish from other Loaders when debugging.

To remain backwards compatible, and inline with how Django's template loaders are defined, this file still defines a `Loader` alias.

tests/test_component_media.py
```diff
@@ -851,40 +840,46 @@ class Media:
...
-            <div>
-                <h1>Parent content</h1>
-                {% component "variable_display" shadowing_variable='override' new_variable='unique_val' %}
-                {% endcomponent %}
-            </div>
-            <div>
-                {% slot 'content' %}
-                    <h2>Slot content</h2>
-                    {% component "variable_display" shadowing_variable='slot_default_override' new_variable='slot_default_unique' %}
+    def _gen_parent_component(self):
```
JuroOravec, line 843: In multiple places in test files I had to wrap top-level component definitions in functions to be called from within tests, so that the state is properly cleared up between tests.

tests/test_templatetags.py
```diff
@@ -10,10 +10,6 @@
 setup_test_config({"autodiscover": False})
 
 
-class SlottedComponent(Component):
```
JuroOravec, line 13: Unused

tests/test_signals.py
```diff
@@ -48,24 +44,33 @@ def wrapper(*args, **kwargs):
 
 @djc_test
 class TestTemplateSignal:
-    class InnerComponent(Component):
-        template_file = "simple_template.html"
+    def gen_slotted_component(self):
```
JuroOravec, line 47: Again wrapping components in functions.

tests/testutils.py
```diff
@@ -37,12 +37,19 @@ def setup_test_config(
                 "OPTIONS": {
                     "builtins": [
                         "django_components.templatetags.component_tags",
-                    ]
+                    ],
+                    'loaders': [
```
JuroOravec, line 41: Previously, we loaded component template string from files ourselves with `Path.read_text()`. This allowed us to bypass django's template loaders.

However, now, to support django loaders, we have to load the files through Django.

Which means that also the common Django settings that we use in tests had to be updated to add our template loader `django_components.template_loader.Loader`

tests/test_templatetags_extends.py
```diff
@@ -11,21 +11,18 @@
 setup_test_config({"autodiscover": False})
 
 
-class SlottedComponent(Component):
-    template_file = "slotted_template.html"
+def gen_slotted_component():
```
JuroOravec, line 14: Again, wrapping components in functions in tests.

src/django_components/component.py
```diff
@@ -2937,8 +3048,30 @@ def _render_impl(
...
+        #
+        #     Then we can simply apply `template_data` to the context in the same layer
+        #     where we apply `context_processor_data` and `component_vars`.
+        with prepare_component_template(component, template_data) as template:
+            # Set `Template._djc_is_component_nested` based on whether we're currently INSIDE
+            # the `{% extends %}` tag.
+            # Part of fix for https://github.com/django-components/django-components/issues/508
+            # See django_monkeypatch.py
+            if template is not None:
+                template._djc_is_component_nested = bool(  # type: ignore[union-attr]
```
JuroOravec, line 3069: This `template._djc_is_component_nested` branch is not new, just moved from [`_prepare_template`](https://github.com/django-components/django-components/pull/1222/files#diff-35c84ebceae50f7149f03bbcaf2b4f1289bc2984bf7859ca521b5dbe7960c396L3409)

src/django_components/component.py
```diff
@@ -2937,8 +3048,30 @@ def _render_impl(
         #       but instead can render one component at a time.
         #############################################################################
 
-        with _prepare_template(component, template_data) as template:
-            component_ctx.template_name = template.name
+        # TODO_v1 - Currently we have to pass `template_data` to `prepare_component_template()`,
```
JuroOravec, line 3051: As the comment says, in v1 we will be able to simplify this logic once we drop support for `get_template_string()`, `get_template_name()`, and `get_template()`.

src/django_components/component.py
```diff
@@ -3089,26 +3222,31 @@ def renderer(root_attributes: Optional[List[str]] = None) -> Tuple[str, Dict[str
...
 
-            # Add necessary HTML attributes to work with JS and CSS variables
-            updated_html, child_components = set_component_attrs_for_js_and_css(
-                html_content=html_content,
-                component_id=render_id,
-                css_input_hash=css_input_hash,
-                css_scope_id=css_scope_id,
-                root_attributes=root_attributes,
-            )
+            if template is not None:
```
JuroOravec, line 3226: Only changed indentation here as `template` now may be `None`

src/django_components/component.py
```diff
@@ -3370,43 +3518,3 @@ def _get_parent_component_context(context: Context) -> Union[Tuple[None, None],
 
     parent_comp_ctx = component_context_cache[parent_id]
     return parent_id, parent_comp_ctx
-
-
-@contextmanager
-def _maybe_bind_template(context: Context, template: Template) -> Generator[None, Any, None]:
```
JuroOravec, line 3376: These helpers were moved to `template.py`.

src/django_components/template.py
```diff
@@ -62,3 +75,363 @@ class MyTemplate(Template):
         template = maybe_cached_template
 
     return template
+
+
+########################################################
+# PREPARING COMPONENT TEMPLATES FOR RENDERING
```
JuroOravec, line 81: This section contains the updated logic for loading  component templates from files.

src/django_components/component_media.py
```diff
@@ -736,6 +738,11 @@ def _resolve_component_relative_files(
     as the component class. If so, modify the attributes so the class Django's rendering
     will pick up these files correctly.
     """
+    if comp_media.resolved_relative_files:
```
JuroOravec, line 741: Component's template file path can be defined relative to the Component's file.
Previously, this relative file resolution was done at the same time as when we also loaded the JS / CSS / HTML from the files.

But to be able to associate `Template` instances with corresponding `Component` classes, we need to resolve the relative paths earlier.

So I added this `resolved_relative_files` so that we can do the path normalization and file loading at different times.

src/django_components/template.py
```diff
@@ -62,3 +76,400 @@ class MyTemplate(Template):
...
+    # This code is taken from `Context.bind_template()` from Django v5.1
+    context.template = template
+    try:
+        yield
+    finally:
+        context.template = None
+
+
+########################################################
+# LOADING TEMPLATES FROM FILEPATH
```
JuroOravec, line 157: This section below contains logic that was previously in `component.py` in `Component._get_template`.

We still support all the different ways of defining a template for a component:
- `template` attribute
- `template_file` attribute
- `get_template_name()`
- `get_template_string()`
- `get_template()`

src/django_components/util/django_monkeypatch.py
```diff
@@ -10,11 +10,61 @@
...
+            component_cls = get_component_from_origin(origin)
+        elif origin is not None and origin.template_name is not None:
+            component_cls = get_component_by_template_file(origin.template_name)
+            if component_cls is not None:
+                set_component_to_origin(origin, component_cls)
+        else:
+            component_cls = None
+
+        if component_cls is not None:
+            # TODO - Apply extensions.on_template_preprocess() here.
```
JuroOravec, line 59: By wrapping `Template.__init__()`, we will be able to allow extensions to pre-process the template string even for templates loaded from a file.

JuroOravec:
Phew, this was a tricky one. @EmilStenstrom would you look at this one?

CHANGELOG.md
```diff
@@ -404,6 +438,51 @@ Summary:
 
     Since `get_context_data()` is widely used, it will remain available until v2.
 
+- `Component.get_template_name()` and `Component.get_template()` are now deprecated. Use `Component.template`,
+`Component.template_file` or `Component.on_render()` instead.
+
+    `Component.get_template_name()` and `Component.get_template()` will be removed in v1.
```
EmilStenstrom, line 444: I think we could be stricter here, and remove these earlier than v1. Since they are blocking pre-processing of templates I think breaking things early is worth it in this case.

JuroOravec: I think it's fine till v1. The way it's implemented in this PR, there is a tradeoff that people who will still use those `get_` methods won't be able to benefit from the extensions that make use of template pre-processing. So I'm hoping that that will be a motivation for them to "upgrade". But it's not a blocker per se.

PR #1224, opened 2025-06-02
refactor: don't inherit media if child set to None

JuroOravec:
Previously, when a child component's template / JS / CSS attributes were set to `None`, the child component still inherited the parent's template / JS / CSS.

After the previous PR (https://github.com/django-components/django-components/pull/1222), setting a component template to `None` is now valid and seen as "this component has no template".

Because of that, the child component will now NOT inherit the parent's template / JS / CSS if the child sets the attribute to `None`.

Before:

```py
class Parent(Component):
    template = "parent.html"

class Child(Parent):
    template = None

# Child still inherited parent's template
assert Child.template == Parent.template
```

After:

```py
class Parent(Component):
    template = "parent.html"

class Child(Parent):
    template = None

# Child does not inherit parent's template
assert Child.template is None
```

### DIFF
--- a/tests/test_component_media.py
+++ b/tests/test_component_media.py
@@ -1239,7 +1239,7 @@ class TestSubclassingMedia:
 
         # `None` means that we will NOT inherit `Media` from `GrandParentComponent`
         class ParentComponent(GrandParentComponent):
-            Media = None
+            Media = None  # type: ignore[assignment]
 
         class ChildComponent(ParentComponent):
             class Media:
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 ⚠️ Major release ⚠️ - Please test thoroughly before / after upgrading.
 
+This is the biggest step towards v1. While this version introduces
```
JuroOravec, line 7: A bit of reassurance that this is a one off big update.

src/django_components/component_media.py
```diff
@@ -41,6 +41,21 @@
...
+# If set to `None`, then this compoenent has no template.
+class Unset:
+    def __bool__(self) -> bool:
+        return False
+
+    def __repr__(self) -> str:
+        return "<UNSET>"
+
+
+UNSET = Unset()
```
JuroOravec, line 61: Since `None` is now a valid value for the component media properties, I used a sentinel value `UNSET` to differentiate between explicit `None` and unset fields.

JuroOravec:
@EmilStenstrom This one's ready for review.

After this, there's only 3 more PRs until v0.140 🔥 🔥

PR #1229, opened 2025-06-02
refactor: Add Node metadata

JuroOravec:
Ok, this is the last PR related to slots and after this I'm [done working on slots](https://github.com/django-components/django-components/issues/433#issuecomment-2826726311) until v1 🤞 

This PR wires up metadata, so we can trace back to their origin when we have a Component or a Slot instance:

- If component comes from `{% component %}` tag, then `Component.node` will point to the `ComponentNode`
- If a slot fill comes from `{% fill %}` tag, then `Slot.fill_node` points to `FillNode`
- If a slot fill comes from `{% component %}` tag (AKA default fill), then `Slot.fill_node` points to that `ComponentNode`.

In extensions:

- In `on_slot_rendered` hook, one can access the original `{% slot %}` tag (aka `SlotNode`) via `ctx.slot_node`.

Moreover each template tag node defined by us (with our `BaseNode`) has access to 2 new properties:
- `template_name` - The name of the `Template` instance inside which the node was defined
- `template_component` - The component class that the `Template` belongs to

In other words, if a Component was defined with a template file:

```py
class MyTable(Component):
    template_file = "path/to/template.html"
```

And the template looked like this:

```py
<div>
  {% slot "content" default / %}
</div>
```

Then `SlotNode` will have the knowledge that it was defined inside a template file named `"path/to/template.html"`
and inside a Component `MyTable`.

Overall, this is a more general approach to what I previously did with `Slot.source`, which distinguished between if the slot was defined in a template or in Python. So I removed the `Slot.source` attribute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -158,12 +158,17 @@ Summary:
     ```py
     def render_to_response(
         context: Optional[Union[Dict[str, Any], Context]] = None,
-        args: Optional[Tuple[Any, ...]] = None,
-        kwargs: Optional[Mapping] = None,
-        slots: Optional[Mapping] = None,
+        args: Optional[Any] = None,
+        kwargs: Optional[Any] = None,
+        slots: Optional[Any] = None,
         deps_strategy: DependenciesStrategy = "document",
-        render_dependencies: bool = True,
+        type: Optional[DependenciesStrategy] = None,  # Deprecated, use `deps_strategy`
+        render_dependencies: bool = True,  # Deprecated, use `deps_strategy="ignore"`
+        outer_context: Optional[Context] = None,
         request: Optional[HttpRequest] = None,
+        registry: Optional[ComponentRegistry] = None,
+        registered_name: Optional[str] = None,
+        node: Optional[ComponentNode] = None,
         **response_kwargs: Any,
     ) -> HttpResponse:
     ```
@@ -1008,6 +1013,13 @@ Summary:
 
     Then, the `contents` attribute of the `BaseNode` instance will contain the string `"Hello, world!"`.
 
+- The `BaseNode` class also has two new metadata attributes:
+
+    - `template_name` - the name of the template that rendered the node.
+    - `template_component` - the component class that the template belongs to.
+
+    This is useful for debugging purposes.
+
 - `Slot` class now has 3 new metadata fields:
 
     1. `Slot.contents` attribute contains the original contents:
@@ -1018,10 +1030,11 @@ Summary:
 
     2. `Slot.extra` attribute where you can put arbitrary metadata about the slot.
 
-    3. `Slot.source` attribute tells where the slot comes from:
+    3. `Slot.fill_node` attribute tells where the slot comes from:
 
-        - `'template'` if the slot was created from `{% fill %}` tag.
-        - `'python'` if the slot was created from string, function, or `Slot` instance.
+        - `FillNode` instance if the slot was created from `{% fill %}` tag.
+        - `ComponentNode` instance if the slot was created as a default slot from a `{% component %}` tag.
+        - `None` if the slot was created from a string, function, or `Slot` instance.
 
     See [Slot metadata](https://django-components.github.io/django-components/0.140/concepts/fundamentals/slots/#slot-metadata).
 
@@ -1046,6 +1059,46 @@ Summary:
     {% component "table" %}
       {% fill "pagination" body=my_slot / %}
     {% endcomponent %}
+    ```
+
+- You can now access the `{% component %}` tag (`ComponentNode` instance) from which a Component
+    was created. Use `Component.node` to access it.
+
+    This is mostly useful for extensions, which can use this to detect if the given Component
+    comes from a `{% component %}` tag or from a different source (such as `Component.render()`).
+    
+    `Component.node` is `None` if the component is created by `Component.render()` (but you
+    can pass in the `node` kwarg yourself).
+
+    ```py
+    class MyComponent(Component):
+        def get_template_data(self, context, template):
+            if self.node is not None:
+                assert self.node.name == "my_component"
+    ```
+
+- Node classes `ComponentNode`, `FillNode`, `ProvideNode`, and `SlotNode` are part of the public API.
+
+    These classes are what is instantiated when you use `{% component %}`, `{% fill %}`, `{% provide %}`, and `{% slot %}` tags.
+
+    You can for example use these for type hints:
+
+    ```py
+    from django_components import Component, ComponentNode
+
+    class MyTable(Component):
+        def get_template_data(self, args, kwargs, slots, context):
+            if kwargs.get("show_owner"):
+                node: Optional[ComponentNode] = self.node
+                owner: Optional[Component] = self.node.template_component
+            else:
+                node = None
+                owner = None
+
+            return {
+                "owner": owner,
+                "node": node,
+            }
     ```
 
 - Component caching can now take slots into account, by setting `Component.Cache.include_slots` to `True`.
--- a/docs/concepts/fundamentals/slots.md
+++ b/docs/concepts/fundamentals/slots.md
@@ -728,7 +728,7 @@ with extra metadata:
 - [`component_name`](../../../reference/api#django_components.Slot.component_name)
 - [`slot_name`](../../../reference/api#django_components.Slot.slot_name)
 - [`nodelist`](../../../reference/api#django_components.Slot.nodelist)
-- [`source`](../../../reference/api#django_components.Slot.source)
+- [`fill_node`](../../../reference/api#django_components.Slot.fill_node)
 - [`extra`](../../../reference/api#django_components.Slot.extra)
 
 These are populated the first time a slot is passed to a component.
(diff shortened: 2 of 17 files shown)


docs/concepts/fundamentals/slots.md
```diff
@@ -738,10 +738,33 @@ still point to the first component that received the slot.
...
+
+You can use this to find the [`Component`](../../../reference/api#django_components.Component) in whose
+template the [`{% fill %}`](../../../reference/template_tags#fill) tag was defined:
+
+```python
+class MyTable(Component):
+    def get_template_data(self, args, kwargs, slots, context):
+        footer_slot = slots.get("footer")
+        if footer_slot is not None and footer_slot.fill_node is not None:
+            owner_component = footer_slot.fill_node.template_component
```
JuroOravec, line 763: The reason for using `Slot.fill_node` over `Slot.source` is that having access to the whole `FillNode` gives access to more metedata.

src/django_components/__init__.py
```diff
@@ -18,6 +18,7 @@
 from django_components.component import (
     Component,
     ComponentInput,
+    ComponentNode,
```
JuroOravec, line 21: Finally exposed the Node classes as public API.

Initially (as in, a year ago), I was reluctant about this, because I was considering making django-components compatible with Jinja. But now I know that because django-components relies on the Context and Template and other such objects, we're tightly coupled to Django's templating, and cannot use Jinja for the Nodes layer.

But that means it's fine to expose these Nodes, because they will be around even in v1 and later versions.

src/django_components/component.py
```diff
@@ -2346,6 +2348,51 @@ def on_render_before(self, context: Context, template: Template) -> None:
     that was used to render the component.
     """
 
+    node: Optional["ComponentNode"]
```
JuroOravec, line 2351: Here I've added `Component.node`.

src/django_components/node.py
```diff
@@ -303,6 +351,133 @@ def render(self, context: Context, name: str, **kwargs: Any) -> str:
         """
         return self.nodelist.render(context)
 
+    # #####################################
+    # Attributes
+    # #####################################
+
+    params: List[TagAttr]
```
JuroOravec, line 358: These are not new fields, just documented the existing attributes of `BaseNode`, since the subclasses like `ComponentNode`, `FillNode`, etc, are now public and show up in the docs.

src/django_components/node.py
```diff
@@ -359,6 +555,8 @@ def parse(cls, parser: Parser, token: Token, **kwargs: Any) -> "BaseNode":
             params=tag.params,
             flags=tag.flags,
             contents=contents,
+            template_name=parser.origin.name if parser.origin else None,
+            template_component=get_component_from_origin(parser.origin) if parser.origin else None,
```
JuroOravec, line 559: And here I've added the `template_name` / `template_component`.

When a template was defined on a Component class (e.g. `Component.template_file`), then the owner component class is available under `template.origin.component_cls`.

When a template is being compiled from string to AST of Nodes, we don't have access to the `Template` instance, but we do have access to its Origin.

That's how we can pass the info on owner component class to the `Node` instances.

src/django_components/slots.py
```diff
@@ -926,6 +945,7 @@ def render(self, context: Context, name: str, **kwargs: Any) -> SafeString:
                 component_id=component_id,
                 slot=slot,
                 slot_name=slot_name,
+                slot_node=self,
```
JuroOravec, line 948: This is where the `on_slot_rendered` extension hook receives the refrence to SlotNode.

The rest of this file is then about passing the fill node to `Slot` instance instead of the old `Slot.source` attribute.

tests/test_extension.py
```diff
@@ -359,10 +368,25 @@ class TestComponent(Component):
...
         assert slot_call.slot_name == "content"
+        assert isinstance(slot_call.slot_node, SlotNode)
+        assert slot_call.slot_node.template_name.endswith("test_extension.py::TestComponent")  # type: ignore
+        assert slot_call.slot_node.template_component == TestComponent
         assert slot_call.slot_is_required is True
         assert slot_call.slot_is_default is True
         assert slot_call.result == "Some content"
 
+    @djc_test(components_settings={"extensions": [SlotOverrideExtension]})
+    def test_on_slot_rendered__override(self):
```
JuroOravec, line 379: `on_slot_rendered` hook was missing a test for when a new string is returned from it.

CHANGELOG.md
```diff
@@ -158,12 +158,17 @@ Summary:
     ```py
     def render_to_response(
         context: Optional[Union[Dict[str, Any], Context]] = None,
-        args: Optional[Tuple[Any, ...]] = None,
-        kwargs: Optional[Mapping] = None,
-        slots: Optional[Mapping] = None,
+        args: Optional[Any] = None,
```
JuroOravec, line 161: Not related to this PR, except for the `node` kwarg at the end. Just realized that the rest of the signature doesn't reflect all the changes.

JuroOravec:
@EmilStenstrom Another one's ready! This one's mostly text, tests, and only a few code changes. And thanks for all your help on the previous ones!

src/django_components/__init__.py
EmilStenstrom, line 1: The public API is huge here. I'm a little worried are taking on way to big of a surface area to support if we have this for v1. Thoughts?

JuroOravec: You're right, altho I'm not sure there's much that can be done about this, other than teach others.

Because when you break it down, then:
- About half of the public API is the core (Component, Slot, Registry),
- The other half for flexibility / extendability - settings, extensions, tag formatter

<details>
<summary> Here's an overview out of curiosity: </summary>
<br/>
Total: 74 entities (+/- 1)
<hr/>

- Component:

	- all_components
	- Component
	- ComponentCache
	- ComponentDebugHighlight
	- ComponentDefaults
	- ComponentFileEntry
	- ComponentInput
	- ComponentMediaInput
	- ComponentMediaInputPath
	- ComponentNode
	- ComponentVars
	- ComponentView
	- get_component_by_class_id
	- get_component_dirs
	- get_component_files
	- get_component_url
	- render_dependencies
	- DependenciesStrategy

- Slots:

	- Slot
	- SlotContent
	- SlotContext
	- SlotFallback
	- SlotFunc
	- SlotInput
	- SlotNode
	- SlotRef
	- SlotResult
	- FillNode

- Registry:

	- all_registries
	- AlreadyRegistered
	- ComponentRegistry
	- register
	- registry
	- RegistrySettings
	- NotRegistered

- Own template tags (BaseNode):

	- BaseNode
	- template_tag
	- TagProtectedError
	
- Extensions:

	- ComponentExtension
	- ExtensionComponentConfig
	- OnComponentClassCreatedContext
	- OnComponentClassDeletedContext
	- OnComponentDataContext
	- OnComponentInputContext
	- OnComponentRegisteredContext
	- OnComponentUnregisteredContext
	- OnRegistryCreatedContext
	- OnRegistryDeletedContext

- Extension CLI commands:

	- CommandArg
	- CommandArgGroup
	- CommandHandler
	- CommandLiteralAction
	- CommandParserInput
	- CommandSubcommand
	- ComponentCommand

- Extension URLs:

	- URLRoute
	- URLRouteHandler

- Settings:

	- ComponentsSettings
	- ContextBehavior

- Tag formatter:

	- ShorthandComponentFormatter
	- TagFormatterABC
	- component_formatter
	- component_shorthand_formatter
	- ComponentFormatter

- Misc:

	- autodiscover
	- cached_template
	- Default
	- DynamicComponent
	- Empty
	- format_attributes
	- import_libraries
	- merge_attributes
	- ProvideNode
	- TagResult
	- types

</details>

I feel like that currently the biggest risks in this department are:
- Me getting hit by the bus (losing intrinsic knowledge)
- Changes in Django's API
- Changes in argparse API

Our own API feels thoroughly tested, so I have confidence in that.

Ofc, that's useless if nobody knows how to work with the codebase, hence the bus factor.

But other than that the project feels well self-encapsulated. But some of the public API directly reflects / exposes Django's or argparse's API, so if they make big changes, it could be a headache for us to upgrade.

I hope / believe that things will get gradually better. WIth v0.140, I will finally feel fine with making blogs / vids / tutorials, because most of what will be shown will be already the same as will be in v1.

One of the things I'd like to do is then go over the old design discussions we had here, and talk about them. (For example the algo that enables infinite component nesting is IMO very elegant visually 😄). So that would hit two rocks with one bird - document the architecture, and promote the library.

EmilStenstrom: One way to think about it is typing stuff vs other stuff. Perhaps types should be in their own namespace?

JuroOravec: What do you mean by types here? I assume it's anything that can be used as a type hint. But I don't like it that much, it feels like in Python there's blurry line between what is type and what is not:
- Crystal clear are type aliases - yes, those are types
- Classes like `Component`, `Slot`, `ComponentNode`, etc? Those can be used for type hints. But also not.
- Then there are data objects that help with auxiliary things. Like the extension data classes e.g. `OnComponentClassCreatedContext`. User shouldn't be creating these themselves, and if they import it, it will be 99% for type hints. but it's still not a type per se.

So IMO split betwen types / non-types would be more confusing.

Along thing thought, other thing could be to split the public API by the domain, e.g.

```py
from django_components.component import Component
from django_components.slots import Slot
from django_components.extensions import OnComponentClassCreatedContext
```

But it feels like that would only add mental overhead. Right now users don't have to remember in which file a var / class / fn is, they simply import the symbol from `django_components`.

Potentially the splitting of the public API could lead to better loading times. But I think there's enough interdependence between the files, that even if we exposed `Component` from `django_components.component`, that would still trigger the loading of the entire `django_components`.

PR #1231, opened 2025-06-03
feat: on_render

JuroOravec:


src/django_components/components/dynamic.py
```diff
@@ -99,23 +99,25 @@ class DynamicComponent(Component):
...
 
-    # TODO: Replace combination of `on_render_before()` + `template` with single `on_render()`
-    #
-    # NOTE: The inner component is rendered in `on_render_before`, so that the `Context` object
+    # NOTE: The inner component is rendered in `on_render`, so that the `Context` object
     # is already configured as if the inner component was rendered inside the template.
     # E.g. the `_COMPONENT_CONTEXT_KEY` is set, which means that the child component
     # will know that it's a child of this component.
-    def on_render_before(self, context: Context, template: Template) -> Context:
+    def on_render(
```
JuroOravec, line 106: With `on_render()` the implementation of the dynamic component is a bit simpler.

src/django_components/dependencies.py
```diff
@@ -249,22 +248,11 @@ def set_component_attrs_for_js_and_css(
...
 
-    # These attributes are set on all tags
-    all_attributes = []
-
-    # We apply the CSS scoping attribute to both root and non-root tags.
-    #
-    # This is the HTML part of Vue-like CSS scoping.
-    # That is, for each HTML element that the component renders, we add a `data-djc-scope-a1b2c3` attribute.
-    # And we stop when we come across a nested components.
-    if css_scope_id:
```
JuroOravec, line 260: Removed mentions of css_scope_id from the codebase, since CSS scoping will be implemented as a separate extensions

src/django_components/extensions/cache.py
```diff
@@ -198,5 +198,8 @@ def on_component_rendered(self, ctx: OnComponentRenderedContext) -> None:
         if not cache_instance.enabled:
             return None
 
+        if ctx.error is not None:
```
JuroOravec, line 201: Had to update other extensions, since `on_component_rendered()` now receives also the error if component rendering failed.

tests/test_component.py
```diff
@@ -1429,260 +1430,555 @@ class Root(Component):
...
+                    Inner end
+                    Middle text
+                    Inner start
+                    Inner end
+                Middle end
+            Outer end
             """,
         )
+
+        assert calls == [
```
JuroOravec, line 1599: Here in this test we check that the order of `on_render_before()`, `on_render()` and `on_render_after()` is correct even with multiple components.

tests/test_component.py
```diff
@@ -1429,260 +1430,555 @@ class Root(Component):
...
-            def on_render_after(self, context: Context, template: Template, html: str) -> None:
-                context["from_on_after"] = ":)"
-                nonlocal context_in_after
-                context_in_after = context
+        with pytest.raises(ValueError, match=re.escape("BROKEN")):
+            SimpleComponent.render()
+
+    @djc_test(
+        parametrize=(
+            ["template", "action", "method"],
```
JuroOravec, line 1770: And here I added tests for checking of `on_render_before()` / `on_render()` / `on_render_after()` behave when:
1. They receive error
2. They render template that contains error
3. They render valid template

vs

- When `on_render` / `on_render_after` should override the HTML
- When `on_render` / `on_render_after` should override the error
- Do nothing

tests/test_component.py
```diff
@@ -1429,260 +1430,555 @@ class Root(Component):
 
 @djc_test
 class TestComponentHook:
-    def test_on_render_before(self):
-        @register("nested")
-        class NestedComponent(Component):
+    def _gen_slotted_component(self, calls: List[str]):
```
JuroOravec, line 1433: I wrapped these component definintions in functions, so that the state is properly cleaned up afterwards.

docs/concepts/fundamentals/render_api.md
```diff
@@ -89,7 +90,7 @@ class Table(Component):
         page: int
         per_page: int
 
-    def on_render_before(self, context: Context, template: Template) -> None:
+    def on_render_before(self, context: Context, template: Optional[Template]) -> None:
```
JuroOravec, line 93: In this file just updated the signature of `on_render_before`

docs/concepts/fundamentals/html_js_css_files.md
```diff
@@ -118,6 +118,92 @@ class Button(Component):
 </button>
 ```
 
+### Dynamic templates
```
JuroOravec, line 121: Mentioned how use `Component.on_render()` to customize what gets rendered - Basically covering what was asked here https://github.com/django-components/django-components/discussions/1204

src/django_components/perfutil/component.py
```diff
@@ -29,28 +31,45 @@
 component_context_cache: Dict[str, "ComponentContext"] = {}
 
 
-class PostRenderQueueItem(NamedTuple):
-    content_before_component: str
-    child_id: Optional[str]
+class ComponentPart(NamedTuple):
```
JuroOravec, line 34: This file's not easy to explain, but in a nutshell it's all about being able to bubble up errors.

So if I am rendering a parent Component and there's an error while the Component's template is being rendered, following happens:

1. We capture the error and set the result to `None`

2. We allow `Component.on_render()` and `Component.on_render_after()` to optionally handle this:
  - E.g. `on_render()` can recover from the error and instead decide to return a new HTML instead - This is how the `ErrorFallback` component will work.
  - These methods can also do the opposite - the HTML rendered OK, but they decide to trigger an error instead.
- Same is done for the `on_component_rendered()` extension hook - it can recover from errors, or raise new errors, etc.
3. Either way, at the end we've got the tuple
     ```py
     (result, error)
     ```

   If there is an error, then the component's rendering is stopped / discarded,
   and the component re-raises the error.

4. When that happens, the cycle repeats - the parent's parent can recover from the error with `on_render()` or `on_render_after()`, or do nothing to re-raise it further.

5. When the error reaches the top component there's no other component to intercept the error, so it is finally raised as a normal error.

src/django_components/perfutil/component.py
```diff
@@ -29,28 +31,45 @@
...
         return (
-            f"PostRenderQueueItem(child_id={self.child_id!r}, parent_id={self.parent_id!r}, "
-            f"grandparent_id={self.grandparent_id!r}, component_name_path={self.component_name_path!r}, "
-            f"content_before_component={self.content_before_component[:10]!r})"
+            f"ComponentPart(child_id={self.child_id!r}, parent_id={self.parent_id!r}, "
+            f"component_name_path={self.component_name_path!r})"
         )
 
 
+class TextPart(NamedTuple):
```
JuroOravec, line 48: Another change here and what the `TextPart` and `ComponentPart` relate to is the following:

1. When a component's template is rendered, it looks it contains plain HTML, and sections owned by child components:

    ```html
    <div>  <!-- Normal HTML-->
       <template data-render-id-ca1b2c3>  <!-- Placeholder for child comp -->
       </template>
    </div>
    ```

2. The HTML above is split to cut out the child component placeholders, and replace them with actual child component renderings

2. Before this PR I've split the HTML into tuples of:

    ```py
    (html_in_front_of_component, component_id)
    ```

    E.g. the example above would be split into:

    ```py
    [
        (
            "<div>  <!-- Normal HTML-->",  # content before child comp
            "ca1b2c3"  # child comp
        ),
        (
            "</div>",  # content after child comp
            None,  # End of template
         )
    ]
    ```

3. Now, in this PR, instead of using 2-member tuples, it returns a simple list. And to know whether we're processing a piece of rendered string, or a component placeholder, we use the `ComponentPart` and `TextPart`:

    ```py
    [
        TextPart("<div>  <!-- Normal HTML-->"),  # content before child comp
        ComponentPart("ca1b2c3"),  # child comp
        TextPart("</div>"),  # content after child comp
    ]
    ```

    When we get `ComponentPart`, we render the nested component, parse its result into `TextPart` and `ComponentPart`, and insert those in place of the original `ComponentPart`.

     E.g. when the nested component is rendered, the stack may look like this:

    ```py
    [
        TextPart("<div>  <!-- Normal HTML-->"),  # content before child comp

        # Child component's parts
        TextPart("<button>"),  # child's before text
        ComponentPart("ce6e6e6"),  # child's child comp
        TextPart("</button>"),  # child's after text

        # Component's parts end
        TextPart("</div>"),  # content after child comp
    ]
    ```

4. With this, I was able to inject the error handling logic by adding additional `ErrorPart`.

    - When we come across a `TextPart`, we simply append it to the result. There's nothing that can raise error.
    - But when handling the `ComponentPart`, we actually render the user's component definition. So that may raise error.

    If user's component raises an error, we insert the `ErrorPart` at the front of the queue. This basically says "STOP everything and handle error!". (see `handle_error()` on [line 263](https://github.com/django-components/django-components/pull/1231/files#diff-aa0763d60fe294562514f1ee18e7c3329454cd13467f385ba403c6fd7873d7b1R263))

    So the queue will go from:

    ```py
    [
        ComponentPart("ca1b2c3"),  # child comp
        TextPart("</div>"),  # content after child comp
    ]
    ```

    To

    ```py
    [
        ErrorPart(Error("Error while rendering child comp")), 
        TextPart("</div>"),  # content after child comp
    ]
    ```

    On the next iteration, we process the `ErrorPart` on [line 334](https://github.com/django-components/django-components/pull/1231/files#diff-aa0763d60fe294562514f1ee18e7c3329454cd13467f385ba403c6fd7873d7b1R334)

    This is where we call parent's `Component.on_render()` and `on_render_after()` with the error.

    If the parent doesn't recover from the error, we insert another `ErrorPart` to the front of the queue, bubbling the error to parent's parent. Thus eventually bubbling to the top.

    If parent DOES recover the error, we will use the string that the parent component returned from `on_render()` / `on_render_after()`. And we will continue rendering the rest of the component tree.

src/django_components/perfutil/component.py
```diff
@@ -187,85 +243,180 @@ def component_post_render(
...
+        try:
+            with component_error_message(full_path[1:]):
+                comp_content, grandchild_component_attrs, on_render_generator = curr_comp_renderer(curr_comp_attrs)
+        # This error may be triggered when any of following raises:
+        # - `Component.on_render()` (first part - before yielding)
+        # - `Component.on_render_before()`
+        # - Rendering of component's template
+        #
+        # In all cases, we want to mark the component as errored, and let the parent handle it.
+        except Exception as err:
```
JuroOravec, line 391: Also important to note that there's actually 2 places where we have to guard for user errors:

1. **Before/during** the render - e.g. when `Component.on_render_before()`, or first part of `Component.on_render()`
2. **After** rendering - this is when `Component.on_render_after()` and the second part of `Component.on_render()` run.

The second one is on [line 301](https://github.com/django-components/django-components/pull/1231/files#diff-aa0763d60fe294562514f1ee18e7c3329454cd13467f385ba403c6fd7873d7b1R301)

src/django_components/perfutil/component.py
```diff
@@ -305,3 +460,44 @@ def get_html_parts(component_id: str) -> List[str]:
...
+) -> Callable[[Optional[str], Optional[Exception]], OnComponentRenderedResult]:
+    if on_render_generator is None:
+        return inner_fn
+
+    def on_component_rendered_wrapper(
+        html: Optional[str],
+        error: Optional[Exception],
+    ) -> OnComponentRenderedResult:
+        try:
+            on_render_generator.send((html, error))
```
JuroOravec, line 477: This passes the fully rendered result / error into the `on_render()` funcition via `yield`:

```py
class Table(Component):
```

src/django_components/component.py
```diff
@@ -3280,16 +3592,31 @@ def renderer(root_attributes: Optional[List[str]] = None) -> Tuple[str, Dict[str
...
-            if template is not None:
-                # Get the component's HTML
-                html_content = template.render(context)
+            # Get the component's HTML
+            # To access the *final* output (with all its children rendered) from within `Component.on_render()`,
+            # users may convert it to a generator by including a `yield` keyword. If they do so, the part of code
+            # AFTER the yield will be called once, when the component's HTML is fully rendered.
+            #
+            # Hence we have to distinguish between the two, and pass the generator with the HTML content
+            html_content_or_generator = component.on_render(context, template)
```
JuroOravec, line 3597: When user includes `yield` keyword in `Component.on_render()`:

```py
class Table(Component):
```

Then the way Python works is that what the `on_render` method returns is actually a [Generator](https://docs.python.org/3/reference/expressions.html#yield-expressions) object.

It is this generator object which allows us to:
1. Receive the result of `template.render(context)` from the middle of the function
2. Pass back into the middle of function the result of rendering (`html, error`)
3. Internally pass the generator around.

JuroOravec:
Here's preview of the updated docs:

Component hooks:

https://youtu.be/DLiLtwbJ1_M

Dynamic template:

Screenshot 2025-06-02 at 19 51 44

JuroOravec:
@EmilStenstrom Another one ready for review. I tried to describe what's going on, but since some changes are in algo-heavy part, let me know if anything's not clear.

And after this PR the only thing that's left is https://github.com/django-components/django-components/issues/1206 🎉

src/django_components/perfutil/component.py
EmilStenstrom: Very elegant!

src/django_components/perfutil/component.py
EmilStenstrom: Again, extremely elegant. Good design! :)

CHANGELOG.md
```diff
@@ -586,6 +586,35 @@ Summary:
     )
     ```
 
+- Component method `on_render_after` was updated to receive also `error` field.
+
+    For backwards compatibility, the `error` field can be omitted until v1.
```
EmilStenstrom, line 591: "For extensions that needs to work with older versions of django-components, the `error` field can be omitted." - I don't think this has anything to do with v1?

CHANGELOG.md
```diff
@@ -1221,6 +1263,34 @@ Summary:
 
 - Passing `Slot` instance to `Slot` constructor raises an error.
 
+- Extension hook `on_component_rendered` now receives `error` field.
+
+    `on_component_rendered` now behaves similar to `Component.on_render_after`:
```
EmilStenstrom, line 1268: What is the difference between these two?

docs/concepts/advanced/hooks.md
```diff
@@ -1,58 +1,333 @@
...
+
+            if error is not None:
+                # The rendering failed
+                print(f"Error: {error}")
+    ```
+
+#### Example: ErrorBoundary
+
+[`on_render()`](../../../reference/api#django_components.Component.on_render) can be used to
+implement React's [ErrorBoundary](https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary).
```
EmilStenstrom, line 214: This will be replaced with our own ErrorBoundary component later on?

docs/concepts/advanced/hooks.md
```diff
@@ -1,58 +1,333 @@
...
+class ErrorFallback(Component):
+    template = """
+      {% slot "content" default / %}
+    """
+
+    def on_render(self, context, template):
+        fallback = self.slots.fallback
+
+        if fallback is None:
+            raise ValueError("fallback slot is required")
```
EmilStenstrom, line 242: Would expect this to be handled by setting the slot to required, but I guess this is a way to show that you can raise in on_render?

src/django_components/perfutil/component.py
JuroOravec: Thanks, happy to hear that :)

docs/concepts/advanced/hooks.md
```diff
@@ -1,58 +1,333 @@
...
+
+        html, error = yield template.render(context)
+
+        if error is not None:
+            return fallback()
```
EmilStenstrom, line 247: Is fallback a callable here? Did not expect. fallback.render()? Returning a slot? Maybe this should be a new return type?

CHANGELOG.md
JuroOravec: No no this one is not about extensions, it's about `Component.on_render_after` which is a component method.

But yeah it can be confusing because there's 4 things being discussed with similar names:

- `Component.on_render_before()`
- `Component.on_render()`
- `Component.on_render_after()`
- `Extension.on_component_rendered()`

Before this PR, `Component.on_render_after()` took 4 arguments, after this PR it will be 5. So if someone defined the method previously, and didn't use `*args` or `**kwargs` at the end, then Python would raise an error

CHANGELOG.md
JuroOravec: `on_component_rendered` is for extension authors while `Component.on_render_after` is for Component authors.

See https://github.com/django-components/django-components/pull/1231/files/c3026d96b195ad805ac46b6ac9c1464e25e2753f#r2127398315.

docs/concepts/advanced/hooks.md
JuroOravec: Yeah this was just a toy example, the actual implementation will look a bit different.

Here I rendered the slot in Python. But in the actual implementation I will use `{% slot %}` template tag.

I actually played with the ErrorFallback implementation today, so you just reminded me to also update this code, thanks!

docs/concepts/advanced/hooks.md
JuroOravec: Yup! The slots accessed from inside the component's Python code are all normalized to `Slot` instances. And these are callable, so one can treat them as functions.

PR #1233, opened 2025-06-04
refactor: deprecate Component.input and add raw_args, raw_kwargs, raw_slots

### LINKED ISSUE #1206: Deprecate Component.input
I think we can keep it for now, it ensures that there's a way to access all the inputs that a component received when rendered.

Without `input`, we would either have to:
- Make some render inputs inaccessible 
- Or set all inputs as Component attributes

Altho, looking at my latest code, all the inputs to `Component.render()` already are set on the Component instance as attributes, except for `deps_strategy`...

On the other hand, the `input.args`, `input.kwargs`, `input.slots` are handy when writing extensions. Because these are plain dicts / lists. Whereas the `args` / `kwargs` / `slots` on the Component instance may be instances of any exotic class that might not be possible to revert back to dict / lists.

One more for removing `input` however - I see that in my code it has already drifted - that I forgot to add some fields to the `input` object that are being passed to `Component.render()`.

So maybe, we could remove `input`, and instead:
1. Add `deps_strategy` as Component instance attribute
2. Add `raw_args`, `raw_kwargs`, `raw_slots` instance attributes, which would hold the plain lists / dicts.

_Originally posted by @JuroOravec in https://github.com/django-components/django-components/pull/1205#discussion_r2105923687_

JuroOravec:
Last one before v0.140 🎉 🎉 @EmilStenstrom thank you for your infinite patience! 😄 

As mentioned in #1206, this marks `Component.input` as deprecated.

In order to avoid using `Component.input`, following Component attributes were added:

- `Component.raw_args` - Same as `input.args` - Positional args. Unlike `Component.args` this stays as plain list even when `Component.Args` is given.
- `Component.raw_kwargs` - Same as above, but for `kwargs`
- `Component.raw_slots` - Same as above, but for `slots`
- `Component.deps_strategy` - Stores the value passed to `Component.render()`, because it was stored only on `input.deps_strategy`.

Closes #1206

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -609,9 +609,9 @@ Summary:
     ```py
     class MyComponent(Component):
         def on_render(self, context, template):
-            assert self.raw_args == [1, 2, 3]
-            assert self.raw_kwargs == {"a": 1, "b": 2}
-            assert self.raw_slots == {"my_slot": "CONTENT"}
+            assert self.args == [1, 2, 3]
+            assert self.kwargs == {"a": 1, "b": 2}
+            assert self.slots == {"my_slot": "CONTENT"}
             assert self.context == {"my_slot": "CONTENT"}
             assert self.deps_strategy == "document"
             assert (self.deps_strategy != "ignore") is True
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -586,6 +586,66 @@ Summary:
     )
     ```
 
+- `Component.input` (and its type `ComponentInput`) is now deprecated. The `input` property will be removed in v1.
+
+    Instead, use attributes directly on the Component instance.
+
+    Before:
+
+    ```py
+    class MyComponent(Component):
+        def on_render(self, context, template):
+            assert self.input.args == [1, 2, 3]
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -586,6 +586,37 @@ Summary:
...
+    ```
+
+    After:
+
+    ```py
+    class MyComponent(Component):
+        def on_render(self, context, template):
+            assert self.raw_args == [1, 2, 3]
+            assert self.raw_kwargs == {"a": 1, "b": 2}
+            assert self.raw_slots == {"my_slot": "CONTENT"}
```
EmilStenstrom: Maybe the recommendation should be self.args/kwargs/slots here instead? I'm thinking of "raw_"-attributes as most usable for extensions, but on_render is for the component author, right?

JuroOravec: Yup, that's good point, updated! I originally used the `raw_` variants because that's the literal equivalent.

JuroOravec:
And we're live! https://github.com/django-components/django-components/releases/tag/0.140.0 🕺

PR #1242, opened 2025-06-07
feat: on_xx_loaded extension hooks

JuroOravec:
This adds 4 new extension hooks `on_template_loaded`, `on_js_loaded`, `on_css_loaded`, and `on_template_compiled`

The first 3 hooks are called when Component's template / JS / CSS is loaded as a string.

The `on_template_compiled` hook is called when Component's template is compiled to a Template.

```py
    def on_template_loaded(self, ctx: OnTemplateLoadedContext) -> Optional[str]:
        return ctx.content + "<!-- Hello! -->"

    def on_js_loaded(self, ctx: OnJsLoadedContext) -> Optional[str]:
        return ctx.content + "// Hello!"

    def on_css_loaded(self, ctx: OnCssLoadedContext) -> Optional[str]:
        return ctx.content + "/* Hello! */"
```

Secondly, to get this working correctly, I had to update the logic for how the `Template` instances are created from `template` / `template_file`. The issue was that when subclassing components where the parent defined `template` / `template_file`, and child didn't define anything, then the child was missing the Template. Instead, we want the child to have it's own instance of template, so as if `template` / `template_file` were defined directly on the child component.

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -32,7 +32,7 @@ from django_components.component_media import ComponentMediaInput, ComponentMedi
 from django_components.component_registry import ComponentRegistry
 from django_components.component_registry import registry as registry_
 from django_components.constants import COMP_ID_PREFIX
-from django_components.context import _COMPONENT_CONTEXT_KEY, make_isolated_context_copy
+from django_components.context import _COMPONENT_CONTEXT_KEY, COMPONENT_IS_NESTED_KEY, make_isolated_context_copy
 from django_components.dependencies import (
     DependenciesStrategy,
     cache_component_css,
@@ -2314,9 +2314,6 @@ class Component(metaclass=ComponentMeta):
         cls.class_id = hash_comp_cls(cls)
         comp_cls_id_mapping[cls.class_id] = cls
 
-        # Make sure that subclassed component will store it's own template, not the parent's.
-        cls._template = None
-
         ALL_COMPONENTS.append(cached_ref(cls))  # type: ignore[arg-type]
         extensions._init_component_class(cls)
         extensions.on_component_class_created(OnComponentClassCreatedContext(cls))
@@ -3517,14 +3514,14 @@ class Component(metaclass=ComponentMeta):
         #     Then we can simply apply `template_data` to the context in the same layer
         #     where we apply `context_processor_data` and `component_vars`.
         with prepare_component_template(component, template_data) as template:
-            # Set `Template._djc_is_component_nested` based on whether we're currently INSIDE
+            # Set `_DJC_COMPONENT_IS_NESTED` based on whether we're currently INSIDE
             # the `{% extends %}` tag.
             # Part of fix for https://github.com/django-components/django-components/issues/508
             # See django_monkeypatch.py
             if template is not None:
-                template._djc_is_component_nested = bool(
-                    context.render_context.get(BLOCK_CONTEXT_KEY)  # type: ignore[union-attr]
-                )
+                comp_is_nested = bool(context.render_context.get(BLOCK_CONTEXT_KEY))  # type: ignore[union-attr]
+            else:
+                comp_is_nested = False
 
             # Capture the template name so we can print better error messages (currently used in slots)
             component_ctx.template_name = template.name if template else None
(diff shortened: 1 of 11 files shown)


src/django_components/component.py
```diff
@@ -3517,14 +3514,14 @@ def _render_impl(
         #     Then we can simply apply `template_data` to the context in the same layer
         #     where we apply `context_processor_data` and `component_vars`.
         with prepare_component_template(component, template_data) as template:
-            # Set `Template._djc_is_component_nested` based on whether we're currently INSIDE
```
JuroOravec, line 3520: Not related to the PR, but now that the behaviour about managing Templates is clear (each Component class caches and reuses its own Template instance), I realized that this code for compatibility with `{% extends %}` isn't ideal, because it modifies the attribute on `Template` instance on each render.

So instead, the flag that is set for the compatibility is set as a context variable, so we don't mutate the Template at render time.

src/django_components/component_media.py
```diff
@@ -43,7 +45,7 @@
 
 
 # These are all the attributes that are handled by ComponentMedia and lazily-resolved
-COMP_MEDIA_LAZY_ATTRS = ("media", "template", "template_file", "js", "js_file", "css", "css_file")
+COMP_MEDIA_LAZY_ATTRS = ("media", "template", "template_file", "js", "js_file", "css", "css_file", "_template")
```
JuroOravec, line 48: Accesing `Component._template` now behaves like the other "media attributes". that is, accessing this attribute triggers lazy loading of the HTML / JS / CSS, if not loaded / resolved yet.

src/django_components/component_media.py
```diff
@@ -611,20 +620,39 @@ def get_template_data(self, args, kwargs, slots, context):
...
+    comp_media.js = js_str
+
+    css_str, _ = _get_asset(comp_cls, comp_media, inlined_attr="css", file_attr="css_file", comp_dirs=comp_dirs)
+    comp_media.css = css_str
+
+    # If `Component.template` or `Component.template_file` were explicitly set on this class,
+    # then Template instance was already created.
+    #
+    # Otherwise, search for Template instance in parent classes, and make a copy of it.
+    if not isinstance(template_obj, Unset):
```
JuroOravec, line 642: This is the new logic that fixes the bug - if the current Component class doesn't define any template, then look up the parent tree to see if anyone defined them. If yes, create a copy of Template. if no, then this child component has no template either.

src/django_components/component_media.py
```diff
@@ -1061,42 +1088,68 @@ class MyComp(Component):
...
-    if full_path is None:
-        # NOTE: The short name, e.g. `js` or `css` is used in the error message for convenience
-        raise ValueError(f"Could not find {inlined_attr} file {asset_file}")
-
-    # NOTE: Use explicit encoding for compat with Windows, see #1074
-    asset_content = Path(full_path).read_text(encoding="utf8")
-
-    # TODO: Apply `extensions.on_js_preprocess()` and `extensions.on_css_preprocess()`
-    # NOTE: `on_template_preprocess()` is already applied inside `load_component_template()`
+    # then there's nothing to resolve. Use it as is.
```
JuroOravec, line 1091: This was refactored so that instead of returning only the content (HTML / JS / CSS) as string, it  now also returns the Template instance.

src/django_components/extension.py
```diff
@@ -763,6 +799,108 @@ def on_component_rendered(self, ctx: OnComponentRenderedContext) -> Optional[str
         """
         pass
 
+    ##########################
+    # Template / JS / CSS hooks
+    ##########################
+
+    def on_template_loaded(self, ctx: OnTemplateLoadedContext) -> Optional[str]:
```
JuroOravec, line 806: Definition of the new hooks

src/django_components/template.py
```diff
@@ -170,30 +170,45 @@ def _maybe_bind_template(context: Context, template: Template) -> Generator[None
 loading_components: List["ComponentRef"] = []
 
 
-def load_component_template(component_cls: Type["Component"], filepath: str) -> Template:
-    if component_cls._template is not None:
-        return component_cls._template
+def load_component_template(
```
JuroOravec, line 173: Previously this function set `Component._template`. Since that attribute is now managed in `component_media.py` , this function only returns the value without assigning it to the class

tests/test_component_media.py
```diff
@@ -1037,73 +1050,121 @@ def test_both_js_and_js_file_none(self):
...
 
         assert TestComp.js is None
         assert TestComp.js_file is None
+        assert TestComp.template is None
+        assert TestComp.template_file is None
+
+        assert TestComp._template is None
+
+        assert isinstance(ParentComp._template, Template)
+        assert ParentComp._template.source == "<h1>parent</h1>"
```
JuroOravec, line 1144: Updated `component_media` tests to also check for `Component._template`.

`Component._template` is still an internal field, but we access it when rendering, so it makes sense to ensure that it's pointing to a correct Template.

EmilStenstrom (approved):
Nice addition!

PR #1250, opened 2025-06-10
refactor: fix wrongly initiated settings

JuroOravec:
Previously, our settings were properly loaded only once we've reached the `AppConfig.ready()` hook in `apps.py`.

This worked well when running the app normally.

But with autoreload, turns out that when the app failed and Django restarted, it called template loaders BEFORE we've reached `AppConfig.ready()`:

1. Our template loader (`src/django_components/template_loader.py`) calls [`get_component_dirs()`](https://github.com/django-components/django-components/blob/2350a6b6c423837266a3959a2a798b3cfa3d6f48/src/django_components/template_loader.py#L28) . 

2. And that in turn needs to access our settings [here](https://github.com/django-components/django-components/blob/2350a6b6c423837266a3959a2a798b3cfa3d6f48/src/django_components/util/loader.py#L46), which were not properly set up yet.

So the fix to this was to load our settings lazily when first accessed the first time, instead of at `apps.py`.

Closes https://github.com/django-components/django-components/issues/1232


src/django_components/app_settings.py
```diff
@@ -853,70 +858,73 @@ def _prepare_context_behavior(self, new_settings: ComponentsSettings) -> Literal
 
         return raw_value
 
-    # TODO REMOVE THE PROPERTIES BELOW? THEY NO LONGER SERVE ANY PURPOSE
```
JuroOravec, line 856: Previously I was thinking of removing these properties. But now we need them to lazily load our settings, so I removed the TODO

docs/scripts/reference.py
```diff
@@ -524,7 +531,7 @@ def gen_reference_commands():
             )
 
 
-def gen_reference_templatetags():
+def gen_reference_template_tags():
```
JuroOravec, line 534: Unrelated to the bug, but renamed the functions / file names in the generated docs so we use full names like "template variables" instead of "template vars"

docs/scripts/reference.py
```diff
@@ -465,7 +466,13 @@ def gen_reference_commands():
             # Add link to source code
             module_abs_path = import_module(cmd_def_cls.__module__).__file__
             module_rel_path = Path(module_abs_path).relative_to(Path.cwd()).as_posix()  # type: ignore[arg-type]
-            obj_lineno = inspect.findsource(cmd_def_cls)[1]
+
+            # NOTE: Raises `OSError` if the file is not found.
+            try:
+                obj_lineno = inspect.findsource(cmd_def_cls)[1]
```
JuroOravec, line 472: Unrelated to the bug, but:
- Our settings are now lazily loaded on access.
- There are some bits when we are generating our API reference documentation that rely on Django settings being configured
- So I had to add `settings.configure()` on [line 1111](https://github.com/django-components/django-components/pull/1250/files#diff-a92bcbf4ada28ed45f9610be9e27a66eb1308eb47fd87836aa018df5ca9b14e0R1111)
- Since Django settings are now configured, it means that also command classes are built for our built-in extensions

Here on the line 472 we automatically generate documentation for all our [CLI commands](https://django-components.github.io/django-components/latest/reference/commands/) from those Command classes.

But `inspect.findsource()` raises error when it is given a class that was created dynamically with `type()`. Because such class was created in-memory, and so it has no source file.

And because we run Django's `settings.configure()`, it means that the list of all our available CLI commands now includes also dynamically generated Command classes for some of our built-in extensions.

So the try / catch serves to ignore those dynamically generated commands.

src/django_components/util/exception.py
```diff
@@ -25,7 +25,7 @@ def component_error_message(component_path: List[str]) -> Generator[None, None,
             if not components:
                 orig_msg = str(err.args[0])
             else:
-                orig_msg = err.args[0].split("\n", 1)[-1]
+                orig_msg = str(err.args[0]).split("\n", 1)[-1]
```
JuroOravec, line 28: Unrelated to the PR, but there was also another occasional bug when inside a component we raised an error and `Error.args[0]` was NOT a string.

For example `KeyError` has as `args[0]` the key which we tried to access. And the key doesn't have to be string. So in those cases this formatting failed.

JuroOravec:
cc @EmilStenstrom

PR #1251, opened 2025-06-10
refactor: in `components ext run`, display only those extensions that actually have subcommands

JuroOravec:
Another small cleanup. The `components ext run` CLI command now displays only those extensions that actually define any subcommands. [See example here](https://github.com/django-components/django-components/pull/1251/files#diff-acd1edd77ed6a5df30cf62e778556047d3a27459db2aa1fac3d6c3b50c29fbe2R209).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@
 
 - Fix bug when error formatting failed when error value was not a string.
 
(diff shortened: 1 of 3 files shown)


JuroOravec:
cc @EmilStenstrom

PR #1283, opened 2025-07-03
refactor: Fix #1277 + Cache components' JS/CSS scripts at class creation

### LINKED ISSUE #1277: Dynamic components are missing a template object which raises an `AttributeError` in tooling when trying to access `.name`
I'm newish to Django so I may not be able to describe this with perfect clarity, but I've noticed the framework throws an `AttributeError` in (at least) a couple cases when using the new [Dynamic Component](https://django-components.github.io/django-components/latest/concepts/fundamentals/rendering_components/?h=pagination#dynamic-components). 

It would seem that various Django tools expect a template object and name attribute to be defined when the class instance for the module is accessed. For example, in the [`response` object](https://docs.djangoproject.com/en/5.2/topics/testing/tools/#django.test.Response.templates) of test responses. (This may or may not be a fair assumption for tools to make. 😅)

---

### Using
- Django: v5.2
- django-components: (>=0.141.0,<0.142.0)
- django-debug-toolbar: (>=5.2.0,<6.0.0)

---

### Examples:

1. When using the Django test helper [`assertTemplateUsed`](https://docs.djangoproject.com/en/5.2/topics/testing/tools/#django.test.SimpleTestCase.assertTemplateUsed)

Image

2. When using Django Debug toolbar

    Error output in browser:
    Image

    Here's the `None` object the toolbar shows in it's template list when the toolbar is expanded:
    Image

JuroOravec:
## Fix: JS/CSS scripts caching

At my work I'm finally getting around to make full use of django_component's  HTML fragment support.

There, there are a few pages that follow the pattern where the majority of the HTML document is static, and only the main content is being replaced with AJAX.

However, one bug(?) that I found yesterday was that, early after the release, I had to visit some of the pages multiple times for them to properly load. The HTML fragments that were being inserted with AJAX contained new Django components, and these components had their own JS/CSS. But JS/CSS file (e.g. from `Component.js_file` or `Component.js`) were missing from the cache, and so the page failed to load.

Today, all is working well, so the problem was possibly with HTTP response caching, and not django-components.

Nevertheless, when looking into the caching, I noticed that our cache for those JS/CSS files (`LocMemCache`) was misconfigured. Instead of being unbounded, it was using the Django's default for max entries in the cache (300).

I'm also thinking that the problem with the AJAX could have been that the JS/CSS files are set to the cache lazily, and so it's hard to say if the entry is already in the cache or not. It's hard to trace the issue.

So one change I made is to populate the cache with JS/CSS files eagerly at component class creation. That way, when I'm using the default `LocMemCache`, I know at all JS/CSS files should be available after the server start.

---

## Fix: Do not send `template_rendered` signal when component has no template

While at it I also fixed https://github.com/django-components/django-components/issues/1277.

Closes https://github.com/django-components/django-components/issues/1277


src/django_components/component.py
```diff
@@ -3695,7 +3696,8 @@ def renderer(
             component.on_render_before(context, template)
 
             # Emit signal that the template is about to be rendered
-            template_rendered.send(sender=template, template=template, context=context)
+            if template is not None:
+                template_rendered.send(sender=template, template=template, context=context)
```
JuroOravec, line 3700: This is the bit that fixes #1277

src/django_components/extensions/dependencies.py
```diff
@@ -0,0 +1,21 @@
...
+    This extension is automatically added to all components.
+    """
+
+    name = "dependencies"
+
+    # Cache the component's JS and CSS scripts when the class is created, so that
+    # components' JS/CSS files are accessible even before having to render the component first.
+    def on_component_class_created(self, ctx: OnComponentClassCreatedContext) -> None:
+        cache_component_js(ctx.component_cls, force=True)
+        cache_component_css(ctx.component_cls, force=True)
```
JuroOravec, line 21: This is where the components' JS/CSS files are cached. This is used for the HTML fragments support, so that the JS/ CSS can be lazily accessed via endpoint by our JS dependency manager from within the browser.

Implemented as an extension, so that 1) this runs for all components, and 2) so that this runs after Django is set up.

tests/test_autodiscover.py
```diff
@@ -50,8 +50,6 @@ class TestImportLibraries:
         }
     )
     def test_import_libraries(self):
-        # Ensure we start with a clean state
-        registry.clear()
```
JuroOravec, line 54: For unknown reason this was throwing errors when running both this test file and `tests/test_component.py`. But luckily it's fine to remove it.

JuroOravec:
cc @EmilStenstrom

JuroOravec:
I'll go ahead and merge / release this, as it's blocking me at work from releasing.

PR #1284, opened 2025-07-03
Add register decorator to Calendar class in Quickstart section of the document.

Antoliny0919:
It would be even better if code for registering the component is included!


Antoliny0919:
Are there any other places besides the README where the same change should be applied?

README.md
```diff
@@ -24,8 +24,9 @@ A component in django-components can be as simple as a Django template and Pytho
 
 ```py
 # components/calendar/calendar.py
-from django_components import Component
+from django_components import Component, register
 
+@register("calendar")
```
Antoliny0919, line 29: Would it be better to add an explanation about adding register and separate it into a distinct code block?

JuroOravec (approved):

PR #1286, opened 2025-07-06
docs: fix parametrising components section next step line position

Antoliny0919:
The phrase indicating the "next step" seemed to be in the wrong position.
I have moved it so that it appears at the end.


JuroOravec:
Thanks for fixing this! 🙏

Issue #1281, opened 2025-07-01
Multiple attributes with the same name not being merged in {% html_attrs %}


JuroOravec:
I came across a case where I wanted to use `{% html_attrs %}` to merge CSS classes coming from two sources by defining `class=...` kwarg twice. I expected that this would merge the two.

```django
  {% html_attrs attrs
    class=btn_class
    class="inline-flex w-full text-sm font-semibold no-underline sm:mt-0 sm:w-auto focus-visible:outline-2 focus-visible:outline-offset-2"
  %}
```

But in reality, only the first `class` definition got rendered (variable `btn_class`). Second definition, with the literal `"inline-flex..."` got ignored.

Here's full component example:

```py
from typing import Literal

from django_components import Component, SlotInput, register
from pydantic import BaseModel

from app_home.constants.theme import ThemeColor, ThemeVariant, get_styling_css

@register("Button")
class Button(Component):
    template_file = "button.html"

    class Kwargs(BaseModel):
        href: str | None = None
        link: bool | None = None
        disabled: bool | None = False
        variant: ThemeVariant | Literal["plain"] | None = "primary"
        color: ThemeColor | None = "default"
        type: str | None = "button"
        attrs: dict | None = None

    class Slots(BaseModel):
        content: SlotInput | None = None
        default: SlotInput | None = None

    def get_template_data(self, args, kwargs: Kwargs, slots: Slots, context):
        if kwargs.variant == "plain":
            btn_class = ""
        else:
            button_classes = get_styling_css(
                kwargs.variant,
                kwargs.color,
                kwargs.disabled,
            )
            btn_class = f"{ button_classes } px-3 py-2 justify-center rounded-md shadow-sm"

        is_link = not kwargs.disabled and (kwargs.href or kwargs.link)

        all_attrs = {
            **(kwargs.attrs or {}),
        }
        if kwargs.disabled:
            all_attrs["aria-disabled"] = "true"

        return {
            "href": kwargs.href,
            "disabled": kwargs.disabled,
            "type": kwargs.type,
            "btn_class": btn_class,
            "attrs": all_attrs,
            "is_link": is_link,
        }
```

```django
{# Based on buttons from https://tailwindui.com/components/application-ui/overlays/modals #}

{% if is_link %}
<a
  href="{{ href }}"
  {% html_attrs attrs
    class=btn_class
[...]

Issue #1258, opened 2025-06-11
Auto-extract defaults from namedtuple, dataclass, pydantic


JuroOravec:
Just a small quality of life improvement.

As I'm using more features from v0.140 in my project, I notice that there's some code duplication when defining kwargs and their defaults:

```py
@register("Autocomplete")
class Autocomplete(Component):
    template_file = "autocomplete.html"
    js_file = "autocomplete.js"

    class Kwargs(BaseModel):
        name: str | None = None
        editable: bool | None = None
        multiple: bool | None = None
        selected_items: list[AutocompleteItem] | None = None
        max_width: int | str | None = None
        placeholder: str | None = None
        attrs: dict | None = None
        js: dict | None = None

    class Defaults:
        name = "value"
        max_width = "300px"
        placeholder = "Search..."
        js = Default(lambda: {})
        editable = True
```

From implementation's perspective this makes sense - we (DJC) cannot assume what class will `Kwargs` class. Maybe it will contain defaults, may it won't.

But in real world usage, I assume that `Kwargs` will be mostly like one of 3 kinds:
- NamedTuple
- Dataclass
- Pydantic model

So I think we could add support for these, such that the above could be condensed into

```py
@register("Autocomplete")
class Autocomplete(Component):
    template_file = "autocomplete.html"
    js_file = "autocomplete.js"

    class Kwargs(BaseModel):
        name: str | None = "value"
        editable: bool | None = True
        multiple: bool | None = None
        selected_items: list[AutocompleteItem] | None = None
        max_width: int | str | None = "300px"
        placeholder: str | None = "Search..."
        attrs: dict | None = None
        js: dict | None = Field(default_factory=lambda: {})
```

Since Pydantic is a 3rd party, I'm thinking:
- `NamedTuple` and dataclass support should be implemented in core DJC
- Pydantic model support should live in [`djc-ext-pydantic`](https://github.com/django-components/djc-ext-pydantic)

EmilStenstrom:
Looks great, love the simplification of the interface!

JuroOravec:
One more thing to make the use of `Kwargs` / `Args` / `Slots` classes easier - when I define these classes, but I don't define any parent class, these should automatically subclass either `NamedTuple` or (frozen) `Dataclass`.

So this:

```py
class Autocomplete(Component):
    class Kwargs:
        name: str | None = None
        editable: bool | None = None
```

Would be automatically treated as:

```py
class Autocomplete(Component):
[...]

Issue #1256, opened 2025-06-10
[docs] Update outdated examples in htmx_components


JuroOravec:
See https://github.com/django-components/django-components/pull/1254

JuroOravec:
I implemented the examples on top of [Citry](https://citry.dev) (the successor to django-components), as one of the example projects in [`examples/demos/htmx`](https://github.com/citry-dev/citry/tree/main/examples/demos/htmx).

Shout out to @dylanjcastillo - The examples you originally wrote for django-components were the inspo for the demo Citry+HTMX integration.

Issue #794, opened 2024-11-30
Django-cotton-like html tag syntax for django-components


JuroOravec:
### Discussed in https://github.com/EmilStenstrom/django-components/discussions/792

<div type='discussions-op-text'>

<sup>Originally posted by **oliverhaas** November 29, 2024</sup>
I (and one of my employees) tinkered with adding django-cotton-like syntax to django-components, because I initially loved the django-cotton syntax, but also wanted some of the functionality of django-components that django-cotton lacks. 

https://gist.github.com/oliverhaas/49a1683c6a0a4d5bac107dbfd72b283c

I would say that this is more than a POC, but it does not all the reasonable features it in principle could.

There are some upsides to the syntax (nicer formatting and normal html css class autocomplete for example), but I'm not sure anymore if the syntax is really that beneficial, since among other things you'll end up with kind of a hybrid of traditional django templates and JSX-like custom html tags. I thought I'd share, in case anyone is interested in this.

</div>

JuroOravec:
## How to differentiate static HTML attribute from dynamic input?

In https://github.com/django-components/djc-core/pull/17#issuecomment-3491233133, I got an idea how to differentiate between HTML attributes and Python attributes (component inputs).

The referenced PR adds Python expressions by demarking it with parentheses, e.g. `abc=( 1 + a )`.

Same thing, parentheses, could be reused to know whether an attribute is a static HTML attribute or component input. so these two would be equivalent:

- Django:
	```django
	{% component "my_table" data=data rows=rows attrs:class="ma-4" / %}
	```
- HTML:
	```html
	<my_table data=( data ) rows=( rows ) class="ma-4" />
	```


Altho one remaining question is - The expression inside the parentheses, should it be JUST python expression? Or should it allow other things like filters and such?

- a) Just Python expressions:
	```html
	<my_table data=( data[0].name ) class="ma-4" />
	```
- b) Also filters and other
	```html
	<my_table data=( data|first|get:"name" ) class="ma-4" />
	```

b) is more Django-friendly - one could use existing filters and template tags inside the parentheses.

a) is what I believe the API will be in v3, when this project is Django-independent. So it would skip a few steps ahead.

I think option b) could be a little confusing, because `a=(...)` will have different behavior in Django and in HTML modes:
- In Django, `()` means python expression `{% card first_name=( name.split(" ")[0] ) %}`
- In HTML, `()` will mean component input. `<card first_name=( (name.split(" ")[0]) ) >`

[...]

PR #1289, opened 2025-07-06
docs: add next step line for Adding slots to Rendering components

Antoliny0919:


JuroOravec:
@Antoliny0919 Thanks again for all the fixes :)

Antoliny0919:
Hello @JuroOravec :)
I'm currently working on developing the [djangosnippets](https://github.com/django/djangosnippets.org) site and have recently integrated django-components into the project.

Yesterday, I started refactoring the existing pagination by using django-components and submitted a PR for it (though the work is still in progress).

I'm very interested in django-components and would love to contribute more actively to the project.

While reading through the documentation, I noticed a few areas that could be improved. I’ll create issues for those suggestions soon.

Are there any issues that would be good for beginners to work on? If so, I’d really appreciate it if you could let me know!

JuroOravec:
@Antoliny0919 That's awesome!

> Re [djangosnippets](https://github.com/django/djangosnippets.org)

I'm migrating my work project to django-components v0.141. After that's done, I'd love to share the generic components (e.g. autocomplete field) and the design patterns I follow to keep things managable and AI-friendly.

Need maybe only 2-3 full days to finish the migrations, but lots is going on right now, so I might not get to it till mid Aug ([see this comment](https://github.com/django-components/django-components/issues/1281#issuecomment-3045674186) at the end).

> Yesterday, I started refactoring the existing pagination by using django-components and submitted a PR for it (though the work is still in progress).

Send me a link and I can provide feedback.

> Re contributing

Absolutely! Also I joined just what, maybe a year ago? (March/April last year). First wanted to improve just a few things, then I got more and more involved.

Do you also have another project where you want to use django-components?

- If no and if you're new to django-components, then #1256 might be good start to get a hang of how to write the components.

- If yes and you already know how to write components, you may want to look into #1281 or #1258. They are both self-isolated, small in scope, and not blocked by anything.

- If you're interested in high-impact work, then after the two above I recommend #794. The idea is to allow people to write the templates as pure HTML, so using `< ... >` instead of Django's `{% ... %}`.
  - This is one of the two missing features because of which people choose other libraries (like [django-cotton](https://github.com/wrabit/django-cotton)) over django-components (the other being template-only components https://github.com/django-components/django-components/issues/1240)
  - What needs to be done here is mostly to explore the designs:
    1. E.g. should we handle only component template tags, or make it work for all template tags?
       - So `<c-button key=val />` -> `{% component "button" key=val / %}`
       - Versus `<d-if>` -> `{% if %}`
       - NOTE: I'd go with both as above, so components with `c-` prefix and other Django template tags with `d-` prefix. Reason for this is that HTML file formatting doesn't work well when a file contains django template tags.
    2. How to differentiate between static HTML attributes vs Django template variables:
       - To make it work well with IDEs, the HTML attributes should be set without anything extra needed. So that means that 
we'd need an extra syntax for Django variables:
       - So `<c-button class="btn" />` (static HTML attribute), which would be converted to `{% component "button" attrs:class="btn" / %}`
       - Versus something like `<c-button d-variant="secondary" />`, which would be converted to `{% component "button" variant="secondary" / %}`
    3. Do we need to wrap everything in quotes when specifying django variables? How would it look like if I wanted to specify a literal list or dict?
       - E.g.  `<c-select d-items="[1, 2, 3]" />` vs `<c-select d-items=[1, 2, 3] />`
       - I'd go with always wrapping in quotes, to ensure that IDE / formatters correctly treat these as HTML attributes.
       - So that means that passing a literal string as a django variable would look like `<c-button d-variant="'secondary'" />`
  - Implementation-wise, this is fairly self-contained - it would be implemented as [an extension](https://django-components.github.io/django-components/latest/concepts/advanced/extensions/) that pre-processes the Django template. So text in, text out. Potentially regexes could suffice for first implementation. So I think doable to have it vibe-coded with LLMs.

Antoliny0919:
Thank you so much for your incredibly kind and thoughtful reply @JuroOravec 🙌
I can really feel your deep passion for `django-components`.

This is actually my first time using `django-components`.
I introduced it to the djangosnippets project, and as I mentioned earlier, I submitted a PR where I refactored the previously implemented pagination using `django-components`.
I haven't written tests yet, but I believe the overall structure is more or less complete.
I'll mention you in the PR(I’d be very grateful if you could take a look and provide a review!)

Also, thank you for sharing so many helpful tips regarding contribution.
Since I’ve only read parts of the documentation so far, I think I’ll need to familiarize myself with the codebase a bit more before I can contribute to the internals of `django-components`.
That said, the tasks you suggested sound really appealing.
The idea of using Django components in a regular HTML tag-like syntax already excites me
I can’t wait to see that become a reality!

Anyway, thank you once again for your generous and encouraging message.
I’ll try to make time to dive deeper into the codebase and hopefully start working on the issues you recommended 😃

PR #1300, opened 2025-07-20
fix: Fix bug where JS and CSS were missing when `{% component %}` tag was inside `{% include %}` tag

JuroOravec:
As the title says - when there was a template with `{% include %}` and the inner included template contained `{% component %}`, then the JS/CSS of that inner component were missing.

The cause of this was the changes introduced in v0.140, where calling `Template.render()` by default also renders the JS/CSS. But the default mode is the smart "document" strategy, which inserts JS/CSS only inside `<body>/<head>` tags, or explicitly into `{% component_dependencies %}` placeholders.

Well and the issue was that the `{% include %}` tag renders the included template by calling `Template.render()` ([source](https://github.com/django/django/blob/a5b0a618c3ff16f9f52a9abdc89c05899ef112b4/django/template/loader_tags.py#L208)).

So the solution was to monkeypatch `{% include %}` tag (`IncludeNode.render()`), to modify it so that it renders the included template without rendering the JS/CSS, leaving JS/CSS rendering to be handled by the parent template.

Closes https://github.com/django-components/django-components/issues/1296


tests/templates/component_inside_include_base.html
```diff
@@ -1,4 +1,4 @@
-{% include "test_cached_component_inside_include_sub.html" %}
```
JuroOravec, line 1: Renamed the `test_cached_component_inside_include` files so that they can be reused also for tests for this bug

src/django_components/util/django_monkeypatch.py
```diff
@@ -210,5 +211,37 @@ def _template_render(self: Template, context: Context, *args: Any, **kwargs: Any
...
+    #
+    # Fix for https://github.com/django-components/django-components/issues/1296
+    if is_cls_patched(include_node_cls):
+        # Do not patch if done so already. This helps us avoid RecursionError
+        return
+
+    orig_include_render = include_node_cls.render
+
+    # NOTE: This implementation is based on Django v5.1.3)
+    def _include_render(self: IncludeNode, context: Context, *args: Any, **kwargs: Any) -> str:
```
JuroOravec, line 239: This is the actual fix - the original `IncludeNode.render()` was wrapped in
```py
with context.update({_STRATEGY_CONTEXT_KEY: "ignore"}):
```

JuroOravec:
cc @EmilStenstrom

EmilStenstrom (approved):
Nice and general fix, as I would expect, well done! :)

PR #1301, opened 2025-07-20
Create devcontainer.json

JuroOravec:
Turns out that there's a way to configure what container should be created when one uses Github's codespace feature (AKA the IDE in browser).

I was looking into this because I wanted to run the codebase in the browser in Python 3.13. It seems that the behaviour of Mypy is a bit different than in previous versions of Python, which led to this error in this pipeline:

https://github.com/django-components/django-components/actions/runs/16400669363/job/46339620229?pr=1300

And on my laptop I currently have Python 3.11.

The container config looks like a good starting point. Down the line we can add some more extensions so it's easier to start for others who might want to submit a small PR or play around.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,30 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.245.0/containers/python-3
+{
+	"name": "Python 3.13",
+	"image": "mcr.microsoft.com/devcontainers/python:3.13-bullseye",
+	// Configure tool-specific properties.
+	"customizations": {
+		// Configure properties specific to VS Code.
+		"vscode": {
+			// Set *default* container specific settings.json values on container create.
+			"settings": { 
+				"python.defaultInterpreterPath": "/usr/local/bin/python",
+				"python.linting.enabled": true
+			},
+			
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,31 @@
...
+				"python.defaultInterpreterPath": "/usr/local/bin/python",
+				"python.linting.enabled": true
+			},
+			
+			// Add the IDs of extensions you want installed when the container is created.
+			"extensions": [
+				"ms-python.python",
+				"ms-python.vscode-pylance",
+				"ms-python.vscode-python-envs",
+				"jurooravec.python-inline-source-2"
```
JuroOravec, line 21: I also added the VSCode extension that handles highlighting of inlined JS/CSS/HTML.

JuroOravec:
cc @EmilStenstrom

NOTE: The [failing pipeline](https://github.com/django-components/django-components/actions/runs/16401011844/job/46340454605?pr=1301) will be fixed in https://github.com/django-components/django-components/pull/1300 so I skipped it here.

JuroOravec:
In the end installing Python 3.13 resulted in like 4-5 min of waiting before the remote container could be used. So I commented it out. See https://github.com/django-components/django-components/pull/1302

The long build time could be maybe fixed with [Github's prebuilds](https://docs.github.com/en/codespaces/prebuilding-your-codespaces/configuring-prebuilds), but I don't want to spend time on configuring that right now.

PR #1339, opened 2025-08-13
feat: render fragments without document strategy

JuroOravec:
This PR makes it possible to insert HTML fragments (AKA components rendered with "fragment" strategy) into DOM that doesn't have our dependency manager JS script yet.

This is achieved by adding a small JS script to all rendered fragments - this script detects if our dep manager is loaded. If it's not, then it additionally fetches and executes that script.

Closes https://github.com/django-components/django-components/issues/1323


src/django_components_js/src/manager.ts
```diff
@@ -258,11 +258,17 @@ export const createComponentsManager = () => {
...
     _loadComponentScripts(data);
-  });
+  };
+  
+  // Initialise the MutationObserver that watches for `<script>` tags with `data-djc` attribute
+  observeScriptTag(onDjcScript);
+
+  // Also consider any embedded scripts at the moment the file is loaded.
+  const existingScripts = document.querySelectorAll<HTMLScriptElement>('script[data-djc]');
+  existingScripts.forEach(onDjcScript);
```
JuroOravec, line 271: @PouriaMzn's solution was close to working. Last missing step was that when we load the dependency manager script, we also have to look for outstanding `<script>` tags with `data-djc` attributes - these describe what JS/CSS needs to be loaded.

tests/e2e/testserver/testserver/views.py
```diff
@@ -239,6 +247,39 @@ def fragment_base_htmx_view(request):
...
+                </button>
+
+                {% component_js_dependencies %}
+            </body>
+        </html>
+    """
+    template = Template(template_str)
+
+    frag = request.GET["frag"]
+    rendered = template.render(Context({"frag": frag, "DJC_DEPS_STRATEGY": "ignore"}))
```
JuroOravec, line 278: Used `DJC_DEPS_STRATEGY: ignore` so that this script does NOT have dependency manager inserted.

tests/e2e/testserver/testserver/views.py
```diff
@@ -239,6 +247,39 @@ def fragment_base_htmx_view(request):
     return HttpResponse(rendered)
 
 
+
+# HTML into which a fragment will be loaded using HTMX
+# This variant doesn't include the component manager script, so that we can test
+# that the fragment can be rendered even when the page wasn't rendered with the "document" strategy.
+def fragment_base_htmx_view__raw(request):
```
JuroOravec, line 253: View that the test visits

tests/test_dependency_rendering_e2e.py
```diff
@@ -539,6 +539,58 @@ async def test_fragment_htmx(self):
 
         await page.close()
 
+    # Fragment where the page wasn't rendered with the "document" strategy
+    @with_playwright
+    async def test_fragment_without_document(self):
```
JuroOravec, line 544: This tests that we can fetch and insert the fragments and they fetch their own JS/CSS

JuroOravec:
cc @EmilStenstrom

PR #1342, opened 2025-08-15
refactor: fix some links, fix restart loop on file change of mkdocs serve

JuroOravec:
After https://github.com/django-components/django-components/issues/1331, I wanted to try to enable link validation so we can get rid of issue like that altogether.

But before the link validation can be enabled, the existing issues need to be resolved.

And it turns out that this whole time I was writing the links in the docs wrong. I was using the relative path as relative to the final rendered HTML. But mkdocs actually wants me to define the paths as relative to the filesystem.

In other words, often I linked to items within the API reference page as `../api#bla.bla`.

But correctly it should be `./api.md#bla.bla`.

I tried to use LLM to help me fix this, but it doesn't get it right all the time. There are also some genuinely broken links. So it needs to be done manually.

After fixing ~4 files and ~150 links, I decided to park this for now, because there's still ~1200 links to fix, and I can imagine it could take up to 10 hrs. And there's more important things to address.

Screenshot 2025-08-15 at 08 16 56

As part of this PR I also fixed the mkdocs dev server. When the docs are being built, there is a script to generate the API refence pages from the source code. Long time ago I modified that script to save the page files using the regular Python API (e.g. `open()`), because I didn't properly understand the how the `mkdocs-gen-files` works.

But since then, whenever the dev server was started, and you modified a file, mkdocs server got into a restarting loop. Because the generated API pages were re-generated on restart, but this was then picked up as a new file change, triggering another restart.

So for the whole time, when I wanted to check for changes, I had to manually stop the server and start it again. This was bearable when writing big chunks of the docs. But when I was now trying to fix the links, I needed to restart the server often, and it got in the way.


docs/scripts/reference.py
```diff
@@ -358,8 +370,7 @@ def gen_reference_tagformatters():
         elif _is_tag_formatter_cls(obj):
             tag_formatter_classes[name] = obj
 
-    out_file.parent.mkdir(parents=True, exist_ok=True)
-    with out_file.open("w", encoding="utf-8") as f:
+    with mkdocs_gen_files.open(out_path, "w", encoding="utf-8") as f:
```
JuroOravec, line 373: The fix for the restarting loop was to use `mkdocs_gen_files.open()` instead of `open()` to file writes go through `mkdocs-gen-files`

mkdocs.yml
```diff
@@ -1,10 +1,11 @@
 ---
+# See https://www.mkdocs.org/user-guide/configuration
 site_name: Django-Components
 site_description: A way to create simple reusable template components in Django.
 site_url: https://django-components.github.io/django-components/
 
 repo_url: https://github.com/django-components/django-components
-repo_name: django-components/django-components
+repo_name: django-components
```
JuroOravec, line 8: Also updated the github project name in the docs site, so it shows only the name, not the org name.

Before:

Screenshot 2025-08-15 at 08 37 12


After:

Screenshot 2025-08-14 at 21 19 17

JuroOravec:
cc @EmilStenstrom

Issue #1325, opened 2025-08-07
Template recursion issue after 0.141.0


SupImDos:
After upgrading to [`0.141.0`](https://github.com/django-components/django-components/releases/tag/0.141.0), certain nesting of components `{% slot %}` and usage of the `{% include %}` / `{% extends %}` tags appears to cause some type of template finding recursion, resulting in a `TemplateDoesNotExist` exception:

```jinja
{% load component_tags %}
{% component "a-outer-component" %}
  {% component "b-inner-component" %}
    <!-- "c-include-template" extends another template -->
    {% include "c-include-template.html" %}
  {% endcomponent %}
{% endcomponent %}
```

results in:

```
TemplateDoesNotExist
An error occured while rendering components a-outer-component > b-inner-component > b-inner-component(slot:b):d-extended-template.html

[...]

Template-loader postmortem
Django tried loading these templates, in this order:
Using engine django:
    django.template.loaders.filesystem.Loader: /<blah>/templates/d-extended-template.html (Skipped to avoid recursion)
```

Full minimal reproducible example available here: https://github.com/SupImDos/django-components-mrp

JuroOravec:
Hey @SupImDos thanks for reporting this.

I've found the bug and found workaround.

TL;DR - We currently render the body of `{% component %}` / `{% endcomponent %}` twice - first time searching for `{% fill %}` tags, second time to actually render the content.

   Turns out that in that particular setup that you shared, Django didn't want to render the extended template the second time, because it has already seen it in the current context. So the fix was to reset the state after the first "search" render.

## Fix

A proper fix that would avoid the double execution, can be then done some other time (https://github.com/django-components/django-components/issues/1326).

But as a workaround, we can wrap the [line 1464 in slots.py](https://github.com/django-components/django-components/blob/0df269adfeb2b0731e24d17f97cc82b88106e99d/src/django_components/slots.py#L1464) with following:

```py
    # NOTE: When we search for `{% fill %}` tags, we also evaluate `{% include %}` and `{% extends %}`
    #       tags if they are within component body (between `{% component %}` / `{% endcomponent %}` tags).
    #       But by doing so, we trigger Django's block/extends logic to remember that this extended file
    #       was already walked.
    #       (See https://github.com/django/django/blob/0bff53b4138d8c6009e9040dbb8916a1271a68d7/django/template/loader_tags.py#L114)
    #
    #       We need to clear that state, otherwise Djagno won't render the extended template the second time
[...]

PR #1344, opened 2025-08-15
fix: fix compat with extends and includes

JuroOravec:
Fix for https://github.com/django-components/django-components/issues/1325

### DIFF
--- a/src/django_components/slots.py
+++ b/src/django_components/slots.py
@@ -1,11 +1,13 @@
 import difflib
 import re
+from contextlib import contextmanager
 from dataclasses import dataclass, field
 from dataclasses import replace as dataclass_replace
 from typing import (
     TYPE_CHECKING,
     Any,
     Dict,
+    Generator,
     Generic,
     List,
     Literal,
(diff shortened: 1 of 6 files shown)


src/django_components/slots.py
```diff
@@ -1667,3 +1671,24 @@ def render_func(ctx: SlotContext) -> SlotResult:
...
+# When we search for `{% fill %}` tags, we also evaluate `{% include %}` and `{% extends %}`
+# tags if they are within component body (between `{% component %}` / `{% endcomponent %}` tags).
+# But by doing so, we trigger Django's block/extends logic to remember that this extended file
+# was already walked.
+# (See https://github.com/django/django/blob/0bff53b4138d8c6009e9040dbb8916a1271a68d7/django/template/loader_tags.py#L114)  # noqa: E501
+#
+# We need to clear that state, otherwise Django won't render the extended template the second time
+# (when we actually render it).
+@contextmanager
+def _extends_context_reset(context: Context) -> Generator[None, None, None]:
```
JuroOravec, line 1688: This is the actual fix - not much to it. I extracted it into its own function so it doens't pollute.

tests/test_templatetags_extends.py
```diff
@@ -1029,3 +1033,43 @@ def get_template_name(self, context):
             </html>
         """
         assertHTMLEqual(rendered, expected)
+
+    # Fix for compatibility with Django's `{% include %}` and `{% extends %}` tags.
+    # See https://github.com/django-components/django-components/issues/1325
+    @djc_test(parametrize=PARAMETRIZE_CONTEXT_BEHAVIOR)
+    def test_nested_component_with_include_and_extends_in_slot(self, components_settings):
```
JuroOravec, line 1040: This tests the same as the example provided in #1325

JuroOravec:
CC @EmilStenstrom

PR #1346, opened 2025-08-18
refactor: replace isort, black and flake8 with ruff

JuroOravec:
Closes https://github.com/django-components/django-components/issues/1299

This replaces flake8, isort, and black with  Ruff:
- `ruff check` command replaces both flake8 and isort
- `ruff format` command replaces black

Ruff also comes with tons of additional rules, including flake8-bugbear, as mentioned in https://github.com/django-components/django-components/issues/1299.

I went through all the files, fixed linter errors, and disabled those Ruff rules that were either not relevant to this project, or which would be hard to implement.

Main changes:
- Updated [pyproject.toml](https://github.com/django-components/django-components/pull/1346/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711) with Ruff
   - [Here](https://github.com/django-components/django-components/pull/1346/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R87) you can see all the disabled rules
   - And [here's a list of rules](https://github.com/django-components/django-components/pull/1346/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R134) that might be good to enable, but which could need some more refactoring
   - Also, [here](https://github.com/django-components/django-components/pull/1346/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R150) I made it such that the code in `src` is more strict, whereas tests and example project can be more lenient
- Updated [pre-commit-config.yml](https://github.com/django-components/django-components/pull/1346/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9)
- Updated docs on [development](https://github.com/django-components/django-components/pull/1346/files#diff-daf94d9ac125e821ddfd6f7341cf6b3c79f5e39d781ddcc32a1c5dcd356419bd)
- Updated [requirements.txt](https://github.com/django-components/django-components/pull/1346/files#diff-5de99c387a95507082f5225072501f952e3dc9a6c839192d0ccd04985005596c) to omit black, flake8, isort
- Updated [tox.ini](https://github.com/django-components/django-components/pull/1346/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449), removing the isort, black, and flake8 environments, and keeping only ruff.

Other than these above, and the comments below, the rest of the code changes is just linting and formatting.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -37,7 +37,7 @@ jobs:
           private-key: ${{ secrets.RELEASE_BOT_APP_PRIVATE_KEY }}
 
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           token: ${{ steps.app-token.outputs.token }}
           fetch-depth: 0
@@ -150,7 +150,7 @@ jobs:
       # Change git authentication to Github Actions, so the rest of the
       # workflow will have lower privileges.
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           fetch-depth: 0
 
--- a/.github/workflows/pr-benchmark-generate.yml
+++ b/.github/workflows/pr-benchmark-generate.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         fetch-depth: 0  # Need full history for ASV
 
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -15,7 +15,7 @@ jobs:
     if: github.repository == 'django-components/django-components'
     steps:
       - name: Checkout the repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Setup python
         uses: actions/setup-python@v5
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -27,7 +27,7 @@ jobs:
       run: |
         git config --global core.longpaths true
 
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
 
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v5
@@ -51,7 +51,7 @@ jobs:
         python-version: ['3.13']
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
 
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v5
@@ -76,7 +76,7 @@ jobs:
         python-version: ['3.13']
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
 
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v5
--- a/docs/overrides/main.html
+++ b/docs/overrides/main.html
@@ -1,0 +1,12 @@
+{#
+  Theme override: inject Google site verification meta tag.
+  Uses documented extrahead block from Material for MkDocs.
+  If token changes, update in mkdocs.yml under extra.google_site_verification.
+#}
+{% extends "base.html" %}
+{% block extrahead %}
+  {{ super() }}
+  {% if config.extra.google_site_verification %}
+    <meta name="google-site-verification" content="{{ config.extra.google_site_verification }}" />
+  {% endif %}
+{% endblock %}
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -80,6 +80,7 @@ extra:
       link: https://github.com/django-components/django-components
     - icon: fontawesome/brands/python
       link: https://pypi.org/project/django-components/
+  google_site_verification: vQA3d50F2ByQxG0eB6b0YoPnYW9gZo8xnd6HKhCyuys
 
 markdown_extensions:
   abbr: {}
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -54,7 +54,7 @@ cssselect2==0.8.0
     # via cairosvg
 defusedxml==0.7.1
     # via cairosvg
-django==4.2.23
+django==4.2.24
     # via hatch.envs.docs
 djc-core-html-parser==1.0.2
     # via hatch.envs.docs
@@ -134,9 +134,9 @@ mkdocs-git-authors-plugin==0.10.0
     # via hatch.envs.docs
 mkdocs-git-revision-date-localized-plugin==1.4.7
     # via hatch.envs.docs
-mkdocs-include-markdown-plugin==7.1.6
+mkdocs-include-markdown-plugin==7.1.7
     # via hatch.envs.docs
-mkdocs-material==9.6.17
+mkdocs-material==9.6.19
     # via hatch.envs.docs
 mkdocs-material-extensions==1.3.1
     # via mkdocs-material
@@ -197,9 +197,9 @@ pyyaml-env-tag==1.1
     # via
     #   mike
     #   mkdocs
-regex==2025.7.34
+regex==2025.9.1
     # via mkdocs-material
-requests==2.32.4
+requests==2.32.5
     # via mkdocs-material
 six==1.17.0
     # via python-dateutil


.github/copilot-instructions.md
```diff
@@ -8,7 +8,7 @@ Always reference these instructions first and fallback to search or bash command
 
 ### Initial Setup
 - Install development dependencies:
-  - `pip install -r requirements-dev.txt` -- installs all dev dependencies including pytest, black, flake8, etc.
+  - `pip install -r requirements-dev.txt` -- installs all dev dependencies including pytest, ruff, etc.
```
JuroOravec, line 11: Updated the instructions to mention only Ruff

docs/scripts/gen_release_notes.py
```diff
@@ -20,7 +20,7 @@ def generate_release_notes():
     # Create the output directory if it doesn't exist
     (ROOT / "docs" / releases_dir).mkdir(parents=True, exist_ok=True)
 
-    with open(changelog_path, "r", encoding="utf-8") as f:
+    with changelog_path.open("r", encoding="utf-8") as f:
```
JuroOravec, line 23: Some rules that Ruff introduces were e.g. to prefer to use `Path` over `os.path`, or in this case `Path.open` instead of `open()`

scripts/supported_versions.py
```diff
@@ -8,21 +9,21 @@
 VersionMapping = Dict[Version, List[Version]]
 
 
-def cut_by_content(content: str, cut_from: str, cut_to: str):
+def cut_by_content(content: str, cut_from: str, cut_to: str) -> str:
```
JuroOravec, line 12: Previously, the linter wasn't running against files outside of `src` I think. So now the linting is applied also to the root Python scripts.

scripts/supported_versions.py
```diff
@@ -1,3 +1,4 @@
+# ruff: noqa: T201, S310
```
JuroOravec, line 1: Here's an example of how to disable rules per file.

scripts/validate_links.py
```diff
@@ -112,9 +115,35 @@
 )
 
 
+@dataclass
+class Link:
```
JuroOravec, line 119: This is a script for checking or modifing URLs across the codebase. You can run it with

```sh
python scripts/validate_links.py
```

In this file I did more than only fixed links - I also refactored the script, adding these data classes, so it's clear what function returns what, and so that I could annotate the functions to address the linter errors.

scripts/validate_links.py
```diff
@@ -236,37 +276,83 @@ def validate_urls(all_urls):
...
+        print(f"[done {done_count + 1}/{total_urls}] {method:<4} {url} ...", end=" ")
         try:
-            resp = requests.get(
-                url, timeout=REQUEST_TIMEOUT, headers={"User-Agent": "django-components-link-checker/0.1"}
-            )
-            url_cache[url] = resp
+            # If there is at least one URL that specifies a fragment in the URL,
+            # we will fetch the full HTML with GET.
+            # But if there isn't any, we can simply send HEAD request instead.
+            if method == "GET":
```
JuroOravec, line 295: This `validate_links` script also checks for correctness of URL fragments (e.g. `#my-fragment` in `example.com/page#my-fragment`). I updated the logic so that, if the URL does NOT contain a fragment, we send only a HEAD request.

src/django_components/__init__.py
```diff
@@ -85,12 +85,8 @@
 
 
 __all__ = [
-    "all_components",
```
JuroOravec, line 88: No changes to the public API, the exports were only re-ordered

src/django_components/attributes.py
```diff
@@ -287,7 +287,7 @@ def normalize_class(value: ClassValue) -> str:
         # `{"class": True, "extra": True}` will result in `class="class extra"`
         res = value
     else:
-        raise ValueError(f"Invalid class value: {value}")
+        raise TypeError(f"Invalid class value: {value}")
```
JuroOravec, line 290: There's a rule that suggest to use `TypeError` when we checked for input *type* (e.g. `isinstance(value, MyCls)`. So error types of some functions changed from `ValueError` to `TypeError`.

src/django_components/commands/startcomponent.py
```diff
@@ -3,9 +3,7 @@
 
 # TODO_REMOVE_IN_V1 - Superseded by `components create`
 class StartComponentCommand(CreateCommand):
-    """
-    **Deprecated**. Use [`components create`](../commands#components-create) instead.
-    """
+    """**Deprecated**. Use [`components create`](../commands#components-create) instead."""
```
JuroOravec, line 6: Linter asks single-line docstrings to be surrounded by `"""`.

There was actually a lot of rules for dostrings that I had to disable, because it didn't fit how we've handled docstrings so far. E.g.
- Ruff asked to put docstrings to ALL methods, classes, and files - I disabled these, as I don't see the point in adding a docstring to e.g. internal methods. I look at docstrings as a public contract - For internals, we can use comments instead.
- Ruff asked to start all docstrings with imperatives, e.g. active verb like "Process this", "Do that". But it doesn't fit in some cases like the one above, which starts with a deprecation warning.

src/django_components/component.py
```diff
@@ -85,14 +84,17 @@
...
+from django_components.component_registry import AlreadyRegistered as AlreadyRegistered  # noqa: PLC0414
+from django_components.component_registry import ComponentRegistry as ComponentRegistry  # noqa: PLC0414,F811
+from django_components.component_registry import NotRegistered as NotRegistered  # noqa: PLC0414
+from django_components.component_registry import register as register  # noqa: PLC0414
+from django_components.component_registry import registry as registry  # noqa: PLC0414
 
 # isort: on
 
+if TYPE_CHECKING:
+    from django.views import View
```
JuroOravec, line 96: One nice thing is that some Ruff rules can identify when an import is used only for typing, and tells me to move that import inside `TYPE_CHECKING`.

There were some issue with circular imports before. And Ruff helping us identify which imports are for types only could help us resolve that.

src/django_components/component_registry.py
```diff
@@ -38,8 +38,6 @@ class AlreadyRegistered(Exception):
     [ComponentRegistry](./api.md#django_components.ComponentRegistry).
     """
 
-    pass
```
JuroOravec, line 41: Also turns out that when you have an empty class:

```py
class MyClass:
    pass
```

And you add docstrings to it

```py
class MyClass:
    """hello"""
    pass
```

Then you can remove the `pass` keyword:

```py
class MyClass:
    """hello"""
```

src/django_components/util/misc.py
```diff
@@ -221,8 +222,5 @@ def format_as_ascii_table(data: List[Dict[str, Any]], headers: List[str], includ
         data_rows.append(data_row)
 
     # Combine all parts into the final table
-    if include_headers:
-        table = "\n".join([header_row, separator] + data_rows)
-    else:
-        table = "\n".join(data_rows)
+    table = "\n".join([header_row, separator, *data_rows]) if include_headers else "\n".join(data_rows)
```
JuroOravec, line 225: One of the rules that Ruff had was to prefer to use ternary operator `X if Y else Z` instead of multi-line `if/else`.

I used it at first, but then disabled the rule, because I don't think it always leads to more clarity. Especially when there are additional comments for each branch. But so you can find in this PR some places that I refactored before I turned it off, like here.

tests/e2e/testserver/testserver/views.py
```diff
@@ -1,11 +1,14 @@
...
 from django.template import Context, Template
 from testserver.components import FragComp, FragMedia
 
-from django_components import types
+if TYPE_CHECKING:
+    from django_components import types
 
 
-def single_component_view(request):
+def single_component_view(_request):
```
JuroOravec, line 11: One of the rules was for unused function arguments - Hence you can see a lot of changes like this, prefixing the unused arguments with `_`.

TBH I'm lukewarm about this - problem with Python is that it relies on the argument name when assigning kwargs. So, in this case, we can't know for sure if some library called this view handler as:

```py
single_component_view(request=request)
```

In which case we'd be cooked (it'd raise error).

But on the other hand, getting a warning when an argument is truly not used in one of our functions, I think that's still worth enabling this rule.

tests/test_finders.py
```diff
@@ -120,13 +122,13 @@ def test_python_and_html_omitted(self):
         collected = do_collect()
 
         # Check that our staticfiles_finder finds the files and OMITS .py and .html files
-        self.assertIn(Path("staticfiles/staticfiles.css"), collected["modified"])
```
JuroOravec, line 123: There was also a rule to use plain `assert` instead of alternatives like `self.assertIn`

tests/test_templatetags_component.py
```diff
@@ -837,20 +837,23 @@ def test_text_outside_fill_tag_is_error_when_fill_tags(self, components_settings
...
+        template_str: types.django_html = """
+            {% load component_tags %}
+            {% component "test" %}
+            {% fill "header" %}{% endfill %}
+        """
         with pytest.raises(
             TemplateSyntaxError,
             match=re.escape("Unclosed tag on line 3: 'component'"),
         ):
-            template_str: types.django_html = """
```
JuroOravec, line 851: One rule that I liked was that, when there was multiple statements inside `with pytest.raises()` block, it asked me to move code out so that there's only one statement.

PR #1356, opened 2025-08-23
Add Google site verification meta tag support in MkDocs

EmilStenstrom:
Introduce support for Google site verification by adding a meta tag in the MkDocs configuration. Update the `mkdocs.yml` to include the verification token.

Relevant issue: https://github.com/django-components/django-components/issues/1355


EmilStenstrom:
Verified works locally: 
image

PR #1370, opened 2025-09-09
fix djc_test is not happy with ComponentsSettings

### LINKED ISSUE #1369: `djc_test` is not happy with `ComponentsSettings`
When you use the recommended `ComponentsSettings`, and you write tests using `@djc_test` decorator, you run into this problem:

```text
TypeError: 'ComponentsSettings' object is not a mapping
```

Lets enable the decorator to use the `ComponentsSettings`

PavelPancocha:
Fix #1369


PavelPancocha:
Sorry, I missed one typing :)

JuroOravec:
Looking good, thanks! I've fixed some unrelated issues with ruff that were blocking this PR.

PR #1374, opened 2025-09-12
fix: fix link to Adding slot page in getting started doc.

Antoliny0919:
It seems that the problem is caused by a line break.

Screenshot 2025-09-12 at 9 24 02 AM


JuroOravec:
Thanks! :)

PR #1386, opened 2025-09-18
docs: add discord link to documentation and README.

Antoliny0919:
I added the Discord link ...

Screenshot 2025-09-18 at 11 11 28 AM

Screenshot 2025-09-18 at 11 11 47 AM

Screenshot 2025-09-18 at 11 12 27 AM


Antoliny0919:
Among the suggestions @JuroOravec provided, I couldn’t find a way to add the discord link to the header navigation, to the right of the github widget.
Do you know of any opensource projects using MkDocs that have a discord link next to the github widget?
If so, I’d appreciate it as a reference.
However, it seems that usually the discord link is only added to the footer.

JuroOravec:
@Antoliny0919 Thanks for looking into this! You're right, there's no default way to add a link there.

In theory it would be possible to insert the Discord link there by overriding and modifying the [`header.html` file](https://github.com/squidfunk/mkdocs-material/blob/97e275c953e5b8255ca24ca166ddd802e23d84b1/src/templates/partials/header.html#L110). (E.g. similarly to how I've overriden the top nav [here with `nav-item.html`](https://github.com/django-components/django-components/blob/master/docs/overrides/partials/nav-item.html)). But the `partials/header.html` is quite big, so IMO no need to override it just to add a Discord link.

JuroOravec:
Note: I've update the link URL, because the one I gave you might expire in 7 days. The new one shouldn't expire

Screenshot 2025-09-18 at 11 24 45

PR #1387, opened 2025-09-18
fix: add required dependency package django-components to requirements-doc.

Antoliny0919:
Since this file uses `django_components`, it seems necessary to add `django_components` to `requirements-docs`.

Currently, if you create a virtual environment and install the dependencies with 
```pip install -r requirements-docs``` and running ```mkdocs serve``` will fail.
(This can be resolved by installing `django_components`.)

### DIFF
--- a/requirements-docs.in
+++ b/requirements-docs.in
@@ -16,5 +16,6 @@ mkdocstrings-python
 pymdown-extensions
 pygments
 pygments-djc
+django-components
 django>=4.2
 djc-core-html-parser>=1.0
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -41,9 +41,15 @@ cssselect2==0.8.0
 defusedxml==0.7.1
     # via cairosvg
 django==5.2.5
+    # via
+    #   -r requirements-docs.in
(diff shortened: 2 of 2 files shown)


Antoliny0919:
Oh, is the reason it wasn’t added to requirements because django-components is meant to be installed directly in editable mode?
If so, it would be great if a guideline were provided for this.
(Guidelines are provided for the development environment, but not for the documentation environment.
It would be helpful to provide this and include an explanation for it.)

JuroOravec:
> Oh, is the reason it wasn’t added to requirements because django-components is meant to be installed directly in editable mode?
> If so, it would be great if a guideline were provided for this.

@Antoliny0919 Yup that's correct. And thanks for raising this, I've added a section on how to run the documentaiton website, see https://github.com/django-components/django-components/pull/1391.

With that I think we can close this PR.

Antoliny0919:
> @Antoliny0919 Yup that's correct. And thanks for raising this, I've added a section on how to run the documentaiton website, see #1391.
> 
> With that I think we can close this PR.

Nice!! Thank you.

PR #1393, opened 2025-09-20
docs: add first sponsor.🙌

Antoliny0919:
We would like to thank our first sponsor, ohne-makler.

### README

Screenshot 2025-09-20 at 11 30 04 AM

### Docs

Screenshot 2025-09-20 at 11 30 24 AM


README.md
```diff
@@ -11,6 +11,15 @@ in modern frontend frameworks like Vue or React.
 
 With `django-components` you can support Django projects small and large without leaving the Django ecosystem.
 
+## Sponsors
+
+<p align="center"><strong>Platinum Sponsors</strong></p>
```
Antoliny0919: Most sponsors have [tiers based on the amount](https://github.com/sponsors/tiangolo) of their contribution.
However, since I’m not exactly sure where we stand, I’ve set it to the highest tier, Platinum, for now.
That said, I wonder if we really need to have tiers at all, considering we don’t have many sponsors yet.

Antoliny0919 (commented):
In addition to the parts added so far, if a django-components People page is created, it would be nice to add a sponsor section there, similar to [FastAPI](https://fastapi.tiangolo.com/fastapi-people/#sponsors).

Personally, I think the [Insiders tab in mkdocs material](https://squidfunk.github.io/mkdocs-material/insiders/) is the best example of sponsor related documentation. However, I’m not sure we need that level of documentation just yet. I feel it’s something we should consider later, once we become independent from Django and start targeting Python.

Antoliny0919:
For now, I think it would be fine to just add a note of thanks in the sponsors section.
like ...

"We would like to thank everyone who sponsors django-components."
"Your support gives the team time to make django-components even better."

README.md
EmilStenstrom: I think we should go without any tiers. Better to keep things simple when we don't yet have a plan :)

EmilStenstrom (approved):
Would remove the second level sponsor heading with platinum in it, otherwise go ahead!

README.md
```diff
@@ -11,6 +11,15 @@ in modern frontend frameworks like Vue or React.
...
+<p align="center">
+  <a href="https://www.ohne-makler.net/" target="_blank"><img
```
JuroOravec: Add the tagline on hover-over + add identifier so they can potentially track traffic from this project.

```suggestion
  <a href="https://www.ohne-makler.net/?ref=django-components" target="_blank" title="Ohne-makler: Sell ​​and rent real estate without an agent"><img
```

docs/overview/welcome.md
```diff
@@ -7,6 +7,15 @@ in modern frontend frameworks like Vue or React.
 
 With `django-components` you can support Django projects small and large without leaving the Django ecosystem.
 
+## Sponsors
+
+<p align="center"><strong>Platinum Sponsors</strong></p>
+<p align="center">
+  <a href="https://www.ohne-makler.net/" target="_blank"><img
```
JuroOravec: Add the tagline on hover-over + add identifier so they can potentially track traffic from this project.

```suggestion
  <a href="https://www.ohne-makler.net/?ref=django-components" target="_blank" title="Ohne-makler: Sell ​​and rent real estate without an agent"><img
```

README.md
```diff
@@ -11,6 +11,15 @@ in modern frontend frameworks like Vue or React.
...
+  src="https://raw.githubusercontent.com/django-components/django-components/master/.github/assets/sponsors/sponsor-ohne-makler.png" height="120"
```
JuroOravec, line 19: Since the image has transparent background, it has black background on dark mode:

Screenshot 2025-09-20 at 12 16 26

Could you replace it with this image?

sponsor-ohne-makler-white-bg

JuroOravec: Also, there's already the top-level `logo` directory. I think we could clean up a bit and put all the images under a single top-level `assets` directory.

```
|- assets
    |- logo
        |- logo-black-on-transparent.svg
        ...
    |- sponsors
        |- sponsor-ohne-makler.png
        ...
```

README.md
JuroOravec: Maybe we could create an issue "Sponsorship excellence" (inspired by the [Documentation excellence](https://github.com/django-components/django-components/issues/804) issue) and cross-post this there. Because the tiers are a good idea for the future, but yeah for now we can simply mention the single sponsor.

In the Sponsorship excellence ticket we can also mention what you've mentioned above about the Insiders programme that Material for Mkdocs have. Because that's a very good idea too. But that one is definitely somehting that can wait for well after v1.

README.md
Antoliny0919: > Also, there's already the top-level `logo` directory. I think we could clean up a bit and put all the images under a single top-level `assets` directory.
> 
> ```
> |- assets
>     |- logo
>         |- logo-black-on-transparent.svg
>         ...
>     |- sponsors
>         |- sponsor-ohne-makler.png
>         ...
> ```

Yes, file cleanup is definitely needed.
I think it would be good to organize it using the structure you suggested.
But wouldn’t it be better to separate this task into its own?

Antoliny0919:
Thank you for the quick review, Emil and Juro :)

README.md
Antoliny0919: Yes, I agree. Since it’s not urgent right now, I think it would be better to proceed with it later, once django-components gains more support.
I’ve created an [issue](https://github.com/django-components/django-components/issues/1394)!

README.md
JuroOravec: Either as part of this PR or separate task, up to you :)

Yes, making a separate task would be best practice, but practically speaking, since not many people are working on codebase at the same time (just you and I, maybe someone else) and it's a tiny change and also this PR is tiny, then I think either way is fine.

Issue #1397, opened 2025-09-22
Add basedpyright


JuroOravec:
As part of updating the project tooling, I explored using pyright instead of mypy as a type checker.

To be exact, I used [basedpyright](https://docs.basedpyright.com/latest/) instead of [pyright](https://github.com/microsoft/pyright). basedpyright project says that unlike Microsoft's pyright, basedpyright is not locked to the VSCode ecosystem. I'm using VSCode, so this doesn't affect me. But since we're a Python project, and people may be using PyCharm, then I think it makes sense to use basedpyright.

Pyright / basedpyright have more nuanced type checker. For example, it can catch the error when a for-loop variable is referenced inside a function that was created in that for loop. In the past year I got bit by that error twice. Mypy doesn't pick on that.

Even when switching from Mypy to Pyright, as an open source project we should test our code against both type checkers. So that our users can choose either one and the type checking will still work for them.

I gave it a go to make django-components to work with basedpyright. But frankly, I would leave it for when django-components is mature, and once we minimize the area of interacting with Django code (DJC v2 or v3).

Because right now django-components interacts a lot with Django code, which is not typed, and it seems that the [typing stubs](https://github.com/typeddjango/django-stubs) are also not always correct/present? Moreover, to build our library on top of Django, we sometimes add custom attributes, like `_djc_patched` when patching Django classes, and pyright doesn't like that. So adding pyright to DJC resulted in a lot of type casting, and a lot of `# type: ignore` comments, making the development feel more sluggish.

[Here's a diff](https://github.com/JuroOravec/django-components/compare/7fad7f9...JuroOravec:django-components:jo-add-basedpyright) of when I tried to implement it.
- NOTE: The linked diff is made against a specific commit, so the link should remain valid even in the future. The changes needed for actual implementation might change in the future, but the linked diff should serve as a good example of what changes are needed.

JuroOravec:
(Opus 4.8)

## Findings from a fresh look (parking this for now)

Revisited this while debugging why a contributor's IDE wasn't resolving `from django.template import ...`. Root cause there was just interpreter/venv selection, but it led to mapping out what a basedpyright gate would actually require. Recording the findings so we don't re-derive them next time. **Status unchanged: still deferred** (the Django-interop `# type: ignore` tax described in the original issue still applies). Note also that mypy remains the source of truth, and basedpyright/pyright are not run in CI/tox today, so anything below is purely additive editor + gate tooling.

[...]

PR #1403, opened 2025-09-23
refactor: allow to use pylance/pyright with the codebase

JuroOravec:
This PR adds settings to disable a lot of pyright warnings and errors, so one can install the (based)pyright extension and be able to write code without having to be floded by errors.

---

After trying out basedpyright with django-components on my local computer (https://github.com/django-components/django-components/issues/1397), I somehow messed up my VSCode setup, and I wasn't able to revert back to my previous setup.
- Note: My original setup was using Mypy for type checking. While type checking worked, the Python syntax highlighting broke, and I couldn't figure out why.

That's when I noticed that Cursor is pushing for people to install their fork of basedpyright. So it seems that, at least in Cursor, (based)pyright is the default type checker for Python. Don't know what's the situation for vanilla VSCode.

As I mentioned in https://github.com/django-components/django-components/issues/1397, there's a lot of warnings and errors when using pyright with DJC codebase, because we interact with Django's API a lot, sometimes adding extra properties, etc, and pyright can't recognize these. We also still support Python 3.8, including type annotations like `List[]`, etc, which pyright marks as deprecated.

Note: This does NOT add pyright type checking to CI. As mentioned in #1397, I think that can wait.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -32,7 +32,7 @@ django==4.2.24
     # via
     #   -r requirements-dev.in
     #   django-template-partials
-django-template-partials==25.1
+django-template-partials==25.2
     # via -r requirements-dev.in
 djc-core-html-parser==1.0.2
     # via -r requirements-dev.in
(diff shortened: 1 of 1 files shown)


JuroOravec:
cc @EmilStenstrom

PR #1405, opened 2025-09-23
refactor: move logos and sponsors to assets dir

JuroOravec:
Follow up from https://github.com/django-components/django-components/pull/1393#discussion_r2365576389


JuroOravec:
cc @Antoliny0919

PR #1414, opened 2025-09-29
refactor: allow to call Component.inject() outside of render

JuroOravec:
There was an error I encountered in https://github.com/django-components/django-components/pull/1411 (see [broken pipeline](https://github.com/django-components/django-components/actions/runs/18004778853/job/51222326796?pr=1411)).

It happens when provided data were accessed from within a component that's inside a forloop:

```py
class MyComponent(Component):
       data = self.inject("my_provide")
       return {"data": data}
```

```django
{% load component_tags %}
{% provide "my_provide" key="hi" data=data %}
   {% for i in range(10) %}
       {% component "my_component" / %}
   {% endfor %}
{% endprovide %}
```

To fix the error, I changed how the provided data is managed.

- Previously, it was done entirely within the Django's Context object. We counted refences to the provided data, and removed it when the reference dropped to 0, or when the `{% provide %}` block ended.

   This was incorrect, because the reference count could drop to 0 even between components of the same depth:

   ```django
	{% load component_tags %}
	{% provide "my_provide" key="hi" data=data %}
     {# <-- 0 comps refer to "my_provide" #}
	   {% for i in range(10) %}
           {# <-- 0 comps refer to "my_provide" #}
	       {% component "my_component" / %}  {# <-- 1 comp refers to "my_provide" #}
           {# <-- 0 comps refer to "my_provide" #}
	   {% endfor %}
	{% endprovide %}
   ```

- So now, instead, the reference counting is done differently:
   - When we reach `{% endprovide %}`, we remove the provided data, UNLESS there's a component instance pointing to it.
   - We use [finalizers](https://docs.python.org/3/library/weakref.html#weakref.finalize) to detect when Component instances are garbage collected. And when they are, only then we check whether we can release the provided data or not.

The side effect of this new approach is that `Component.inject()` will now work also outside of rendering flow

 ```py
 comp = None

 class MyComponent(Component):
         nonlocal comp
         comp = self

 template_str = """
     {% load component_tags %}
     {% provide "my_provide" key="hi" data=data %}
         {% component "my_component" / %}
     {% endprovide %}
 """
 template = Template(template_str)
 rendered = template.render(Context({}))

 assert comp is not None

 injected = comp.inject("my_provide")
 assert injected.key == "hi"
 assert injected.data == "data"
 ```

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -19,7 +19,7 @@ from typing import (
     Union,
     cast,
 )
-from weakref import ReferenceType, WeakValueDictionary, finalize
+from weakref import ReferenceType, WeakValueDictionary, finalize, ref
 
 from django.forms.widgets import Media as MediaCls
 from django.http import HttpRequest, HttpResponse
@@ -61,9 +61,10 @@ from django_components.perfutil.component import (
     ComponentRenderer,
     OnComponentRenderedResult,
     component_context_cache,
+    component_instance_cache,
     component_post_render,
 )
-from django_components.perfutil.provide import register_provide_reference, unregister_provide_reference
+from django_components.perfutil.provide import register_provide_reference, unlink_component_from_provide_on_gc
 from django_components.provide import get_injected_context_var
 from django_components.slots import (
     Slot,
@@ -102,9 +103,11 @@ COMP_ONLY_FLAG = "only"
 if sys.version_info >= (3, 9):
     AllComponents = List[ReferenceType[Type["Component"]]]
     CompHashMapping = WeakValueDictionary[str, Type["Component"]]
+    ComponentRef = ReferenceType["Component"]
 else:
     AllComponents = List[ReferenceType]
     CompHashMapping = WeakValueDictionary
+    ComponentRef = ReferenceType
 
 
 OnRenderGenerator = Generator[
@@ -516,7 +519,7 @@ class ComponentMeta(ComponentMediaMeta):
 # Internal data that are made available within the component's template
 @dataclass
 class ComponentContext:
-    component: "Component"
+    component: ComponentRef
     component_path: List[str]
     template_name: Optional[str]
     default_slot: Optional[str]
@@ -525,6 +528,12 @@ class ComponentContext:
     # shares this dictionary for storing callbacks that are called from within `component_post_render`.
     # This is so that we can pass them all in when the root component is passed to `component_post_render`.
     post_render_callbacks: Dict[str, Callable[[Optional[str], Optional[Exception]], OnComponentRenderedResult]]
+
+
+def on_component_garbage_collected(component_id: str) -> None:
+    """Finalizer function to be called when a Component object is garbage collected."""
+    unlink_component_from_provide_on_gc(component_id)
+    component_context_cache.pop(component_id, None)
 
 
 class Component(metaclass=ComponentMeta):
@@ -2314,6 +2323,9 @@ class Component(metaclass=ComponentMeta):
         self.registry = default(registry, registry_)
         self.node = node
 
+        # Run finalizer when component is garbage collected
+        finalize(self, on_component_garbage_collected, self.id)
+
         extensions._init_component_instance(self)
 
     def __init_subclass__(cls, **kwargs: Any) -> None:
@@ -2940,7 +2952,7 @@ class Component(metaclass=ComponentMeta):
 
         As the `{{ message }}` is taken from the "my_provide" provider.
         """
-        return get_injected_context_var(self.name, self.context, key, default)
+        return get_injected_context_var(self.id, self.name, key, default)
 
     @classmethod
     def as_view(cls, **initkwargs: Any) -> ViewFn:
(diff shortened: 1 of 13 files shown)


src/django_components/component.py
```diff
@@ -3608,10 +3643,6 @@ def on_component_rendered(
                 error = new_error
                 html = None
 
-            # Remove component from caches
-            del component_context_cache[render_id]
-            unregister_provide_reference(render_id)
```
JuroOravec, line 3613: These are now removed when the `component` instance is garbage collected

src/django_components/extension.py
```diff
@@ -255,18 +256,25 @@ class Cache(CacheExtension.ComponentConfig):
...
+        This attribute holds the owner [`Component`](./api.md#django_components.Component) instance
+        that this extension is defined on.
+        """
+        component = self._component_ref()
+        if component is None:
+            raise RuntimeError("Component has been garbage collected")
+        return component
 
     def __init__(self, component: "Component") -> None:
-        self.component = component
```
JuroOravec, line 269: The methods inside extensions component code can access the current component via `self.component`:

```py
class MyComp(Component):
    class MyExt:
        def abc(self):
            return self.component.kwargs["some_arg"]
```

But this prevented Components from being garbage collected. So now extensions keep only a weak ref to the component. When `self.component` is accessed, we evaluate the weak ref to get the actual Component instance.

src/django_components/perfutil/component.py
```diff
@@ -30,6 +30,15 @@
...
 component_context_cache: Dict[str, "ComponentContext"] = {}
 
+# ComponentID -> Component instance mapping
+# This is used so that we can access the component instance from inside `on_component_rendered()`,
+# to call `Component.on_render_after()`.
+# These are strong references to ensure that the Component instance stays alive until after
+# `on_component_rendered()` has been called.
+# After that, we release the reference. If user does not keep a reference to the component,
+# it will be garbage collected.
+component_instance_cache: Dict[str, "Component"] = {}
```
JuroOravec, line 40: Other thing which was preventing component instance from being garbage collected is the `on_component_rendered` function defined inside `component.py` inside `Component._render_impl()` [here](https://github.com/django-components/django-components/blob/52a2d7a4b5458a1867355b174430fe7538bab4ef/src/django_components/component.py#L3622).

Currently the inner `on_component_rendered` function needs to access the component instance, but at a later point in time then the rest of `Component.render()`.

Previously we accessed the component simply by accessing it from the outer scope:

```py
def outer():
    my_var = 1
    def inner():
        print(my_var)

    return inner
```

But the problem with this approach is that Python's garbage collection then kept the strong reference inside `inner()` alive for longer than needed. My guess is that maybe because the `inner` function had different lifespan or was referenced in other places.

So to solve around this, the `on_component_rendered` function (AKA `inner`) now accesses the component instance via global dictionary.

And the entries in the global dictionary are cleaned up when the component instance is garbage collected:

```py
# Pseudocode
comps = {}

def on_garbage_collect(comp_id):
    comps.pop(comp_id,None)

class Component:
	def _render_impl():
	    comp_id = gen_id()
	    comps[comp_id] = self

        finalize(self, on_garbage_collect, comp_id)
	
	    def on_component_rendered():
	        comp = comps[comp_id]
	        print(comp)

	    return inner
```

src/django_components/perfutil/provide.py
```diff
@@ -77,77 +81,103 @@
...
-            provide_cache.pop(provide_id)
-            provide_references.pop(provide_id)
+    # Remove this component from all provide references it was subscribed to
+    for provide_id in component_provides_ids.values():
+        references_to_this_provide = provide_references.get(provide_id)
+        if references_to_this_provide:
+            references_to_this_provide.discard(component_id)
+
+
+def unlink_component_from_provide_on_gc(component_id: str) -> None:
```
JuroOravec, line 172: This is the logic that runs when a Component instance is being garbage collected

src/django_components/provide.py
```diff
@@ -102,24 +102,22 @@ def render(self, context: Context, name: str, **kwargs: Any) -> SafeString:
...
     """
     Retrieve a 'provided' field. The field MUST have been previously 'provided'
     by the component's ancestors using the `{% provide %}` template tag.
     """
-    # NOTE: For simplicity, we keep the provided values directly on the context.
-    # This plays nicely with Django's Context, which behaves like a stack, so "newer"
-    # values overshadow the "older" ones.
-    internal_key = _INJECT_CONTEXT_KEY_PREFIX + key
+    # NOTE: `component_provides` is defaultdict. Use `.get()` to avoid making an empty dictionary.
+    providers = component_provides.get(component_id)
```
JuroOravec, line 115: Previously the provided data was stored for the duration of the context.

But to make the provide/inject work even outside of rendering, the provided data is stored for the lifetime of the Component instance. Our Component instances may now outlive the Django's Context object.

As such, when we call `Component.inject()`, we now access the provided data using the component instance as the key.

src/django_components/slots.py
```diff
@@ -828,7 +832,12 @@ def render(self, context: Context, name: str, **kwargs: Any) -> SafeString:
             if parent_index is not None:
                 ctx_id_with_fills = context.dicts[parent_index][_COMPONENT_CONTEXT_KEY]
                 ctx_with_fills = component_context_cache[ctx_id_with_fills]
-                slot_fills = ctx_with_fills.component.raw_slots
+                parent_component = ctx_with_fills.component()
```
JuroOravec, line 835: Another part that was preventing Component instances from being garbage collected was when we render the slots. These parts now also have only weak references to their owner Component instance.

tests/templates/inject.html
```diff
@@ -1,5 +1,5 @@
 {% load component_tags %}
 <div>
-  {% component "injectee" %}
+  {% component "injectee17" %}
```
JuroOravec, line 3: As mentioned in https://github.com/django-components/django-components/issues/1413, there's now race conditions sometimes, as the finalizers for test A now run in parallel with the execution of test B, which sometimes causes problems.

As a simple workaround, I updated the tests where we rely on garbage collecion, so that each test defines a unique component name, so that the delayed finalizers aren't deleting incorrect entries.

I described the proper solution in https://github.com/django-components/django-components/issues/1413

tests/test_templatetags_provide.py
```diff
@@ -652,19 +664,143 @@ class ParentComponent(Component):
...
             """,
         )
-        self._assert_clear_cache()
+        _assert_clear_cache()
+
+    # TODO - Enable once globals and finalizers are scoped to a single DJC instance")
+    #        See https://github.com/django-components/django-components/issues/1413
+    @pytest.mark.skip("#TODO")
+    @djc_test(parametrize=PARAMETRIZE_CONTEXT_BEHAVIOR)
+    def test_provide_component_inside_forloop(self, components_settings):
```
JuroOravec, line 673: Here are tests for the interaction of `{% provide %}` and `{% for %}`

tests/test_templatetags_provide.py
```diff
@@ -758,40 +894,125 @@ def get_template_data(self, args, kwargs, slots, context):
...
+    @djc_test(
+        parametrize=(
+            ["components_settings"],
+            [
+                [{"context_behavior": "isolated"}],
+            ],
+            ["isolated"],
+        )
+    )
+    def test_inject_called_outside_rendering__persisted_ref(self, components_settings):
```
JuroOravec, line 922: And here are tests that verify that `Component.inject()` can be used even outside the render flow.

JuroOravec:
@EmilStenstrom Your thoughts on this one?

EmilStenstrom:
@JuroOravec To be frank, this is too much to wrap my head around. Feel free to merge!

PR #1419, opened 2025-09-30
ci: check supported versions once a week

JuroOravec:
Updated the Python script that instructs where and how to update the info on supported Python and Django versions.

The script now has 2 commands - "generate" (old behaviour) and "check" (new).

The "check" command is used from within a Github workflow - It parses the current state of the documentation, and compares it against the latest Python / Django versions. And if there is a change, it creates a new Github issue.

This workflow runs once a week on Sunday night.

Demo issue created by the bot can be seen here - https://github.com/django-components/django-components/issues/1418

### Check command when all OK:

Screenshot 2025-09-30 at 14 06 48

### Check command when mismatch but no GITHUB_TOKEN:

Screenshot 2025-09-30 at 14 02 51

### Check command when mismatch and GITHUB_TOKEN:

Screenshot 2025-09-30 at 14 26 37

### Check command when the issue already exists:

Screenshot 2025-09-30 at 14 36 11


.pre-commit-config.yaml
```diff
@@ -1,10 +1,10 @@
 repos:
```
JuroOravec, line 1: Updated pre-commit to match the checks we run in CI, as mentioned in https://github.com/django-components/django-components/issues/1395#issuecomment-3350413821.

JuroOravec:
cc @EmilStenstrom

EmilStenstrom (approved):
Looks good to me!

(why does every PR have to be 500+ lines?! :D)

PR #1420, opened 2025-09-30
ci: Discord integration

JuroOravec:
Adds integration with Discord:

- Send message to the `announcements` channel when a new release is published
- Send message to the `development` channel when a new issue, PR, or discussion is opened

~Blocked by https://github.com/django-components/django-components/pull/1419~


JuroOravec:
@EmilStenstrom These are the workflows to cross-post to Discord. AI generated, haven't tested yet, but I checked the event triggers and they seem correct.

PR #1425, opened 2025-10-02
refactor: fix multiple yield in on_render + prepare for scoped CSS

JuroOravec:
Follow up to https://github.com/django-components/django-components/pull/1421 as I realized some thing didn't work properly there. No public API change.


src/django_components/perfutil/component.py
```diff
@@ -521,40 +530,12 @@ def on_item(curr_item: Union[ErrorPart, TextPart, ComponentPart]) -> None:
...
             full_path = [*curr_item.full_path, curr_comp_name]
+            generators_by_component_id[component_id] = on_render_generator
 
             # This is where we actually render the component
-            #
-            # NOTE: [1:] because the root component will be yet again added to the error's
-            # `components` list in `_render_with_error_trace` so we remove the first element from the path.
-            try:
-                with component_error_message(full_path[1:]):
-                    comp_content, extra_child_component_attrs, on_render_generator = curr_comp_renderer(
```
JuroOravec, line 537: The main change in this PR is that:
- Previously, when rendering a component, there were 2 ways how the rendered template was handled.
    1. Here you can see `comp_content` - That was the result of calling `Component.on_render()`.
    2. Then there is `on_render_generator` - This is the remainder if  `Component.on_render()` contained `yield` keyword.

- Now, it's all normalized to a generator, whether `on_render()` contains `yield` or not.

The problem was that while the `comp_content` was post-processed by us, the follow-up HTML from the generator was NOT.

Example:

In this example, we have some component which renders fallback content on error.

But the problem was that, BEFORE, the error message HTML would not be tagged with `data-djc-id` HTML attribute.

AFTER this PR, the `data-djc-id` attribute is added in both cases.

```py
class TableWithFallback(Component):
    template = "..."

        if error:
            return '<p class="error-msg">Failed to render table</p>'
```

Will output on error:

```html
<p data-djc-id-c123456 class="error-msg">Failed to render table</p>
```

src/django_components/perfutil/component.py
```diff
@@ -443,17 +441,28 @@ def finalize_component(item_id: QueueItemId, error: Optional[Exception], full_pa
...
 
+                # Allow to optionally override/modify the intermediate result returned from `Component.on_render()`
+                # and by extensions' `on_component_intermediate` hooks.
+                on_component_intermediate = component_tree_context.on_component_intermediate_callbacks[
+                    item_id.component_id
+                ]
+                # NOTE: [1:] because the root component will be yet again added to the error's
+                # `components` list in `_render_with_error_trace` so we remove the first element from the path.
+                with component_error_message(full_path[1:]):
+                    new_html = on_component_intermediate(new_html)
```
JuroOravec, line 452: Another important thing about this PR is that this finally opens the doors for implementing the scoped CSS feature.

It does so by defining a place where the "intermediate" result can be modified.

The "intermediate" result is the component's rendered template which hasn't had its child components rendered yet, and has only placeholders. E.g.

```html
<div>
  <h2>...</h2>
  <template djc-render-id="a1b3cf"></template>
  <span>...</span>
  <template djc-render-id="f3d3cf"></template>
</div>
```

This is needed for the CSS scoping feature, because this intermediate results contains only the HTML defined directly by the component at hand. So we can simply walk over the elements in the HTML and add a special attribute to implement CSS scoping:

```html
<div data-css-scope-123>
  <h2 data-css-scope-123>...</h2>
  <template djc-render-id="a1b3cf"></template>
  <span data-css-scope-123>...</span>
  <template djc-render-id="f3d3cf"></template>
</div>
```

src/django_components/component.py
```diff
@@ -541,6 +542,27 @@ def __del__(cls) -> None:
         extensions.on_component_class_deleted(OnComponentClassDeletedContext(comp_cls))
 
 
+# Internal data that's shared across the entire component tree
+@dataclass
+class ComponentTreeContext:
```
JuroOravec, line 547: I also extracted out data that's shared across the entire component tree.

src/django_components/component.py
```diff
@@ -3645,26 +3669,70 @@ def _render_impl(
...
 
+        # This callback is called with the value that was yielded from `Component.on_render()`.
+        # It may be called multiple times for the same component, e.g. if `Component.on_render()`
+        # contains multiple `yield` keywords.
+        def on_component_intermediate(html_content: Optional[str]) -> Optional[str]:
+            # HTML attributes passed from parent to current component.
+            # NOTE: Is `None` for the root component.
+            curr_comp_attrs = component_tree_context.component_attrs.get(render_id, None)
+
+            if html_content:
```
JuroOravec, line 3720: Extensions like CSS scoping will be able to modify the "intermediate" component result here. In fact, the this `if html_content` branch could be actually extracted out as an internal extension too.

JuroOravec:
cc @EmilStenstrom

PR #1428, opened 2025-10-03
refactor: use lamba with yield in Component.on_render()

JuroOravec:
One more update on the `Component.on_render()` hook.

⚠️ This slightly modifies the API as we discussed. ⚠️ 

So in the past I suggested to `yield` inside `on_render()` as a way to access the fully rendered tree:

```py
class MyTable(Component):
```

But the problem is that `template.render(context)` can throw error - E.g. if `MyTable` incorrectly calls another Django template tag inside of its template.

So then users are forced to write a workaround like this:

```py
class MyTable(Component):
        try:
            intermediate = template.render(context)
            html, error = yield intermediate
        except Exception as e:
            html, error = None, e
```

Instead, we can allow people to yield a function instead of just a string. We then do the error handling for the users behind the scene. So users will simply need to write:

```py
class MyTable(Component):
        html, error = yield lambda: template.render(context)
```

### DIFF
--- a/src/django_components/perfutil/component.py
+++ b/src/django_components/perfutil/component.py
@@ -1,6 +1,6 @@
 import re
 from collections import deque
-from typing import TYPE_CHECKING, Callable, Deque, Dict, List, NamedTuple, Optional, Set, Tuple, Union
+from typing import TYPE_CHECKING, Callable, Deque, Dict, List, Literal, NamedTuple, Optional, Set, Tuple, Union
 
 from django.utils.safestring import mark_safe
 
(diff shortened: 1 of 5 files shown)


src/django_components/perfutil/component.py
```diff
@@ -590,22 +602,55 @@ def _call_generator(
...
+        #     intermediate = template.render(context)
+        # except Exception as err:
+        #     result = None
+        #     error = err
+        # else:
+        #     result, error = yield intermediate
+        # ```
+        if callable(new_result):
+            try:
+                new_result = new_result()
```
JuroOravec, line 638: This is where we evaluate the lambda function if we are given one.

JuroOravec:
cc @EmilStenstrom

PR #1433, opened 2025-10-03
feat: add ErrorFallback

JuroOravec:
Add ErrorFallback (originally mentioned in https://github.com/django-components/django-components/issues/1085) built-in component. Available from within the template under `error_fallback`.

This works similarly to React's ErrorBoundary - It wraps the rest of the template. If the thing that's inside throws error, ErrorFallback captures that, and renders the fallback content instead.

Either pass the fallback as a kwarg:

```django
{% component "error_fallback" fallback="Oops, something went wrong" %}
    {% component "table" / %}
{% endcomponent %}
```

Or use the full `fallback` slot:

```django
{% component "error_fallback" %}
    {% fill "content" %}
        {% component "table" / %}
    {% endfill %}
    {% fill "fallback" data="data" %}
        <p>Oops, something went wrong</p>
        {% button href="/report-error" %}
            Report error
        {% endbutton %}
    {% endfill %}
{% endcomponent %}
```

Screenshot 2025-10-03 at 17 11 15
Screenshot 2025-10-03 at 17 14 32

### DIFF
--- a/src/django_components/apps.py
+++ b/src/django_components/apps.py
@@ -19,6 +19,7 @@ class ComponentsConfig(AppConfig):
         from django_components.autodiscovery import autodiscover, import_libraries
         from django_components.component_registry import registry
         from django_components.components.dynamic import DynamicComponent
+        from django_components.components.error_fallback import ErrorFallback
         from django_components.extension import extensions
         from django_components.util.django_monkeypatch import (
             monkeypatch_include_node,
@@ -66,6 +67,7 @@ class ComponentsConfig(AppConfig):
 
         # Register the dynamic component under the name as given in settings
         registry.register(app_settings.DYNAMIC_COMPONENT_NAME, DynamicComponent)
+        registry.register("error_fallback", ErrorFallback)
 
         # Let extensions process any components which may have been created before the app was ready
         extensions._init_app()
--- a/src/django_components/components/error_fallback.py
+++ b/src/django_components/components/error_fallback.py
@@ -1,0 +1,156 @@
+from typing import NamedTuple, Optional, cast
+
+from django.template import Context, Template
+from django.template.exceptions import TemplateSyntaxError
+
+from django_components import Component, OnRenderGenerator, SlotInput, types
+
+
+class ErrorFallback(Component):
+    """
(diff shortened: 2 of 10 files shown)


src/django_components/components/error_fallback.py
```diff
@@ -0,0 +1,156 @@
...
+            return fallback_kwarg
+        elif fallback_slot is not None:
+            # Render the template second time, this time with the error
+            # So that we render the fallback slot with proper access to the outer context and whatnot.
+            with context.push({"error": error}):
+                return template.render(context)
+        else:
+            return ""
+
+    # TODO - Once we don't have to pass Context to the slot, we can remove the template
```
JuroOravec, line 148: Not related to the feature it self, but just highlighting this - Right now we can access the slots as functions from within component instance when it's being rendered.

But currently we have to be careful what context (Django's `Context`) we pass to the slot. Hence why I'm rendering the slots like below, reusing the same template, and only triggering the different branch. Because the `{% slot %}` template tags know how to render the slots correctly.

I mentioned it before somewhere else, but an improvement for this would be not allow to pass any Context to the slot functions. Then we wouldn't need this workaround with a branched template, and we could simply call:

```py
result = self.slots.fallback({ "error": error })
```

JuroOravec:
Btw I noticed that this PR didn't trigger the Discord integration ([see pipeline](https://github.com/django-components/django-components/actions/runs/18228321993/job/51918810372?pr=1433)). My guess is it's because this PR was opened from a fork, so it didn't have access to repository secrets. the URL to post to Discord is stored as a secret.

I'll fix this in the next PR

Screenshot 2025-10-04 at 09 01 45

src/django_components/apps.py
```diff
@@ -66,6 +67,7 @@ def ready(self) -> None:
 
         # Register the dynamic component under the name as given in settings
         registry.register(app_settings.DYNAMIC_COMPONENT_NAME, DynamicComponent)
+        registry.register("error_fallback", ErrorFallback)
```
JuroOravec, line 70: I didn't add a setting to change the name under which the ErrorFallback is accessed from within the templates. We can add it later if need be.

JuroOravec:
cc @EmilStenstrom

PR #1435, opened 2025-10-04
refactor: update workflow deps, update docs, and smaller images

JuroOravec:
Follow up for https://github.com/django-components/django-components/pull/1411 with small improvements


docs/css/style.css
```diff
@@ -41,23 +41,23 @@ h6 {
...
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  font-size: 18px;
 }
 
 .user .avatar-wrapper {
-  width: 8em;
-  height: 8em;
+  width: 6em;
```
JuroOravec, line 51: Made the people's profile pictures slightly smaller

.github/workflows/maint-docs-people.yml
```diff
@@ -5,20 +5,24 @@ name: Django Components People
 
 on:
   schedule:
+    # At 18:00 on day-of-month 1.
     - cron: "0 18 1 * *"
 
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
JuroOravec, line 12: This will allow us to trigger the action to test it so we don't have to wait for a whole month.

JuroOravec: Also this workflow file was in `.github/actions/`, which I didn't notice at first. So I moved it to `.github/workflows/` instead.

Based on checking ChatGPT and docs, my understanding is that the `actions` directory is for defining local [Github Actions](https://docs.github.com/en/actions/reference/workflows-and-actions/metadata-syntax).

E.g. "actions" in Github are the pieces of code that we call from within "workflows" like so:

```yml
jobs:
  myjob:
    steps:
      - uses: actions/checkout@v5      # <--- Action

      - name: Set up Python
        uses: actions/setup-python@v6  # <--- Action
        with:
          python-version: '3.13'
          python-version: "3.13"
```

JuroOravec:
cc @Antoliny0919

PR #1436, opened 2025-10-04
docs: self-contained examples

JuroOravec:
This PR makes it easier to add [public examples](https://django-components.github.io/django-components/latest/examples/) in the documentation site.

All examples are defined in a single place - `docs/examples`. The examples have special folder layout:
 
```
|- docs/
  |- examples/
    |- <example_name>/
      |- component.py - The component definition
      |- page.py      - The page view for the example
      |- test_example_<example_name>.py - Tests
      |- README.md    - Component documentation
      |- images/      - Images used in README
```

This allows us to keep the examples in one place, yet define, test, and document them.

**Dev server previews**

There's a script in `sampleproject/examples/utils.py` that picks up the `component.py` and `page.py` files and imports them. In `sampleproject/examples/url.py` we then automatically create URLs for live demos, e.g. `http://localhost:8000/examples/<example_name>`.

To see all available examples, one can go to `http://localhost:8000/examples/`.

**Tests**

I'm using the file format `test_example_<example_name>.py` to define tests for the example. These tests are then picked up when running pytest.

Closes https://github.com/django-components/django-components/issues/1415

Screenshot 2025-10-04 at 23 06 16

Screenshot 2025-10-04 at 17 26 11

Screenshot 2025-10-04 at 16 05 25

Screenshot 2025-10-04 at 16 55 08


sampleproject/examples/components/form/form.py
```diff
@@ -5,8 +5,6 @@
 
 @register("form")
 class Form(Component):
-    template_file = "form.html"
```
JuroOravec, line 8: Because the examples are defined in `docs/examples`, this was causing problem when I tried to test the components if they at the same time defined separate HTML or CSS files. I didn't want to have to think how to configure the template directories correctly. So instead I inlined the HTML/JS/CSS inside Python.

docs/examples/form/test_example_form.py
```diff
@@ -1,41 +1,25 @@
...
-    parametrize=PARAMETRIZE_CONTEXT_BEHAVIOR,
-    components_settings={
-        "dirs": [
-            Path(__file__).parent / "components",
-            # Include the directory where example components are defined
-            Path(__file__).parent.parent / "sampleproject/examples/components",
-        ],
-    },
-)
+@pytest.mark.django_db
```
JuroOravec, line 17: Tests were tricky too, and I was running into issues with how Django is set up.

In an ideal world, each test file would run in its own Python process, so they could configure Django however each file wants. Currently it's all shared because Django settings are global.

There's one set of settings used for normal tests. But because these tests in `docs/examples` are located elsewhere then the rest of tests, I couldn't import the test helper that would set up these common settings.

One workaround I found was to use `@pytest.mark.django_db`. That comes from [pytest-django](https://pytest-django.readthedocs.io/en/latest/).

But it's kind of weird that when I use `@pytest.mark.django_db` and run tests only from `docs/examples`, then they all get skipped. And it's only when the tests in `docs/examples` are ran together with the rest of tests from `tests/` that it actually all works.

So what that means is that in CI everything should work alright. But when working on tests in `docs/examples`, then I found I had to run a command that ran 1) the examples tests, AND 2) at least one file from `tests/`:

```sh
tox -e py39 -- docs/examples tests/test_component.py
```

pyproject.toml
```diff
@@ -216,5 +217,6 @@ reportIncompatibleVariableOverride = "none"
 [tool.pytest.ini_options]
 testpaths = [
     "tests",
+    "docs/examples",
```
JuroOravec, line 227: This line here ensures that pytest picks up the tests in `docs/examples`

sampleproject/examples/urls.py
```diff
@@ -1,9 +1,57 @@
...
+
+# For each example in `docs/examples/*`, register a URL pattern that points to the example's view.
+# The example will be available at `http://localhost:8000/examples/<example_name>`.
+# The view is the first Component class that we find in example's `page.py` module.
+#
+# So if we have an example called `form`:
+# 1. We look for a module `examples.dynamic.form.page`,
+# 2. We find the first Component class in that module (in this case `FormPage`),
+# 3. We register a URL pattern that points to that view (in this case `http://localhost:8000/examples/form`).
+def get_example_urls():
```
JuroOravec, line 19: This auto-registers URLs on the dev server for previewing the examples, like `http://localhost:8000/examples/form`

sampleproject/examples/views.py
```diff
@@ -0,0 +1,122 @@
+from django.http import HttpRequest
+from django.utils.safestring import mark_safe
+
+from django_components import Component, types
+
+from .utils import discover_example_modules
+
+
+class ExamplesIndexPage(Component):
```
JuroOravec, line 9: The examples index page which shows all the examples.

This is part of the dev server (`sampleproject`)

http://localhost:8000/examples

Screenshot 2025-10-04 at 17 26 11

tests/test_attributes.py
```diff
@@ -11,7 +11,7 @@
 
 from .testutils import PARAMETRIZE_CONTEXT_BEHAVIOR, setup_test_config
 
-setup_test_config({"autodiscover": False})
```
JuroOravec, line 14: Unrelated, but I modified `setup_test_config` to set autodiscovery to False by default in tests, so I could remove a lot of these.

JuroOravec:
cc @EmilStenstrom

PR #1439, opened 2025-10-06
refactor: fix djc not working when not among built-ins

JuroOravec:
Fix for https://github.com/django-components/django-components/pull/1390

Also fixes another bug that was introduced in v0.142.0 where I forgot to add `{% load component_tags %}` inside the `ErrorFeedback`'s template.

### DIFF
--- a/src/django_components/components/error_fallback.py
+++ b/src/django_components/components/error_fallback.py
@@ -148,6 +148,7 @@ class ErrorFallback(Component):
     # TODO - Once we don't have to pass Context to the slot, we can remove the template
     #        and render the slots directly.
     template: types.django_html = """
+        {% load component_tags %}
         {% if not error %}
             {% slot "content" default / %}
         {% else %}
--- a/src/django_components/util/django_monkeypatch.py
+++ b/src/django_components/util/django_monkeypatch.py
@@ -285,6 +285,7 @@ def monkeypatch_inclusion_init(inclusion_node_cls: Type[Node]) -> None:
         original_init(self, func, takes_context, args, kwargs, filename, *future_args, **future_kwargs)  # type: ignore[misc]
 
         orig_func = self.func
+
         def new_func(*args: Any, **kwargs: Any) -> Any:
             result = orig_func(*args, **kwargs)
             result["_DJC_INSIDE_INCLUSION_TAG"] = True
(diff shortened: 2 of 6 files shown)


src/django_components/util/django_monkeypatch.py
```diff
@@ -196,6 +197,12 @@ def _template_render(self: Template, context: Context, *args: Any, **kwargs: Any
         if not COMPONENT_COMMENT_REGEX.search(result.encode("utf-8")):
             return result
 
+        # Don't post-process if this template was rendered with Django's InclusionNode.
+        # Fix for https://github.com/django-components/django-components/issues/1390
+        # NOTE: Lenght of 2 means a Context with single layer (+ layer with defaults)
+        if "_DJC_INSIDE_INCLUSION_TAG" in context and len(context.dicts) == 2:
```
JuroOravec, line 203: The fix was to skip post-processing when a template was rendered by Django's `inclusion_tag` / `InclusionNode`.

The challenge, tho, was that the `InclusionNode` does not leave any state that we could use to identify that a particular Template was rendered by the `InclusionNode`.

So I decided to monkeypatch Django's `InclusionNode` class, so that it passes also `{"_DJC_INSIDE_INCLUSION_TAG": true}` to the template's context.

That way, inside `Template.render()`, we can check for that flag, and skip post-processing if it's present.

---

It's annoying that it's yet another Django class that needs to be patched, I'd rather ask people to not mix inclusion tags and our components.

But then again, making these compatible will make it easier for people to eventually migrate from Django's syntax (include, extends/blocks, custom tags) to components.

src/django_components/components/error_fallback.py
```diff
@@ -148,6 +148,7 @@ def on_render(
     # TODO - Once we don't have to pass Context to the slot, we can remove the template
     #        and render the slots directly.
     template: types.django_html = """
+        {% load component_tags %}
```
JuroOravec, line 151: Forgetting this broke v0.142.0

tests/test_dependency_rendering.py
```diff
@@ -631,3 +631,43 @@ class SimpleOuter(SimpleComponent):
             <div data-djc-id-ca1bc44>Another</div>
             """,
         )
+
+    # See https://github.com/django-components/django-components/pull/1390
+    def test_dependencies_with_component_in_inclusion_tag(self):
```
JuroOravec, line 636: Test taken from #1390

JuroOravec:
cc @EmilStenstrom

PR #1449, opened 2025-10-09
feat: Auto-set View.public when HTTP methods defined

### LINKED ISSUE #1446: Make Component.View.public optional
I find that having to set `Component.View.public` is bit of a hindrance - easy to forget, and harder to test because you get the error only at runtime.

The initial idea was so that flag was so that we don't spend too much time registering all component classes as View with `urlpatterns`. With the flag we know the exact few classes that need to be registered.

But maybe there could be a more convenient way.

**Option 1) - Check for overrides** - We could check if any of the methods (like `View.post` or `View.get`) have been overriden. And if so, set `Component.View.public` to `True` (if it's not explicitly set).

**Option 2) - Set component to public when calling `get_component_url()` on that Component** - Also an option. But I don't like it as much. Could be trickier. And if I'm calling `get_component_url()` on a component, I assume it must must have defined some HTTP methods, so option 1 applies anyway.

RohanDisa:
feat: Auto-set Component.View.public when HTTP methods are overridden

Automatically sets `Component.View.public = True` when users define HTTP
methods (get, post, put, etc.) without explicitly setting public, removing
the need to manually set this flag.

Implementation:
- Added auto-detection logic in `_init_component_class()` after View wrapper creation
- Checks user-defined View classes for HTTP method overrides
- Respects explicit `public` settings (both True and False)
- Only inspects user code, ignores framework base classes

Tests added:
- test_auto_public_set_for_overridden_http_method
- test_auto_public_set_with_multiple_http_methods 
- test_default_public_false_when_no_methods_define
- test_explicit_public_true_is_preserved

Fixes #1446

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ _06 Oct 2025_
 
 ## v0.142.0
 
+⚠️ This version is broken, please update to v0.142.1 ⚠️
+
 _05 Oct 2025_
 
 #### Feat
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ license = { text = "MIT" }
 [project.urls]
 Homepage = "https://github.com/django-components/django-components/"
 Documentation = "https://django-components.github.io/django-components/"
-Changelog = "https://django-components.github.io/django-components/latest/release_notes/"
+Changelog = "https://django-components.github.io/django-components/latest/releases/"
 Issues = "https://github.com/django-components/django-components/issues"
 Donate = "https://github.com/sponsors/EmilStenstrom"
 


RohanDisa:
Hi @JuroOravec, just wanted to check in to see if there’s been any update or if there’s anything I can help with to move this forward. Thanks!

JuroOravec:
Hey @RohanDisa thanks for ping, and sorry for silence I was away Fri-Sun.

Thanks for the help! The idea was good, tho there was more work that had to be done - I moved the logic to `src/django_components/extensions/view.py`, and simplified by it by simply checking for the methods instead of walking over all base classes. Then I also reviewed the docs where we mentioned the `Component.View.public`.

I'm not sure what's the best practice for making these kind of large changes to PRs, so I made a new PR (https://github.com/django-components/django-components/pull/1451), and mentioned you as the co-author ([see commit](https://github.com/django-components/django-components/commit/485027b308c0088567172c566e60d1817caf04c8)).

I'll release this today / tomorrow as v0.142.3

PR #1450, opened 2025-10-10
docs: fix changelog issues / inconsistencies

KyeRussell:
- Amend release notes for v0.142.0 to include the 'do not use' warning that's included in the notes on the GitHub release.
- Amend `pyproject.toml` to point to correct changelog URL - this is affecting what's fed through to PyPI, Renovate, etc.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ license = { text = "MIT" }
 [project.urls]
 Homepage = "https://github.com/django-components/django-components/"
 Documentation = "https://django-components.github.io/django-components/"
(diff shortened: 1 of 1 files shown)


dalito:
In addition to the change, there is also the possibility to [yank a version in pypi](https://docs.pypi.org/project-management/yanking/). Then no one can install it anymore.

JuroOravec:
@KyeRussell Thanks for these!

@dalito Thanks I'll do that

JuroOravec:
Ah, forgot I don't have access to the project on PyPi. @EmilStenstrom, would you yank v0.142.0?

EmilStenstrom:
@JuroOravec It's now yanked!

PR #1451, opened 2025-10-12
refactor: make Component.View.public optional

JuroOravec:
As the title says, this makes the `Component.View.public = True` optional. If user defines any HTTP handlers like `Component.View.get`, then the component will be automatically made public.

Before:

```py
class MyTable(Component):
    class View:
        public = True

        def get(self, request):
            return self.render_to_response()

url = get_component_url(MyTable)
```

After:

```py
class MyTable(Component):
    class View:
        def get(self, request):
            return self.render_to_response()

url = get_component_url(MyTable)
```

One can still explicitly expose/hide the component with explicit `Component.View.public = True/False`.

Closes https://github.com/django-components/django-components/issues/1446

Supersedes https://github.com/django-components/django-components/pull/1449

Screenshot 2025-10-12 at 18 47 51

### DIFF
--- a/tests/test_component.py
+++ b/tests/test_component.py
@@ -1243,7 +1243,7 @@ class TestComponentRender:
                 def get(self, request):
                     how = "via GET request"
 
-                    return self.component_cls.render_to_response(
+                    return self.component_cls.render_to_response(  # type: ignore[attr-defined]
                         context=RequestContext(request),
                         kwargs={"how": how},
                     )
--- a/tests/test_component_view.py
+++ b/tests/test_component_view.py
@@ -98,7 +98,7 @@ class TestComponentAsView:
 
             class View:
                 def get(self, request, *args, **kwargs) -> HttpResponse:
-                    return self.component_cls.render_to_response(kwargs={"variable": "GET"})
+                    return self.component_cls.render_to_response(kwargs={"variable": "GET"})  # type: ignore[attr-defined]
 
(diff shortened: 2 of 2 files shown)


tests/test_component.py
```diff
@@ -1240,7 +1240,7 @@ class Thing(Component):
             def get_template_data(self, args, kwargs, slots, context):
                 return {"how": kwargs.pop("how")}
 
-            class View(ComponentView):
```
JuroOravec, line 1243: Removed the cases where `Component.View` subclassed from `ComponentView` as that's not necessary

src/django_components/extensions/view.py
```diff
@@ -306,4 +335,25 @@ def on_component_class_deleted(self, ctx: OnComponentClassDeletedContext) -> Non
 def _is_view_public(view_cls: Optional[Type[ComponentView]]) -> bool:
     if view_cls is None:
         return False
-    return getattr(view_cls, "public", False)
+
+    # Allow users to skip setting `View.public = True` if any of the HTTP methods
+    # are defined. Users can still opt-out by explicitly setting `View.public` to `True` or `False`.
+    public = getattr(view_cls, "public", None)
```
JuroOravec, line 341: Instead of the implementation in https://github.com/django-components/django-components/pull/1449, this one doesn't check all the base classes of `Component.View`. Instead it only checks whether final `Component.View` defines any non-default HTTP handlers.

JuroOravec:
cc @EmilStenstrom @RohanDisa

PR #1458, opened 2025-10-17
refactor: make djc compatible with django-template-partials v23.3 and later

JuroOravec:
Add `_render()` method to `TemplateProxy` from `django-template-partials`. This makes django-components compatible with `django-template-partials` as old as v23.3.

I also added a section to the compatibility page on compatibility with other packages, where we can document the situation.

Closes https://github.com/django-components/django-components/issues/1455

Screenshot 2025-10-17 at 10 45 14

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -12,8 +12,6 @@ asv==0.6.5
     # via -r requirements-dev.in
 asv-runner==0.2.1
     # via asv
-backports-asyncio-runner==1.2.0
-    # via pytest-asyncio
 build==1.3.0
     # via asv
 cachetools==6.2.0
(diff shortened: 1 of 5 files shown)


JuroOravec:
cc @EmilStenstrom

tests/test_integration_template_partials.py
```diff
@@ -28,6 +28,7 @@
 setup_test_config()
 
 
+# Test compatibility with django-template-partials.
```
JuroOravec, line 31: Added this because for a second I forgot that there already were tests for django-template-partials because I couldn't find them by the package name

requirements-dev.txt
```diff
@@ -12,8 +12,6 @@ asv==0.6.5
     # via -r requirements-dev.in
 asv-runner==0.2.1
     # via asv
-backports-asyncio-runner==1.2.0
```
JuroOravec, line 15: Unrelated, but pinning the exact version of `backports-asyncio-runner` was preventing installing dev deps on python 3.12

PR #1466, opened 2025-10-20
refactor: make it optional having to specify parent class of Args, Kwargs, Slots, etc

JuroOravec:
This PR makes it optional for our users to specify the base classes of `Args`, `Kwargs`, `Slots` (etc) classes. 

Before:

```py
class Button(Component):
    class Kwargs(NamedTuple):  # <--------
        variable: str
        maybe_var: Optional[int] = None

    def get_template_data(self, args, kwargs: Kwargs, slots, context):
        ...
```

After:

```py
class Button(Component):
    class Kwargs:  # <--------
        variable: str
        maybe_var: Optional[int] = None

    def get_template_data(self, args, kwargs: Kwargs, slots, context):
        ...
```

Now, when the `Kwargs` or other class  doesn't extend anything else, we automatically convert it to a `NamedTuple` behind the scenes.

Part of #1258

### DIFF
--- a/src/django_components/util/misc.py
+++ b/src/django_components/util/misc.py
@@ -314,5 +314,11 @@ def convert_class_to_namedtuple(cls: Type[Any]) -> Type[Tuple[Any, ...]]:
     else:
         defaults_list = []
     tuple_cls = namedtuple(cls.__name__, field_names, defaults=defaults_list)  # type: ignore[misc]  # noqa: PYI024
+
+    # `collections.namedtuple` doesn't allow to specify annotations, so we pass them afterwards
     tuple_cls.__annotations__ = cls.__annotations__
+    # Likewise, `collections.namedtuple` doesn't allow to specify class vars
+    for name, value in class_attrs.items():
+        setattr(tuple_cls, name, value)
+
     return tuple_cls


docs/concepts/fundamentals/typing_and_validation.md
```diff
@@ -270,34 +268,63 @@ class Button(Component):
 
 ## Custom types
```
JuroOravec, line 269: This docs section was updated to describe the new feature.

docs/examples/form_grid/component.py
```diff
@@ -114,7 +114,7 @@ def prepare_form_grid(slots: Dict[str, Slot]):
             # Case: Component user didn't explicitly define how to render the label
             #       We will create the label for the field automatically
             label = FormGridLabel.render(
-                kwargs=FormGridLabel.Kwargs(field_name=field_name),
+                kwargs=FormGridLabel.Kwargs(field_name=field_name),  # type: ignore[call-arg]
```
JuroOravec, line 117: One drawback of omitting `NamedTuple` as parent class is that mypy no longer knows how these classes should behave when being instantiated.

Of course, users may want to subclass `Kwargs` with `NamedTuple` (or other), so that calling `Kwargs(...)` works correctly with mypy.

But within this codebase I wanted to show the minimal examples, so instead in some tests and code examples I've added `type: ignore[call-arg]`.

src/django_components/component.py
```diff
@@ -506,6 +513,34 @@ def __new__(mcs, name: str, bases: Tuple[Type, ...], attrs: Dict) -> Type:
...
+        #     class Kwargs(NamedTuple):
+        #         ...
+        # ```
+        # Can be simplified to:
+        # ```py
+        # class MyTable(Component):
+        #     class Kwargs:
+        #         ...
+        # ```
+        for data_class_name in ["Args", "Kwargs", "Slots", "TemplateData", "JsData", "CssData"]:
```
JuroOravec, line 530: Here is the actual implementation, inside Component's metaclass. The rest of the file is just updating hte documentation.

src/django_components/extension.py
```diff
@@ -265,16 +273,27 @@ def component(self) -> "Component":
...
-        component = self._component_ref()
+        component: Optional[Component] = None
+        if self._component_ref is not None:
+            component = self._component_ref()
         if component is None:
             raise RuntimeError("Component has been garbage collected")
         return component
 
-    def __init__(self, component: "Component") -> None:
+    def __init__(self, component: "Optional[Component]") -> None:
```
JuroOravec, line 288: Changes in this file are not related to the main feature. But as I was trying to finaly finish the Storybook intergration, this was one of the issues I came across.

The gist is that the methods defined for the Storybook integration do not run during the normal rendering process. Instead, the code simply finds all Component classes, and for each calls `Component.Storybook.generate()`.

But to keep the API clean, `Component.Storybook.generate()` is not a class method. And that means that `Component.Storybook` needs to be instantiated.

Previously instantiating `Component.Storybook` required a dummy `Component` instance. Now it can be more appropriate by passing in `None`.

src/django_components/extensions/dependencies.py
```diff
@@ -16,6 +16,14 @@ class DependenciesExtension(ComponentExtension):
 
     # Cache the component's JS and CSS scripts when the class is created, so that
     # components' JS/CSS files are accessible even before having to render the component first.
+    #
+    # This is important for the scenario when the web server may restart in a middle of user
```
JuroOravec, line 20: Also not related to the main feature. I just noticed that I forgot why this code was here, so documented it.

src/django_components/util/misc.py
```diff
@@ -267,3 +269,50 @@ def format_as_ascii_table(
 def is_generator(obj: Any) -> bool:
     """Check if an object is a generator with send method."""
     return hasattr(obj, "send")
+
+
+def convert_class_to_namedtuple(cls: Type[Any]) -> Type[Tuple[Any, ...]]:
```
JuroOravec, line 274: Another part of the implementation of the main feature. Turns out that creating NamedTuples dynamically from another class is more complicated than would seem at first. I wanted to use NamedTuples because there was some comment / thread where I read that NamedTuples may be up to 3x faster than dataclasses.

But one can't simply use `NamedTuple` with other parents:

```py
class X(NamedTuple, Another):
    x: int = 1
```

Also you can't further subclass the subclass of `NamedTuples`:

```py
class Another(NamedTuple):
    x: int = 1

class X(Another):
    y: str
```

Also, when using `typing.NamedTuple` as a function, you can't pass in defaults:

```py
my_class = typing.NamedTuple("MyClass", [("x", int), ("y", str)])
```

I tried setting the defaults ([`_field_defaults`](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._field_defaults)) manually, but Python wasn't picking that up.

One option was to define the NamedTuple with a class syntax as a string, and then evaluate that string. But that had 2 problems - 1) security risk, and 2) we'd need to import all the types used in annotations:

```py
my_cls_str = """
from typing import NamedTuple

from path.to.custom import CustomClass

class MyClass(NamedTuple):
    x: int
    y: str
    z: CustomClass
"""
my_cls = eval(my_cls_str, {}, {})
```

Lastly I managed to get it working using `collections.namedtuple`. This function doesn't define the field annotations, but it is able to handle defaults. Just one annoying thing was how it sets defaults the same way as Python functions do - it assigns the defaults from the back of the function signature.

So if I have NamedTuple with 3 fields - `x`, `y`, and `z`, and I set defaults to `["hello", 123]`:

```py
my_cls = namedtuple("MyClass", ["x", "y", "z"], defaults=["hello", 123])
```

then this is the same as writing:

```py
class MyClass(NamedTuple):
    x: int
    y: str = "hello"
    z: int = 123
```

One caveat is that `collections.namedtuple()` doesn't set the annotations, so I had also set `__annotations__` at the end.

---

Also, one thing I didn't know - when defining `NamedTuple`, the fields it recognizes are only those that are annotated. If a field is NOT annotated, it's treated as a CLASS VARIABLE:

```py
class MyClass(NamedTuple):
    x: int            # Required field
    y: str = "hello"  # Optional field (with default)
    z = 123           # Class var!
```

So at the end I also assigned the class variables to the new namedtuple class.

---

Surprisingly LLMs were really bad at solving this, so I'm hoping this comment gets into the next training set.

JuroOravec:
cc @EmilStenstrom, lot's of docs change, but the actual code is fairly small.

EmilStenstrom:
@JuroOravec Great little API update!

PR #1467, opened 2025-10-21
feat: allow to set comp defaults on Kwargs class

JuroOravec:
Allow to define component defaults on `Component.Kwargs`. And add new helper `get_component_defaults()` to get the final, merged defaults.

Before, the defaults had to be defined on a separate `Component.Defaults` class:

```python
class ProfileCard(Component):
    class Kwargs:
        user_id: int
        show_details: bool

    class Defaults:
        show_details = True
```

Now, django-components can detect the defaults from `Component.Kwargs` and apply
them. So you can merge `Component.Kwargs` with `Component.Defaults`:

```python
class ProfileCard(Component):
    class Kwargs:
        user_id: int
        show_details: bool = True
```

NOTE: This applies only when `Component.Kwargs` is a NamedTuple or dataclass.

Closes https://github.com/django-components/django-components/issues/1258

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -530,6 +530,7 @@ class ComponentMeta(ComponentMediaMeta):
         # NOTE: Using dataclasses with `slots=True` could be faster than using NamedTuple,
         #       but in real world web pages that may load 1-2s, data access and instantiation
         #       is only on the order of milliseconds, or about 0.1% of the overall time.
+        #       See https://github.com/django-components/django-components/pull/1467#discussion_r2449009201
         for data_class_name in ["Args", "Kwargs", "Slots", "TemplateData", "JsData", "CssData"]:
             data_class = attrs.get(data_class_name)
             # Not a class


CHANGELOG.md
```diff
@@ -1,10 +1,64 @@
...
+    class ProfileCard(Component):
+        class Kwargs:
+            user_id: int
+            show_details: bool = True
+    ```
+
+    NOTE: This applies only when `Component.Kwargs` is a NamedTuple or dataclass.
 
-- Simpler syntax for defining component inputs.
+- New helper `get_component_defaults()`:
```
JuroOravec, line 33: I made `get_component_defaults` public so that it is possible to get component's defaults from the outside. This will be needed for language server to be able to infer variables and inputs inside templates.

E.g. if we had a template like so:

```django
<ul>
  {% for item in items %}
    <li>
        {% component "profile_page" item=item icon_size=24 editable=False / %}
    </li>
  {% endfor %}
</ul>
```

Then we could show a linter error in the template if an input was missing, because:

1. With the Rust template AST parser we can statically identify which params were passed to the component
2. We can use `get_component_defaults()` on the component class `ProfilePage`, to get the defaults
3. Thus we can combine 1) and 2) to know what all params will be available.
4. We can then check `ProfilePage.Args`, `ProfilePage.Kwargs`, and `ProfilePage.Slots` to decide if the component is missing any inputs / slots. And if so, mark it as an error.

Moreover, if we knew statically:
1. What variables were passed to the templates
2. How `{% for %}` works

Then we could infer also types of inputs (e.g. `item`), and thus check typing.

docs/getting_started/parametrising_components.md
```diff
@@ -255,6 +255,57 @@ class Calendar(Component):
         }
 ```
 
+### 6. Add input validation
```
JuroOravec, line 258: Added section on adding `Component.Kwargs` to Getting Started.

While intially the component typing / validation feature felt niche (back when it used generics). Now the API is I think really simple, and so I think it can be part of the "MVP"

src/django_components/component.py
```diff
@@ -527,6 +527,9 @@ def __new__(mcs, name: str, bases: Tuple[Type, ...], attrs: Dict) -> Type:
         #     class Kwargs:
         #         ...
         # ```
+        # NOTE: Using dataclasses with `slots=True` could be faster than using NamedTuple,
+        #       but in real world web pages that may load 1-2s, data access and instantiation
+        #       is only on the order of milliseconds, or about 0.1% of the overall time.
```
JuroOravec, line 532: On this comment, consider these benchmarks (based on [this gist](https://gist.github.com/wolph/02fae0b20b914354734aaac01c06d23b))

Instantiation:

bench_inst_local

Attribute access:

bench_access_local

What this tells us is that on Python 3.11, `NamedTuples` are about 2x slower than using dataclasses.

I looked into this as I was wondering if I should have chosen a different data type for `Component.Args`, `Kwargs`, `Slots`, etc.

However, the benchmark ran those instantiations / attribute access about 5 million times. On the other hand, in real world, I expect that large web pages might do the instatiations / accesses maybe 5-10K times . So 500-1000x less.

Assuming that a large web page loads 1-2 seconds, that means that maybe ~0.1-0.2% of the time is spent on data access/creation.

JuroOravec:
Some screenshots

Screenshot 2025-10-21 at 17 58 34
Screenshot 2025-10-21 at 17 23 10
Screenshot 2025-10-21 at 17 18 33
Screenshot 2025-10-21 at 17 06 19

JuroOravec:
cc @EmilStenstrom

PR #1468, opened 2025-10-21
refactor: add support for python 3.14

JuroOravec:
Updated code / docs to support Python 3.14 as per the `supported_versions.py` script.

Also updated djc-core-html-parser to v1.0.3, to ensure it handles Python 3.14. See https://github.com/django-components/djc-core-html-parser/issues/10

Closes https://github.com/django-components/django-components/issues/1464

### DIFF
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -71,7 +71,7 @@
     // The Pythons you'd like to test against.  If not provided, defaults
     // to the current version of Python used to run `asv`.
     "pythons": [
-        "3.14"
+        "3.13"
     ],
 
     // The list of conda channel names to be searched for benchmark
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -34,10 +34,18 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
         cache: "pip"
-    - name: Install dependencies
+    - name: Install dependencies (Python 3.8)
+      if: matrix.python-version == '3.8'
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install -r requirements-ci-python38.txt
```
JuroOravec, line 41: Because Python 3.8 is no longer supported for so long, the 3rd party dependencies that we use during testing are starting to diverge - Py38 needs different deps than py314.

As a simplest approach, I've made a copy of `requirements-ci.txt`, and made it specific for python 3.8.

Based on the conditional, we install either the deps for 3.8, or for the rest of the Pythons.

And same logic is in `tox.ini`, where py38 installs an old version of playwright.

JuroOravec:
cc @EmilStenstrom

PR #1484, opened 2025-11-09
refactor: raise when template data overwrites variables from context_processors

JuroOravec:
When constructing the variables available to the context, the data comes from 2 sources:
- from `Component.get_template_data()` (or old `get_context_data`)
- from Django's context processors

Currently, when these conflict, the context processors silently overwrite the component's own data.

To improve user experience, this PR turns this into an error, so users are made aware of this conflict.

Closes https://github.com/django-components/django-components/issues/1482


JuroOravec:
cc @EmilStenstrom

mpboom:
Thanks for the fix @JuroOravec <3 . However, do you have any policy surrounding the versioning here? This is released as a patch release, but it's quite a breaking change 🫣

PR #1511, opened 2025-12-01
Security Fix in CICD: move PR fields into env vars 

barakharyati:
Replaced inline ${{ github.* }} values in the run step with environment variables. This prevents bash from interpreting untrusted PR metadata (e.g., title, user login) as shell syntax when running under pull_request_target. No behavior changes—only safer variable handling.

### DIFF
--- a/.github/workflows/pr-discord.yml
+++ b/.github/workflows/pr-discord.yml
@@ -26,7 +26,7 @@ jobs:
             --arg url "$URL" \
             --arg user "$USER" \
             --arg repo "$REPO" \
-            --arg number "#$NUMBER" \
+            --arg number "$NUMBER" \
             --arg timestamp "$TS" \
             '{
               "username": "GitHub Pull Requests",


EmilStenstrom:
@barakharyati Tell me more about this fix. Are you an author of the library that we use? How did you find it?

barakharyati:
> @barakharyati Tell me more about this fix. Are you an author of the library that we use? How did you find it?

@EmilStenstrom   Thanks for the follow-up!

I'm not an author of the library. I found the issue while reviewing GitHub workflows for security risks related to pull_request_target usage.  any untrusted PR metadata (like the title or username) that is passed directly into a run step can allow shell injection.

Your workflow was functionally correct — the fix simply moves those PR-controlled fields into environment variables before entering the shell. Using "$VAR" prevents bash from interpreting malicious characters, while jq --arg safely JSON-escapes them. No logic or behavior was changed, just a hardening step.

Happy to share more details or if needed

EmilStenstrom:
@barakharyati Mind fixing the variable prefixed with "#" as copilot suggests?

barakharyati:
> @barakharyati Mind fixing the variable prefixed with "#" as copilot suggests?

@EmilStenstrom  The "#" was part of the original workflow, so I kept it to ensure the PR remains as minimal and non-disruptive as possible.
If you prefer adjusting it, I’m happy to update it in a follow-up commit.

EmilStenstrom:
@barakharyati Better get it right than minimal! :) Thanks!

barakharyati:
@EmilStenstrom  
changed  please review

EmilStenstrom:
@barakharyati Thank you!

EmilStenstrom:
@barakharyati Merged.

barakharyati:
Hi @EmilStenstrom 

Can you enable the github vulnerability report that we can register it? And for future private disclosure? 

You can find more info here

https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing-information-about-vulnerabilities/privately-reporting-a-security-vulnerability

Thank you

EmilStenstrom:
@barakharyati Enabled. Thanks again.

PR #1523, opened 2026-01-07
feat: add possibility to customize component url

alexandreMartinEcl:
## Some background

We've been using django-components for a couple years now, for [Contexte](https://www.contexte.com/) website.
Using components with extension `class View` seemed quite promising and we had our try on it.

We wanted to call a form as a component, so that the `View` would inherit from django `SingleObjectMixin`. This django view would expect a url like `/components_url/<int:pk>/`.

Overriding `def get_object()` is a possibility, as well as registering ourselves the component url `"components/ext/view/components/{MyComponent.class_id}/<int:pk>"`.

## Suggestion

I'd love avoiding overriding the django view method, so I'd suggest to leave the developer the possibility to customize the component path like this

```
class MyComponent(Component):
    class View(ComponentView):
        @classmethod
        def get_route_path(cls) -> str:
            return f"{super().get_route_path()}<int:pk>/"
...
component_url = get_component_url(type(self), args=[pk])
```

## In the codebase

```
def get_component_url(
    component: Union[Type["Component"], "Component"],
    ...
    args: Optional[Any] = None,
    kwargs: Optional[Any] = None,
) -> str:
    ...
    url = django.urls.reverse(route_name, args=args, kwargs=kwargs)
    return format_url(url, query=query, fragment=fragment)
```
```
class ComponentView(ExtensionComponentConfig, View):
    ...
    @classmethod
    def get_route_path(cls, ) -> str:
        return f"components/{cls.component_cls.class_id}/"


class ViewExtension(ComponentExtension):
    ...
    # Create URL route on creation
    def on_component_class_created(self, ctx: OnComponentClassCreatedContext) -> None:
        ...
        route_name = _get_component_route_name(comp_cls)
        ...
```

### DIFF
--- a/.github/workflows/discussion-discord.yml
+++ b/.github/workflows/discussion-discord.yml
@@ -11,14 +11,20 @@ jobs:
       - name: Post to Discord (development channel)
         env:
           DISCORD_WEBHOOK_URL: ${{ secrets.DISCORD_WEBHOOK_DEVELOPMENT }}
+          TITLE: ${{ github.event.discussion.title }}
+          URL: ${{ github.event.discussion.html_url }}
+          USER: ${{ github.event.discussion.user.login }}
+          REPO: ${{ github.repository }}
+          NUMBER: ${{ github.event.discussion.number }}
+          TS: ${{ github.event.discussion.created_at }}
         run: |
           payload=$(jq -n \
-            --arg title "${{ github.event.discussion.title }}" \
-            --arg url "${{ github.event.discussion.html_url }}" \
-            --arg user "${{ github.event.discussion.user.login }}" \
-            --arg repo "${{ github.repository }}" \
-            --arg number "#${{ github.event.discussion.number }}" \
-            --arg timestamp "${{ github.event.discussion.created_at }}" \
+            --arg title "$TITLE" \
+            --arg url "$URL" \
+            --arg user "$USER" \
+            --arg repo "$REPO" \
+            --arg number "#$NUMBER" \
+            --arg timestamp "$TS" \
             '{
               "username": "GitHub Discussions",
               "avatar_url": "https://github.githubassets.com/images/modules/logos_page/GitHub-Mark.png",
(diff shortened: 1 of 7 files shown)


JuroOravec:
Hi @alexandreMartinEcl, great suggestion!

> We wanted to call a form as a component, so that the `View` would inherit from django `SingleObjectMixin`. This django view would expect a url like `/components_url/<int:pk>/`.
> 
> Overriding `def get_object()` is a possibility, as well as registering ourselves the component url `"components/ext/view/components/{MyComponent.class_id}/<int:pk>"`.

Do I understand right that for your case you need the URL to contain the `<int:pk>` route param, and that's why you want to modify the generated URL?

Because there's an important nuance, that only the trailing end of the URL path can be modified ([see example](https://github.com/django-components/django-components/pull/1523/files#diff-5370e45fff2717ed8f26914ecedafa84fb4a4aaf7d9cfbc2003ed06c0bca47caR342)). But the changes in the PR suggest you are aware of this.

What I'm talking about is that all component URL paths are prefixed with `/components/ext/view/`.

This is due to the architecture of django-components:
- the View feature is implemented as an "extension"
- Extensions can introduce new URL paths
- django-components' internals then register these paths, prefixing them with `/components/ext/<extension_name>/`

But if I understand right, than for your case you don't need to change the prefix, just the suffix, and so this change is fine.

alexandreMartinEcl:
Good morning @JuroOravec 
Glad you appreciate the suggestion

For a minute you made my doubt about if my changes were making it clear enough for developers that `get_route_path` result would still be prefixed.
But given that `get_component_url` is still aware of this and is recommended, I guess it should be alright

JuroOravec (approved):

JuroOravec:
Ok, anyway, looking good!

I tried to update the PR with updated docs, docstrings and changelog, but I wasn't able to push it to this PR, even though it says "Maintainers are allowed to edit this pull request" on the side. Not sure what's up with that. So I'll make a follow up PR with those additional changes.

PR #1529, opened 2026-01-15
refactor: Move used github variables to be defined in env section

JuroOravec:
Follow up for https://github.com/django-components/django-components/pull/1511.

This applies the same thing - accessing github metadata in the `env` section, instead of directly in the script - but for the remaining scripts, so things are consistent.


JuroOravec:
cc @EmilStenstrom

EmilStenstrom (approved):

PR #1536, opened 2026-01-20
feat: use Rust-based tag parser (djc-core)

JuroOravec:
This replaces the Python tag parsing logic with Rust implementation from `djc-core` (https://github.com/django-components/djc-core/pull/14). As can be seen in the [benchmarks](https://github.com/django-components/django-components/pull/1536#issuecomment-3774301104), this makes `django-components` ~17% faster 🎉 

**NOTE: As I mentioned also in https://github.com/django-components/djc-core/pull/14, this is a LEGACY implementation, meaning that in Citry this will be simplified.**

### DIFF
--- a/src/django_components/util/template_tag.py
+++ b/src/django_components/util/template_tag.py
@@ -9,6 +9,7 @@ from typing import (
     Any,
     Callable,
     Dict,
+    Iterable,
     List,
     Mapping,
     NamedTuple,
@@ -22,7 +23,7 @@ from django.template import NodeList, Variable, VariableDoesNotExist
 from django.template.base import Parser, Token
 from django.template.exceptions import TemplateSyntaxError
 from djc_core.safe_eval import safe_eval
-from djc_core.template_parser import ParserConfig, TagAttr, compile_tag, parse_tag
+from djc_core.template_parser import GenericTag, ParserConfig, TagAttr, compile_tag, parse_tag
 from djc_core.template_parser.parse import TagUnion
 
 from django_components.expression import DynamicFilterExpression
@@ -61,11 +62,12 @@ def parse_template_tag(
     # Otherwise, depending on the end_tag, the tag may be:
     # 2. Block tag - With corresponding end tag, e.g. `{% slot %}...{% endslot %}`
     # 3. Inlined tag - Without the end tag, e.g. `{% html_attrs ... %}`
-    is_inline = parsed_tag_info.is_self_closing
+    is_inline = isinstance(parsed_tag_info, GenericTag) and parsed_tag_info.is_self_closing
 
     tag_config = tag_parser_config.get_tag(tag)
     tag_allowed_flags = (tag_config and tag_config.get_flags()) or set()
-    remaining_attrs, flags = _extract_flags(parsed_tag_info.attrs, tag_allowed_flags)
+    attrs = parsed_tag_info.attrs if isinstance(parsed_tag_info, GenericTag) else ()
+    remaining_attrs, flags = _extract_flags(attrs, tag_allowed_flags)
 
     def _parse_tag_body(parser: Parser, end_tag: str, inline: bool) -> Tuple[NodeList, Optional[str]]:
         if inline:
@@ -133,7 +135,7 @@ def _extract_contents_until(parser: Parser, until_blocks: List[str]) -> str:
 
 
 def _extract_flags(
-    attrs: List[TagAttr],
+    attrs: Iterable[TagAttr],
     allowed_flags: Set[str],
 ) -> Tuple[List[TagAttr], Dict[str, bool]]:
     found_flags: Set[str] = set()
@@ -156,8 +158,8 @@ def _extract_flags(
 
 def resolve_template_string(
     context: Mapping[str, Any],
-    source: str,
-    token: Tuple[int, int],
+    _source: str,
+    _token: Tuple[int, int],
     filters: Mapping[str, Callable],
     tags: Mapping[str, Callable],
     expr: str,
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -31,7 +31,7 @@ classifiers = [
 ]
 dependencies = [
     'Django>=4.2',
-    'djc-core-html-parser>=1.0.3',
+    'djc-core>=1.2.2',
```
JuroOravec, line 34: The Rust package was renamed to `djc-core`

src/django_components/node.py
```diff
@@ -56,9 +60,23 @@ def __new__(
...
+        # we'll be able to pass this metadata to the parser.
+        allowed_flags = attrs.get("allowed_flags")
+        if allowed_flags:
+            if tag_name in allowed_flags:
+                raise ValueError(
+                    f"Node {name}'s `tag` attribute ('{tag_name}') cannot be the same as one of its `allowed_flags`."
+                )
+
+            tag_config = TagConfig(tag=TagSpec(tag_name, set(allowed_flags)), sections=[])
+            parser_config.set_tag(tag_config)
```
JuroOravec, line 78: The way the Rust-based tag parser works is that it receives a config object that defines which tags can have which flags.

This is how we can define that e.g.:
- `{% component %}` and its `only` flag
- `{% slot %}` and its `required` and `default` flags

This config contains info on ALL template tags created by us / our users.

So there is a single global `parser_config`. And whenever a new Node class is created (e.g. `ComponentNode`, `SlotNode`, etc), then its tag name and allowed flags are added to the `parser_config`.

src/django_components/node.py
```diff
@@ -88,11 +106,21 @@ def __new__(
...
         cls._signature = validation_signature
 
+        # This runs when the node's template tag is being rendered.
         @functools.wraps(orig_render)
         def wrapper_render(self: "BaseNode", context: Context) -> str:
             trace_node_msg("RENDER", self.tag, self.node_id)
 
-            resolved_params = resolve_params(self.tag, self.params, context)
+            if self._params_resolver is None:
+                self._params_resolver = compile_tag_params_resolver(
```
JuroOravec, line 115: When we want to render a Node or a component, the first time we do so we compile a function to convert the args / kwargs inside `{% ... %}` into Python objects.

On subsequent calls, we reuse this function (stored in `_params_resolver`)

src/django_components/util/tag_parser.py
```diff
@@ -1,914 +0,0 @@
-# ruff: noqa: S105
-"""
-Parser for Django template tags.
```
JuroOravec, line 3: This was the old implementation that we've replaced.

tests/test_expression.py
```diff
@@ -408,232 +409,6 @@ def get_template_data(self, args, kwargs, slots, context):
 
 
 class TestSpreadOperator:
-    @djc_test(parametrize=PARAMETRIZE_CONTEXT_BEHAVIOR)
-    def test_component(self, components_settings):
```
JuroOravec, line 412: These removed tests were duplicates of the same, just using a different tag name. Not needed.

tests/test_tag_parser.py
```diff
@@ -1,1505 +1,3374 @@
+"""
+This file is defined both in django-components and djc_core.template_parser to ensure compatibility.
```
JuroOravec, line 2: As it says here, this test file was copied from `djc-core`. See https://github.com/django-components/djc-core/pull/24

src/django_components/expression.py
```diff
@@ -40,20 +36,25 @@ class DynamicFilterExpression:
     Then we treat it as a regular template and pass it as string.
     """
 
-    def __init__(self, parser: Parser, expr_str: str) -> None:
-        if not is_dynamic_expression(expr_str):
```
JuroOravec, line 44: I could remove the checks that decide whether a string is a nested template (AKA "dynamic expression"). Because this is now decided by the Rust logic and the result is stored on the AST.

src/django_components/expression.py
```diff
@@ -103,161 +104,3 @@ def __init__(self, wrapped_node: Node) -> None:
...
-
-    if not isinstance(value, str) or not value or len(value) < MIN_EXPR_LEN:
-        return False
-
-    # Is not wrapped in quotes, or does not contain any tags
-    return bool(DYNAMIC_EXPR_RE.match(value))
-
-
-# TODO - Move this out into a plugin?
-def process_aggregate_kwargs(params: List["TagParam"]) -> List["TagParam"]:
```
JuroOravec, line 147: `is_aggregate_key`, `process_aggregate_kwargs` and `_check_kwargs_for_agg_conflict` were all moved [here](https://github.com/django-components/django-components/pull/1536/files#diff-56d60a8b1d11d8ff86b912ddde6d8ee539f71a1fae917bc39a1fd049ea22705cR312)

src/django_components/util/template_tag.py
```diff
@@ -206,80 +135,342 @@ def _extract_contents_until(parser: Parser, until_blocks: List[str]) -> str:
...
+
+    expr_resolver = python_expression_cache[code]
+    return expr_resolver(context)
+
+
+class CompiledTagFn(Protocol):
+    def __call__(self, context: Mapping[str, Any]) -> Tuple[List[Any], List[Tuple[str, Any]]]: ...
+
+
+def compile_tag_params_resolver(
```
JuroOravec, line 246: This file contains too many changes to succintly summarize it. But this part is notable.

So how the compilation with `compile_tag()` works is that:
1. It takes the AST of a parsed tag (e.g. `{% component ... %}`
2. It also takes "resolvers". So functions that will decide how to process a value when it's a Python expression, or when it's a Django variable or filter, etc.

This way, the Rust code does all the heavy lifting, and here we only decide how to transform e.g. variable names to their values.

But `compile_tag()` does NOT return a value. It returns a FUNCTION.

This function then takes a CONTEXT to return the actual results.

To summarize:
- For variables and translations, we use Django's `Variable` class
- For filters, we look up a filter from a dict by the name and call the filter function
- For nested templates, we use our `DynamicFilterExpression` class.
- For Python expression, we use `safe_eval()` from `djc-core` to generate a function that we can call to evaluate the expression.

JuroOravec:
@EmilStenstrom Tagging for visibility. Also this makes `django-components` ~17% faster 🎉

EmilStenstrom:
@JuroOravec Awesome addition!

PR #1537, opened 2026-01-20
refactor: changelog, linting, type fixes

JuroOravec:
Cleanup after https://github.com/django-components/django-components/pull/1536


.github/workflows/tests.yml
```diff
@@ -3,8 +3,8 @@ name: Run tests
 on:
   push:
     branches:
-      - 'master'
```
JuroOravec, line 6: Just whitespace or single/double quote changes

PR #1538, opened 2026-01-21
refactor: Remove component input pre-validation + print template position when {% component %} errors

JuroOravec:
## Background

Improve performance by skipping input validation. This comes back to https://github.com/django-components/django-components/issues/935

When there is a template tag:

```django
{% component "table" 123 key=val / %}
```

Then behind the scene we transform that a function call kind of like this:

```py
ComponentNode.render(context, "table", 123, key=val)
```

So you see that:
1. the inputs to the `{% component %}` tag are treated as function arguments
2. We insert the `context` argument at the front.

Well, the issue was that some tags, like `{% slot %}` or `{% provide %}` have only specific args / kwargs they accept.
If they receive more or less than they need, they raise an error, same way a Python function raises an error when it gets wrong input.

**But because we also pass in the `context` argument, which is NOT visible to the user, the error message could be cryptic.**

In the past I addressed this by validating the inputs BEFORE passing the args / kwargs to the Node (`BaseNode.render()`).

But as mentioned in https://github.com/django-components/django-components/issues/935, this had a big performance penalty.

## Changes

1. So this PR finally comes back to this issue. Instead of validating the inputs *BEFORE* calling `BaseNode.render()`, we instead "ask for forgiveness" - We don't do validation beforehand, but if `BaseNode.render()` throws an error, we modify the error message to show the correct arguments counts.

2. Next I went a step further and added similar logic to what I implemented when working on the Rust-based template parser.

    So now, when you call `{% component %}`, `{% slot %}`, etc, and it raises an error, you will be shown also the position of the errored tag within the template:

    Screenshot 2026-01-21 at 12 21 31

Closes https://github.com/django-components/django-components/issues/935

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,14 @@
     {% mytag 'John' x=123 x=456 %}
     ```
 
+- All template tags (`{% component %}`, `{% slot %}`, etc.) now include the exact tag (as found in the template) in the error message when an error occurs:
+
+    ```py
+    TypeError: Error in mytag: missing 1 required keyword-only arguments: 'msg'
+        1 | {% mytag 'John' %}
+            ^^^^^^^^^^^^^^^^^^
+    ```
+
 ## v0.144.0
 
 #### Feat


src/django_components/node.py
```diff
@@ -121,95 +125,31 @@ def wrapper_render(self: "BaseNode", context: Context) -> str:
                 )
 
             args, kwargs = self._params_resolver(context)
-
-            # Template tags may accept kwargs that are not valid Python identifiers, e.g.
```
JuroOravec, line 125: The entire removed block was just about validating the tag inputs before calling `BaseNode.render()`. Now, instead, we don't validate before, but post-process the error msg.

src/django_components/node.py
```diff
@@ -725,3 +665,226 @@ def decorator(fn: Callable) -> Callable:
...
+# {% component "my_comp" name="John" age=20 %}
+# ```
+#
+# Then we call the `render()` method of the `BaseNode` class, which is the actual implementation of the template tag.
+#
+# But we hide from the user the fact that we pass `context` as the first argument to the `render()` method.
+#
+# Note: Python doesn't count `self` in error messages (it's passed automatically), so we only need
+# to subtract 1 for `context`, not 2 for `self` and `context`.
+def _modify_typeerror_message(error: TypeError, current_file: str) -> None:
```
JuroOravec, line 694: One shortcoming of modifying the error after it has already happened is that we need to be careful to distinguish between:
1. Error raise directly by calling `BaseNode.render()` - e.g. too few or too many args.
2. Error coming from WITHIN the `render()` method.

Read the comments above. Internally we pass also `context` argument to `render()`, e.g.
```py
BaseNode.render(context, *args, **kwargs)
```

But externally, when user defines a `{% component %}` or other node in the template, they only see their own arguments:
```django
{% component "table" data=data / %}
```
So this `_modify_typeerror_message` function modifies the resulting `TypeError`, so that the error messages like "takes 2 arguments but received 5" are corrected to match the number of arguments that user provides from within the template.

src/django_components/node.py
```diff
@@ -725,3 +665,226 @@ def decorator(fn: Callable) -> Callable:
...
+    """Modify TypeError messages in place to omit the `context` argument."""
+    error_msg = str(error)
+
+    # Try to figure out if the error occurred when calling the `BaseNode.render()` method
+    # (the one to which we supplied the `self` and `context`), or in a nested function,
+    # by inspecting the traceback.
+    # If error occured in a nested function, we do NOT modify this error message.
+    tb = error.__traceback__
+    should_modify = True
+    if tb is not None:
```
JuroOravec, line 704: To try to guess if the error was raised by the `BaseNode.render()` itself, here we check the traceback of the raised Error.

src/django_components/node.py
```diff
@@ -725,3 +665,226 @@ def decorator(fn: Callable) -> Callable:
...
+    - The lines containing the error (start_index to end_index)
+    - Up to 2 following lines
+    - Underlined code with ^^^ characters
+
+    **Example:**
+
+    ```
+    TypeError: Error in mytag3: missing 1 required keyword-only argument: 'msg' and 'mode'
+
+        1 | {% mytag3 'John' %}
```
JuroOravec, line 791: This here is the function that prints out the template tag (as found in the template) and appends it to the error when `BaseNode.render()` throws an error.

Because `BaseNode.render()` is the logic shared by all our template tags, that means that now all - `{% component %}`, `{% slot %}`, etc - will print out their template position when their throw an error.

src/django_components/util/template_tag.py
```diff
@@ -433,235 +432,6 @@ def is_aggregate_key(key: str) -> bool:
     )
 
 
-# For details see https://github.com/django-components/django-components/pull/902#discussion_r1913611633
-# and following comments
-def validate_params(
```
JuroOravec, line 438: Removed the pre-validation logic. No longer needed

tests/test_expression.py
```diff
@@ -410,7 +410,7 @@ def get_template_data(self, args, kwargs, slots, context):
 
 class TestSpreadOperator:
     @djc_test(parametrize=PARAMETRIZE_CONTEXT_BEHAVIOR)
-    def test_later_spreads_do_not_overwrite_earlier(self, components_settings):
+    def test_later_spreads_can_overwrite_earlier(self, components_settings):
```
JuroOravec, line 413: This test previously tested that when there was a spread AND a literal kwarg both with the same key, that it raised an error.

With the removal of the pre-validation, this no longer happens, which broke this test.

However, in the docs we already say that you can have the same key multiple times, and that only the last instance counts.

So I updated this to match what we say in the docs.

EmilStenstrom:
Performance! Woop woop!

PR #1541, opened 2026-01-21
fix: do not recursively transform dataclasses to dict

JuroOravec:
This PR fixes the following bug:

There was bug where, when you defined `Kwargs`, `Args`, `Slots`, etc, as `@dataclass`,
  and one of its fields was a dataclass object again, the nested dataclass was incorrectly
  converted to a dictionary, instead of being left as a dataclass instance.

```py
from dataclasses import dataclass
from django_components.util.misc import to_dict

@dataclass
class User:
    name: str

class MyTable(Component):
    @dataclass
    class Kwargs:
        user: User
        count: int

    def get_template_data(self, args, kwargs: Kwargs, slots, context):
        # INCORRECT!!! Should be `User(name="John")`
        assert kwargs.user == { "name": "John" }
```

### DIFF
--- a/sampleproject/examples/urls.py
+++ b/sampleproject/examples/urls.py
@@ -49,7 +49,7 @@ def get_example_urls():
             urlpatterns.append(path(url_pattern, view_class.as_view(), name=view_name))
             print(f"Registered URL: {url_pattern} -> {view_class.__name__}")
 
-        except Exception as e:  # noqa: BLE001
+        except Exception as e:
             print(f"Failed to register URL for {example_name}: {e}")
             raise
 
--- a/src/django_components/util/misc.py
+++ b/src/django_components/util/misc.py
@@ -1,7 +1,7 @@
 import re
 import sys
 from collections import namedtuple
-from dataclasses import asdict, fields, is_dataclass
+from dataclasses import fields, is_dataclass
 from hashlib import md5
 from importlib import import_module
 from inspect import getmembers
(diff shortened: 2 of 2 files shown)


src/django_components/util/misc.py
```diff
@@ -163,7 +163,10 @@ def to_dict(data: Any) -> dict:
     if hasattr(data, "_asdict"):  # Case: NamedTuple
         return data._asdict()
     if is_dataclass(data):  # Case: dataclass
-        return asdict(data)  # type: ignore[arg-type]
+        # NOTE: This is same as `asdict`, but without recursing into nested dataclasses
+        return {
+            f.name: getattr(data, f.name) for f in fields(data)
```
JuroOravec: This is the fix. As it turned out, `dataclasses.asdict` actually recursively transforms dataclasses to dicts

sampleproject/examples/urls.py
```diff
@@ -34,7 +34,8 @@ def get_example_urls():
             view_class = None
             for attr_name in dir(module):
                 attr = getattr(module, attr_name)
-                if issubclass(attr, Component) and attr_name != "Component" and attr_name.endswith("Page"):
+                is_component_sublass = isinstance(attr, type) and issubclass(attr, Component)
+                if is_component_sublass and attr_name != "Component" and attr_name.endswith("Page"):
```
JuroOravec, line 38: Unrelated to the bug, this was meant to go with the next PR. There was a bug with the logic that generates the examples in `/examples` in the local `sampleproject`.

PR #1545, opened 2026-01-22
tests: run E2E tests also against firefox and webkit

JuroOravec:
https://github.com/django-components/django-components/issues/1544 exposed that different browsers may behave differently, and that Firefox is currently broken.

This PR updates E2E (Playwright) tests to run also against Firefox and Webkit (Safari). Previously, we tested only against Chromium (Chrome, Edge).

The tests for this PR are expected to fail, as we finally capture the error found in https://github.com/django-components/django-components/issues/1544:

(Notice that all the failing tests end with `[firefox]`)

Screenshot 2026-01-22 at 17 08 34


JuroOravec:
To reiterate, the CI tests are failing, but this is expected. See [this job](https://github.com/django-components/django-components/actions/runs/21255949516/job/61170399058?pr=1545), which fails on the Firefox-specific tests:

Screenshot 2026-01-22 at 17 25 55

PR #1546, opened 2026-01-22
refactor: rename client-side global `Components` to `DjangoComponents` to avoid conflict on Firefox

JuroOravec:
Closes https://github.com/django-components/django-components/issues/1544

Beside just the tests, I also tried running the demo project locally on Firefox and checking the example that loads HTML fragments with AlpineJS / HTMX / vanilla. And it worked, too.

### DIFF
--- a/tests/test_dependencies.py
+++ b/tests/test_dependencies.py
@@ -369,7 +369,7 @@ class TestRenderDependencies:
             <thead>
             <script>
                 (() => {
-                    if (!globalThis.Components) {
+                    if (!globalThis.DjangoComponents) {
                         const s = document.createElement('script');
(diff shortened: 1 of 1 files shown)


tests/test_dependency_manager.py
```diff
@@ -426,7 +444,7 @@ async def test_raises_if_input_hash_not_registered(self, browser_name):
 
         with pytest.raises(
             Error,
-            match=re.escape("Error: [Components] 'my_comp': Cannot find input for hash 'input-abc'"),
```
JuroOravec, line 429: Interestingly, Playwright formats the error (from running a JS-side code) diffently on Firefox and Chromium.

PR #1549, opened 2026-01-23
refactor: drop support for Python 3.8 and 3.9, and update types

JuroOravec:
Part of https://github.com/django-components/django-components/issues/1514 (The other part will be dropping Django 5.1 and adding Django 6.0).

This PR drops support for Python 3.8 and 3.9 and updates the codebase:

- Updated types:
  - `Optional[X]` -> `X | None`
  - `Union[X, Y]` -> `X | Y`
  - `List[X]` -> `list[X]`
  - `Dict[X]` -> `dict[X]`
  - `Set[X]` -> `set[X]`
  - `Tuple[X]` -> `tuple[X]`
  - `Type[X]` -> `type[X]`
- Added `X: TypeAlias` annotations to variables that defined types
  - E.g. `MyType: TypeAlias = X | Y | Z`
- Removed a lot of branching logic that checked for >=3.10.
  - E.g. when using weak refs, the typing was available only for >=3.10.

I've also updated the `supported_versions.py` script - the script that decides which Django / Python version we support - to NOT support unsupported Python versions. Because our strategy until now was to support all Python types that Django does.


docs/community/development.md
```diff
@@ -370,6 +358,13 @@ Head over to [Dev guides](./devguides/dependency_mgmt.md) for a deep dive into h
 
 The `scripts/supported_versions.py` script manages the supported Python and Django versions for the project.
 
+The script determines supported versions by:
```
JuroOravec, line 361: Mentioned the new behaviour of `scripts/supported_versions.py` in the docs.

pyproject.toml
```diff
@@ -21,19 +21,13 @@ classifiers = [
...
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.8",
-    "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
     "Programming Language :: Python :: 3.14",
 ]
-dependencies = [
```
JuroOravec, line 32: The rest of the changes in this file (this line including) are due to formatting.

requirements-ci-python38.txt
JuroOravec, line 1: Removed the `requirements.txt` file that was used specifically for Python3.8

tox.ini
```diff
@@ -51,25 +46,6 @@ deps =
   django-template-partials
 commands = pytest {posargs}
 
-# TODO: Delete this when Django 4.2 reaches end of life
-[testenv:py38-django42]
```
JuroOravec, line 55: Also here removed the 3.8-specific dependencies

JuroOravec:
It is a lot of changes. I went through them twice. I was mainly using AI tab completion for the typing changes (so I saw exactly what has changed). But I noticed that there were maybe 3 instances where the AI autocomplete was incorrect, so I had to double check.

The error that AI tab completion was sometimes introducing is that it converted typed fields like:

```py
class Data:
    X: Optional[Y]
```

into:

```py
class Data:
    X: Y | None = None
```

There's a subtle difference - the latter field has a default value set.

JuroOravec:
I've also updated [the `master` branch rules](https://github.com/django-components/django-components/settings/rules/4279617), so that CI test jobs for Python 3.8 and 3.9 (e.g.  `"Run tests / build (3.8, windows-latest)"`) are not required.

Screenshot 2026-01-23 at 16 27 08

JuroOravec:
cc @EmilStenstrom tagging for visibility

PR #1550, opened 2026-01-23
refactor: reuse browsers across playwright tests for faster tests

JuroOravec:
After I had updated Playwright tests to run also against Firefox and Webkit ([in addition to Chromium](https://github.com/django-components/django-components/pull/1545)), now the CI test suite takes up to 20 mins ([see here](https://github.com/django-components/django-components/actions/runs/21290986762/job/61292365767)).

This PR should speed it up by reusing the browser instances. Before, the browser was re-opened and then closed before/after each test.


JuroOravec:
Ok so looks like it went down from 19 min to 12 min ([see after](https://github.com/django-components/django-components/actions/runs/21295018640/job/61299522808?pr=1550))

PR #1551, opened 2026-01-23
refactor: drop support for Django 5.1 and add support for Django 6.0

JuroOravec:
As the title says. See the Django versions lifecycle here https://endoflife.date/django

Closes https://github.com/django-components/django-components/issues/1514


JuroOravec:
cc @EmilStenstrom

Issue #1395, opened 2025-09-22
Use uv for package management


JuroOravec:
After replacing `flake8` with `ruff`, I was also thinking of updating other parts of the toolbelt. I was intrigued by `uv`, as it's popular right now.

Right now we use pip + `requirements.txt` files. We update these `requirements.txt` files in two ways:
- manually
- with pip-compile from [pip-tools](https://pypi.org/project/pip-tools/) (not documented in your docs)

After looking into the documentation, it felt like it would bring a bit more overhead compared to maintaining just plain requirements.txt - for contributors who haven't used `uv` yet, you'll have to read up some more documentation to get started. After getting familiar with `uv`, it's about the same overhead as pip-compile.

But there are some clear benefits to `uv` too:
- Installation and virtual env setup should be [3-10x faster](https://www.reddit.com/r/programming/comments/1dpr61b/never_thought_id_be_saying_this_but_switching/). We should install also [`tox-uv`](https://github.com/tox-dev/tox-uv), so we can see the speedup in environments managed in tox, and thus in CI tests too. Based on the advertized speeds, CI test jobs could be overall 60% faster.

   <img width="1164" height="479" alt="Image" src="https://github.com/user-attachments/assets/53b2b923-1541-415e-9f5a-4e4a9f288cb0" />

- Another thing I found useful is that `uv` supports monorepo structure with [workspaces](https://docs.astral.sh/uv/concepts/projects/workspaces/). I don't think we should migrate to workspaces right now, but it's nice to have a tool that *can* do so in the future.

TODO:
- [ ] Install `uv`
- [ ] Move the different environment dependencies (dev, docs, prod) from `requirements.txt` and `requirements.in` to `uv`
- [ ] Update documentation on development to talk about using `uv` for installing dependencies.
- [ ] Add section to development.md about how to update dependencies.
- [ ] Install `tox-uv` and configure `tox` to use it

JuroOravec:
While we're at it, we could also move the current `tox.ini` inside `pyproject.toml`, so there's fewer config files lying around. See https://tox.wiki/en/stable/config.html.

(And unrelated, but maybe also remove [`.pre-commit-config.yaml`](https://github.com/django-components/django-components/blob/master/.pre-commit-config.yaml)? It doesn't seem to be used.)

dalito:
[...]

PR #1552, opened 2026-01-23
refactor: replace requirements.txt files with uv

JuroOravec:
Closes https://github.com/django-components/django-components/issues/1395

Unlike what I originally suggested in #1395, I've kept tox config in `tox.ini`. No harm in that, and also don't want to bloat `pyproject.toml` too much. Right now it already contains package metadata, dependencies for all environments, and config for linters/typecheckers.

I also kept the demo `sampleproject` as is, still using `requirements.txt`. My reasoning is that the demo project should have minimal barrier to entry. Forcing people to install `uv` just to check out the project is needless overhead.

I tried going through the flow locally. But I want to test it also on this PR, opening a github codespace, to ensure clean slate.

One interesting observation about using `uv` is that when you switch between environments (e.g. `uv sync --group dev` followed by `uv sync --group docs`), then it deletes the previously-installed dependencies and keeps only what the last group uses.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,6 +6,7 @@
 requires =
   tox>=4
   tox-uv>=1  # Use uv for dependency management
+  tox-gh-actions>=3  # Run only those envs as set in [gh-actions]
 
 envlist =
   py{310,311}-django{42,52}


.github/workflows/tests.yml
```diff
@@ -55,18 +58,19 @@ jobs:
...
         with:
           python-version: ${{ matrix.python-version }}
-          cache: "pip"
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
-          python -m pip install -r requirements-docs.txt
-          # Install your package locally
-          python -m pip install -e .
```
JuroOravec, line 69: Also a nice thing is that, supposedly, `uv` also installs the local package automatically.

pyproject.toml
```diff
@@ -38,9 +39,85 @@ Changelog = "https://django-components.github.io/django-components/latest/releas
...
 
+###########################################
+# ENVIRONMENT DEPENDENCIES
+###########################################
+
+[dependency-groups]
+
+# Dev dependencies - install with `uv sync --group dev`
+dev = [
+    "django",
```
JuroOravec, line 50: Below are dependencies that were previously in `requirements-dev.txt`, `requirements-docs.txt` and `requirements-ci.txt`

JuroOravec:
Ok I just checked starting a new project and installing uv / dependencies from scratch, and everything still works :)

JuroOravec:
cc @EmilStenstrom This replaces the `requirements.txt` files and introduces `uv` 🎉

uv.lock
dalito, line 1: Do you really want to put the lock file in git? Typically it is not done for a library.

JuroOravec: Hm, why not? Not familiar with best practices around uv / lock files in Python, but in JS and NPM the `package-lock.json` still makes sense - the dependencies set in `package.json` / `pyproject.toml` may be vague, whereas the lock file says "this is the exact combination of dependencies that work".

Imagine following:
1. I have `"Django>=4.2"` in `pyproject.toml`. Everything works with Django `4.2.0`.
2. Django releases `4.2.1`, but it's broken.
3. If I don't use the lock file, and I install `django-components` locally, it will automatically download `4.2.1` because it's the latest that fits the `>=4.2` constraint. Because `4.2.1` is broken, I will now have a broken project that worked yesterday, and I don't know what changed.
4. OTOH, if I DO use a lock file, when I install project dependencies, it will still install `4.2.0`. And my local project will keep working until I explicitly bump the lock file.

Side note, but if you are concerned that the lock file would be included in the built wheel, I just checked and it isn't.

dalito: With the lock file you will test in CI exactly the locked set of dependencies whereas you maybe want to test always the up-to-date set of dependencies for a library.

JuroOravec: We do have dependabot for that already:)

PR #1554, opened 2026-01-24
refactor: move lint and coverage to own jobs to speed up CI

JuroOravec:
Speed up CI tests by running linting and coverage separately, ~and caching playwright browsers~.

Tests themselves take about 3 min, which is not bad.

However, one suite of tests can still take up to 10 minutes, because:
1. Windows runners spend ~4 mins downloading browsers for E2E tests. ~3 mins just waiting before they start downloading, and then ~1 min to actually download the browsers. (maybe they are downloading the entire time?).
2. These same runners then take ~3 mins to run the tests. (~1.5 min per Python-Django combination)
3. The latest Python version then also runs coverage (another 1.5-2 mins) and linting, but only if all other check have passed.

So this is an attempt to break it down and speed it up by:
1. ~Caching browser binaries (potentially unblocking 3-4 min)~
2. Breaking off coverage and linting into standalone jobs, so they can ran in parallel with the other tests.

If this works, most of the jobs should be finished in ~3-4 mins.

---

**UPDATE: SEE https://github.com/django-components/django-components/pull/1554#issuecomment-3794644942**

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,32 +43,13 @@ jobs:
         run: |
           uv sync --group ci
 
-      # Cache Playwright browsers (~500MB) to speed up CI, because it takes ~3 mins to download them on Windows.
-      # Cache persists across all jobs and workflow runs until the cache key changes
-      # Cache key includes OS and uv.lock hash, so it invalidates when dependencies change
+      # NOTE: WINDOWS ONLY - Installing system dependencies for Playwright browsers takes ~3-4 mins.
+      # See https://github.com/django-components/django-components/pull/1554
       # See https://playwright.dev/docs/browsers#managing-browser-binaries
-      - name: Cache Playwright browsers
-        uses: actions/cache@v4
-        id: playwright-cache
-        with:
-          path: |
-            ~/.cache/ms-playwright
-            C:\Users\runneradmin\AppData\Local\ms-playwright
-          key: playwright-${{ runner.os }}-${{ hashFiles('uv.lock') }}
-          restore-keys: |
-            playwright-${{ runner.os }}-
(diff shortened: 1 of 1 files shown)


JuroOravec:
I've also added the new split-off jobs - `lint` and `coverage` - to be required to be passed in order for PR for `master` to pass.

This can be configured here - https://github.com/django-components/django-components/settings/rules/4279617

Screenshot 2026-01-24 at 13 36 33
Screenshot 2026-01-24 at 13 35 18

JuroOravec:
Ok, the caching of browsers was working, but unfortunately that was NOT the bottleneck.

Turns out that the bottleneck on Windows is installing the system dependencies, which takes ~3.5 min.

Downloading browsers is the fast part.

So to speed it up, I guess we'd have to have a testing container with the playwright browsers and dependendencies pre-installed?

Not worth the hassle, but I might be wrong.

---

And for future reference (because this was the first time I worked with Github cache action), below is the showcase that the implemented cache worked ([see code here](https://github.com/django-components/django-components/pull/1554/commits)). But for now, I'm removing the caching logic from the PR.

Ubuntu:
- Before (no cache) - https://github.com/django-components/django-components/actions/runs/21315096385/job/61356722685?pr=1554
- After (reuse cached) - https://github.com/django-components/django-components/actions/runs/21315096385?pr=1554

Windows:
- Before (no cache) - https://github.com/django-components/django-components/actions/runs/21315631691/job/61357942636?pr=1554
- After - https://github.com/django-components/django-components/actions/runs/21315631691/job/61358303268?pr=1554

JuroOravec:
cc @EmilStenstrom This splits linting and coverage tests into own jobs to speed up the CI checks.

PR #1555, opened 2026-01-24
refactor: re-install docs deps to fix docs job

JuroOravec:
After moving project dependencies from `requirements.txt` to uv + lock file (see https://github.com/django-components/django-components/pull/1552), the publishing of docs is broken ([see broken job](https://github.com/django-components/django-components/actions/runs/21311905713/job/61349337603))

The error says `error: Failed to spawn: mike`, which hints that the dependencies are not installed by the time we call `mike`. This should fix it.

### DIFF
--- a/.github/workflows/release-docs.yml
+++ b/.github/workflows/release-docs.yml
@@ -4,10 +4,10 @@ on:
   push:
     tags:
       # for versions 0.### (before 1.0.0)
-      - '0.[0-9]+'
+      - "0.[0-9]+"
       # after 1.0.0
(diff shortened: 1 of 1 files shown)


.github/workflows/release-docs.yml
```diff
@@ -164,6 +163,11 @@ jobs:
           git config user.name github-actions
           git config user.email github-actions@github.com
 
+      - name: Install dependencies
```
JuroOravec, line 166: This is the actual change, the rest is formatting.

PR #1556, opened 2026-01-24
refactor: fix people script and job, and update to use uv

JuroOravec:
The job and the script that populates the the "People page" (that lists all contributors) were broken. See https://github.com/django-components/django-components/actions/runs/20643241826/job/59277751806

This PR fixes it. I've tested the script locally and included the updated statistics in this PR.

### DIFF
--- a/docs/scripts/people.py
+++ b/docs/scripts/people.py
@@ -12,7 +12,7 @@ from datetime import datetime
 from pathlib import Path
 from typing import Any
 
-import httpx
+import requests
 import yaml  # type: ignore[import-untyped]
 from github import Github
 from pydantic import BaseModel, SecretStr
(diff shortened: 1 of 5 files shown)


docs/scripts/people.py
```diff
@@ -31,6 +31,9 @@
     "copilot-swe-agent",
 }
 
+# NOTE: Relative to project root
+PEOPLE_PAGE_PATH = Path("docs/community/people.yml")
```
JuroOravec, line 35: The script was broken because it was using a path relative to this `people.py` file, and either this script or the target file moved in the meantime. I changed it to be relative to the root dir, so there's one less moving part.

docs/scripts/people.py
```diff
@@ -12,7 +12,7 @@
 from pathlib import Path
 from typing import Any
 
-import httpx
+import requests
```
JuroOravec, line 15: Also replaced `httpx` with `requests`. We now install all the dev deps to run this script. And among the dev deps there is already `requests`, so no need to keep both.

pyproject.toml
```diff
@@ -71,6 +72,8 @@ dev = [
     "virtualenv==20.35.4",
     "typing-extensions>=4.12.2",
     "pathspec",
+    "PyGithub",
+    "pyyaml",
```
JuroOravec, line 76: These are dev dependdencies used solely by the `people.py` script

PR #1558, opened 2026-01-25
docs: fix links errors

JuroOravec:
There was ~1000 links which weren't broken, but they were emitting warnings from `mkdocs`.

The warnings were caused because the relative links were defined as relative to the final URL. But `mkdocs` prefers the relative links to be written as relative to the filesystem.

So there were changes like:
```
../../concepts/fundamentals/slots -> ../concepts/fundamentals/slots.md
```

Used LLMs to process the errors to fix the links, then went through them myself and clicked on them to confirm they are working.

The code changes are commented below. The rest is just documentation links fixes

### DIFF
--- a/src/django_components/extensions/debug_highlight.py
+++ b/src/django_components/extensions/debug_highlight.py
@@ -98,7 +98,7 @@ class ComponentDebugHighlight(ExtensionComponentConfig):
         },
     )
     ```
-    """  # noqa: E501
+    """
 
     # TODO_v1 - Remove `DEBUG_HIGHLIGHT_COMPONENTS` and `DEBUG_HIGHLIGHT_SLOTS`
     #           Instead set this as plain boolean fields.


docs/scripts/reference.py
```diff
@@ -1092,35 +1095,40 @@ def _format_command_args(cmd_parser: ArgumentParser, cmd_path: Sequence[str] | N
...
-    return inspect.isclass(obj) and getattr(obj, "_extension_command_api", False)
+    return _is_class(obj) and getattr(obj, "_extension_command_api", False)
 
 
 def _is_extension_url_api(obj: Any) -> bool:
-    return inspect.isclass(obj) and getattr(obj, "_extension_url_api", False)
+    return _is_class(obj) and getattr(obj, "_extension_url_api", False)
+
+
+def _is_class(obj: Any) -> bool:
```
JuroOravec, line 1129: One annoying thing is that when this script fails, `mkdocs` doesn't loudly announce that. So if found that this script wasn't working, because when you pass `Generator` type alias to `issubclass()`, it throws an error.

So this `_is_class()` filters out generic aliases.

scripts/validate_links.py
```diff
@@ -160,7 +160,7 @@ def load_gitignore(root: Path) -> pathspec.PathSpec:
             patterns = f.read().splitlines()
     # Add additional ignored paths
     patterns += IGNORED_PATHS
-    return pathspec.PathSpec.from_lines("gitwildmatch", patterns)
+    return pathspec.PathSpec.from_lines("gitignore", patterns)
```
JuroOravec, line 163: There was a warning that `gitwildmatch` is deprecated. Tho the warning remained even after this change, so it might be that mkdocs internally also uses `PathSpec`?

PR #1559, opened 2026-01-25
refactor: set links to be warnings in mkdocs

JuroOravec:
Just cleaning out my local branch.

Mkdocs has a "strict" mode for CI, where the CLI process errors when there are any warnings.

Unfortunately there's still about 15 errors remaining, and these are related to various non-standard things we do in the documentation. Also `mkdocs` doesn't seem to offer any way to silence/ignore individual errors.

So I'm not enabling the strict mode, but at least updating the rest of the config.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -12,6 +12,9 @@ dev_addr: "127.0.0.1:9000"
 site_dir: site
 docs_dir: docs
 
+# TODO - There's about 15 errors preventing strict mode from being enabled.
+# strict: true
+
 watch:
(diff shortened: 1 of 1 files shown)


mkdocs.yml
```diff
@@ -160,7 +166,7 @@ plugins:
   - mkdocstrings:
       handlers:
         python:
-          import:
+          inventories:
```
JuroOravec, line 169: Unrelated, but the `import` keyword was deprecated.

PR #1560, opened 2026-01-25
feat: add on_extension_created hook

JuroOravec:
Closes https://github.com/django-components/django-components/issues/1239

This introduces a new extension hook, and moves autodiscovery logic into a new extension that uses that hook.

Screenshot 2026-01-25 at 14 16 49

### DIFF
--- a/src/django_components/extension.py
+++ b/src/django_components/extension.py
@@ -1160,7 +1160,6 @@ class ExtensionManager:
 
         self.extensions = self._init_extensions(extensions)
         self._initialized = True
-        self.on_extension_created()
 
         # Populate the `urlpatterns` with URLs specified by the extensions
         # TODO_V3 - Django-specific logic - replace with hook
@@ -1200,6 +1199,9 @@ class ExtensionManager:
 
         # Rebuild URL resolver cache to be able to resolve the new routes by their names.
         self._lazy_populate_resolver()
+
+        # Only after Django URL resolver is populated, it's safe to call the `on_extension_created` hook.
+        self.on_extension_created()
 
         # Flush stored events
         #


src/django_components/__init__.py
```diff
@@ -40,6 +40,9 @@
     ExtensionComponentConfig,
     OnComponentRegisteredContext,
     OnComponentUnregisteredContext,
+    OnCssLoadedContext,
+    OnExtensionCreatedContext,
+    OnJsLoadedContext,
```
JuroOravec, line 45: Turns out I forgot to expose `OnCssLoadedContext` and `OnJsLoadedContext` before

src/django_components/app_settings.py
```diff
@@ -788,46 +785,33 @@ def _get_settings(self) -> ComponentsSettings:
...
-                DefaultsExtension,
-                DependenciesExtension,
-                ViewExtension,
-                DebugHighlightExtension,
-            ],
-        ) + list(extensions)
-
-        # Extensions may be passed in either as classes or import strings.
-        extension_instances: list[ComponentExtension] = []
-        for extension in extensions:
```
JuroOravec, line 817: This logic was moved to `extension.py`. This forloop creates instances for given extensions classes/import paths

src/django_components/apps.py
```diff
@@ -41,12 +42,6 @@ def ready(self) -> None:
         # NOTE: Delete when Django 5.2 reaches end of life
         monkeypatch_template_proxy_cls()
 
-        # Import modules set in `COMPONENTS.libraries` setting
-        import_libraries()
```
JuroOravec, line 45: These lines with `import_libraries` and `autodiscover` were moved to a separate extension.

JuroOravec:
cc @EmilStenstrom

PR #1561, opened 2026-01-25
refactor: fix error with multi-line strings

JuroOravec:
Closes https://github.com/django-components/django-components/issues/1255

### DIFF
--- a/tests/test_expression.py
+++ b/tests/test_expression.py
@@ -456,6 +456,7 @@ class TestTemplateExpression:
             """,
         )
 
+
 class TestSpreadOperator:
     @djc_test(parametrize=PARAMETRIZE_CONTEXT_BEHAVIOR)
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers = [
     "Programming Language :: Python :: 3.14",
 ]
 # Package dependencies
-dependencies = ["Django>=4.2", "djc-core>=1.2.2", "typing-extensions>=4.12.2"]
+dependencies = ["Django>=4.2", "djc-core>=1.3.0", "typing-extensions>=4.12.2"]
```
JuroOravec, line 31: The bug had to be fixed upstream in Rust code. While there I also dropped support for py 3.8 and 3.9 and updated to use `uv` in `djc-core`.

PR #1564, opened 2026-01-25
feat: add Component.parent, .root, and .ancestors to traverse the comp tree

JuroOravec:
Closes https://github.com/django-components/django-components/issues/1252

Add `Component.parent`, `Component.root` and `Component.ancestors` properties to be able to walk / access parent components.

- `Component.parent` - `None` if component doesn't have a parent.
- `Component.root` - Always defined. `self` if component doesn't have a parent.
- `Component.ancestors` - Iterator for walking up the component tree. Does not contain `self`.

Screenshot 2026-01-25 at 20 41 21

### DIFF
--- a/src/django_components/component.py
+++ b/src/django_components/component.py
@@ -3621,12 +3621,16 @@ class Component(metaclass=ComponentMeta):
 
         # Set parent and root as direct attributes on the component instance.
         # This creates strong references that keep parent/root alive as long as children are alive.
-        if parent_id is not None:
+        # NOTE: `parent_id` may be not found in `component_instance_cache` if we are rendering
+        #       an orphaned slot function (AKA slot function that we've taken out of the render context)
+        if parent_id is not None and parent_id in component_instance_cache:
             parent_component = component_instance_cache[parent_id]
         else:
             parent_component = None
 
-        if root_id is not None and root_id != render_id:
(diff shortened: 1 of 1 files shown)


src/django_components/component.py
```diff
@@ -3520,6 +3644,8 @@ def _render_impl(
             registry=registry,
             registered_name=registered_name,
             node=node,
+            parent=parent_component,
```
JuroOravec, line 3651: Parent and root instances are set directly to the `Component` instance. This way, as long as the child components exist, the parent ones will exist too and won't be garbage collected.

src/django_components/component.py
```diff
@@ -3558,22 +3684,6 @@ def _render_impl(
             {BLOCK_CONTEXT_KEY: context.render_context.get(BLOCK_CONTEXT_KEY, BlockContext())},  # type: ignore[union-attr]
         )
 
-        # We pass down the components the info about the component's parent.
-        # This is used for correctly resolving slot fills, correct rendering order,
-        # or CSS scoping.
-        parent_id, parent_comp_ctx = _get_parent_component_context(context)
```
JuroOravec, line 3564: This was just moved higher up

JuroOravec:
cc @EmilStenstrom

PR #1570, opened 2026-01-29
feat: JS variables + define E2E test endpoints in test files

JuroOravec:
## Related issues

- [X] https://github.com/django-components/django-components/discussions/1456
   - TODO - Update

- [X] Closes https://github.com/django-components/django-components/issues/622 (JS / CSS variables)
- [X] Part of https://github.com/django-components/django-components/issues/1567 (Co-localizing E2E endpoints in test files)
   - What remains is to update other E2E files to also use the `server()` pattern

- Part of https://github.com/django-components/django-components/issues/836
   - The only thing that remains is to:
     1. Wrap JS in a self-executing function to avoid polluting global scope
     2. Allow to export JS / CSS with `collectstatic`.

## Description

Pass data from Python to JavaScript with `Component.get_js_data()`.

`Component.get_js_data()` returns a dictionary. This data is automatically
serialized to JSON and made available to your component's JavaScript code.

In your JavaScript file, access these variables using the `$onComponent()` callback function.

JS variables are automatically scoped to each component instance. Different
instances of the same component can have different JS variable values.

**Example:**

```python
from django_components import Component, register

@register("product_card")
class ProductCard(Component):
    template_file = "product_card.html"
    js_file = "product_card.js"
    css_file = "product_card.css"

    def get_js_data(self, args, kwargs: Kwargs, slots, context):
        product = Product.objects.get(id=kwargs.product_id)
        return {
            "product_id": kwargs.product_id,
            "price": float(product.price),
            "api_endpoint": f"/api/products/{kwargs.product_id}/",
        }
```

```javascript
// product_card.js

// Access component JS variables in $onComponent callback
$onComponent(({ product_id, price, api_endpoint }, ctx) => {
  const containerEl = ctx.els[0];
  containerEl.querySelector(".add-to-cart")
    .addEventListener("click", () => {
      fetch(api_endpoint, {
        method: "POST",
        headers: { "Content-Type": "application/json" },
        body: JSON.stringify({ action: "add_to_cart", price: price }),
      });
    });
});
```

### DIFF
--- a/docs/concepts/fundamentals/html_js_css_variables.md
+++ b/docs/concepts/fundamentals/html_js_css_variables.md
@@ -223,8 +223,8 @@ If you render multiple instances of the same component with different JS data, e
 
 ```python
 # Render multiple instances
-{% component "InteractiveMap" location_id=1 %}{% endcomponent %}
-{% component "InteractiveMap" location_id=2 %}{% endcomponent %}
+{% component "Calendar" date=some_date %}{% endcomponent %}
+{% component "Calendar" date=other_date %}{% endcomponent %}
 ```
 
 Each instance will have its `$onComponent()` callback called with the data specific to that instance's `get_js_data()` return value.
--- a/src/django_components/dependencies.py
+++ b/src/django_components/dependencies.py
@@ -999,7 +999,6 @@ def _process_dep_declarations(
     # NOTE: No exec script for the "simple"|"prepend"|"append"|"ignore" mode,
     #       as they are NOT using the dependency manager.
     if strategy == "document":
-        # This script is loaded EARLY to ensure we load the JS/CSS from various sources
         exec_script = _gen_exec_script(
             output_type="json",
             script_kind="core",  # Indicate not to mess with this script
@@ -1008,18 +1007,6 @@ def _process_dep_declarations(
             css_tags__fetch_in_client=[],
             js_urls__mark_loaded_in_client=js_urls__mark_loaded_in_client,
             css_urls__mark_loaded_in_client=css_urls__mark_loaded_in_client,
-            comp_calls=[],
-            comp_js_vars=[],
-        )
-        # This script is loaded LATE to trigger component calls after everything else is loaded
-        calls_script = _gen_exec_script(
-            output_type="json",
-            script_kind="core",  # Indicate not to mess with this script
-            script_origin_class_id=None,
-            js_tags__fetch_in_client=[],
-            css_tags__fetch_in_client=[],
-            js_urls__mark_loaded_in_client=[],
-            css_urls__mark_loaded_in_client=[],
             comp_calls=comp_calls,
             comp_js_vars=[],
         )
@@ -1038,10 +1025,8 @@ def _process_dep_declarations(
             comp_calls=comp_calls,
             comp_js_vars=[],
         )
-        calls_script = None
     else:
         exec_script = None
-        calls_script = None
 
     # Core scripts without which the rest wouldn't work
     core_script_tags: list[Script | Style] = []
@@ -1074,8 +1059,6 @@ def _process_dep_declarations(
         # JS variables
         # Loaded after `Component.js`, because `Component.js` defines the variables callbacks
         *component_js_vars__inline,
-        # This triggers component calls after everything else is loaded
-        *([calls_script] if calls_script else []),
     ]
 
     final_styles = cast(
--- a/src/django_components_js/src/manager.ts
+++ b/src/django_components_js/src/manager.ts
@@ -40,7 +40,7 @@ export interface TagJson {
  * Usage:
  *
  * ```js
- * DjangoComponents.registerComponent("table", async (data, { id, name, els }) => {
+ * DjangoComponents.manager.registerComponent("table", async (data, { id, name, els }) => {
  *   ...
  * });
  * ```
@@ -48,34 +48,42 @@ export interface TagJson {
  * Multiple callbacks can be registered for the same component:
  *
  * ```js
- * DjangoComponents.registerComponent("table", async (data, { id, name, els }) => {
+ * DjangoComponents.manager.registerComponent("table", async (data, { id, name, els }) => {
  *   // First callback
  * });
- * DjangoComponents.registerComponent("table", async (data, { id, name, els }) => {
+ * DjangoComponents.manager.registerComponent("table", async (data, { id, name, els }) => {
  *   // Second callback - will be called after the first
  * });
  * ```
  *
  * ```js
- * DjangoComponents.registerComponentData("table", "3d09cf", () => {
+ * DjangoComponents.manager.registerComponentData("table", "3d09cf", () => {
  *   return JSON.parse('{ "abc": 123 }');
  * });
  * ```
  *
  * ```js
- * DjangoComponents.callComponent("table", 12345, "3d09cf");
+ * DjangoComponents.manager.callComponent("table", 12345, "3d09cf");
  * ```
  *
  * ```js
- * DjangoComponents.loadJs({ tag: 'script', attrs: { src: '/abc/def' }, content: 'console.log("Hello, world!");' });
+ * DjangoComponents.manager.loadJs({
+ *   tag: 'script',
+ *   attrs: { src: '/abc/def' },
+ *   content: 'console.log("Hello, world!");'
+ * });
  * ```
  *
  * ```js
- * DjangoComponents.loadCss({ tag: 'link', attrs: { href: '/abc/def', rel: 'stylesheet' }, content: '' });
+ * DjangoComponents.manager.loadCss({
+ *   tag: 'link',
+ *   attrs: { href: '/abc/def', rel: 'stylesheet' },
+ *   content: ''
+ * });
  * ```
  *
  * ```js
- * DjangoComponents.markScriptLoaded("js", '/abc/def');
+ * DjangoComponents.manager.markScriptLoaded("js", '/abc/def');
  * ```
  */
 export const createComponentsManager = () => {
@@ -476,7 +484,7 @@ export const createComponentsManager = () => {
    * @param compFn - The callback function to call when the component is instantiated.
    *
    * @example
-   * DjangoComponents.registerComponent("table", async (data, { id, name, els }) => {
+   * DjangoComponents.manager.registerComponent("table", async (data, { id, name, els }) => {
    *   ...
    * });
    */
@@ -495,7 +503,7 @@ export const createComponentsManager = () => {
 
   /**
    * @example
-   * DjangoComponents.registerComponentData("table", "a1b2c3", () => {{
+   * DjangoComponents.manager.registerComponentData("table", "a1b2c3", () => {{
    *   return JSON.parse('{ "a": 2 }');
    * }});
    */
--- a/src/django_components_js/src/index.ts
+++ b/src/django_components_js/src/index.ts
@@ -4,6 +4,9 @@ import { unescapeJs } from './utils';
 
 export type * from './manager';
 
+// TODO_v1: Make `DjangoComponents` THE manager object
+//          - keep `createComponentsManager` as one of manager's methods (used in tests)
+//          - remove `unescapeJs` (unused)
 export const DjangoComponents = {
   manager: createComponentsManager(),
   createComponentsManager,
--- a/tests/test_component_js.py
+++ b/tests/test_component_js.py
@@ -41,7 +41,7 @@ class TestJsVariables:
                 "cssTags__toFetch": [],
                 "jsTags__toFetch": [],
                 "componentJsVars": [],
-                "componentJsCalls": []}</script>
+                "componentJsCalls": [["VGVzdENvbXBvbmVudF9hNjdmOWY=", "Y2ExYmMzZQ==", "NTQ0MGU0"]]}</script>
             <script>
                 DjangoComponents.manager.registerComponent("TestComponent_a67f9f", ({ message }) => {
                     console.log(message);
@@ -57,12 +57,6 @@ class TestJsVariables:
                         "componentJsCalls": []});
                 })();
             </script>
-            <script type="application/json" data-djc>{"cssUrls__markAsLoaded": [],
-                "jsUrls__markAsLoaded": [],
-                "cssTags__toFetch": [],
-                "jsTags__toFetch": [],
-                "componentJsVars": [],
-                "componentJsCalls": [["VGVzdENvbXBvbmVudF9hNjdmOWY=", "Y2ExYmMzZQ==", "NTQ0MGU0"]]}</script>
             """,  # noqa: E501
             rendered,
         )
@@ -122,7 +116,8 @@ class TestJsVariables:
                 "cssTags__toFetch": [],
                 "jsTags__toFetch": [],
                 "componentJsVars": [],
-                "componentJsCalls": []}</script>
+                "componentJsCalls": [["VGVzdENvbXBvbmVudFJlZ2lzdGVyZWRfY2U0NWQ0", "Y2ExYmM0MQ==", "YTZmZjgy"],
+                    ["VGVzdENvbXBvbmVudFJlZ2lzdGVyZWRfY2U0NWQ0", "Y2ExYmM0Mg==", "YzBjYWEx"]]}</script>
             <script>
                 DjangoComponents.manager.registerComponent("TestComponentRegistered_ce45d4", ({ value }) => {
                     console.log('Value:', value);
@@ -148,13 +143,6 @@ class TestJsVariables:
                         "componentJsCalls": []});
                 })();
             </script>
-            <script type="application/json" data-djc>{"cssUrls__markAsLoaded": [],
-                "jsUrls__markAsLoaded": [],
-                "cssTags__toFetch": [],
-                "jsTags__toFetch": [],
-                "componentJsVars": [],
-                "componentJsCalls": [["VGVzdENvbXBvbmVudFJlZ2lzdGVyZWRfY2U0NWQ0", "Y2ExYmM0MQ==", "YTZmZjgy"],
-                    ["VGVzdENvbXBvbmVudFJlZ2lzdGVyZWRfY2U0NWQ0", "Y2ExYmM0Mg==", "YzBjYWEx"]]}</script>
             """,  # noqa: E501
             rendered,
         )
@@ -196,7 +184,7 @@ class TestJsVariables:
                 "cssTags__toFetch": [],
                 "jsTags__toFetch": [],
                 "componentJsVars": [],
-                "componentJsCalls": []}</script>
+                "componentJsCalls": [["VGVzdENvbXBvbmVudF8yZTY4MDY=", "Y2ExYmMzZQ==", "N2FjZmI5"]]}</script>
             <script>
                 DjangoComponents.manager.registerComponent("TestComponent_2e6806", ({ lat, lng, zoom, markers }) => {
                     console.log(`Map at ${lat}, ${lng}, zoom: ${zoom}`);
@@ -214,12 +202,6 @@ class TestJsVariables:
                         "componentJsCalls": []});
                 })();
             </script>
-            <script type="application/json" data-djc>{"cssUrls__markAsLoaded": [],
-                "jsUrls__markAsLoaded": [],
-                "cssTags__toFetch": [],
-                "jsTags__toFetch": [],
-                "componentJsVars": [],
-                "componentJsCalls": [["VGVzdENvbXBvbmVudF8yZTY4MDY=", "Y2ExYmMzZQ==", "N2FjZmI5"]]}</script>
             """,
             rendered,
         )
@@ -251,18 +233,12 @@ class TestJsVariables:
                 "cssTags__toFetch": [],
                 "jsTags__toFetch": [],
                 "componentJsVars": [],
-                "componentJsCalls": []}</script>
+                "componentJsCalls": [["VGVzdENvbXBvbmVudF9hOGI3NjY=", "Y2ExYmMzZQ==", null]]}</script>
             <script>
                 DjangoComponents.manager.registerComponent("TestComponent_a8b766", () => {
                     console.log('No data');
                 });
             </script>
-            <script type="application/json" data-djc>{"cssUrls__markAsLoaded": [],
-                "jsUrls__markAsLoaded": [],
-                "cssTags__toFetch": [],
-                "jsTags__toFetch": [],
-                "componentJsVars": [],
-                "componentJsCalls": [["VGVzdENvbXBvbmVudF9hOGI3NjY=", "Y2ExYmMzZQ==", null]]}</script>
             """,
             rendered,
         )
@@ -307,7 +283,7 @@ class TestJsVariables:
                 "cssTags__toFetch": [],
                 "jsTags__toFetch": [],
                 "componentJsVars": [],
-                "componentJsCalls": []}</script>
+                "componentJsCalls": [["VGVzdENvbXBvbmVudF84MTFlNzA=", "Y2ExYmMzZQ==", "YmEyZDkx"]]}</script>
             <script>
                 DjangoComponents.manager.registerComponent("TestComponent_811e70", ({ user_id, api_key }) => {
                     fetch(`/api/users/${user_id}`, {
@@ -317,15 +293,14 @@ class TestJsVariables:
             </script>
             <script type="text/javascript">
                 (function() {
-                    DjangoComponents.manager._loadComponentScripts({"cssUrls__markAsLoaded": [], "jsUrls__markAsLoaded": [], "cssTags__toFetch": [], "jsTags__toFetch": [], "componentJsVars": [["VGVzdENvbXBvbmVudF84MTFlNzA=", "YmEyZDkx", "eyJ1c2VyX2lkIjogMTIzLCAiYXBpX2tleSI6ICJzZWNyZXQta2V5In0="]], "componentJsCalls": []});
+                    DjangoComponents.manager._loadComponentScripts({"cssUrls__markAsLoaded": [],
+                        "jsUrls__markAsLoaded": [],
+                        "cssTags__toFetch": [],
+                        "jsTags__toFetch": [],
+                        "componentJsVars": [["VGVzdENvbXBvbmVudF84MTFlNzA=", "YmEyZDkx", "eyJ1c2VyX2lkIjogMTIzLCAiYXBpX2tleSI6ICJzZWNyZXQta2V5In0="]],
+                        "componentJsCalls": []});
                 })();
             </script>
-            <script type="application/json" data-djc>{"cssUrls__markAsLoaded": [],
-                "jsUrls__markAsLoaded": [],
-                "cssTags__toFetch": [],
-                "jsTags__toFetch": [],
-                "componentJsVars": [],
-                "componentJsCalls": [["VGVzdENvbXBvbmVudF84MTFlNzA=", "Y2ExYmMzZQ==", "YmEyZDkx"]]}</script>
             """,  # noqa: E501
             rendered,
         )


.github/workflows/tests.yml
```diff
@@ -43,6 +43,17 @@ jobs:
         run: |
           uv sync --group ci
 
+      # TODO: DELETEME
+      # Workaround: Yarn Debian repo switched GPG key (Jan 2026); old key in images breaks apt-get update.
+      # See https://github.com/yarnpkg/yarn/issues/9218
+      # Remove this step once CI images ship with the updated Yarn APT key or Yarn repo is no longer enabled by default.
+      - name: Disable broken third-party apt repos (e.g. Yarn) before Playwright install
```
JuroOravec, line 50: Not related to this feature. But while preparing the PR, I came across this error. [See this thread](https://github.com/yarnpkg/yarn/issues/9218#issuecomment-3823880280).

It's a temporary measure, but added it to unblock this PR

docs/examples/fragments/component.py
```diff
@@ -17,8 +18,15 @@ class Kwargs:
         </div>
     """
 
+    def get_js_data(self, args, kwargs: Kwargs, slots, context):
```
JuroOravec, line 21: Updated this example to showcase JS variables

docs/community/development.md
```diff
@@ -186,6 +186,77 @@ pytest
 tox -e py310
 ```
 
+### E2E Test Server Configuration
```
JuroOravec, line 189: Part of this PR is this internal feature. I was just too comfortable to split it as separate PR.

The issue was that:
1. For E2E tests, we need to define a Django server that serves the HTML files / fragments that we need to test the behaviour in th browser.
2. But this "Django test server" is defined in its own directory, so also all the HTML templates and components are defined there.
3. So updating E2E tests is always messy, because you need to update files across multiple locations, and actively keep in your mind which endpoints have been defined, what are their names, etc.

So the solution to this issue is to allow to define the endpoints, templates and components for the test server directly within the original test file where those endpoints are actually accesed (see below in this markdown file).

So now each test file (`test_*.py` ) can define a module-level `server()` function. And this function must return a dictionary of `{ "/my/endpoint": handler() }`.

---

This logic is defined in [tests/e2e/testserver/testserver/server_discovery.py](https://github.com/django-components/django-components/pull/1570/changes#diff-5b02bd269945b9e4eed6acf10c102d419c1dd5aab311bc2cf95d973c30ed6227).

How it works is that:
1. `pytest` has a `testpaths` setting in `pyproject.toml` that defines the tests directories.
2. A script takes this `testpaths` value, and searches for `test_*` files within there.
3. For each test file, it checks if the file contains the module-level `server` function.
4. If it does contain `server` func, the test file is properly imported and the `server` function is called, returning a dictionary of new endpoints.
5. The new endpoints are added to the Django test server's `urlpatterns` in [tests/e2e/testserver/testserver/urls.py](https://github.com/django-components/django-components/pull/1570/changes#diff-4fddf34ecc2cfc04a88fc97d346789546f00010cef4c12eeaefb1bb72119ba7e)

The script uses Python AST parser to check for the `server` function, which means that `server` can be a variable, function, lambda func, class, ... basically anything as long as it's called `server`.

---

There will be a separate PR in which I'll refactor the rest of the test files to use this `server` pattern.

docs/concepts/fundamentals/html_js_css_variables.md
```diff
@@ -147,6 +147,214 @@ The difference is that:
...
+    $onComponent(async (data, { id, name, els }) => {
+      // Second callback - will be called after the first
+    });
+    ```
+
+!!! info
+
+    Components' `$onComponent()` callbacks are called in the order in which the components are found in the HTML.
+
+### `$onComponent()` callback function
```
JuroOravec, line 253: **Important: The special callback to access JS variables from in JS code is called `$onComponent()` instead of `$onLoad()`.**

Just a rename, the overall flow remains.

I realized that `$onLoad` is a bit cryptic, it doesn't explain itself right away - "On loading of what? page? third party package?".  Whereas `$onComponent` links back to components directly.

src/django_components/dependencies.py
```diff
@@ -59,43 +240,57 @@
...
+# Matches: $onComponent( with optional whitespace before the opening parenthesis
+_ONCOMPONENT_PATTERN = re.compile(r"\$onComponent\s*\(")
+
+
+def _transform_oncomponent_calls(js_content: str, comp_cls_id: str) -> str:
+    """
+    Replace `$onComponent(` with `DjangoComponents.manager.registerComponent("comp_cls_id", `
+    so that $onComponent is just syntactic sugar for `registerComponent()`.
+    """
+    return _ONCOMPONENT_PATTERN.sub(f'DjangoComponents.manager.registerComponent("{comp_cls_id}", ', js_content)
```
JuroOravec, line 293: One important part of how `$onComponent(cb)` works is that in reality it's just a syntactic sugar for

```js
DjangoComponents.manager.registerComponent(comp_class_id, cb)
```

Also important, `$onComponent()` is expanded into the code above ONLY in the code in `Component.js` / `Component.js_file`. It doesn NOT apply to JS files in `Component.Media.js`.

## Background

Previously, I was thinking of defining `$onComponent` as an actual JS function. e.g. that we'd prepend users' JS script with the definition of `$onComponent`

```js
const $onComponent = (cb) => DjangoComponents.manager.registerComponent("MyTable__c123456", cb);

// Component.js  here...
```

But the problem was that `$onComponent` is specific to each component class. So if 3 components classes were used in a  template, there'd be 3 definitions of `$onComponent` in the browser.

So at the very least I had to wrap the definition of `$onComponent` so it's not a global.

```js
(async () => {
const $onComponent = (cb) => DjangoComponents.manager.registerComponent("MyTable__c123456", cb);

// Component.js  here...

})()
```

**But this would make it DIFFICULT to support [JS modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) `<script type="module">`**

Because JS modules can contain import statements, and import statements CANNOT be nested. So this would fail:

```js
// ❌ This will cause a SyntaxError:
// "An import declaration can only be used at the top level of a module."
(async () => {
    import { User } from './api.js'; 
})()
```

So instead, I went with simply expanding
```js
$onComponent(cb)
```
into
```js
DjangoComponents.manager.registerComponent("CompClass__123456", cb)
```

This way we don't have to wrap anything, so it can still work wirh JS modules.

src/django_components/dependencies.py
```diff
@@ -110,11 +305,22 @@ def cache_component_js(comp_cls: type["Component"], force: bool) -> None:
     if not force and _is_script_in_cache(comp_cls, "js", None):
         return
 
+    # Transform `$onComponent(` calls to registerComponent calls before caching
+    transformed_js = _transform_oncomponent_calls(comp_cls.js, comp_cls.class_id)
```
JuroOravec, line 309: Here is where we expand `$onComponent(` into `DjangoComponents.manager.registerComponent(`. It's done before the JS script is cached. So that when this script is requested from the browser, we can just take it from the cache and serve it.

src/django_components/dependencies.py
```diff
@@ -561,103 +811,122 @@ def _process_dep_declarations(content: bytes, strategy: DependenciesStrategy) ->
...
-    all_parts: list[bytes] = []
-
-    def on_replace_match(match: "re.Match[bytes]") -> bytes:
-        all_parts.append(match.group("data"))
-        return b""
-
-    content = COMPONENT_COMMENT_REGEX.sub(on_replace_match, content)
+    from django_components.component import get_component_by_class_id  # noqa: PLC0415
 
-    # NOTE: Python's set does NOT preserve order, so both set and list are needed
```
JuroOravec, line 573: Cleaned up this logic, added more description, etc. But the flow remains the same. The flow is:

1. Given a rendered HTML, figure out which component instances were used by extracting the `<!-- _RENDERED -->` comments that the components left behind.
2. These comments contain metadata about component class ID, component instance ID, and whether the instance defined any JS/CSS variables.
3. We then retrieve the Component class by the ID, and get its `Component.js`, `Component.css`, and `Component.Media` scripts (if defined).
4. If JS/CSS variables were defined, we also them as `<script>` and `<style>` tags.
5. Now that we have all the JS/CSS scripts, we sort them into those that should be 1) inlined into the HTML, and 2) loaded later from within the browser using our dependency manager (`manager.ts`).
   - This is because things need to be included/excluded differently if we're rendering a base page (`document`) vs an HTML fragment (`fragment`).
6. Lastly, we insert the `<script>`, `<style>`, `<link>` tags into the HTML, either at the default locations (CSS in `<head>` and JS in `<body>`), or where user has put `{% component _js_dependencies %}` tags

src/django_components/dependencies.py
```diff
@@ -848,103 +1254,160 @@ def _prepare_tags_and_urls(
...
-            f"Could not find {script_type.upper()} for component '{comp_cls.__name__}' (id: {comp_cls.class_id})",
-        )
+    if cached_data is None:
+        return None
 
-    if script_type == "js":
-        content = wrap_component_js(comp_cls, content)
-    elif script_type == "css":
-        content = wrap_component_css(comp_cls, content)
+    # TODO_V1: Keep only new format in v1. Since scripts may be stored in
```
JuroOravec, line 1303: We store in the cache the scripts to generate CSS / JS variables. Before, these were stored as plain strings in the cache. Now those entries are serialized JSONs that deserialize back to `Script` / `Style ` instances.

Well the issue is that caches may be independent services that persist even when the server restarts. So to keep this logic backwards compatible, when we are accessing the JS/CSS from the cache, we first try to handle it as the new JSON format, and fallback to the old plain string format.

src/django_components/dependencies.py
```diff
@@ -41,6 +45,183 @@
 DEPS_STRATEGIES = ("document", "fragment", "simple", "prepend", "append", "ignore")
 
 
+ScriptKind: TypeAlias = Literal["component", "variables", "core", "extra"]
+
+
+@dataclass
+class Script:
```
JuroOravec, line 52: One new and important thing - Previously we handled the `<script>` and `<style>` tags in `dependencies.py` as strings. Now, there are `Script` and `Style` classes.

You can see usage here:
- https://github.com/django-components/django-components/pull/1570/changes#diff-371d604ce82d2a01458065f1b137fd981b12af1b0b5c5fb2c1f7f5b45c20003fR313
- https://github.com/django-components/django-components/pull/1570/changes#diff-371d604ce82d2a01458065f1b137fd981b12af1b0b5c5fb2c1f7f5b45c20003fR1152
- https://github.com/django-components/django-components/pull/1570/changes#diff-371d604ce82d2a01458065f1b137fd981b12af1b0b5c5fb2c1f7f5b45c20003fR1348

One thing that I still have actively in mind is that eventually there needs to be a way for people to easily configure whether to render the JS scripts as scripts vs modules, add nonce for CSP (see https://github.com/django-components/django-components/issues/932), and other.

Handling `<script>` and `<style>` strings as `Script` and `Style` objects makes this much, much easier:
- If user wants to modify the content of JS scripts, they can modify `Script.content`, without needing to use regex or similar.
- Or if user needs to add nonce, they can add a `nonce` attribute to all Script/Style objects with `Script.attrs`

The Script/Style objects are then serialized to string `<script>` / `<style>` / `<link>` tags at the very end of the process. See [`Script.render()`](https://github.com/django-components/django-components/pull/1570/changes#diff-371d604ce82d2a01458065f1b137fd981b12af1b0b5c5fb2c1f7f5b45c20003fR115) and [`Style.render()`](https://github.com/django-components/django-components/pull/1570/changes#diff-371d604ce82d2a01458065f1b137fd981b12af1b0b5c5fb2c1f7f5b45c20003fR194)

src/django_components/dependencies.py
```diff
@@ -561,103 +811,122 @@ def _process_dep_declarations(content: bytes, strategy: DependenciesStrategy) ->
...
+        component_js_vars__inline,
+        component_css_vars__inline,
+        component_js_vars__fetch_in_client,
+        component_css_vars__fetch_in_client,
+        component_js_vars__mark_loaded_in_client,
+        component_css_vars__mark_loaded_in_client,
+    ) = _prepare_scripts_and_urls(variables_data, strategy)
+
+    #############################
+    # Legacy code start
```
JuroOravec, line 915: **Important for Script/Style objects to work**

Internally we now use the Script/Style objects ([see this comment](https://github.com/django-components/django-components/pull/1570#discussion_r2749602642)).

But when users want to modify how the JS/CSS are generated, the current public API asks users to interact with `<script>` and `<style>`  as strings.

So as a compromise until v1 (Citry), we keep the same flow for backwards compatibilty, but internally parse the `<script>` and `<style>` strings into `Script/Style` objects.

The logic for parsing is:
- [`_postprocess_media_tags`](https://github.com/django-components/django-components/pull/1570/changes#diff-371d604ce82d2a01458065f1b137fd981b12af1b0b5c5fb2c1f7f5b45c20003fR1102)
- [`_parse_html_tag_attrs`](https://github.com/django-components/django-components/pull/1570/changes#diff-371d604ce82d2a01458065f1b137fd981b12af1b0b5c5fb2c1f7f5b45c20003fR1176)

---

Details:

Right now we support modifying how to render a tag by piggy-backing on Django's `Media` class ([See our docs](https://django-components.github.io/django-components/0.146.0/concepts/fundamentals/secondary_js_css_files/#media-class)).

So users can subclass the Django's `Media` class, override the `render_js()/render_css()` methods, and assign the `Media` class to a Component.

src/django_components/dependencies.py
```diff
@@ -131,34 +337,41 @@ def cache_component_js(comp_cls: type["Component"], force: bool) -> None:
...
 
     # Generate and cache a JS script that contains the JS variables.
-    if not _is_script_in_cache(comp_cls, "js", input_hash):
-        _cache_script(
-            comp_cls=comp_cls,
-            script="",  # TODO - enable JS vars
-            script_type="js",
-            input_hash=input_hash,
+    if not _is_script_in_cache(comp_cls, "js", variables_hash):
+        js_vars_script = _gen_exec_script(
```
JuroOravec, line 351: This is the JS script that gets inserted into the page that sends the JSONified data from `get_js_data()`.

The generated script will look like [this](https://github.com/django-components/django-components/pull/1570/changes#diff-371d604ce82d2a01458065f1b137fd981b12af1b0b5c5fb2c1f7f5b45c20003fR1486):

```js
(function() {
    DjangoComponents.manager._loadComponentScripts(exec_script_content);
})();
```

Where `exec_script_content` has [this format](https://github.com/django-components/django-components/pull/1570/changes#diff-371d604ce82d2a01458065f1b137fd981b12af1b0b5c5fb2c1f7f5b45c20003fR1432):

```py
exec_script_data = {
    "loadedCssUrls": map_to_base64(loaded_css_urls),
    "loadedJsUrls": map_to_base64(loaded_js_urls),
    "toLoadCssTags": map_to_base64(to_load_css_tags),
    "toLoadJsTags": map_to_base64(to_load_js_tags),
    # For the URLs that are to be marked as "already loaded", we format them just as URLs,
    # NOT an entire HTML tag. Because we don't care about the other HTML attributes.
    "cssUrls__markAsLoaded": map_to_base64(extract_urls(css_urls__mark_loaded_in_client)),
    "jsUrls__markAsLoaded": map_to_base64(extract_urls(js_urls__mark_loaded_in_client)),
    # But for the `<script>/<style>/<link> tags that we want to dynamically load in browser
    # we pass JSON objects with tag, attrs, and content fields. The browser will construct
    # the HTML elements from this data.
    "cssTags__toFetch": map_to_base64(render_tags(css_tags__fetch_in_client)),
    "jsTags__toFetch": map_to_base64(render_tags(js_tags__fetch_in_client)),
    # NOTE: Component call data contains only hashes and IDs. But since this info is taken
    # from the rendered HTML, which could have been tampered with, it's better to escape these to base64 too.
    "componentJsVars": [map_to_base64(js_vars) for js_vars in comp_js_vars],
    # NOTE: Component call data contains only hashes and IDs. But since this info is taken
    # from the rendered HTML, which could have been tampered with, it's better to escape these to base64 too.
    "componentJsCalls": [
        [
            to_base64(call.comp_cls_id),
            to_base64(call.comp_id),
            # `None` (converted to `null` in JSON) means that the component has no JS variables
            to_base64(call.js_input_hash) if call.js_input_hash is not None else None,
        ]
        for call in comp_calls
    ],
}
```

This data will be then procesed by the our client-side dependency manager [here in `manager.ts`](https://github.com/django-components/django-components/pull/1570/changes#diff-5b8acf408bfdc9d80635c5efe6b1b636da344349820e975c948e0214cf736b98R614).

src/django_components/dependencies.py
```diff
@@ -666,177 +935,327 @@ def get_component_media(comp_cls_id: str) -> Media:
...
+            *(js_tags__fetch_in_client if strategy == "document" else []),
         ],
+        key=lambda x: x.url if x.url else "",
     )
 
-    # NOTE: No exec script for the "simple" mode, as that one is NOT using the dependency manager
-    if strategy in ("document", "fragment"):
+    # NOTE: No exec script for the "simple"|"prepend"|"append"|"ignore" mode,
+    #       as they are NOT using the dependency manager.
+    if strategy == "document":
```
JuroOravec, line 1001: I have split these for clarity, so it's clear that for fragments we really don't include any JS/CSS scripts, only commands to the dependency manager to load the needed JS/CSS.

src/django_components/dependencies.py
```diff
@@ -666,177 +935,327 @@ def get_component_media(comp_cls_id: str) -> Media:
...
+            *css_tags__fetch_in_client,
         ],
     )
 
-    return (content, final_script_tags.encode("utf-8"), final_css_tags.encode("utf-8"))
-
-
-href_pattern = re.compile(r'href="([^"]+)"')
-src_pattern = re.compile(r'src="([^"]+)"')
+    return (content, final_scripts, final_styles)
```
JuroOravec, line 1080: You can see here that before we returned serialized strings with all the `<script>` and `<style>` tags combined. Now we return lists of Script and Style objects.

And instead, the serialization now happens outside of this function, [here in `render_dependencies()`](https://github.com/django-components/django-components/pull/1570/changes#diff-371d604ce82d2a01458065f1b137fd981b12af1b0b5c5fb2c1f7f5b45c20003fR671).

It is there, right before serializng the Scripts and Styles, where we could insert a plugin insertion point to allow plugins to modify how the `<script>` and `<style>` tags are generated.

src/django_components/dependencies.py
```diff
@@ -954,18 +1430,64 @@ def map_to_base64(lst: Sequence[str]) -> list[str]:
     #
     # NOTE 2: Convert to Base64 to avoid any issues with `</script>` tags in the content
     exec_script_data = {
```
JuroOravec, line 1415: As a reminder, this is the JSON that gets sent to the browser for our dependency manager to pick it up and process the metadata.

This JSON can be used to send info to the browser like:
- Which JS/CSS scripts to fetch and insert into the DOM
- Which JS/CSS to "mark as loaded", meaning that future requests to load it will be ignored
- JS variables - passing the JSON payload from `get_js_data()`, the class ID, and the payload hash.
- "component calls" AKA command to trigger `$onComponent()` for a specific component instance.

All of this data is base64-encoded, because it may contain user-defined data which could break the HTML.

And all of this data is static - no `exec()` or anything. The logic for processing is defined in `manager.ts`.

For example, when a `componentJsCalls` entry is pushed to the browser, the `manager.ts` adds it to the queue, checks if everything that this component calls needs is there, and if so, calls the user-defined `$onComponent()` callback.

The data from `get_js_data()` is deserialized anew every time we call `$onComponent()`, to ensure that multiple `$onComponent()` callbacks have their own separate data objects and don't interfere with each other.

src/django_components_js/src/manager.ts
```diff
@@ -34,6 +45,17 @@ export type ScriptType = 'js' | 'css';
  * });
  * ```
  *
+ * Multiple callbacks can be registered for the same component:
```
JuroOravec, line 48: As it says here, there can now multiple `$onComponent()` callbacks. This is mainly useful so that extensions could add custom JS logic without interfering with user's JS.

src/django_components_js/src/manager.ts
```diff
@@ -1,29 +1,40 @@
...
+  TEl extends HTMLElement = HTMLElement,
+> = (data: TData, ctx: ComponentContext<TEl>) => MaybePromise<any>;
 
 export type DataFn = () => object;
 
-export type ScriptType = 'js' | 'css';
+export type ScriptType = "js" | "css";
+
+/** Queue for component calls that are waiting for dependencies to be registered */
+export interface PendingCall {
```
JuroOravec, line 23: Ok now a bit more about how the `manager.ts` works to safely and orderly call `$onComponent()`:

In my initial implementation, our manager called `$onComponent()` as soon as the command arrived. If there was something missing, like the JS data from `get_js_data()`, then we simply raised an error.

However, when working around with the HTML fragments with JS variables, it can get tricky to ensure that all the parts are loaded in the correct order.

Now instead, when we receive a call command (`callComponent()`), we put the command to a queue, and wait until other parts have been provided (`registerComponent()` and `registerComponentData()`). And only after everything is set we finally call `$onComponent()` callback.

### Breakdown of `$onComponent`

To successfully call `$onComponent()`, we need 3 things:
1. `manager.registerComponent()` - AKA register the `$onComponent()` callback
2. `manager.registerComponentData()` - (IF JS VARIABLES) register a function that generates the JS variables on the JS side
3. `manager.callComponent()` - Actually call the `$onComponent()` callback.

`$onComponent()` is a syntactic sugar for `manager.registerComponent()`. So when we run JS code that contains `$onComponent()`, that's when we call `registerComponent()`.

`registerComponentData()` is sent from the server only if the Python component defined any JS data from `get_js_data()`.

And `callComponent()` is sent from the server if the JS code contained any `$onComponent()`.

### Component call queue

The queue also preserves order. The order of `$onComponent()` calls should match the order in which the Python components were rendered in the HTML.

So in the example below, the order of calls of `$onComponent()` should be:
1. table
2. image
```html
<div>
  {% component "table" / %}
  {% component "image" / %}
</div>
```

To preserve the order, the items in the queue are **BLOCKING**.

If Component `table` is still waiting for its JS data to arrive, the Component `image` won't be executed either.

Only after `table` has everything ready, and we call the `$onComponent()` callback. Only then we look at the `image` component and consider whether we can call it or whether we should wait.

src/django_components_js/src/manager.ts
```diff
@@ -59,76 +81,115 @@ export type ScriptType = 'js' | 'css';
...
+  // The key has format `compClsId:compId:jsVarsHash`.
+  // The value is for error (if any), otherwise null.
+  //
+  // - If the key is not present in the Map, then the promise has not ended yet.
+  // - If the key is present and the value is null, then the promise has ended successfully.
+  // - If the key is present and the value is non-null, then the promise has ended in error.
+  //
+  // This way, when we are checking if a pending comp call is ready, we can simply check
+  // this Map instead of trying to inspect the promise itself.
+  const promiseCompletionStatus = new Map<string, Error | null>();
```
JuroOravec, line 133: [This comment](https://github.com/django-components/django-components/pull/1570#discussion_r2749717449) said that before we can call `$onComponent()`, we have to wait for `registerComponent` and `registerComponentData`.

In reality, there's one more thing we have to wait for. And that's the JS dependencies from `Component.Media.js`.

This is because those JS scripts in `Component.Media.js` act as dependencies. And the code in `Component.js` may rely on those dependencies.

In browsers, the `<script>` tag (aka [`HtmlScriptElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement)) has an `onload` hook that it called when the script's JS code is done executing.

This way we can detect when all the JS scripts from `Component.Media.js` are loaded - We set up `onload` hooks for all of them, to resolve a promise when `onload` is called. And this Promise tells us when the dependencies are loaded. ([See here in `loadJs()`](https://github.com/django-components/django-components/pull/1570/changes#diff-5b8acf408bfdc9d80635c5efe6b1b636da344349820e975c948e0214cf736b98R192))

This then leads us to `promiseCompletionStatus`:
- JavaScript doesn't offer any way to synchronously inspect whether a Promise is finished (resolved or rejected).
- But when checking whether an enqueued call is ready by checking its Promise (for `Component.Media.js`), this is done synchrously.
- So as a solution, `promiseCompletionStatus` is for capturing the results of finished Promises, so we can check if a given Promise for given component has already finished, and whether successfully or not.

src/django_components_js/src/manager.ts
```diff
@@ -140,58 +201,296 @@ export const createComponentsManager = () => {
...
+  };
+
+  /**
+   * Start the warning interval if not already started.
+   *
+   * The interval periodically logs warnings for component calls that are still waiting for dependencies.
+   *
+   * The interval checks for blocked calls every 5 seconds.
+   */
+  const _startWarningInterval = (): void => {
```
JuroOravec, line 365: One problem with the new way how the calls to `$onComponent()` are queued is that, in theory, component calls can get stuck if they never receive what they need.

So as a catch-all, when there are pending component calls, we start an interval. The interval checks every 5 seconds and reports are many component calls are waiting.

Once the queue is emptied, the interval is cleared.

In practice we should never see this log message from the interval. But if the browser loses internet connection mid-loading, that's when this message would show up in the devtools.

src/django_components_js/src/manager.ts
```diff
@@ -140,58 +201,296 @@ export const createComponentsManager = () => {
...
+      }
+    }, 5000); // Check every 5 seconds
+  };
+
+  /**
+   * Process pending component calls in order.
+   * Executes all consecutive unblocked calls from the start of the queue,
+   * stopping at the first blocked call.
+   */
+  const _processPendingCalls = async (): Promise<void> => {
```
JuroOravec, line 393: This function flushes the queue.

src/django_components_js/src/manager.ts
```diff
@@ -200,82 +499,216 @@ export const createComponentsManager = () => {
...
     componentInputs[key] = dataFactory;
+
+    // Check if any pending calls are now unblocked
+    return _processPendingCalls();
   };
 
-  const callComponent = (name: string, compId: string, inputHash: string): MaybePromise<any> => {
-    const initFn = components[name];
-    if (!initFn) throw Error(`[DjangoComponents] '${name}': No component registered for that name`);
+  const queueComponentCall = (
```
JuroOravec, line 522: This adds the component call to the queue.

src/django_components_js/src/manager.ts
```diff
@@ -200,82 +499,216 @@ export const createComponentsManager = () => {
...
+    });
 
-    const dataKey = `${name}:${inputHash}`;
-    const dataFactory = componentInputs[dataKey];
-    if (!dataFactory) throw Error(`[DjangoComponents] '${name}': Cannot find input for hash '${inputHash}'`);
+    return callPromise;
+  };
 
-    const data = dataFactory();
+  const callUnblockedComponent = async (
```
JuroOravec, line 568: And this is logic that runs when we know that a component call is "unblocked", and we can actually call the `$onComponent` callback.

src/django_components_js/src/manager.ts
```diff
@@ -200,82 +499,216 @@ export const createComponentsManager = () => {
...
 
-  // Also consider any embedded scripts at the moment the file is loaded.
-  const existingScripts = document.querySelectorAll<HTMLScriptElement>('script[data-djc]');
+  // Also search for any already-embedded scripts at the moment the file is loaded.
+  const existingScripts = document.querySelectorAll<HTMLScriptElement>("script[data-djc]");
   existingScripts.forEach(onDjcScript);
 
   return {
-    callComponent,
+    callComponent: queueComponentCall,
```
JuroOravec, line 713: This is intentional - while the public API of the JS manager has a `callComponent()` method, internally this first jsut enqueues the call, and only after all is ready we actually call the component.

src/django_components_js/src/errorHandling.ts
```diff
@@ -12,6 +12,7 @@ export function callWithErrorHandling(fn: Fn, args?: any[]) {
     return args ? fn.apply(null, args) : fn();
   } catch (err) {
     logError(err);
+    throw err;
```
JuroOravec, line 15: Small changes in this file to ensure that errors are not swallowed up in the browser. So when the `$onComponent()` callback raises an error, we log the error, but we also let it propagate

docs/concepts/fundamentals/html_js_css_variables.md
```diff
@@ -147,6 +147,214 @@ The difference is that:
...
+});
+```
+
+The component context contains the following properties:
+
+- `id` - `string`: The unique ID of the component instance (e.g. `"c1a2b3c"`).
+- `name` - `string`: The name of the component class (e.g. `"MyComponent"`).
+- `els` - `HTMLElement[]`: The list of DOM elements of the component instance.
+
+### Integrating with other JavaScript libraries
```
JuroOravec, line 273: In the docs I also included an example how to use `$onComponent` to integrate with AlpineJS

tests/test_dependency_manager.py
```diff
@@ -439,27 +443,42 @@ async def test_raises_if_input_hash_not_registered(self, browser: "Browser", bro
...
-            // Should raise Error
-            manager.callComponent(compName, compId, inputHash);
+            // `callComponent()` should be blocked until we register component data.
+            let isFinished = false;
+            const resultPromise = manager.callComponent(compName, compId, inputHash);
+            resultPromise.then(() => {
+                isFinished = true;
+            });
+
+            // After a short while the component call should be still blocked
```
JuroOravec, line 457: Because the calling of `$onComponent()` is now managed as a queue, the tests were updated to assert that the order remains and that items are blocking until we define all relevant parts

tests/test_component_js_e2e.py
```diff
@@ -0,0 +1,624 @@
+"""
+End-to-end tests for JS variables feature.
+
+These tests verify that JS variables from `get_js_data()` work correctly
```
JuroOravec, line 4: There's 2 test files:
- `tests/test_component_js.py` - Tests just the generated HTML - whether JS scripts are inserted correctly, etc. with and without JS variables.
- `tests/test_component_js_e2e.py` - Similar, tests fragments and base documents, with and without JS variables, and using browser to confirm it all works together.

JuroOravec:
cc @EmilStenstrom - Another big milestone :)

Everything is as we talked, except I renamed `$onLoad` to `$onComponent` so it's clearer what it does.

EmilStenstrom:
@JuroOravec Awesome milestone!

PR #1572, opened 2026-01-30
refactor: fix issue with context processors running twice

JuroOravec:
Fix for https://github.com/django-components/django-components/issues/1569

### DIFF
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -4,7 +4,7 @@ from typing import cast
 
 import pytest
 from django.http import HttpRequest
-from django.template import Context, RequestContext, Template, engines
+from django.template import Context, RequestContext, Template
 from pytest_django.asserts import assertHTMLEqual, assertInHTML
 
 from django_components import Component, register, registry, types
@@ -979,6 +979,10 @@ class TestContextProcessors:
                     },
                 },
             ],
+            # Use "django" context behavior so the component receives the same context
+            # (with MY_LIST from context processors), not an isolated copy that can
+            # miss the context-processor layer when run after test_benchmark_django.
+            "COMPONENTS": {"context_behavior": "django"},
         },
     )
     def test_context_processor_called_once_with_extends(self):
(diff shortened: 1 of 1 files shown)


JuroOravec:
Uffff, this was a difficult one... The positive is that LLM was able to fairly quickly find the issue that caused the original error as mentioned in https://github.com/django-components/django-components/issues/1569) ([see commit](https://github.com/django-components/django-components/pull/1572/changes/4318c1c89a73f1b179b8c8a7e46cf94e5dfd58bf#diff-16631b440644497efc002bed55935c78cd0e9781ed80e616e64766c8266a05e7R138)).

BUT, what was a much, much harder problem was how Django (dis)manages their settings. Basically even when using `@override_settings()`, there were some artifacts that were leaking out.

When I ran tests for the `tests/test_context.py` file itself, everything worked.

But when I ran `tests/test_context.py` together with `tests/test_benchmark_django.py`, all of a sudden the fix didn't work anymore.

What eventually worked was when the failing new test created it's own insteance of Django backend. It's somewhet annoying, because creating a new instance of Django backend, it's then hard to tell if this test still reflects a usage in the wild... But the test was passing (when ran alone) even BEFORE creating the backend instance, so at least at this moment in time we know it works.

It seems it boils down to Django settings (or other parts that depend on the settings) having some caching mechanism which doesn't get properly cleared up even with `@override_settings()`.

Just makes me want to really get rid of Django as fast as possible, and release Citry as independent package. Because this is ridiculous and such a horrible time sink.

This is the summary that LLM provided:


> ## Cause of the failure
> 
> The failure came from **two** places:
> 
> 1. **`test_benchmark_django.py` configures Django at import time**  
>    When it runs first (or is imported first), it calls `settings.configure(...)` with its own `TEMPLATES` (no `context_processors`). That can leave the global template engine and `Engine.get_default()` cache in a state that later tests still use.
> 
> 2. **`context_behavior` from the benchmark**  
>    The benchmark does `settings.COMPONENTS["context_behavior"] = CONTEXT_MODE` (e.g. `"isolated"`). With **isolated** context behavior, components get `make_isolated_context_copy(context)`, which does not carry over the context-processor layer (e.g. `MY_LIST` from `list_context_processor_1569`). So the component saw a context without `MY_LIST`.
> 
> ## Changes made
> 
> ### **`tests/test_context.py` – `test_context_processor_called_once_with_extends`**
> 
> - **Use a dedicated backend**  
>   The test now builds a `DjangoTemplates` backend from the test’s own TEMPLATES (with `list_context_processor_1569`), instead of relying on `engines["django"]`. That avoids interference from the global engine cache and from other tests (like the benchmark).
> 
> - **Force `context_behavior: "django"` for this test**  
>   The test’s `django_settings` now include `"COMPONENTS": {"context_behavior": "django"}` so the component receives the full request context (including context processor data like `MY_LIST`) instead of an isolated copy.

joeyjurjens:
@JuroOravec A bit of topic, but what does this mean?

> Just makes me want to really get rid of Django as fast as possible, and release Citry as independent package. Because this is ridiculous and such a horrible time sink.

Will Django compatibility be gone? Or will Django compatibility built upon Citry?

EmilStenstrom:
@joeyjurjens Django compatibility will NOT go away! The idea behind Citry is to move out the core to a fast Rust-based library, and have django-components be the compatibility layer that connects all this to Django. The core advantage of this is raw speed, today we're slower than pure django templates, but we want to be faster.

We're not about to throw away all the current users of the library, don't worry about that! :)

JuroOravec:
Yes, sorry for lack of context. @EmilStenstrom thanks for clarifying. @joeyjurjens It's the latter, that "Django compatibility will be built upon Citry".

PR #1577, opened 2026-02-02
feat: on_dependencies extension hook + scope JS code

JuroOravec:
Part of https://github.com/django-components/django-components/issues/1472, part of https://github.com/django-components/django-components/issues/932, Part of https://github.com/django-components/django-components/issues/836

This PR exposes a way for extensions to modify the JS/CSS `<script>` / `<style>` tags that we insert into the rendered HTML.

Inside the `on_dependencies()` extension hook, one can access and modify the list of scripts / styles to be rendered.

Scripts / styles are exposed as `Script` and `Style` objects. So users don't have to parse the raw `<script>` strings. Setting a new attribute is a simple as:

```py
my_script.attrs["type"] = "module"
```

Both `Script` and `Style` have an `url` attribute:
- When set on `Script`, it renders as `src` attr, `<script src="...">`
- When set on `Style`, it renders as `href` attr, `<link rel="stylesheet" href="...">`

Alternatively `Script` and `Style` may have `content` set, in which case:
- `Script` with content is rendered as `<script>{{ content }}</script>`
- `Style` with content is rendered as `<style>{{ content }}</style>`

This design allows users to e.g.:
- Define JS scripts as `type="module"` (see https://github.com/django-components/django-components/issues/1472)
- Add CSP nonce to JS/CSS scripts (see https://github.com/django-components/django-components/issues/932)

While this PR adds an extension hook, in the next PR I will add the same thing but as a `Component` method, in order to modify only JS/CSS of a single component class.

**Example**

```python
from django_components import ComponentExtension, OnDependenciesContext, Script, Style

    name = "my_extension"

    def on_dependencies(self, ctx: OnDependenciesContext):
        scripts = list(ctx.scripts)
        styles = list(ctx.styles)

        # Modify existing scripts and styles
        for script in scripts:
            if script.kind == "extra":
                script.wrap = False
        for style in styles:
            if style.kind == "extra":
                style.attrs["media"] = "print"

        # Add new scripts (inline content)
        scripts.append(
            # <script type="module">console.log('my custom script');</script>
            Script(
                content="console.log('my custom script');",
                attrs={"type": "module"},
                wrap=False,
            )
        )
        styles.append(
            # <style nonce="1234567890">body { background-color: red; }</style>
            Style(
                content="body { background-color: red; }",
                attrs={"nonce": "1234567890"},
            )
        )
        # Add new scripts (external URL)
        scripts.append(
            # <script src="/static/analytics.js"></script>
            Script(
                url="/static/analytics.js",
                content=None,
            )
        )
        styles.append(
            # <link rel="stylesheet" href="/static/print.css" media="print">
            Style(
                url="/static/print.css",
                content=None,
                attrs={"media": "print"},
            )
        )
        return (scripts, styles)
```

---

Secondly, because `<script>` and `<style>` tags are now handled as python objects, we're able to build on that and make it possible to smartly prevent JS scripts to polluting the global JS state (see https://github.com/django-components/django-components/issues/836).

The idea is that when there is a JS script, we wrap it in a function scope so that any variable assignments are local to that scope instead of assigned globally.

So if user defines:
```py
class Table(Component):
    js = """
    const items = [1, 2, 3];
    """
```
Then we convert it to:
```js
(() => {
    const items = [1, 2, 3];
})()
```

This way, to actually write variables to the global scope in JS, users have to use `globalThis` (or `window`).

```js
(() => {
    globalThis.items = [1, 2, 3];
})()
```


tests/__snapshots__/test_benchmark_djc.ambr
```diff
@@ -8332,7 +8347,9 @@
           document.addEventListener('alpine:init', () => {
               AlpineComposition.registerComponent(Alpine, ProjectOutputForm);
           });
-      </script><script>
+      
+  })();</script><script>(function() {
```
JuroOravec, line 8351: As can be seen from the tests, the JS scripts from `Component.js` are now by default wrapped in a scope.

tests/test_component_js_e2e.py
```diff
@@ -427,6 +459,30 @@ async def test_document_no_vars(
 
         await page.close()
 
+    @with_playwright
+    async def test_component_js_variable_not_in_global_scope(
```
JuroOravec, line 463: Added E2E test that checks that indeed the variables set in JS scripts will not pollute global scope. So now, instead, users should explicitly use `globalThis.xxx = ...` to assign to global scope.

src/django_components/dependencies.py
```diff
@@ -45,75 +46,107 @@
...
+        "text/javascript1.5",
+        "text/jscript",
+        "text/livescript",
+        "text/x-ecmascript",
+        "text/x-javascript",
+    }
+)
+
+
+def _script_type_should_wrap(attrs: dict[str, str | bool]) -> bool:
```
JuroOravec, line 83: JS scripts are optionally wrapped in a scope:
```js
(() => {
  // Original code here
})()
```

This is applied mirroring the logic from what MDN defines as "classic JS":
- `type` attribute must be either empty or omitted
- or if `type` is set, it must be one of the JS MIME types above.

In all other cases, the JS is NOT wrapped, because it may be:
- JS module `type="module"`
- JSON data, e.g. `type="importmap"`
- Other non-JS data

src/django_components/dependencies.py
```diff
@@ -45,75 +46,107 @@
...
+
+    # "text/javascript", etc
+    if type_str == "" or type_str in _JAVASCRIPT_MIME_TYPES:  # noqa: SIM103
+        return True
+    # "importmap", "module", "speculationrules", etc are not classic JS
+    return False
+
+
+@dataclass
+class Dependency:
```
JuroOravec, line 112: `Dependency` is a parent class of `Script` and `Style`. `Dependency` can be used for typing to replace `Script | Style`.

All 3 are now exported from `django_components`, so that they can be used for typing or for creating new entries within the `on_dependencies()` hook.

src/django_components/dependencies.py
```diff
@@ -666,10 +811,17 @@ def my_view(request):
     else:
         content_ = cast("bytes", content)
 
-    content_, js_dependencies, css_dependencies = _process_dep_declarations(content_, strategy)
+    content_, scripts, styles = _process_dep_declarations(content_, strategy)
 
-    js_deps_bytes = "".join([script.render() for script in js_dependencies]).encode("utf-8")
-    css_deps_bytes = "".join([style.render() for style in css_dependencies]).encode("utf-8")
+    scripts, styles = extensions.on_dependencies(
```
JuroOravec, line 816: Here is the entrypoing for the extension. Just before we convert `Script` and `Style` objects to `<script>` and `<style>` strings.

JuroOravec:
Few screenshots

Screenshot 2026-02-02 at 14 40 15
Screenshot 2026-02-02 at 14 39 50
Screenshot 2026-02-02 at 14 22 57

JuroOravec:
cc @EmilStenstrom

PR #1578, opened 2026-02-02
feat: allow Script and Style objects to be used in Component.Media.js/css

JuroOravec:
Closes https://github.com/django-components/django-components/issues/1472

Follow up to https://github.com/django-components/django-components/pull/1577, this allows `Script` and `Style` objects to be defined in `Component.Media.js/css`. This open us a way for users to easily define extra dependencies with custom attributes, e.g.:

```python
from django_components import Component, Script, Style

class Table(Component):
    class Media:
        js = [
            Script(
                url="/path/to/module.js",
                attrs={
                    "type": "module",
                }
            )
        ]
        css = [
            Style(
                url="/path/to/module.js",
                attrs={
                    "type": "module",
                }
            )
        ]
```

### DIFF
--- a/src/django_components/dependencies.py
+++ b/src/django_components/dependencies.py
@@ -1299,7 +1299,7 @@ def _postprocess_media_tags(
                 # External script with src
                 if url in url_to_obj:
                     continue
-                script_obj = Script(
+                script_obj: Script | Style = Script(
                     kind="extra",
                     url=url,
                     content=None,


src/django_components/dependencies.py
```diff
@@ -152,6 +152,10 @@ def render(self) -> SafeString:
         attrs_prefix = " " + attrs_str if attrs_str else ""
         return mark_safe(f"<{tag_name}{attrs_prefix}>{content}</{tag_name}>")  # type: ignore[return-value]
 
+    def __html__(self) -> SafeString:
```
JuroOravec, line 155: Had to add this so that `Script` and `Style` can behave as if wrapped with Django's `mark_safe()`.

What this does is that if Django's `Media` class comes across a `Script` or `Style`, it will not try to wrap it in `<script>` / `<style>` tags, but instead will assume that this object is an already rendered string.

src/django_components/dependencies.py
```diff
@@ -1260,48 +1285,74 @@ def _postprocess_media_tags(
...
-            )
-
-        # Skip duplicates (keep first occurrence)
-        if url in url_to_obj:
-            continue
-
-        # Create Script or Style object with all parsed attributes
         if script_type == "js":
-            script_obj: Script | Style = Script(
+            url = parsed_attrs.pop(attr_name, None)
```
JuroOravec, line 1297: Until now, the entries in `Component.Media.js/css` were expected to be URLs. Or even if it was an `<script>` or `<link>` tags, they were expected to have a URL.

This was so that in the client-side these JS/CSS scripts could be marked as loaded by the dependency manager.

However, adding support for `Style` / `Script` objects in `Component.Media.js/css` changes this.

Now user may define a `Script` / `Style` without URL, just content:

```python
class Table(Component):
    class Media:
        js = Script(content"console.log('hello!')")
```

So this change makes that possible. Everything stays the same, except:
- Previously we raised error if we came across a `<script>` or `<link>` tags without URL.
- Now we handle them, using the script's body to detect duplicates.

JuroOravec:
cc @EmilStenstrom

PR #1586, opened 2026-02-03
feat: add Component.on_dependencies hook

JuroOravec:
Closes https://github.com/django-components/django-components/issues/932

Follow up to https://github.com/django-components/django-components/pull/1577, this PR adds a `Component.on_dependencies()` hook.

This allows to modify the JS / CSS (`<script>` / `<style>` tags) that come from particular Component. This includes:
- JS / CSS from `Component.js/css`
- Scripts for JS / CSS variables (if this component instance had variables)
- JS / CSS from `Component.Media.js/css`.

The hook receives lists of [`Script`](https://django-components.github.io/django-components/0.148.0/reference/api#django_components.Script) and [`Style`](https://django-components.github.io/django-components/0.148.0/reference/api#django_components.Style).

In side the function you can modify the list of scripts and styles.

Or to totally replace the values, return a new tuple of `(new_scripts, new_styles)` 

Return `None` to leave them unchanged.

**Example:**

```python
from django_components import Component, Script, Style

class MyButton(Component):
    @classmethod
    def on_dependencies(cls, scripts: list[Script], styles: list[Style]):
        # Add a nonce to every inline style for this component
        for style in styles:
            if style.content and "nonce" not in style.attrs:
                style.attrs["nonce"] = get_current_nonce()
        return (scripts, styles)
```

### DIFF
--- a/src/django_components/extensions/defaults.py
+++ b/src/django_components/extensions/defaults.py
@@ -195,7 +195,7 @@ def _apply_defaults(kwargs: dict, defaults: list["ComponentDefaultField"]) -> No
     """
     for default_field in defaults:
         # Defaults are applied only to missing or `None` values
-        given_value = kwargs.get(default_field.key, None)
+        given_value = kwargs.get(default_field.key)
         if given_value is not None:
             continue
 
--- a/tests/test_benchmark_django.py
+++ b/tests/test_benchmark_django.py
@@ -2284,7 +2284,7 @@ def expansion_panel(context: Context, data: ExpansionPanelData):
             "content_attrs": data.content_attrs,
             "icon_position": data.icon_position,
             "init_data": init_data,
-            "panel_id": data.panel_id if data.panel_id else False,
+            "panel_id": data.panel_id or False,
             "expand_icon_data": expand_icon_data,
             "slot_header": data.slot_header,
             "slot_content": data.slot_content,
--- a/tests/test_benchmark_djc.py
+++ b/tests/test_benchmark_djc.py
@@ -2197,7 +2197,7 @@ class ExpansionPanel(Component):
             "content_attrs": content_attrs,
             "icon_position": icon_position,
             "init_data": init_data,
-            "panel_id": panel_id if panel_id else False,
+            "panel_id": panel_id or False,
         }
 
     template: types.django_html = """
--- a/tests/test_component.py
+++ b/tests/test_component.py
@@ -2347,7 +2347,7 @@ class TestComponentHook:
 
         class SimpleComponent(Component):
             def on_render(self, context: Context, template: Template):
-                _html, error = yield lambda: broken_template()
+                _html, error = yield broken_template
                 error.args = ("ERROR MODIFIED",)
 
         with pytest.raises(
--- a/tests/test_tag_parser.py
+++ b/tests/test_tag_parser.py
@@ -258,7 +258,7 @@ def plain_variable_value(
         token_val = token(content, start_index, line, col)
         value_token = token_val
     # Extract root variable name (everything before first dot, or entire content if no dot)
-    root_var = content.split(".")[0] if "." in content else content
+    root_var = content.split(".", maxsplit=1)[0] if "." in content else content
     used_var_start_index = start_index + len(spread) if spread is not None else start_index
     used_var_col = col + len(spread) if spread is not None else col
     used_var_token = token(root_var, used_var_start_index, line, used_var_col)
--- a/uv.lock
+++ b/uv.lock
@@ -548,7 +548,7 @@ wheels = [
 
 [[package]]
 name = "django-components"
-version = "0.146.0"
+version = "0.147.0"
 source = { editable = "." }
 dependencies = [
     { name = "django", version = "5.2.10", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.12'" },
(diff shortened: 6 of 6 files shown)


JuroOravec:
The rendering seems to be a bit slower after this PR, 3-4%. There's two potential causes of this:
- For backwards compatibility, we now doubly render the JS/CSS from `Component.Media`:
   - Media -> `<script>` string -> `Script` obj -> `<script>` string
- Second reason may be that the collection of JS/CSS dependencies may be less efficient now:
   - e.g. now we create a `Script` object for `Component.js` for each Component instance, so that the `on_dependenies` hook always receives the same input. And only at the end we remove the duplicates.
   - Previously, we created only a single `Script` object for `Component.js`, no matter how many instances there were.

cc #14

src/django_components/component.py
```diff
@@ -2085,6 +2085,37 @@ def on_render_after(self, context, template, result, error):
             ```
         """
 
+    @classmethod
+    def on_dependencies(
```
JuroOravec, line 2089: Here is the new `Component.on_dependencies` method

tests/test_dependencies.py
```diff
@@ -642,6 +643,67 @@ class Media:
         rendered = template.render(Context({}))
         assertInHTML("<style>.x { color: red; }</style>", rendered)
 
+    def test_component_on_dependencies_return_value_used(self):
```
JuroOravec, line 646: Here are the tests for the new method

CHANGELOG.md
```diff
@@ -83,6 +106,12 @@
             ]
     ```
 
+#### Refactor
+
+- `Component.Media.js/css` now render BEFORE `Component.js/css`, instead of after.
```
JuroOravec, line 111: Slight change, the JS/CSS in `Component.Media.js/css` now loads BEFORE `Component.js/css`.

This makes sense when you consider `Media.js/css` as dependencies, e.g. tailwind, Alpinejs, etc.

src/django_components/dependencies.py
```diff
@@ -833,7 +862,13 @@ def my_view(request):
     else:
         content_ = cast("bytes", content)
 
-    content_, scripts, styles = _process_dep_declarations(content_, strategy)
+    # TODO - Strategies should be only "document" | "fragment" | "simple".
+    #        "prepend" | "append" should be NOT strategies, but instead optional insertion positions.
```
JuroOravec, line 866: As the comment say, in the v1 the "prepend" and "append" strategies may be removed and replaced with a different setting, e.g. `insertion`. Fundamentally there are only 3 strategies to handle JS/CSS in our code:
- document - Insert all `<script>` and `<style>` directly into the HTML. Enables fragments.
- simple -  Same as "document" but does NOT enable fragments.
- fragment - Dynamically fetch `<script>` and `<style>` once fragment inserted into DOM.

src/django_components/dependencies.py
```diff
@@ -974,7 +1009,7 @@ def _pre_loader_js() -> Script:
 # 6. And lastly, we generate a JS script that will load / mark as loaded the JS and CSS
 #    as categorized in previous step.
 def _process_dep_declarations(
```
JuroOravec, line 1011: This function is where the main change sits. This function extracts the `<!-- _RENDERED -->` comments from an HTML string, looks up the Component classes and JS/CSS variables, collects their JS/CSS scripts, and puts it all together.

The function interface hasn't changed. Nor the final output.

But what changed was that previously, the JS/CSS scripts were prepared by their type:
1. First all `Component.js/css` from all used components
2. Then all JS/CSS variables
3. Then all `Component.Media` from all used components

But that doesn't play well with the `Component.on_dependencies()` hook. Because there we want to access all JS/CSS that belongs ONLY to that component.

So instead, now we prepare the JS/CSS scripts per component:
1. Component X - `Component.js/css`, JS/CSS vars, `Component.Media`
2. Component Y - `Component.js/css`, JS/CSS vars, `Component.Media`
3. etc...

src/django_components/dependencies.py
JuroOravec, line 52: How does the user know which JS/CSS to touch and which one to leave be?

This was added already earlier PR, but each `Script` and `Style` object has a `kind` field with one of the 4 values:
- `"core"`: Required for Django Components library to work.
- `"component"`: Dependency from a component's `Component.js` or `Component.css`.
- `"variables"`: Dependency from a component's JS/CSS variables.
- `"extra"`: Any other dependencies, e.g. from `Component.Media.js/css`.

This way, users can do some smart post-processing:
- e.g. modify only the `Component.js/css` of components by targetting only `kind == "component"`
- or modify only `"extra"`

src/django_components/dependencies.py
```diff
@@ -1203,160 +1245,252 @@ def get_component_media(comp_cls_id: str) -> Media | None:
...
+        *core_styles,
+        # CSS from `Media.css`, `Component.css` and CSS variables (if not fragment)
+        *all_css__inline,
+    ]
 
     return (content, final_scripts, final_styles)
 
 
-# Detect duplicates by URLs, extract URLs, and sort by URLs
+def _get_instance_scripts_and_styles(
```
JuroOravec, line 1267: This is a new function that collects all JS/CSS scripts for a single Component. These will be then passed to `Component.on_dependencies()`.

JuroOravec:
cc @EmilStenstrom

Issue #1587, opened 2026-02-04
Error 'Unset' object has no attribute '_render' in multi-threaded environments


oliverhaas:
We were getting the above error in our production environment with django-components 0.147.

I think in `_load_media()`, `resolved_template=True` is set before the base class early-return, leaving `_template` as `UNSET` while marked resolved. Thread 2 could `read _template` while Thread 1 was still loading.

I think this might have been introduced in #1547, but I'm just guessing since it's just the larger recent change in that area.

I'll open a PR in a second to illustrate what I mean. Not sure if my thinking and understanding is fully correct, but I'm fairly certain it's a race condition one way or another.

PR #1588, opened 2026-02-04
fix: race condition where _template remains UNSET

oliverhaas:
See #1587.

Basically the idea is here to change the order of the logic a little bit, so the idempotent loading might happen twice, but a race condition does not lead to an error.

Test is just for illustration. I would not merge the test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,17 @@
             ]
     ```
 
+#### Fix
+
+- Fixed race condition where `Component._template` was not set correctly when rendering multiple components in parallel.
+
+    See [#1587](https://github.com/django-components/django-components/issues/1587)
+
+    ```python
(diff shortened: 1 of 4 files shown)


JuroOravec:
@oliverhaas Thanks for looking into this and fixing it! The only Django project I worked on was single thread WSGI app, and I didn't work with multithreaded Python programs yet, so issues like this can fall through the cracks. So I appreciate your help.

I've fixed linter errors + bit of cleanup. The test does actually capture the error, so I think it's fine keeping it :)

oliverhaas:
> The only Django project I worked on was single thread WSGI app, and I didn't work with multithreaded Python programs yet, so issues like this can fall through the cracks. 

No worries.

I'm actually not happy about our production setup. Threading causes a couple annoying issues, but also without it we have other downsides.

PR #1589, opened 2026-02-04
refactor: set deps_strategy="ignore" when nested in another component

### LINKED ISSUE #1463: feat: Make deps_strategy optional when rendering component inside another one
This was inspired by one Reddit thread where they mentioned [Flask's `current_app`](https://stackoverflow.com/a/31279205/9788634). 

## Context

In my work project it was not uncommon to use a pattern where I generated some HTML inside `Component.get_template_data()`. I had to do it because the Table component I had implemented accepted table cells as plain strings. So I had to pre-render the HTML and then insert the rendered result inside the table:

```py
class MyPage(Component):
        table_rows = [
            {
                "name": item.name,
                "action": Icon.render(
                    kwargs={
                        name="pencil-outline",
                        prepend="Edit",
                    },
                    deps_strategy="ignore",  # <-------- required to ensure correct JS/CSS handling
                ),
        ]

        return {
            "rows": table_rows,
        }
```

The problem is that one has to pass in `deps_strategy="ignore"`, so that if `Icon` component defined and JS / CSS, then to ensure that the JS/CSS is processed with the rest of the template.

`deps_strategy` feels like a niche / advanced setting which can be easy to forget (happened even to me).

## Solution

I think it's safe to assume that when one component is being rendered in another like the `Icon` above, that this component is meant to be inserted into the encapsulating component tree.

So here is where the inspo from Flask, Django, or even Vue's `getCurrentInstance()` comes in:

- Internally we'd track the "component context" that we're in (as a stack).
  - (NOTE: Initially this would be a global state, would become a part of the `Components()` instance once https://github.com/django-components/django-components/issues/1413 is implemented)
- So when I call `MyPage.render()`, or `{% component "my_page" / %}`, we'll set the current component context to this MyPage instance.
- Then, when we reach `Icon.render()`, we'd check if we're already inside a component context.
  - If yes, `deps_strategy` will default to `"ignore"` (new behaviour)
  - If no, `deps_strategy` will default to `"document"` (old behaviour)

What we'd achieve is that we could drop the need for setting `deps_strategy` inside the nested components:

```py
class MyPage(Component):
        table_rows = [
            {
                "name": item.name,
                "action": Icon.render(
                    kwargs={
                        name="pencil-outline",
                        prepend="Edit",
                    },
                ),
        ]

        return {
            "rows": table_rows,
        }
```

And so one wouldn't have to deal with the "deps strategies" and whatnot until one needs to do some advanced stuff.

JuroOravec:
Closes https://github.com/django-components/django-components/issues/1463

This PR removes the need to pass `deps_strategy="ignore"` when rendering one component inside another:

Before:

```py
class Outer(Component):
        content = Inner.render(deps_strategy="ignore")
        return {"content": content}
```

After:

```py
class Outer(Component):
        content = Inner.render()  # no deps_strategy needed!
        return {"content": content}
```

Screenshot 2026-02-04 at 11 58 55

### DIFF
--- a/src/django_components/component_render.py
+++ b/src/django_components/component_render.py
@@ -1,6 +1,8 @@
 import re
 from collections import deque
 from collections.abc import Callable, Generator, Mapping
+from contextlib import contextmanager
+from contextvars import ContextVar
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Literal, NamedTuple, TypeAlias, cast
 from weakref import ReferenceType, WeakKeyDictionary, ref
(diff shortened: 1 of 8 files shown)


src/django_components/component_render.py
```diff
@@ -172,43 +233,44 @@ def render_with_error_trace(
...
-                registry=registry,
-                registered_name=registered_name,
-                node=node,
-            )
-        except Exception as e:
-            # Clean up if rendering fails
-            component_instance_cache.pop(render_id, None)
-            raise e from None
+        render_id = gen_component_id()
+        with _render_stack(deps_strategy) as deps_strategy_with_default:
```
JuroOravec, line 243: When we call `Component.render()`, we pass through this context manager, which tracks the depth, and thus whether we're nested in a component or not.

JuroOravec:
Note: this won't be an issue in Citry. The issue with django-components is that we have to pre-render parts of the template before we put together the final template tree. In Citry we'll be able to pass the component tree of nodes around, and render it only once, at the very end.

PR #1606, opened 2026-03-20
fix: memory leaks in CacheExtension.render_id_to_cache_key and context_processors_data WeakKeyDictionary

oliverhaas:
See#1607 

1. Change dict read to .pop() in on_component_rendered so entries are removed after use. Also clean up on the error path.
2. Remove context_processors_data WeakKeyDictionary in favor of setting the context on the request itself.

Tests for reproduction mostly, can be dropped, but I would keep them.


EmilStenstrom:
@oliverhaas How come you closed this?

oliverhaas:
@EmilStenstrom Sorry I kinda messed up my fork while trying out some stuff and then cleaning up. Had to create a new one... See https://github.com/django-components/django-components/pull/1613.

Issue #1607, opened 2026-03-20
Memory leaks in CacheExtension and context_processors_data


oliverhaas:
I think I've found two memory leaks:

1. `CacheExtension.render_id_to_cache_key`. Entries are never removed. Two paths: `on_component_rendered` reads the key but never pops it, and `on_component_input` stores the key before checking the cache. 
2. `context_processors_data` WeakKeyDictionary. The latter never collects entries because Django's context processors put the request into the cached value, creating a strong ref cycle back to the weak key. 

Actually visible in our production env, but I'm guessing many will never really notice this, since it's not big and required quite a lot of requests/renders.

See possible fixes in #1606.

EmilStenstrom:
This is now live in 0.149, thanks @oliverhaas!

PR #1613, opened 2026-04-04
fix: memory leaks in CacheExtension.render_id_to_cache_key and context_processors_data WeakKeyDictionary

oliverhaas:
See accidentally messed up previous PR #1606 and issue #1607.


EmilStenstrom:
@JuroOravec Mind having a look?

PR #1619, opened 2026-04-17
{% cache %} compatibility

jonathan-s:
@EmilStenstrom @JuroOravec 

One annoyance / issue that I stumbled upon was that you get a key error when using django's template cache. The library should really be compatible with django's template tags. 

The approach taken here is that we override django's `cache` template tag to make it component aware. That way it works for django-components and it will continue to work for normal django stuff. 

One commit also renames an e2e test such that you can do `pytest tests --ignore-glob='*e2e*'` which is quite handy when you want to run all tests that are not e2e. 

Checking out b6f7af4d9926789620f92b343ff300507e378630 and you will see the tests failing. If you add in `cache` in the load tag, you will see 1 failing test out of the 3. 

Would really appreciate if this could be patched in :)


jonathan-s:
@EmilStenstrom Thanks a lot!

PR #1623, opened 2026-04-24
[codex] Speed up test matrix

EmilStenstrom:
## Summary

This splits the slow full test run into focused CI lanes so the compatibility matrix stops paying for browser setup and heavyweight snapshot tests on every Python/Django/OS combination.

Changes include:

- Run the main tox test env with `pytest-xdist`, excluding browser E2E and benchmark snapshot tests.
- Keep `tests/test_templatetags_provide.py` serial because it has shared cache assertions that are not xdist-safe.
- Add dedicated `e2e` and `benchmark_snapshot` tox envs.
- Make the Django E2E test server opt-in instead of an autouse fixture.
- Add `DJC_TEST_BROWSERS`, defaulting to all browsers locally while CI runs Chromium in the E2E lane.
- Reduce Windows matrix coverage to smoke coverage while keeping full Linux Python/Django coverage.
- Stop installing Playwright browsers in non-E2E CI jobs.

## Timing

Measured locally before the change:

- `uv run pytest -q`: `1325 passed, 7 skipped`, `real 120.50s`

Measured locally after the split:

- main non-E2E/non-benchmark run: `1211 passed, 6 skipped`, `real 9.69s`
- Chromium E2E lane: `37 passed`, `real 11.55s`
- benchmark snapshot lane: `4 passed`, `real 1.20s`

The comparable local split is about `22.44s` total versus `120.50s` before. CI should see a larger practical win because Playwright browser installation now only happens in the E2E job.

## Validation

- `uv run tox -e py314-django52`
- `uv run tox -e e2e,benchmark_snapshot`
- `uv run ruff check tests/conftest.py tests/e2e/utils.py`
- `uv run ruff format --check tests/conftest.py tests/e2e/utils.py`
- `uv run mypy tests/conftest.py tests/e2e/utils.py`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,8 +14,21 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
-        os: [ubuntu-latest, windows-latest]
+        include:
+          - python-version: "3.10"
+            os: ubuntu-latest
+          - python-version: "3.11"
+            os: ubuntu-latest
+          - python-version: "3.12"
+            os: ubuntu-latest
+          - python-version: "3.13"
+            os: ubuntu-latest
+          - python-version: "3.14"
+            os: ubuntu-latest
+          - python-version: "3.10"
+            os: windows-latest
+          - python-version: "3.14"
+            os: windows-latest
 
     steps:
       # Configure git to handle long paths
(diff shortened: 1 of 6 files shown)


EmilStenstrom:
@JuroOravec Thoughts?

JuroOravec:
@EmilStenstrom Great idea! I remember that in the past there was a Firefox-specific bug w.r.t. to `Window.Components`. So I think there is value in testing other browsers. But instead of it being part of main CI checks, it can be e.g. a weekly pipeline. See https://github.com/django-components/django-components/pull/1652

PR #1630, opened 2026-05-01
Fix stale registry finalizer

### LINKED ISSUE #1598: `weakref.finalize` causes registry corruption
The `finalize` callback in `ComponentRegistry.register()` can unregister the *wrong* component class, leaving the registry permanently corrupted.

### Root cause

`component_registry.py:360`:

```python
finalize(entry.cls, lambda: self.unregister(name) if name in self._registry else None)
```

This finalizer captures `name` (e.g., `"button"`) and `self` (the registry) in a closure. When the component class object is garbage collected, the callback checks whether `name` is still in the registry and, if so, unregisters it.

**The callback does not verify that the class currently registered under `name` is the same class being collected.** If a different class has been registered under the same name (which happens whenever a module is re-imported), the stale finalizer unregisters the *new* class.

### The chain of events

Here is the exact sequence, with code references:

#### Phase 1: Django startup

- Component modules are imported via autodiscovery
- `Button_v1` class is created → `__init_subclass__` sets `class_id` via `hash_comp_cls()` (`component.py:3409-3413`)
- `@register("button")` calls `ComponentRegistry.register("button", Button_v1)` (`component_registry.py:317-368`)
- Finalizer `F1` is attached to `Button_v1`: *"when `Button_v1` is GC'd, call `registry.unregister("button")`"* (line 360)
- `LOADED_MODULES` is empty because `IS_TESTING` is `False` (`autodiscovery.py:108-109`)

#### Phase 2: First `@djc_test` test — setup

- `_setup_djc_global_state()` sets `IS_TESTING = True` (`testing.py:441-442`)
- `extensions._initialized = False` then `extensions._init_app(...)` (`testing.py:451-452`)
- This triggers `AutodiscoveryExtension.on_extension_created()` which calls `autodiscover()` (`extensions/autodiscovery.py`)
- `autodiscover()` calls `importlib.import_module()` for each component module (`autodiscovery.py:103`)
- Modules are already in `sys.modules`, so `import_module()` is a no-op — no new classes created
- **But** `IS_TESTING` is now `True`, so module names are appended to `LOADED_MODULES` (`autodiscovery.py:108-109`)

#### Phase 3: First `@djc_test` test — cleanup

- `_clear_djc_global_state()` removes modules from `sys.modules` (`testing.py:547-548`):
  ```python
  for mod in LOADED_MODULES:
      sys.modules.pop(mod, None)
  LOADED_MODULES.clear()
  ```
- This removes `"ui.components.button.button"` from `sys.modules`
- `gc.collect()` runs (`testing.py:566-567`) — but `Button_v1` is NOT collected because the registry still holds a strong reference to it via `entry.cls`
- `IS_TESTING` is set back to `False` (`testing.py:576`)

#### Phase 4: Between tests —  next `@djc_test` setup

- `autodiscover()` is called (via `_setup_djc_global_state()` → `extensions._init_app()`)
- Module `"ui.components.button.button"` is NOT in `sys.modules` (it was removed in Phase 3)
- `importlib.import_module()` **re-executes the module**, creating a NEW class `Button_v2`
- `Button_v2.class_id` == `Button_v1.class_id` (same `md5(module + qualname)` — see `util/misc.py:135`, `hash_comp_cls()`)
- `@register("button")` calls `registry.register("button", Button_v2)` (`component_registry.py:317-368`)
- `class_id` matches (line 345), so no `AlreadyRegistered` error — the entry is **silently overwritten** (line 357)
- New finalizer `F2` is attached to `Button_v2` (line 360)
- **Old finalizer `F1` is still alive**, attached to `Button_v1` — it was never cancelled
- `Button_v1` has **lost its last strong reference** (the registry entry now points to `Button_v2`; the module was removed from `sys.modules`)

#### Phase 5: Finalizer fires

- At some later point — during `gc.collect()` in another `@djc_test` cleanup, or during normal garbage collection triggered by memory allocation — Python's GC collects `Button_v1`
- `F1` fires: `registry.unregister("button") if "button" in registry._registry else None`
- `"button"` IS in `registry._registry` — but it maps to `Button_v2`, not `Button_v1`
- `registry.unregister("button")` executes, **removing `Button_v2`**
- **The `"button"` component is now gone from the registry permanently**

#### Phase 6: Subsequent tests fail

- Any test that renders a template using `{% component "button" %}` raises:
  ```
  django_components.component_registry.NotRegistered: The component "button" is not registered
  ```
- Calling `autodiscover()` again does NOT help if the module is already back in `sys.modules` (from the re-import in Phase 4), because `importlib.import_module()` is a no-op for cached modules

#### Option A: Detach old finalizers on re-registration

Store the finalizer object and call `.detach()` before overwriting a registry entry:

```python
# component_registry.py
class ComponentRegistry:
    def __init__(self, ...):
        ...
        self._finalizers: dict[str, weakref.finalize] = {}  # NEW

    def register(self, name, component):
        if existing_component and existing_component.cls.class_id != component.class_id:
            raise AlreadyRegistered(...)

        # Cancel old finalizer before overwriting
        old_finalizer = self._finalizers.pop(name, None)
        if old_finalizer is not None:
            old_finalizer.detach()

        entry = self._register_to_library(name, component)
        ...
        self._registry[name] = entry

        self._finalizers[name] = finalize(
            entry.cls,
            lambda: self.unregister(name) if name in self._registry else None,
        )
```

This is the most robust approach — it guarantees at most one finalizer exists per registered name.

#### Option B: Skip module removal for pre-existing modules in `_clear_djc_global_state`

Only remove modules from `sys.modules` that were **actually loaded during the test** (i.e., not present before the test started):

```python
# testing.py, in _clear_djc_global_state
# Instead of removing ALL LOADED_MODULES from sys.modules,
# only remove modules that were not present at test start.
# This requires saving sys.modules.keys() in _setup_djc_global_state.
```

This avoids the re-import chain that creates the conditions for the bug, but doesn't fix the underlying finalizer problem.

#### Option C: Combine Options A & B

Ensure that only one finalizer is persent while also only removing modules from sys.modules that are not there before tests start. This has the nice properties of fixing finalizers while also having fewer import lib calls during testing.

### Environment

- django-components 0.148.0 (likely affects all versions with the `finalize` call on line 360 of `component_registry.py`)
- Python 3.14.0
- pytest 9.0.2

EmilStenstrom:
## Summary
- Avoid keeping registered component classes alive through registry finalizer callbacks
- Prevent stale finalizers from unregistering newer replacement component classes
- Add a regression test for reimported component replacement

Fixes #1598

## Validation
- uv run pytest tests/test_registry.py

### DIFF
--- a/src/django_components/component_registry.py
+++ b/src/django_components/component_registry.py
@@ -357,7 +357,15 @@ class ComponentRegistry:
         self._registry[name] = entry
 
         # If the component class is deleted, unregister it from this registry.
-        finalize(entry.cls, lambda: self.unregister(name) if name in self._registry else None)
+        # Use the object ID so the finalizer does not keep the class alive.
+        registered_cls_object_id = id(entry.cls)
+
+        def unregister_component() -> None:
+            current_entry = self._registry.get(name)
+            if current_entry is not None and id(current_entry.cls) == registered_cls_object_id:
+                self.unregister(name)
+
+        finalize(entry.cls, unregister_component)
 
         extensions.on_component_registered(
             OnComponentRegisteredContext(
--- a/tests/test_registry.py
+++ b/tests/test_registry.py
@@ -1,3 +1,5 @@
+import gc
+
 import pytest
 from django.template import Context, Engine, Library, Template
 from pytest_django.asserts import assertHTMLEqual
@@ -35,6 +37,16 @@ class MockComponent2(Component):
 class MockComponentView(Component):
     def get(self, request, *args, **kwargs):
         pass
+
+
+def gen_reimported_component():
+    return type(
+        "ReimportedComponent",
+        (Component,),
+        {
+            "__module__": "tests.components.reimported_component",
+        },
+    )
 
 
 @djc_test
@@ -131,6 +143,22 @@ class TestComponentRegistry:
             custom_registry.register(name="testcomponent", component=MockComponentView)
         except AlreadyRegistered:
             pytest.fail("Should not raise AlreadyRegistered")
+
+    def test_stale_component_finalizer_does_not_unregister_replacement(self):
+        custom_registry = ComponentRegistry()
+
+        component_v1 = gen_reimported_component()
+        custom_registry.register(name="testcomponent", component=component_v1)
+
+        component_v2 = gen_reimported_component()
+        assert component_v1 is not component_v2
+        assert component_v1.class_id == component_v2.class_id
+
+        custom_registry.register(name="testcomponent", component=component_v2)
+        del component_v1
+        gc.collect()
+
+        assert custom_registry.get("testcomponent") is component_v2
 
     def test_simple_unregister(self):
         custom_registry = ComponentRegistry()


elliottneilclark:
Thank you @EmilStenstrom

Issue #1597, opened 2026-02-25
Better traceback errors


jonathan-s:
This kind of errors are not really helpful in identifying where things have gotten wrong. I'm pretty sure it's somewhere in the template itself, but the traceback doesn't tell me at all where I should look for the error. 

Update. 
At the end of the day the error was generated by the following code that was assigned to a variable in the template. Note that there is a comma instead of a colon in the code.  

```python
reverse_lazy("story-lock-unlock", kwargs={"story_sqid", obj.sqid})
```

```
  File "/Users/jonathan/projects/clients/3tablets/rugela-web/venv/lib/python3.13/site-packages/django/core/handlers/wsgi.py", line 124, in __call__
    response = self.get_response(request)
  File "/Users/jonathan/projects/clients/3tablets/rugela-web/venv/lib/python3.13/site-packages/django/core/handlers/base.py", line 140, in get_response
    response = self._middleware_chain(request)
  File "/Users/jonathan/projects/clients/3tablets/rugela-web/venv/lib/python3.13/site-packages/django/core/handlers/exception.py", line 57, in inner
    response = response_for_exception(request, exc)
  File "/Users/jonathan/projects/clients/3tablets/rugela-web/venv/lib/python3.13/site-packages/django/core/handlers/exception.py", line 141, in response_for_exception
    response = handle_uncaught_exception(
        request, get_resolver(get_urlconf()), sys.exc_info()
    )
  File "/Users/jonathan/projects/clients/3tablets/rugela-web/venv/lib/python3.13/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/Users/jonathan/projects/clients/3tablets/rugela-web/venv/lib/python3.13/site-packages/debug_toolbar/middleware.py", line 137, in __call__
    response = toolbar.process_request(request)
  File "/Users/jonathan/projects/clients/3tablets/rugela-web/venv/lib/python3.13/site-packages/debug_toolbar/panels/__init__.py", line 230, in process_request
    return self.get_response(request)
           ~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/Users/jonathan/projects/clients/3tablets/rugela-web/venv/lib/python3.13/site-packages/debug_toolbar/panels/__init__.py", line 230, in process_request
    return self.get_response(request)
           ~~~~~~~~~~~~~~~~~^^^^^^^^^
[...]

PR #1646, opened 2026-05-29
fix: pass origin to StringifiedNode

JuroOravec:
Follow up to #1597 and #1635.

As I'm reading through older work while I was away, I found the two issues #1597 and #1635.

After a review I found that they resolved the issue, but didn't resolve the fundamental problem that the StringifiedNode wasn't using the Origin reference, which the Django's debug mode expects.

----

When a multi-node template expression (e.g. `value="prefix {{ x }}"` passed to a
component) raised while Django template debug mode was enabled, the internal
`StringifiedNode` wrapping the expression carried no template `origin`. Django's
debug-mode traceback annotator (`Node.render_annotated`) reads `origin` to point
the error at the failing source line, so without it the annotator could not
locate the line, and in some cases masked the original exception entirely.

#1635 patched this defensively by copying `origin` off the wrapped node when
present. This PR makes the origin reliably available by threading the host
template's origin from the render context all the way down to the expression
nodes, so the annotator always has a real source to compare against.

## What changed

- `resolve_template_string()` now reads the host template's origin from the live
  render `Context` and passes it to `TemplateExpression`. This is the only
  render-time call site with access to the context, so it's where the origin is
  injected.
- `TemplateExpression.__init__` accepts an optional `origin` and stamps it onto
  the `Parser`, so every parsed node (including `StringifiedNode`) inherits a
  real origin.
- `StringifiedNode` keeps a last-resort fallback: if a node still has no origin
  (e.g. an expression built directly in code rather than via
  `resolve_template_string`), it borrows the origin from
  `context.render_context.template` at render time.
- `origin` defaults to `None`, so direct constructors (and existing callers)
  remain backward-compatible.


src/django_components/expression.py
```diff
@@ -101,13 +108,28 @@ class StringifiedNode(Node):
     def __init__(self, wrapped_node: Node) -> None:
         super().__init__()
         self.wrapped_node = wrapped_node
-        if hasattr(wrapped_node, "token"):
-            self.token = wrapped_node.token
-        if hasattr(wrapped_node, "origin"):
-            self.origin = wrapped_node.origin
```
JuroOravec, line 107: The fix from https://github.com/django-components/django-components/pull/1635 is preserved, just the IF statements were removed.

pyproject.toml
```diff
@@ -55,7 +55,7 @@ dev = [
     "pytest-django",
     "pytest-asyncio==1.3.0",
     "syrupy",
-    "ruff",
+    "ruff==0.15.15",
```
JuroOravec, line 58: Unrelated to the StringifiedNode work. In the last PR i pushed (https://github.com/django-components/django-components/pull/1645), I accidentally broke ruff checks. While looking into it, I was getting different errors locally then what I got in the CI. The reason was because locally I was on v0.15 while CI was on v0.15.15.

So pinned the version of ruff to avoid this discrepancy

src/django_components/util/template_tag.py
```diff
@@ -156,10 +156,16 @@ def resolve_template_string(
     tags: Mapping[str, Callable],
     expr: str,
 ) -> Any:
+    # During rendering, `context.template` is the host template (set via `Context.bind_template`).
+    # Pass its origin so nested expression nodes can be annotated against the correct source file
+    # when an error is raised. Guarded with `getattr` so a plain-dict context (no `.template`)
+    # degrades gracefully to `None`.
+    origin = getattr(getattr(context, "template", None), "origin", None)
```
JuroOravec, line 163: I was skeptical whether the context object indeed contains `template` attr. Opus 4.8 traced it and confirmed that `resolve_template_string` indeed receives the Django's `Context` inst and that `context.template` points to the owner Template.

```
BaseNode.wrapper_render(self, context)                       node.py:116
   context = live Django Context, bound to the host template
             via Context.bind_template() inside Template._render()
   context.template = template from bind_template().
        │
        ▼
self._params_resolver(context)                               node.py:128
        ▼
resolver(context) → compiled_tag(context)                    template_tag.py:264
        ▼
compiled_func(context)   ← Rust-generated, see below         djc_core/compile.py
        ▼
template_string(context, source, token, filters, tags, expr) ← SAME context object
        ▼
resolve_template_string(context, ...)                        template_tag.py:151
```

Issue #1650, opened 2026-05-30
[v3] Cache should store the RenderObject, not a string — fixes `{% cache %}` freeze issues


JuroOravec:
(Opus 4.8)

## TL;DR

Django's `{% cache %}` works around components in v1 (fixed in [#1619](https://github.com/django-components/django-components/pull/1619)), but it stores a finished HTML string. That string carries some values that were only meant to be valid for one render. When the cache replays the string later, those values are reused as if they were stable, and they aren't.

Two quiet failure modes follow:

1. **Same cached fragment used in multiple spots on one page** (e.g. an unparameterized sidebar embedded in both the header and footer) → each component's JS event handler ends up attached more than once. The duplicate count grows with the number of embeddings, so a click handler can quietly fire 2, 3, or 4 times per click.
2. **Server restart while the fragment cache survives** (Redis / memcached fragment cache + per-process JS/CSS variable cache) → after the restart, the cached fragment references JS/CSS variables that the new process no longer has. They silently vanish. A component that needed user-specific data (timezone, locale, anything from `get_js_data` / `get_css_data`) renders with that data missing. No errors logged.

Both come from the same root cause: the cache stores a string that has already had per-instance identifiers and per-call variable-hashes burned into it.

There's no clean fix at the string-caching level — we walked through two attempts and both have real downsides (see "Why we can't fix this cleanly in v1" below).

But in v3 ([#1004](https://github.com/django-components/django-components/issues/1004#issuecomment-3521869570)), `Component(...)` is planned to return a `RenderObject` — a small Python object describing **what** to render, similar to React's `RenderElement`. If `{% cache %}` stores that object instead of a finished string, both failure modes disappear by construction: each `.render()` call mints fresh identifiers, and the JS/CSS data is owned by the object instead of looked up via a fragile process-local cache.

This issue captures the design rationale, the concrete reproducers, and the open questions so we have all the context when v3 lands.

## Background: how `{% cache %}` works in v1 today

If you're new to djc internals or coming back after a break, here's just enough to follow the rest.

### What `{% cache %}` does in plain Django

Django's built-in `{% cache TIMEOUT KEY %}...{% endcache %}` tag renders the body once, stores the resulting HTML string in the cache backend (Redis, memcached, locmem, whatever), and on subsequent renders returns the stored string instead of re-rendering. Standard fragment caching.

### Why that didn't "just work" with components before #1619

[...]

PR #1651, opened 2026-05-30
{% cache %} compat follow-ups: docs, callback leak fix, regression tests

JuroOravec:
(Juro)

As I read through #1619, I explored and found:
- 1 bug
- few edge cases that are still problematic with the compatibility with Django's `{% cache %}` tag.

I captured the "proper" solution here https://github.com/django-components/django-components/issues/1650.

But since the proper design would be possibly only in v3, right now I only documented what to be aware of when using the `{% cache %}` tag with components.

---

(Opus 4.8)

## Summary

Follow-ups to #1619 (Django `{% cache %}` tag compatibility with components):

- **Docs.** New "Using Django's `{% cache %}` tag with components" section in [docs/concepts/advanced/component_caching.md](docs/concepts/advanced/component_caching.md). Covers the three real gotchas users can hit (same cached fragment embedded twice on a page → handlers attached more than once; JS/CSS variables silently lost after a deploy with a persistent fragment cache; `vary_on` hygiene), and documents the process-global patch of Django's built-in `{% cache %}` tag.

- **Fix.** Bounded callback-dict accumulation in `_assemble_cached_fragment`: the two synthetic callbacks added to `tree.on_component_intermediate_callbacks` / `tree.on_component_rendered_callbacks` per `{% cache %}` invocation are now popped via `try/finally`, so a template-level loop with many `{% cache %}` calls doesn't leave one dead-lambda pair per iteration in the (still-live) tree's callback dicts.

- **Tests.** Two regression tests:
  - `test_cache_embedded_twice_on_same_page_freezes_inner_render_id` — pins the documented "frozen `data-djc-id`" behavior so a future change is forced to discuss it.
  - `test_error_in_cache_body_does_not_poison_cache` — verifies that an error mid-cache-body doesn't write a partial value (the second render must run the component again, not return a poisoned hit).

- **Docs/comments.** Rewrote the [cache_tag.py](src/django_components/cache_tag.py) module docstring (the old text described an earlier, abandoned implementation approach), and annotated the `DjcCacheNode.render` body to mark which lines are copied verbatim from Django's `CacheNode.render` (Django 5.2) and which are djc-specific. Helps future maintainers re-diff against upstream when Django bumps.

The architectural fix that would dissolve both documented limitations entirely (cache `RenderObject` instead of the rendered string) is tracked in #1650 for v3.

### DIFF
--- a/src/django_components/cache_tag.py
+++ b/src/django_components/cache_tag.py
@@ -2,33 +2,64 @@
 Compatibility shim that makes Django's built-in {% cache %} tag work correctly
 inside component templates.
 
-When component_tags is loaded, this module re-registers the "cache" tag with a
-subclass of Django's CacheNode whose render() clears _COMPONENT_CONTEXT_KEY before
-rendering the nodelist on a cache miss.  The fix is transparent: templates continue
-to use {% cache %} exactly as before; no tag rename is required.
+Importing this module overwrites Django's "{% cache %}" tag with `DjcCacheNode`,
+a subclass of Django's `CacheNode`. Templates continue to use {% cache %} exactly
+as before; no tag rename is required.
 
 Background
 ----------
-django-components uses a two-pass render.  Pass 1 (bottom-up): each nested
+django-components uses a two-pass render. Pass 1 (bottom-up): each nested
 {% component %} tag detects _COMPONENT_CONTEXT_KEY in the Django Context, stores
-its renderer in a module-level dict keyed by a fresh render ID, and returns only a
-placeholder `<template djc-render-id="...">`.  Pass 2 (root assembly): the root
-component replaces each placeholder with the renderer's output.
+its renderer in a process-local dict keyed by a fresh render ID, and returns only
+a placeholder `<template djc-render-id="...">`. Pass 2 (root assembly):
+the root component walks the queue and replaces every placeholder with the
+renderer's output.
 
 When {% cache %} wraps a {% component %} inside a component template:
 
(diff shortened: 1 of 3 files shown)


src/django_components/cache_tag.py
```diff
@@ -41,43 +72,76 @@
...
+            component_tree_context=component_ctx.tree,
+            on_component_tree_rendered=lambda html: html,
+        )
+    finally:
+        component_ctx.tree.on_component_intermediate_callbacks.pop(render_id, None)
+        component_ctx.tree.on_component_rendered_callbacks.pop(render_id, None)
 
 
 class DjcCacheNode(CacheNode):
     def render(self, context: Context) -> str:
```
JuroOravec, line 126: Added commnets here to it's clearer which parts were copied from Django and which are ours.

src/django_components/cache_tag.py
```diff
@@ -41,43 +72,76 @@
...
+    # `parent_render_id=None` makes this act as a render root, processing
+    # placeholders inline. `on_component_tree_rendered` MUST stay a no-op:
+    # a normal root passes `_render_dependencies` here, which resolves
+    # `<!-- _RENDERED -->` into `<script>`/`<link>` tags. Doing that here
+    # would break dep aggregation at the outer real root.
+    #
+    # Pop the callbacks in `finally` so a template-level loop with many
+    # {% cache %} invocations doesn't leave one set of dead lambdas per
+    # iteration in the (still-live) tree's callback dicts.
+    try:
```
JuroOravec, line 111: One code change - `component_post_render()` is wrapped in `try/finally`, and callbacks are removed after teh code is done.

PR #1652, opened 2026-05-30
ci: restore cross-browser E2E coverage via weekly workflow

JuroOravec:
(Opus 4.8)

## Summary

[#1623](https://github.com/django-components/django-components/pull/1623) split the test matrix into lanes and limited the E2E lane to **Chromium only** on every PR for speed. That dropped Firefox and WebKit coverage entirely. Past browser-specific bugs (e.g. a Firefox-only quirk in how `window.Components` is populated) show this coverage has value, so this PR reintroduces it - but on a **weekly cadence** instead of paying for it on every PR.

## Changes

- **New workflow `.github/workflows/tests-cross-browser.yml`**
  - Runs `tox -e e2e` weekly (Mondays 06:00 UTC) and on manual `workflow_dispatch`.
  - Matrix over `chromium`, `firefox`, `webkit`. `fail-fast: false` keeps each browser's red/green independent for easy triage.
  - **Chromium is included as a baseline cell**: if chromium passes and firefox/webkit fail, the bug is browser-specific; if chromium also fails, it's a generic E2E regression the PR-time lane would have caught.
  - A `notify_failure` job files or updates a tracking issue (label `ci-cross-browser-failure`) **on scheduled failures only**, so manual dispatch runs don't open noisy issues. One open issue per outage: it comments on an existing open issue if present, otherwise creates one.

- **`tox.ini`** - `[testenv:e2e]` now reads `DJC_TEST_BROWSERS` via `passenv` instead of hardcoding chromium with `set_env`, so the caller controls which browsers run. (`set_env` would have overridden the caller and silently broken the cross-browser run.)

- **`.github/workflows/tests.yml`** - the PR-time `e2e` lane now sets `DJC_TEST_BROWSERS=chromium` explicitly at the step level, preserving its fast single-browser default.

- **Docs** - `docs/community/development.md` documents the test markers, the CI lanes, and how to select browsers locally with `DJC_TEST_BROWSERS`. Added an explanatory comment to the `pytest_collection_modifyitems` hook in `tests/conftest.py`.

## Notes

- No change to what runs on PRs - they stay Chromium-only and just as fast.
- Notification is intentionally scheduled-only; we can revisit alerting (Slack/Discord) separately.

### DIFF
--- a/.github/workflows/tests-cross-browser.yml
+++ b/.github/workflows/tests-cross-browser.yml
@@ -1,0 +1,117 @@
+name: Cross-browser E2E tests
+
+# Weekly run of the E2E suite against Chromium, Firefox, and WebKit.
+#
+# The PR-time `e2e` lane in tests.yml only runs Chromium for speed. This
+# workflow is the safety net that catches browser-specific regressions
+# (e.g. a Firefox-only quirk in how `window.Components` is populated).
+#
+# Chromium is included as a baseline cell so that, when a failure shows up,
+# you can tell whether the bug is browser-specific (chromium passes, the
+# other(s) fail) or a generic E2E regression (chromium also fails — likely
+# the same thing the PR-time lane would have caught).
+
+on:
+  schedule:
+    # Mondays at 06:00 UTC.
+    - cron: "0 6 * * 1"
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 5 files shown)


EmilStenstrom:
@JuroOravec Sounds like a good compromise.

